PR #6, opened 2022-05-24
Remove unnecessary steps from cargo-deny job

scouten-adobe:


scouten-adobe:
Down from 8m to 2m! 🎉

PR #14, opened 2022-05-24
Improve documentation

gpeacock:
Some documentation cleanup work

## Checklist


scouten-adobe (changes requested):
Looks good, but I'd like to make a few editing changes. Mind holding this until this evening when I can focus on this PR?

gpeacock:
This makes the Ingredient, Thumbnail and DataHash assertions crate private. As a result, some of the APIs that are not used internally are now unused code. We could strip out that code, but I expect at some point we may make these public as a lower level API. Hence the allow-unused additions.

mauricefisher64 (dismissed):
👍

sdk/src/assertion.rs
```diff
@@ -202,10 +202,12 @@ impl fmt::Debug for AssertionData {
     }
 }
 
-/// Standard Assertion data types.  Each assertion type will
-/// contain its AssertionData.  For the User Assertion type we
-/// allow a String to set the label. The AssertionData contains
-/// the data payload for the assertion and the version number for its schema (if supported).
+/// Internal Assertion structure
```
scouten-adobe, line 205: As we discussed on Slack, not sure what to do with this since `AssertionData` is not exposed from the crate.

scouten-adobe (changes requested):
I've made the edits I want to make here; would still like to see the dead-code attributes removed before merging.

sdk/src/assertion.rs
gpeacock: I changed the references to AssertionData to not be in public comments, since it is not exposed

gpeacock:
I discussed this with Maurice. Since the plan is to eventually make these Assertions public again at some point, we would rather not delete all the code just to make Clippy happy for now. I make the allow(dead_code) statements at the module level with comments as to why we are doing it.

sdk/src/assertion.rs
gpeacock: I just removed the public comments about AssertionData so they don't show up in the docs - also corrected this so that the first line is only one line and not a big paragraph.

scouten-adobe (approved):

PR #29, opened 2022-05-31
Gpeacock/bin make tests

gpeacock:
converts make_tests into a scriptable engine

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,2 @@
 [workspace]
-members = ["sdk", "c2patool"]
+members = ["sdk", "c2patool", "make_tests"]
--- a/make_tests/cargo.toml
+++ b/make_tests/cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "make_tests"
+version = "0.1.0"
+authors = ["Gavin Peacock <gpeacock@adobe.com>"]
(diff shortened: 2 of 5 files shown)


scouten-adobe:
How does this relate to #28? Same comments about naming apply here as there.

scouten-adobe (changes requested):
Same concern about naming as in #28.

make_test_images/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "make_test_images"
+version = "0.1.0"
```
scouten-adobe, line 3: If this is to live within the same repo as SDK in the long term, then I'd like to have the version number match. This would require an update to .github/workflows/ci.yml.

scouten-adobe (changes requested):
See comment on Cargo.toml. Otherwise, all is resolved.

scouten-adobe:
Actually, I totally misspoke above. Cargo.toml files all get edited by this action step:

https://github.com/contentauth/c2pa-rs/blob/main/.github/workflows/publish.yml#L74

… and no action will be required to apply that to the new make_test_images target.

scouten-adobe (approved):

PR #32, opened 2022-06-02
Export top level signing functions, hide other signature details

gpeacock:
Export top level signing functions, hide other signature details

## Checklist


sdk/src/manifest.rs
```diff
@@ -570,7 +570,7 @@ pub(crate) mod tests {
 
     use crate::{
         assertions::{c2pa_action, Action, Actions},
-        openssl::temp_signer::get_signer,
+        openssl::temp_signer::get_temp_signer,
```
mauricefisher64, line 573: Why is it called temp_signer?  There is nothing temp about it if you pass the correct certificates.

sdk/src/openssl/temp_signer.rs
```diff
@@ -262,7 +262,7 @@ pub fn get_rsa_signer<P: AsRef<Path>>(
 /// # Panics
 ///
 /// Can panic if unable to invoke OpenSSL executable properly.
-pub fn get_signer_by_alg<P: AsRef<Path>>(
+pub fn get_temp_signer_by_alg<P: AsRef<Path>>(
```
mauricefisher64, line 265: Same as above two comments

sdk/src/openssl/temp_signer.rs
```diff
@@ -62,7 +62,7 @@ use crate::{
 /// # Panics
 ///
 /// Can panic if unable to invoke OpenSSL executable properly.
-pub fn get_signer<P: AsRef<Path>>(path: P) -> (RsaSigner, PathBuf) {
+pub fn get_temp_signer<P: AsRef<Path>>(path: P) -> (RsaSigner, PathBuf) {
```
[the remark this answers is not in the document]
gpeacock: this is the internal one, in temp_signer. I renamed this one because if I promote them both to the top level, I think the one that takes existing certs should be called get_signer and this one that makes temporary certs should be the temp one.

sdk/src/manifest.rs
gpeacock: The temp_signers are the functions used for tests, where you give it a single path and it generates temp signatures in that folder. I'm not sure if we need to export the by-alg version in this case.

The regular get_signer and get_signer_from_files use existing certs.

mauricefisher64 (approved):
👍

PR #37, opened 2022-06-07
Make most jumbf_io functions crate private and move Store dependencies to Store

gpeacock:
Make most jumbf_io functions crate private and move Store dependencies to Store

## Checklist


sdk/src/jumbf_io.rs
```diff
@@ -67,28 +55,6 @@ pub fn get_assetio_handler(ext: &str) -> Option<Box<dyn AssetIO>> {
     }
 }
 
-#[cfg(target_arch = "wasm32")]
-pub fn get_assetio_handler(ext: &str) -> Option<Box<dyn AssetIO>> {
```
mauricefisher64, line 71: Where did this function go?

sdk/src/jumbf_io.rs
```diff
@@ -67,28 +55,6 @@ pub fn get_assetio_handler(ext: &str) -> Option<Box<dyn AssetIO>> {
...
-        "c2pa" => Some(Box::new(C2paIO {})),
-        "jpg" | "jpeg" => Some(Box::new(JpegIO {})),
-        "png" => Some(Box::new(PngIO {})),
-        _ => None,
-    }
-}
-
-// TODO [scouten]: Find a cleaner way to opt in or out of PDF IO.
-#[cfg(not(target_arch = "wasm32"))]
-pub fn get_cailoader_handler(asset_type: &str) -> Option<Box<dyn CAILoader>> {
```
mauricefisher64, line 82: And this function?

sdk/src/jumbf_io.rs
gpeacock: These were duplicates only created so that we could include pdf support, but we removed that, so they were really just entirely duplicate functions, so there was no longer a need to have wasm versions.

sdk/src/jumbf_io.rs
gpeacock: see above

mauricefisher64 (approved):
👍

PR #41, opened 2022-06-13
Remove need for using OpenSSL to generate certs by using pre-generated certs

mauricefisher64:
User can use the c2pa_rs without the need generate certs via opensll

## Checklist
- [ x] All applicable changes have been documented.
- [ x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.


sdk/src/manifest.rs
```diff
@@ -693,7 +693,8 @@ pub(crate) mod tests {
         let test_output = dir.path().join("wc_embed_test.jpg");
 
         //embed a claim generated from this manifest
-        let (signer, _) = get_temp_signer(&dir.path());
+        let cert_dir = fixture_path("certs");
```
gpeacock, line 696: It may make sense to have this "certs" path defined as a constant in test_utils or something, at least for  all the unit tests. Or we could have another dedicated "certs_path() function that just returns this path, since it is used so often.

sdk/src/openssl/ec_validator.rs
```diff
@@ -198,28 +191,4 @@ mod tests {
         let validator = EcValidator::new("es256");
         assert!(!validator.validate(&signature, &data, &pub_key).unwrap());
     }
-
-    #[test]
```
gpeacock, line 202: Is this no longer a valid / needed test?

sdk/src/openssl/temp_signer.rs
```diff
@@ -102,33 +96,20 @@ pub fn get_ec_signer<P: AsRef<Path>>(
...
-    let (key_name, ec_key_name) = match alg {
-        "es256" => ("ec256_key", "prime256v1"),
-        "es384" => ("ec384_key", "secp384r1"),
-        "es512" => ("ec512_key", "secp521r1"),
+    match alg {
+        "es256" | "es384" | "es512" => (),
         _ => {
             panic!("Unknown EC signer alg {:#?}", alg);
         }
-    };
```
gpeacock, line 112: I may want to use this code in c2patool at some point, but I think not having it in the sdk is a good idea.

gpeacock (approved):
A couple of suggestions included

sdk/src/openssl/ec_validator.rs
mauricefisher64: All the new certs are cert chains so this is now redundant.

scouten-adobe (changes requested):
Stand by … I'm reviewing this PR, too.

sdk/src/openssl/temp_signer.rs
```diff
@@ -287,129 +265,3 @@ pub fn get_temp_signer_by_alg<P: AsRef<Path>>(
...
-    //println!("sign_cert_path = {:#?}", sign_cert_path);
-
-    let mut pem_key_path = sign_cert_path.clone();
-    pem_key_path.set_extension("pem");
-    //println!("pem_key_path = {:#?}", pem_key_path);
-
-    (sign_cert_path, pem_key_path)
-}
-
-// The .x509 directory at the root of this repo is flagged
```
scouten-adobe, line 304: If this is no longer true, can we remove the .x509 exclusion from .gitignore?

sdk/tests/fixtures/certs/ed25519.pem
```diff
@@ -0,0 +1,3 @@
+-----BEGIN PRIVATE KEY-----
```
scouten-adobe, line 1: I'd like to have a README at the root of this new certs directory explaining that these are throwaway keys. There will likely be some question about checking in private keys into a public GitHub repo.

scouten-adobe (changes requested):
A couple minor tweaks; otherwise :+1:.

sdk/src/openssl/temp_signer.rs
mauricefisher64: I think so.

sdk/tests/fixtures/certs/ed25519.pem
mauricefisher64: 👍

scouten-adobe (approved):

PR #45, opened 2022-06-15
Adjust temp signer reserved sizes to account for large timestamps.

mauricefisher64:
Also misc small make_test fixes

## Checklist
- [ x] All applicable changes have been documented.
- [x ] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

### DIFF
--- a/make_test_images/src/make_test_images.rs
+++ b/make_test_images/src/make_test_images.rs
@@ -39,6 +39,7 @@ fn fixture_path(file_name: &str) -> PathBuf {
     path.push(file_name);
     path
 }
+
 /// Defines an operation for creating a test image
 #[derive(Debug, Deserialize)]
 pub struct Recipe {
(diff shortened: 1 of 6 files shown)


make_test_images/src/make_test_images.rs
```diff
@@ -67,7 +68,7 @@ pub struct Config {
     /// The signing algorithm to use
     pub alg: String,
     /// A url to a time authority if desired
-    pub ta: Option<String>,
+    pub tsa: Option<String>,
```
gpeacock, line 71: We decided somewhere that we are using ta instead of tsa, and for this value, I used ta_url in c2pa_tool, so it should be changed to ta_url here as well.

scouten-adobe (approved):

gpeacock (changes requested):

make_test_images/src/make_test_images.rs
mauricefisher64: I left that to you upcoming PR.  Do you want me to change it?

gpeacock: I thought you were going to change tsa to ta in the config file, but if we are going to rename the field, I'd rather rename it once.

gpeacock: changed to tsa_url consistently with most of the source code.

PR #47, opened 2022-06-15
Remove make_test_images from code coverage

scouten-adobe:
make_test_images is not really part of the SDK. IMHO we should not include it in coverage report.

## Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - "make_test_images/**/*"


gpeacock:
It is important for us to always run unit tests on this code in order to keep it up to date. It has ~75% coverage and it would be good to get that higher. So I like having coverage reports on, but we could live without them.

gpeacock (approved):
I don't agree, but I approve.

scouten-adobe:
Closing for now; will revisit another time.

PR #48, opened 2022-06-15
Remove self-signed end-entity cert support

mauricefisher64:
Self-signed end-entity certs are not allowed per the spec.  Remove support for those since we a now generating proper cert chains.

## Checklist
- [x ] All applicable changes have been documented.
- [ x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.


scouten-adobe (approved):

sdk/src/manifest_store.rs
```diff
@@ -257,7 +257,7 @@ mod tests {
         assert!(manifest_store.validation_status().is_none());
         let manifest = manifest_store.get_active().unwrap();
         assert!(!manifest.ingredients().is_empty());
-        assert_eq!(manifest.issuer().unwrap(), "Some Company");
+        assert_eq!(manifest.issuer().unwrap(), "C2PA Test Signing Cert");
```
gpeacock, line 260: This change might break some test cases, I'll check. And it will display in all the test images we generate when you view them on verify.

gpeacock: Yes it will break this test case:
    fn manifest_report_image() {
        ...
        assert_eq!(manifest.issuer().unwrap(), "Some Company");

gpeacock (changes requested):

gpeacock (approved):

sdk/src/manifest_store.rs
mauricefisher64: That is why there are new test images.

mauricefisher64: I am not sure I understand the first comment. "Some Company" would also show on the verify site.  I think the new change is more clear.

PR #57, opened 2022-06-20
Support for asynchronous signing of claims

mauricefisher64:
This PR restores the AsyncSigner trait and async signing.

## Checklist


sdk/src/cose_sign.rs
```diff
@@ -148,6 +162,73 @@ pub fn cose_sign(signer: &dyn Signer, data: &[u8], box_size: usize) -> Result<Ve
     Ok(c2pa_sig_data)
 }
 
+/// Returns signed Cose_Sign1 bytes for "data".  The Cose_Sign1 will be signed with the algorithm from `Signer`.
+#[cfg(feature = "async_signer")]
+pub async fn cose_sign_async(
```
mklein13, line 167: Can the cose_sign functionality be moved to a private method to avoid code duplication

sdk/src/cose_sign.rs
```diff
@@ -15,35 +15,18 @@ use crate::time_stamp::{cose_timestamp_countersign, make_cose_timestamp};
...
-    */
-
-    let alg = signer.alg().ok_or(Error::UnsupportedType)?;
-
-    let alg_id = match alg.as_ref() {
+fn build_unprotected_header(
+    data: &[u8],
+    alg: &str,
+    certs: Vec<Vec<u8>>,
+    ta_url: Option<String>,
```
scouten-adobe, line 26: Granted, this is an internal API, but is there any reason for this to consume the `String`? Could this be `Option<&str>` instead?

sdk/src/cose_sign.rs
```diff
@@ -15,35 +15,18 @@ use crate::time_stamp::{cose_timestamp_countersign, make_cose_timestamp};
...
+    ocsp_val: Option<Vec<u8>>,
```
scouten-adobe, line 27: Same for this. Could be `Option<&[u8]>`.

sdk/src/signer.rs
```diff
@@ -76,10 +76,29 @@ use async_trait::async_trait;
 #[async_trait]
 pub trait AsyncSigner: Sync {
     /// Returns a new byte array which is a signature over the original.
-    async fn sign(&self, data: &[u8]) -> Result<Vec<u8>>;
+    async fn sign(&self, data: Vec<u8>) -> Result<Vec<u8>>;
```
scouten-adobe, line 79: Why does this need to consume `data`?

If that _does_ need to remain in place, this PR should be marked with the `(MINOR)` tag in the title since it's an API breaking change.

sdk/src/store.rs
```diff
@@ -1977,79 +2008,67 @@ pub mod tests {
...
+        let mut claim_capture = Claim::new("capture", Some("claim_capture"));
+        create_capture_claim(&mut claim_capture).unwrap();
+
+        // Test generate JUMBF
+        // Get labels for label test
+        let claim1_label = claim1.label().to_string();
+        let capture = claim_capture.label().to_string();
+        let claim2_label = claim2.label().to_string();
+
+        /*
```
scouten-adobe, line 2042: Why the change in comment style?

scouten-adobe (changes requested):

sdk/src/store.rs
```diff
@@ -1977,79 +2008,67 @@ pub mod tests {
...
-            let sig = claim.signature_val();
-
-            assert_eq!(&sig[0..19], b"invalid signature\0\0");
-        }
-    */
+    #[cfg(feature = "async_signer")]
+    #[actix::test]
+    async fn test_jumbf_generation_async() {
+        let signer =
+            crate::openssl::temp_signer_async::AsyncSignerAdapter::new("ps256".to_string());
```
gpeacock, line 2015: We should make AsyncSignerAdaptor a pub(crate) thing in the root of the sdk.

sdk/src/cose_sign.rs
scouten-adobe: This was marked resolved, but I don't see any comment or code change that would resolve it.

sdk/src/cose_sign.rs
scouten-adobe: Same here; no comment or code change to address this comment.

sdk/src/signer.rs
scouten-adobe: This was not addressed.

scouten-adobe (changes requested):
Looks like my earlier feedback on this PR was not addressed.

scouten-adobe (approved):

PR #65, opened 2022-07-03
Oxidize openssl

YannBonnemay:

In this PR is try to remove some of the the Openssl dependance and replace it with Ring and Rustls. The motivation was some conversations with E. Scouten and A. Parsons (and also simple curiosity).

The uprooting of Openssl is not complete : I focused on signing and verifying, but left out the specific topic of ocsp checks (except for the simple "chain order" ones).

Of note :

- Rustls/Ring do not support signing/validating in all cases. As an example, you cannot verify RSA PSS with it: <https://github.com/briansmith/ring/issues/1353>, even if you can sign with it You can sign with it (<https://github.com/briansmith/ring/issues/205>). It can be a little bit confusing (<https://github.com/briansmith/ring/issues/850>).

- I did not implement OCSP checks. it seemed there was Some Assembly Required (https://github.com/rustls/rustls/issues/31, https://github.com/briansmith/webpki/issues/26).

Nice side effect : no more algorithm specific (rsa/ec/ed) structs : only one rustls signer and one validator. I think this could in turn lead to some interesting refacto in the rest of the code, but left it at that for now. 
I put the changes behind a with_rustls feature flag (default is on).

All in all, this wants to show how oxidizing Openssl in c2pa could look like, and try to be a draft for such PRs : this is a demo PR that has not been extensively tested.

## Checklist
- [] This PR represents a single feature, fix, or change.


sdk/tests/fixtures/rustls/es256_private_key.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
YannBonnemay, line 1: That mass of test files felt a little loud, so I put them in their own directory

sdk/src/error.rs
```diff
@@ -224,6 +224,22 @@ pub enum Error {
     #[cfg(feature = "file_io")]
     OpenSslError(#[from] openssl::error::ErrorStack),
 
+    #[error(transparent)]
```
YannBonnemay, line 227: Here I added as need arised as I was a bit unsure on what are the policies on c2pr error management.

sdk/src/cose_validator.rs
```diff
@@ -1001,6 +1014,58 @@ pub mod tests {
         }
     }
 
+    #[test]
+    #[cfg(feature = "with_rustls")]
+    fn test_verify_cose_good_with_rustls() {
```
YannBonnemay, line 1019: These use the new fixtures.

sdk/tests/scripts/generate_fixtures.py
```diff
@@ -0,0 +1,197 @@
...
+        "private_key": os.path.join(fixtures_dir, "%s_private_key.pem" % test_id),
+        "public_key": os.path.join(fixtures_dir, "%s_public_key.pem" % test_id),
+        "public_key_der": os.path.join(fixtures_dir, "%s_public_key.der" % test_id),
+        "signed_data": os.path.join(fixtures_dir, "%s_signed_data.sign" % test_id),
+        "fixtures_data": os.path.join(fixtures_dir, "../data.data"),
+    }, **FIXTURES_SPECIFICS[test_id]}
+
+def generate_fixtures_rsa(data):
+    cmds = [
+        "openssl genpkey -algorithm {algorithm} -out {private_key}".format(**data), # generate private key
```
YannBonnemay, line 75: The actual test fixtures generation is actually a side effect, I think : this was documentation for me.

sdk/src/openssl/ec_signer.rs
```diff
@@ -243,6 +243,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(not(feature = "with_rustls"), ignore)]
```
YannBonnemay, line 246: These are `feature`d  out when `with_rustls`, so I had to add that here.

sdk/src/rustls/common.rs
```diff
@@ -0,0 +1,122 @@
...
+    iter,
+};
+
+pub struct AlgorithmData {
+    pub rustls_id: rustls::SignatureScheme,
+    pub verification_alg: &'static dyn ring::signature::VerificationAlgorithm,
+    // This is because the Ring verification algorithm expects the key in a SubjectPublicKey, Openssl gives it in a SubjectPublicKeyInfo.
+    // We can ask Openssl to give us that with RSAPublicKey_out, but that's (i) only for RSA, (ii) not what happens in existing tests.
+    // https://github.com/briansmith/ring/issues/881#issuecomment-592749266
+    pub spk_offset: usize,
```
YannBonnemay, line 28: Flagging that hack here. As the structures are well defined, it should be safe to just fast forward to the bytes where the usable part is.

Still.

sdk/src/rustls/temp_signer.rs
```diff
@@ -0,0 +1,468 @@
...
+// This function panics if unable to invoke openssl as expected.
+//
+fn create_x509_key_pair(
+    sign_cert_path: &Path,
+    pem_key_path: &Path,
+    has_priv_key: bool,
+    rsa_padding_mode: Option<&str>,
+    sha_mode: Option<&str>,
+) {
+    let mut openssl = Command::new("openssl");
```
YannBonnemay, line 106: I took care of the openssl uses in the Rust code, but and did not change these calls to it.

sdk/src/rustls/validator.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+impl CoseValidator for Validator {
+    fn validate(&self, sig: &[u8], data: &[u8], pkey: &[u8]) -> Result<bool> {
+        let algorithm_data = &get_algorithm_data(&self.alg).map_err(|_err| Error::CoseSignature)?;
+
+        let public_key = signature::UnparsedPublicKey::new(
+            algorithm_data.verification_alg,
+            &pkey[algorithm_data.spk_offset..],
```
YannBonnemay, line 36: Offset used here (see comment in common.rs)

sdk/src/cose_validator.rs
```diff
@@ -1017,25 +1082,25 @@ pub mod tests {
     #[test]
     fn test_verify_ec_good() {
         // EC signatures
-        let mut validator = get_validator("es384").unwrap();
+        // let validator = get_validator("es384").unwrap();
 
-        let sig_es384_bytes = include_bytes!("../tests/fixtures/sig_es384.data");
-        let data_es384_bytes = include_bytes!("../tests/fixtures/data_es384.data");
-        let key_es384_bytes = include_bytes!("../tests/fixtures/key_es384.data");
+        // let sig_es384_bytes = include_bytes!("../tests/fixtures/sig_es384.data");
```
YannBonnemay, line 1087: This one is a little mystery - I can't have these tests pass using the Ring backent, and I can't seem to be able to validate them with command line openssl (`openssl dgst -sha256 -verify ../tests/fixtures/key_es384.data -signature ../tests/fixtures/sig_es384.data ../tests/fixtures/data_es384.data`).

My generated tests, otoh, work ok with both Openssl and Ring validation.

Left commented version to flag the oddity.

mauricefisher64: Do you handle IEEE P1363 signature format?  That is what is spec'd by C2PA

sdk/src/validator.rs
```diff
@@ -59,24 +60,28 @@ impl CoseValidator for DummyValidator {
 // • PS512 (RSASSA-PSS using SHA-512 and MGF1 with SHA-512)
 // • RS256	RSASSA-PKCS1-v1_5 using SHA-256
 // • RS384	RSASSA-PKCS1-v1_5 using SHA-384
-// • RS512	RSASSA-PKCS1-v1_5 using SHA-512
+// • RS512	RSASSA-PKCS1-v1_5 using SHA-512 // validation not supported by Ring
 // • ED25519 Edwards Curve ED25519
```
mauricefisher64, line 64: All RSAxxx have been deprecated by C2PA so it OK to remove these.

sdk/tests/scripts/generate_fixtures.py
mauricefisher64: We now provide sample certs for all signature types as to remove the need for a developer just trying the SDK from needing OpenSLL to generate certs.

sdk/src/cose_validator.rs
YannBonnemay: Complete oversight on my part. Should be possible to re-add it, inasmuch as the packing/unpacking seems to happens outside of openssl (seems to be the case?).

sdk/src/validator.rs
YannBonnemay: Excellent, I'll remove that.

sdk/tests/scripts/generate_fixtures.py
YannBonnemay: Added to the chopping block.

sdk/src/rustls/rustls_signer.rs
```diff
@@ -0,0 +1,208 @@
...
+        signcert: &[u8],
+        pkey: &[u8],
+        alg: String,
+        tsa_url: Option<String>,
+    ) -> Result<Self> {
+        // Get signcerts, the certificates vector
+        let signcerts = get_certificates(signcert);
+        let pkey = get_private_key(&alg, pkey)?;
+
+        check_chain_order(&signcerts);
```
mauricefisher64, line 83: Any reason to call this twice

sdk/tests/fixtures/rustls/es256_private_key.pem
YannBonnemay: Removed

sdk/tests/scripts/generate_fixtures.py
YannBonnemay: New tests were a god send. Thanks!

sdk/src/rustls/common.rs
YannBonnemay: Removed

sdk/src/rustls/validator.rs
YannBonnemay: Removed

sdk/src/cose_validator.rs
YannBonnemay: IEEE P1363 managemen added

sdk/src/rustls/rustls_signer.rs
YannBonnemay: None, removed

mauricefisher64:
OCSP is about to be enabled so if you can find a solution for that it would be great.  Also trust list support is also coming so if you have a way to validate a certificate chain against a set of trust anchors that would be good.  Does this PR work with WASM?

PR #71, opened 2022-07-12
Update XMP Toolkit to 0.5.0

scouten-adobe:


sdk/src/embedded_xmp.rs
```diff
@@ -22,20 +24,28 @@ use xmp_toolkit::{OpenFileOptions, XmpFile, XmpFileError, XmpMeta};
...
-    m.set_property("http://purl.org/dc/terms/", "provenance", manifest_uri);
-    f.put_xmp(&m);
+    f.put_xmp(&m).map_err(xmp_write_err)?;
     f.close();
 
     Ok(())
 }
+
+fn xmp_write_err(_err: XmpError) -> crate::Error {
+    Error::XmpWriteError
```
gpeacock, line 50: Maybe add an error log for the actual error so we have some way to debug the source error?
error!("{:?}",err)

sdk/src/embedded_xmp.rs
```diff
@@ -22,20 +24,28 @@ use xmp_toolkit::{OpenFileOptions, XmpFile, XmpFileError, XmpMeta};
...
+pub(crate) fn add_manifest_uri_to_file<P: AsRef<Path>>(path: P, manifest_uri: &str) -> Result<()> {
+    XmpMeta::register_namespace("http://purl.org/dc/terms/", "dcterms").map_err(xmp_write_err)?;
+
+    let mut f = XmpFile::new().map_err(xmp_write_err)?;
+
+    f.open_file(path, OpenFileOptions::default().for_update())
+        .map_err(xmp_write_err)?;
 
-    let mut f = XmpFile::new();
+    let mut m = match f.xmp() {
```
gpeacock, line 35: Does this work?
let mut m = f.xmp().unwrap_or_else(XmpMeta::new).map_err(xmp_write_err)?;

gpeacock (changes requested):

sdk/src/embedded_xmp.rs
scouten-adobe: Tried that. It would have worked if both `f.xmp()` and `XmpMeta::new` returned `Result`, but they do not.

`f.xmp()` is an `Option` because a file not having XMP is a legitimate, non-error condition. And I really don't like the ergonomics of `Result<Option<xxx>>`, so `f.xmp()` folds error conditions into a `None` response.

sdk/src/embedded_xmp.rs
scouten-adobe: Good suggestion. Added.

dkozma (approved):

sdk/src/embedded_xmp.rs
gpeacock: ok, just checking

gpeacock (approved):

sdk/src/embedded_xmp.rs
mauricefisher64: How do you create an empty XmpFile or is that not possible?

PR #72, opened 2022-07-12
(MINOR) Move crate-level functions for creating signers to new public `create_signer` mod

scouten-adobe:
Turns out it is legal in Rust to publicly export a function as a top-level entity in a crate, but such a function does not appear in the documentation for the crate.

For that reason, we're moving the `get_signer` and `get_signer_from_files` functions into a new `create_signer` module.

## Checklist


gpeacock:
We are currently not documenting anything behind a feature, because that is the default used when publishing via cargo docs. To fix that we need to add this to the c2pa.rs Cargo.toml file:

```
[package.metadata.docs.rs]
all-features = true
```
If we make this change, do we still need this PR?

scouten-adobe:
The missing-doc issue addressed via #73, but I would like to discuss this PR as a possibly-preferred API pattern. (A preliminary local test suggests rustdoc doesn't handle this case well, even with the new flags introduced in #73.)

sdk/src/create_signer.rs
```diff
@@ -0,0 +1,109 @@
...
+///
+/// # Arguments
+///
+/// * `signcert` - Signing certificate
+/// * `pkey` - Private key
+/// * `alg` - Format for signing. Must be one of the supported
+///    formats (`rs256`, `rs384`, `rs512`, `ps256`, `ps384`, `ps512`,
+///    `es256`, `es384`, `es512`, or `ed25519`).
+/// * `tsa_url` - Optional URL for a timestamp authority
+pub fn from_keys(
```
gpeacock, line 42: Is there a reason this is gated on the file_io feature? It's ok since you can't sign anything without file_io for now, but  just wondering.

gpeacock (approved):
I'm ok with this. I would be cleaner if Signer was a struct and these were constructor methods on it, but since Signer is a trait, this makes sense.

scouten-adobe:
@gpeacock agreed that this would be cleaner if we could make these static methods on `Signer` itself, but I've re-checked and did not find a way to do that. Merging as is.

PR #76, opened 2022-07-15
(MINOR) Introduce a new `SigningAlg` enum

scouten-adobe:
This PR replaces almost all instances of the `&str` parameters that identify signing algorithms (`"ps256"`, etc.) with a new `SigningAlg` enum and pushes any conversion to/from the existing string identifiers to a relatively few API edges.

## Checklist


scouten-adobe:
I'm heading out for the evening; will circle back and figure out the WASM failure in the morning.

sdk/src/wasm/webcrypto_validator.rs
```diff
@@ -331,40 +329,40 @@ pub async fn validate_async(alg: &str, sig: &[u8], data: &[u8], pkey: &[u8]) ->
...
-                "SHA-512".to_string(),
-                0,
-                pkey.to_vec(),
-                sig.to_vec(),
-                data.to_vec(),
-            )
-            .await
-        }
-        "es256" => {
+        // "rs256" => {
```
gpeacock, line 332: should this be deleted?

make_test_images/src/make_test_images.rs
```diff
@@ -36,6 +37,7 @@ fn get_signer_with_alg(alg: &str) -> c2pa::Result<Box<dyn Signer>> {
     signcert_path.push(format!("../sdk/tests/fixtures/certs/{}.pub", alg));
     let mut pkey_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     pkey_path.push(format!("../sdk/tests/fixtures/certs/{}.pem", alg));
+    let alg: SigningAlg = alg.parse().map_err(|_| c2pa::Error::UnsupportedType)?;
```
gpeacock, line 40: I was wondering how this would work. We will need to do the same in c2patool. I'm presuming that due to the string mapping, when we convert one of these values to json, we just get the string representation. It needs to round trip well in serde.

gpeacock (approved):
I'd been thinking we should do this, thanks for taking it on!

make_test_images/src/make_test_images.rs
scouten-adobe: If you have a SigningAlg instance, you can call `to_string()` on it and it will round-trip. Off the top of my head (not actually tested), this should work:

```rust
let alg = SigningAlg::Ps256;
assert_eq!(alg.to_string(), "ps256");
```

sdk/src/wasm/webcrypto_validator.rs
scouten-adobe: The issue is that – by design – the `SigningAlg` enum does not provide the possibility to describe any algorithms that are not officially supported by C2PA. The whole point of this PR is to make it a compile-time error to attempt to specify any other signing algorithm.

That's why the Rs*** algorithms can't really be supported in this model.

scouten-adobe: Marking as resolved per Slack conversation with @mauricefisher64.

PR #77, opened 2022-07-15
Use rsa crate for RSA-PSS verification in Wasm

dkozma:
After doing some inspecting with @mauricefisher64, it looks like WebCrypto is incompatible with some of the RSA-PSS-specific parameters in the new properly-formed test certs. This swaps out WebCrypto to use the [`rsa` crate](https://crates.io/crates/rsa) for RSA-PSS processing. It also includes:

- Removing support for the `RSASSA-PKCS1-v1_5` algorithm since it is no longer supported as of the 1.0 spec
- Updating the RSA-PSS key/sig/data fixture files and naming them more appropriately

## Checklist


sdk/Cargo.toml
```diff
@@ -74,22 +73,28 @@ url = "2.2.2"
 ureq = "2.4.0"
 instant = "0.1.0"
 openssl = { version = "0.10.31", features = ["vendored"], optional = true }
+sha2 = "0.9.5"
```
scouten-adobe, line 76: Is there a reason why we aren't using the same version of `sha2` in WASM and non-WASM builds?

scouten-adobe (approved):

sdk/Cargo.toml
dkozma: I thought it broke a bunch of tests when I was developing but it seems good when I try it now - updating.

mauricefisher64 (approved):
👍

PR #79, opened 2022-07-16
(MINOR) `IngredientOptions` allow override of hash and thumbnail generation; image library is now a default feature

gpeacock:
IngredientOptions is now a trait with three methods to override title, hash and thumbnail generation
The blake3 hash is no longer built in, if a hash is desired it can be added via ingredientOptions.
add_thumbnails feature now controls if the image crate is included in a build, by default it is enabled
updated to latest image library
added additional file type/mime mappings for ingredients

## Checklist


gpeacock:
This is required for allowing PS to not generate thumbnails via the cai-helper
Also allows for removal or replacement of the default thumbnail generation code.
Removes blake3 ingredient hash from sdk, developers can now supply their own if needed.

sdk/src/manifest.rs
```diff
@@ -746,6 +782,7 @@ pub(crate) mod tests {
 
         assert_eq!(manifest.format(), "image/jpeg");
         assert_eq!(manifest.title(), Some("wc_embed_test.jpg"));
+        #[cfg(feature = "add_thumbnails")]
```
scouten-adobe, line 785: Again, test the feature-disabled case?

sdk/src/utils/mod.rs
```diff
@@ -17,6 +17,7 @@ pub(crate) mod hash_utils;
 #[allow(dead_code)] // for wasm build
 pub(crate) mod patch;
 #[cfg(feature = "file_io")]
```
scouten-adobe, line 19: I think there's a `cfg(all(feature(...)` syntax that might be shorter here.

scouten-adobe (changes requested):

sdk/src/ingredient.rs
```diff
@@ -794,7 +841,9 @@ mod tests_file_io {
 
         println!("ingredient = {}", ingredient);
         assert_eq!(ingredient.title(), "libpng-test.png");
+        #[cfg(feature = "add_thumbnails")]
```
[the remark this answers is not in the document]
gpeacock: It took some doing, but I added no-default-features test cases

sdk/src/ingredient.rs
```diff
@@ -828,14 +878,15 @@ mod tests_file_io {
...
-        assert_eq!(&ingredient.title, PRERELEASE_JPEG);
-        assert_eq!(&ingredient.format, "image/jpeg");
-        assert!(ingredient.thumbnail.is_some());
-        assert!(ingredient.provenance.is_some());
-        assert!(ingredient.manifest_data.is_none());
-        assert!(ingredient.validation_status.is_some());
+        assert_eq!(ingredient.title(), PRERELEASE_JPEG);
+        assert_eq!(ingredient.format(), "image/jpeg");
+        #[cfg(feature = "add_thumbnails")]
+        assert!(ingredient.thumbnail().is_some());
```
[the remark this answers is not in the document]
gpeacock: added test case

sdk/src/manifest.rs
scouten-adobe: What should `manifest.thumbnail()` return when `add_thumbnails` feature is disabled?

sdk/src/ingredient.rs
```diff
@@ -649,17 +660,45 @@ impl std::fmt::Display for Ingredient {
...
 }
 
-#[derive(Default)]
-/// This defines optional actions when creating [`Ingredient`]s from files.
-pub struct IngredientOptions {
-    /// This allows setting the title for the ingredient. (If `None`, then the default behavior is to use the file's name.)
-    pub title: Option<&'static str>,
+/// This defines optional operations when creating [`Ingredient`] structs from files.
+#[cfg(feature = "file_io")]
+pub trait IngredientOptions {
```
scouten-adobe, line 665: Disregard my previous comment (which I've now deleted). Allowing for overriding of hash and thumbnail generation  makes sense and justifies the change from struct to trait.

sdk/src/ingredient.rs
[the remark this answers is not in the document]
scouten-adobe: I actually had something much simpler in mind: I was thinking you could add something like ...

```rust
#[cfg(not(feature = "add_thumbnails")]
assert!(ingredient.thumbnail().is_none()); // or whatever outcome is intended
```

… within the same test.

scouten-adobe (changes requested):

sdk/src/utils/mod.rs
gpeacock: done

sdk/src/ingredient.rs
gpeacock: updated

sdk/src/manifest.rs
gpeacock: It returns the same things it does when enabled. The add_thumbnails feature only adds the ability to auto generate jpeg and png thumbnails. What manifest. thumbnail() returns depends on the test case. In some of these test cases, thumbnails will not be generated with add_thumbnails disabled, so we need to test for None in those cases only.

make_test_images/src/make_test_images.rs
```diff
@@ -93,6 +93,26 @@ impl Default for Config {
     }
 }
 
+/// Generate a blake3 hash over the image in path using a fixed buffer
+fn blake3_hash(path: &Path) -> Result<String> {
```
mauricefisher64, line 97: Just a side note.  I added blake3 support to hash_utils so it will be generally available.

sdk/src/ingredient.rs
```diff
@@ -649,17 +660,45 @@ impl std::fmt::Display for Ingredient {
...
+        None
+    }
+
+    /// Returns an optional thumbnail image representing the asset
+    ///
+    /// The first value is the content type of the thumbnail, i.e. image/jpeg
+    /// The second value is bytes of the thumbnail image
+    /// The default is to have no thumbnail, so you must provide an override to have a thumbnail image
+    fn thumbnail(&self, _path: &Path) -> Option<(String, Vec<u8>)> {
+        #[cfg(feature = "add_thumbnails")]
```
mauricefisher64, line 687: This can now be written as:
if cfg!(feature  = "add_thumbnails") {
do something
} else {
do something else 
}

gpeacock: Ah is that new? I didn't know that was an option. thanks.

make_test_images/src/make_test_images.rs
```diff
@@ -93,6 +93,26 @@ impl Default for Config {
...
+    use std::fs::File;
+    use std::io::Read;
+    // Hash an input incrementally.
+    let mut hasher = blake3::Hasher::new();
+    const BUFFER_LEN: usize = 1024 * 1024;
+    let mut buffer = [0u8; BUFFER_LEN];
+    let mut file = File::open(path)?;
+    loop {
+        let read_count = file.read(&mut buffer)?;
```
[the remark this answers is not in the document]
gpeacock: I'll replace this with your code once everything gets merged

make_test_images/src/make_test_images.rs
gpeacock: I'll call that instead of what I wrote when we get things merged.

mauricefisher64 (approved):
👍

make_test_images/src/make_test_images.rs
gpeacock: i think that means we will have a Blake3 implementation in in the SDK. The question is if I should use it by default or make it a public API to to called only when desired.

sdk/src/ingredient.rs
gpeacock: updated to use this syntax

make_test_images/src/make_test_images.rs
gpeacock: I'l integrate with the new implementation after this PR

PR #84, opened 2022-07-20
Optimize performance of large assets

mauricefisher64:
This is a series of optimizations that primarily benefit large assets;

-  Interleaved IO for Sha2 and Blake3 hashing
-  patch API for asset handlers to directly insert manifest without needing to copy files
- remove unnecessary file copies
- no more memory buffers for large assets

## Checklist


scouten-adobe (approved):

sdk/src/ingredient.rs
```diff
@@ -407,8 +406,6 @@ impl Ingredient {
         if let Some(opt_title) = options.title(path) {
             ingredient.title = opt_title;
         }
-        // read the file into a buffer for processing
-        let buf = std::fs::read(path).map_err(wrap_io_err)?;
```
gpeacock, line 411: I used to use the same buffer for hash and load_from_memory, so this all works out fine now.

sdk/src/manifest.rs
```diff
@@ -644,7 +644,8 @@ impl Manifest {
         // convert the manifest to a store
         let mut store = self.to_store()?;
         // sign and write our store to to the output image file
-        store.save_to_asset(source_path.as_ref(), signer, dest_path.as_ref())?;
+        // set source and dest to same path to avoid an unnecessary copy since we already copied above
+        store.save_to_asset(dest_path.as_ref(), signer, dest_path.as_ref())?;
```
gpeacock, line 648: There's problem here. We only copy the source to the dest above if the dest does not already exist. If it does, then the two are different files. So if we want this optimization we have to check for the case where dest is a different file.

sdk/src/ingredient.rs
mauricefisher64: The problem was when I loaded a 40GB file this buf had trouble on some systems.  We no longer do that.

sdk/src/manifest.rs
gpeacock: FYI, I have a plan to redo the command line options on c2patool so that this kind of thing will be clear.

sdk/Cargo.toml
```diff
@@ -24,6 +24,7 @@ bmff = []  # Work in progress support for BMFF-based containers
 file_io = ["openssl"]
 serialize_thumbnails = []
 xmp_write = ["xmp_toolkit"]
+no_interleaved_io = ["file_io"]
```
gpeacock, line 27: I'm not sure we need this feature, I suppose it is possible someone might want to build a single threaded version, but I think it is more likely that we will want to add threading in more places.

gpeacock (changes requested):
Some comments and one requested change for manifest.embed()

sdk/src/ingredient.rs
gpeacock: yes, I'm just saying this is fine now since I didn't need the buffer anymore anyway.

sdk/src/manifest.rs
gpeacock: maybe change the value for source_path only if you copy the source to the dest?

sdk/src/ingredient.rs
mauricefisher64: Got, I wasn't sure if you wanted to keep it.

sdk/Cargo.toml
mauricefisher64: It may be the case for some environments.

sdk/src/manifest.rs
mauricefisher64: If the only reason you are doing the copy is to get the info.  Why no just set the info from the source when the dest does not exit.  Then you have no copy code here:

 if !dest_path.as_ref().exists() {
         // first add the information about the target file
        self.set_asset_from_path(source_path.as_ref());
 } else {
        // first add the information about the target file
        self.set_asset_from_path(dest_path.as_ref());
}

mauricefisher64: I changed  to logic to  make this more clear

PR #87, opened 2022-07-21
Add support for remotely generated CoseSign1 signatures

mauricefisher64:
In order to support Claims signed by a remote service we must be able to accept the results and embed those.  This PR
provides a new trait RemoteSigner that returns a complete CoseSign1 object instead of just the signature value. 

## Checklist


sdk/src/store.rs
```diff
@@ -1440,6 +1440,32 @@ impl Store {
         }
     }
 
+    /// Embed the claims store as jumbf into an asset using an CoseSign box generated remotely. Updates XMP with provenance record.
+    #[cfg(feature = "async_signer")]
+    pub async fn save_to_asset_remotely_signed(
```
gpeacock, line 1445: May want to change this to  save_to_asset_remote_signed to match  Manifest.embed_remote_signed.

gpeacock (approved):

sdk/src/store.rs
mauricefisher64: Sure

scouten-adobe (approved):
Made a few doc format tweaks, but with that I'm good

PR #89, opened 2022-07-22
Fix bad response for case when there is no timestamp

mauricefisher64:
If there is no timestamp that is not an error.  

## Checklist


gpeacock (approved):

scouten-adobe (changes requested):
Can you add a unit test for the no-timestamp case, so we can detect any future regression?

Issue #92, opened 2022-07-25
Bad example in Crate docs, `from_file` not implemented in ManifestStore


ronald-das:
Hi, while going through the example in https://docs.rs/c2pa/latest/c2pa/#example-reading-a-manifeststore  
Am getting not implemented error (E0599)
```
error[E0599]: no function or associated item named `from_file` found for struct `ManifestStore` in the current scope
 --> src/main.rs:4:41
  |
4 |     let manifest_store = ManifestStore::from_file("tests/fixtures/C.jpg")?;
  |                                         ^^^^^^^^^ function or associated item not found in `ManifestStore`

```
My cargo.toml file
```
[dependencies]
c2pa = "0.11.0"
```

ronald-das:
After going through the struct documentation https://docs.rs/c2pa/latest/c2pa/struct.ManifestStore.html#method.from_file , `file_io` feature needs to enabled for it to work.

New Cargo.toml file
```
[dependencies]
c2pa = {version="0.11.0", features=["file_io"]}
```

scouten-adobe:
@ronald-das thanks for bringing this to our attention. I'm reopening the ticket. I think we should add a note in the README to mention that `file_io` feature is needed. @crandmck or @gpeacock, would one of you please add this?

crandmck:
Should we just change the `Cargo.toml` file in "Usage" section to be 
```
c2pa = {version="0.11.0", features=["file_io"]}
```
or should we qualify it by just adding something like "If you want to read or write a manifest file, add this to your `Cargo.toml`..." ?

crandmck:
Addressed (partially, at least) by #93.

PR #93, opened 2022-07-29
(IGNORE) Clarify dependency requirements

crandmck:

Clarify when you need to add dependencies to Cargo file, per #92.

## Checklist

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -24,7 +24,7 @@
 //! EXCEPTION: If you are building for WASM, do not add this dependency. 
 //! For example:
 //!
-//! ```
+//! ```text
 //! c2pa = {version="0.11.0", features=["file_io"]}
 //! ```
 //!
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -20,6 +20,12 @@
 //! This library supports reading, creating and embedding C2PA data
 //! with JPEG and PNG images.
 //!
+//! To read or write a manifest file, you must add the `file_io` dependency to your Cargo.toml; for example:
+//!
+//! ```
+c2pa = {version="0.11.0", features=["file_io"]}
+//! ```
+//!
 //! # Example: Reading a ManifestStore
 //!
 //! ```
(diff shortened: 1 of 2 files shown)


sdk/src/lib.rs
```diff
@@ -20,6 +20,14 @@
 //! This library supports reading, creating and embedding C2PA data
 //! with JPEG and PNG images.
 //!
+//! To read or write a manifest file, you must add the `file_io` dependency to your Cargo.toml.
+//! EXCEPTION: If you are building for WASM, do not add this dependency. 
+//! For example:
+//!
+//! ```
```
scouten-adobe, line 27: This is probably what is failing because it's being parsed as a Rust code example that should be tested. I'll need to get back to you on the proper syntax for excluding this from being executed in this way.

scouten-adobe: Since line 28 isn't Rust code, you'll need to tell the doc test engine to not attempt to run it as such. Per https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes, you can change line 27 to (three backticks)`text` and that should pass.

crandmck: Did that. S/be fixed I hope.

crandmck:
I think this addresses the issue, but not sure it covers everything. 

For example, @gpeacock told me:
> async_signer is currently required for embed_async_signed and embed_remote_signed. I don’t think any of the other features are needed for methods, but they gate functionality.

Is it sufficient that the API docs state that, e.g. https://docs.rs/c2pa/latest/c2pa/struct.Manifest.html#method.embed_async_signed ?

Do we need to mention anywhere adding `bmff` feature to dependencies?

scouten-adobe:
Looks like `rustfmt` balked at a trailing space on line 11 in lib.rs; otherwise, looks good to go.

sdk/src/lib.rs
crandmck: @scouten-adobe Hm, it's still failing...  If nothing else, I could put it inline, e.g. `c2pa = {version="0.11.0", features=["file_io"]}` unless there is a better way.

gpeacock (approved):

PR #101, opened 2022-08-08
Added support for external manifests

mauricefisher64:
This PR adds the ability to generate external manifests and detached manifests.  Reading external/detached in manifests will be a separate PR

## Checklist

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -25,7 +25,6 @@ file_io = ["openssl"]
 serialize_thumbnails = []
 xmp_write = ["xmp_toolkit"]
 no_interleaved_io = ["file_io"]
-external_manifest = ["xmp_write", "file_io"]
 
 # The diagnostics feature is unsupported and might be removed.
 # It enables some low-overhead timing features used in our development cycle.
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -80,13 +80,13 @@ tempfile = "3.1.0"
 thiserror = ">= 1.0.20, < 1.0.32"
 time = ">= 0.2.23"
 twoway = "0.2.1"
+url = "2.2.2"
 uuid = { version = "0.8.1", features = ["serde", "v4", "wasm-bindgen"] }
 x509-parser = "0.11.0"
 x509-certificate = "0.12.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 ring = "0.16.20"
-url = "2.2.2"
 ureq = "2.4.0"
 image = { version = "0.24.2", optional = true }
 instant = "0.1.0"
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -25,6 +25,7 @@ file_io = ["openssl"]
 serialize_thumbnails = []
 xmp_write = ["xmp_toolkit"]
 no_interleaved_io = ["file_io"]
+external_manifest = ["xmp_write", "file_io"]
 
 # The diagnostics feature is unsupported and might be removed.
 # It enables some low-overhead timing features used in our development cycle.
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "c2pa"
-version = "0.11.1"
+version = "0.11.3"
 description = "Rust SDK for C2PA (Coalition for Content Provenance and Authenticity) implementors"
 authors = ["Maurice Fisher <mfisher@adobe.com>", "Gavin Peacock <gpeacock@adobe.com>", "Eric Scouten <scouten@adobe.com>", "Leonard Rosenthol <lrosenth@adobe.com>", "Dave Kozma <dkozma@adobe.com>"]
 license = "MIT OR Apache-2.0"
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -233,10 +233,10 @@ pub enum AssertionStoreJsonFormat {
     OrderedListNoBinary, // list of Assertions as json objects omitting binaries results
 }
 
-// Remote manifest options
+/// Remote manifest options. Use 'set_remote_manifest' to generate external manifests.
 #[derive(Clone, Debug, PartialEq)]
-pub(crate) enum RemoteManifest {
-    NoRemote,       // No external manifest
+pub enum RemoteManifest {
+    NoRemote,       // No external manifest (default)
     SideCar,        // Manifest will be saved as a side car file, output asset is untouched.
     Remote(String), // Manifest will be saved as a side car file, output asset will contain remote reference
 }
@@ -430,15 +430,17 @@ impl Claim {
         self.update_manifest
     }
 
-    pub(crate) fn set_remote_manifest(&mut self, remote_url: Option<&str>) -> Result<()> {
-        if let Some(u) = remote_url {
-            let parsed = url::Url::parse(u)
-                .map_err(|_e| Error::BadParam("remote url is badly formed".to_string()))?;
-            self.remote_manifest = RemoteManifest::Remote(parsed.to_string());
-        } else {
-            self.remote_manifest = RemoteManifest::SideCar;
+    pub fn set_remote_manifest(&mut self, remote_manifest: RemoteManifest) -> Result<()> {
+        match remote_manifest {
+            RemoteManifest::NoRemote | RemoteManifest::SideCar => {
+                self.remote_manifest = remote_manifest
+            }
+            RemoteManifest::Remote(r) => {
+                url::Url::parse(&r)
+                    .map_err(|_e| Error::BadParam("remote url is badly formed".to_string()))?;
+                self.remote_manifest = RemoteManifest::Remote(r)
+            }
         }
-
         Ok(())
     }
 
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -11,10 +11,11 @@
 // specific language governing permissions and limitations under
 // each license.
 
+use std::{collections::HashMap, fmt, path::Path};
+
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Map, Value};
-use std::{collections::HashMap, fmt, path::Path};
 use uuid::Uuid;
 
 use crate::{
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -238,8 +238,8 @@ pub enum AssertionStoreJsonFormat {
 #[derive(Clone, Debug, PartialEq)]
 pub(crate) enum RemoteManifest {
     NoRemote,       // No external manifest
-    SideCar,        // manifest will be saved as side car file. Output asset is untouched.
-    Remote(String), // manifest will be saved as side car file. Output asset will contain remote reference.
+    SideCar,        // Manifest will be saved as a side car file, output asset is untouched.
+    Remote(String), // Manifest will be saved as a side car file, output asset will contain remote reference
 }
 
 impl Default for RemoteManifest {
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -12,7 +12,6 @@
 // each license.
 
 use chrono::{DateTime, Utc};
-use core::convert::AsRef;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Map, Value};
 use std::{collections::HashMap, fmt, path::Path};
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -265,7 +265,6 @@ impl Claim {
     /// Create a new claim.
     /// vendor: name used to label the claim (unique instance number is automatically calculated)
     /// claim_generator: User agent see c2pa spec for format
-    // #[cfg(not(target_arch = "wasm32"))]
     pub fn new(claim_generator: &str, vendor: Option<&str>) -> Self {
         let urn = Uuid::new_v4();
         let l = match vendor {
@@ -310,7 +309,6 @@ impl Claim {
     /// Create a new claim with a user supplied GUID.
     /// user_guid: is user supplied guid conforming the C2PA spec for manifest names
     /// claim_generator: User agent see c2pa spec for format
-    // #[cfg(not(target_arch = "wasm32"))]
     pub fn new_with_user_guid(claim_generator: &str, user_guid: &str) -> Self {
         Claim {
             remote_manifest: RemoteManifest::NoRemote,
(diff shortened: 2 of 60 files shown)


sdk/Cargo.toml
```diff
@@ -25,6 +25,7 @@ file_io = ["openssl"]
 serialize_thumbnails = []
 xmp_write = ["xmp_toolkit"]
 no_interleaved_io = ["file_io"]
+external_manifest = ["xmp_write", "file_io"]
```
scouten-adobe, line 28: Please also add a description of this new feature in README.

sdk/src/claim.rs
```diff
@@ -288,6 +307,38 @@ impl Claim {
         }
     }
 
+    /// Create a new claim with a user supplied GUID.
+    /// user_guid: is user supplied guid conforming the C2PA spec for manifest names
+    /// claim_generator: User agent see c2pa spec for format
+    // #[cfg(not(target_arch = "wasm32"))]
```
scouten-adobe, line 313: Why is this commented out? Can we remove before merging the PR?

sdk/src/store.rs
```diff
@@ -186,6 +186,20 @@ impl Store {
         Ok(claim_label)
     }
 
+    /// Add a new Claim to this Store. [`remote_manifest_path`] will be embedded in the
```
scouten-adobe, line 189: May not matter since `Store` isn't part of the public API, but in doc comments the `[]` wrapper should only be used when referring to another item that has its own doc entry (i.e. another public struct or enum, but not an argument to this function). The backtick wrappers are sufficient here.

sdk/src/store.rs
```diff
@@ -1386,26 +1401,76 @@ impl Store {
...
+
+        let jumbf_bytes = self.start_save(asset_path, &temp_file, signer.reserve_size())?;
 
         let pc = self.provenance_claim().ok_or(Error::ClaimEncoding)?;
         let sig = self.sign_claim(pc, signer, signer.reserve_size())?;
         let sig_placeholder = self.sign_claim_placeholder(pc, signer.reserve_size());
 
-        match self.finish_save(jumbf_bytes, output_path, sig, &sig_placeholder) {
-            Ok(v) => {
+        // get correct output path for remote manifest
```
scouten-adobe, line 1456: I ran out of time, so I'll stop reviewing here and publish as-is. I'll pick up from here later.

scouten-adobe:
Adding @dyro as reviewer as he is likely to work with this as well. I'll publish a work-in-progress review now and continue from my marker when I have more time later today.

scouten-adobe (changes requested):

sdk/src/store.rs
```diff
@@ -186,6 +186,20 @@ impl Store {
         Ok(claim_label)
     }
 
+    /// Add a new Claim to this Store. [`remote_manifest_path`] will be embedded in the
+    /// final asset once saved.  If [`remote_manifest_path`] is None an external side car file
+    /// will be generated instead. The function will return the label of the claim.
+    pub fn commit_claim_external(
```
gpeacock, line 192: Why do we need this vs just calling claim.set_remote_manifest prior to calling commit_claim?

sdk/Cargo.toml
mauricefisher64: I will remove for now since I am not gating anything by this yet.  When reading is enabled it will be used.

sdk/src/claim.rs
mauricefisher64: Does not seem to apply so I will remove.

sdk/src/store.rs
mauricefisher64: 👍

sdk/src/store.rs
```diff
@@ -1386,26 +1401,76 @@ impl Store {
...
         &mut self,
         asset_path: &Path,
         signer: &dyn Signer,
-        output_path: &Path,
-    ) -> Result<()> {
-        let jumbf_bytes = self.start_save(asset_path, output_path, signer.reserve_size())?;
+        dest_path: &Path,
+    ) -> Result<Vec<u8>> {
+        // set up temp dir, contents auto deleted
+        let td = tempfile::TempDir::new()?;
```
gpeacock, line 1442: FYI, I remember someone reporting that sometimes TempDir doesn't get deleted so we should keep an eye on that.

gpeacock: If we had some way of knowing when we can do manifest updates in memory vs in the file, we could avoid the need to use temporary storage here. Something to consider for the future.

sdk/src/store.rs
```diff
@@ -1468,39 +1575,52 @@ impl Store {
...
+                    embedded_xmp::add_manifest_uri_to_file(dest_path, &provenance)?;
+                } else {
+                    return Err(Error::XmpWriteError);
+                }
+                dest_path.to_path_buf()
+            }
+            crate::claim::RemoteManifest::SideCar => dest_path.with_extension(MANIFEST_STORE),
+            crate::claim::RemoteManifest::Remote(url) => {
+                let d = dest_path.with_extension(MANIFEST_STORE);
+                embedded_xmp::add_manifest_uri_to_file(dest_path, &url)?;
```
gpeacock, line 1607: I think you need the feature= "xmp_write" here with a failure if it is not enabled.

sdk/src/utils/hash_utils.rs
```diff
@@ -29,7 +29,7 @@ use sha2::{Digest, Sha256, Sha384, Sha512};
 
 use crate::{Error, Result};
 
-const MAX_HASH_BUF: usize = 1024 * 1024 * 1024; // cap memory usage to 1GB
+const MAX_HASH_BUF: usize = 256 * 1024 * 1024; // cap memory usage to 256GB
```
gpeacock, line 32: I think you mean 256MB not 256GB

sdk/src/claim.rs
```diff
@@ -288,6 +306,37 @@ impl Claim {
...
     }
 
+    /// Create a new claim with a user supplied GUID.
+    /// user_guid: is user supplied guid conforming the C2PA spec for manifest names
+    /// claim_generator: User agent see c2pa spec for format
+    pub fn new_with_user_guid(claim_generator: &str, user_guid: &str) -> Self {
+        Claim {
+            remote_manifest: RemoteManifest::NoRemote,
+            box_prefix: "self#jumbf".to_string(),
+            root: jumbf::labels::MANIFEST_STORE.to_string(),
```
gpeacock, line 316: I think we should use a different label, like MANIFEST_STORE_EXTENSION to refer to the extension, as this one is used for the internal c2pa label use case.

gpeacock (changes requested):

sdk/src/store.rs
scouten-adobe: @gpeacock the primary concern that I'm aware of is in the event of a panic (which is common in test code; hopefully less so otherwise), stack unwinding is not 100% predictable, so a temp-dir might not get cleaned up.

That said, I haven't seen that report, so maybe there's some other failure mode I'm not familiar with.

mauricefisher64: Read the temp dir docs.  They do all they can to address this.  The temp dir changes are in response to a request from Gavin to not corrupt the destination file.  To do that we have to write somewhere else, i.e. a temp location.

sdk/src/store.rs
mauricefisher64: Interesting add_manifest_uri_to_file is not behind that feature flag

sdk/src/utils/hash_utils.rs
mauricefisher64: Yup

sdk/src/claim.rs
mauricefisher64: I had that :) it got removed somewhere along the way.

mauricefisher64: Someone tell me where a label like that should live now.

sdk/src/store.rs
mauricefisher64: It is the Rust way.  Just setting a field param does not really indicate the side affects.  The different commit makes it more clear.   I am not attached this either way.

scouten-adobe (approved):

sdk/src/claim.rs
```diff
@@ -381,6 +430,23 @@ impl Claim {
         self.update_manifest
     }
 
+    pub(crate) fn set_remote_manifest(&mut self, remote_url: Option<&str>) -> Result<()> {
```
gpeacock, line 433: This would make more sense taking a RemoteManifest parameter without the Option, or be split into two methods, set_sidecar_manifest, and set_remote_manifest(url).
RemoteManifest could have a RemoteManifest::from_url() method.

mauricefisher64: Split maybe, take the argument not so much.  There is another state that I don't want passed in since you can get into unintended situations.

sdk/src/store.rs
gpeacock: The Option parameter None isn't very Rusty either. It doesn't describe adding a sidecar. It would make more sense taking a RemoteManifest parameter without the option. I'd argue that there a lot of things in a claim that can change what the commit operation does.

sdk/src/store.rs
gpeacock: I don't know that there is anything we need to do at this point, I was just noting that I had this code in claim_tool and the issue was brought up by AdobeStock. It may have been under test conditions.

sdk/src/claim.rs
gpeacock: I don't think we have a place for it. I'd just define it at the top of the file for now. Eventually we should add a utils.format file that defines the things that deal with file formats and extensions. That tends to be scattered through the code now, and is getting more complicated as we add types.

sdk/src/claim.rs
gpeacock: So I can give a remote url OR generate a sidecar, but if I don't generate a sidecar, how do I get the manifest data to post to service referenced by that Url? The claim:from_jumbf() method requires a signer which seems iffy it is a remote signer. Would it always work after committing the claim, or are there cases where we need to write to an actual file?

mauricefisher64: Either remote option generates a sidecar.  The only difference is the embedded case has a URL embedded in the output asset.

mauricefisher64: See the unit tests at the bottom of store.rs

sdk/src/store.rs
gpeacock: The entire embeded_xmp module is behind the "xmp-write" flag. I've verified that this won't build with if that flag is disabled.
```
cd sdk && cargo test --features="file_io" --no-default-features 
   Compiling c2pa v0.11.3 (/Users/gpeacock/dev/c2pa-rs/sdk)
error[E0433]: failed to resolve: use of undeclared crate or module `embedded_xmp`
    --> sdk/src/store.rs:1614:17
     |
1614 |                 embedded_xmp::add_manifest_uri_to_file(dest_path, &url)?;
     |                 ^^^^^^^^^^^^ use of undeclared crate or module `embedded_xmp`

```

sdk/src/store.rs
```diff
@@ -2712,4 +2818,158 @@ pub mod tests {
...
+        // compare returned to external
+        assert_eq!(saved_manifest, loaded_manifest);
+
+        // load the jumbf back into a store
+        let mut validation_log = OneShotStatusTracker::default();
+        let restored_store = Store::from_jumbf(&loaded_manifest, &mut validation_log).unwrap();
+        let pc = restored_store.provenance_claim().unwrap();
+
+        let mut asset_reader = std::fs::File::open(op.clone()).unwrap();
+        let handler = get_cailoader_handler("png").unwrap();
```
gpeacock, line 2918: You can use:  
let ext_ref = XmpInfo::from_source(asset_reader, "png").provenance.unwrap();

sdk/src/store.rs
```diff
@@ -2712,4 +2818,158 @@ pub mod tests {
...
+
+        // start with base url
+        let fp = format!("file:/{}", sidecar.to_str().unwrap());
+        let url = url::Url::parse(&fp).unwrap();
+
+        let url_string: String = url.into();
+
+        // set claim for side car with remote manifest embedding generation
+        claim
+            .set_remote_manifest(RemoteManifest::Remote(url_string.clone()))
```
gpeacock, line 2897: maybe something like this?

fn set_remote_url<S: Into<String>(url_string: S)

 let url = url::Url::parse(&fp).unwrap();
claim.set_remote_url(url);

sdk/src/claim.rs
```diff
@@ -288,6 +306,37 @@ impl Claim {
         }
     }
 
+    /// Create a new claim with a user supplied GUID.
+    /// user_guid: is user supplied guid conforming the C2PA spec for manifest names
+    /// claim_generator: User agent see c2pa spec for format
+    pub fn new_with_user_guid(claim_generator: &str, user_guid: &str) -> Self {
```
gpeacock, line 312: These parameters could be <S: Into<String>> and user_guid.into(). so we don't force cloning.

sdk/src/store.rs
```diff
@@ -2712,4 +2818,158 @@ pub mod tests {
...
+
+        let mut asset_reader = std::fs::File::open(op.clone()).unwrap();
+        let handler = get_cailoader_handler("png").unwrap();
+        let xmp = handler.read_xmp(&mut asset_reader).unwrap();
+        let ext_ref = extract_provenance(&xmp).unwrap();
+
+        assert_eq!(ext_ref, url_string);
+
+        // make sure this manifest goes with this asset
+        for dh_assertion in pc.data_hash_assertions() {
```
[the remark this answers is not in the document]
mauricefisher64: Let me think about how to do that.

sdk/src/store.rs
mauricefisher64: Done

sdk/src/store.rs
```diff
@@ -1386,26 +1388,76 @@ impl Store {
...
+        match remote_type {
+            crate::claim::RemoteManifest::NoRemote => Store::move_or_copy(source, dest)?,
+            crate::claim::RemoteManifest::SideCar | crate::claim::RemoteManifest::Remote(_) => {
+                // make correct path names
+                let source_asset = source;
+                let source_cai = source_asset.with_extension(MANIFEST_STORE_EXT);
+                let dest_cai = dest.with_extension(MANIFEST_STORE_EXT);
+
+                Store::move_or_copy(&source_cai, &dest_cai)?; // copy manifest
+                Store::move_or_copy(source_asset, dest)?; // copy asset
```
gpeacock, line 1414: potential case where first copy works and second fails, would leave only half of the data moved, should we delete it in that case? Don't need to do anything now, just something to think about.

gpeacock (approved):

sdk/src/store.rs
mauricefisher64: 👍  That is always an issue with move APIs.  We don't have transactions.

PR #106, opened 2022-08-11
Updates Manifest API to support remote manifests

gpeacock:
Updates Manifest api to support remote and external manifests
Manifest embed functions now return Ingredient instead of Store
Added Manifest label and remote_url get/set methods.
Additional unit tests for user supplied manifest label and remote manifest creation

## Checklist


sdk/src/manifest.rs
```diff
@@ -931,7 +971,11 @@ pub(crate) mod tests {
 
         let signer = temp_signer();
 
-        let store1 = manifest.embed(&output, &output, &signer).expect("embed");
+        let c2pa_data = manifest.embed(&output, &output, &signer).expect("embed");
+        let mut validation_log = DetailedStatusTracker::new();
+
+        let store1 = Store::load_from_memory("c2pa", &c2pa_data, true, &mut validation_log)
```
mauricefisher64, line 977: How does this work.  This function expect the bytes of a file.  You could set veify to false then it would not need to check the source asset bytes.  You could use this function instead which does not need a file since it does not run the verify step: load_cai_from_memory.

gpeacock: I think it works because it is a c2pa file which, by definition, has no asset to verify. But I can set verify to false or switch to the other method is that is more appropriate in this case.

PR #108, opened 2022-08-11
Support validating remote and external manifest stores

mauricefisher64:
## Changes in this pull request
The PR adds the ability to validate external .c2pa files and fetching and validating remote manifest stores.  Includes
small amount of code cleanup.

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #109, opened 2022-08-11
(IGNORE) Fix new Clippy errors from Rust 1.63

scouten-adobe:


scouten-adobe:
Included in #108. Closing.

PR #114, opened 2022-08-19
Make sure there is  a single manifest store in the asset

mauricefisher64:
Make sure there is only one manifest store in the asset when we read it.  

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/jpeg_io.rs
+++ b/sdk/src/asset_handlers/jpeg_io.rs
@@ -147,6 +147,8 @@ impl CAILoader for JpegIO {
     fn read_cai(&self, asset_reader: &mut dyn CAIRead) -> Result<Vec<u8>> {
         let mut buffer: Vec<u8> = Vec::new();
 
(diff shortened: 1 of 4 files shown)


sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -184,14 +186,20 @@ impl CAILoader for JpegIO {
                                 buffer.append(&mut raw_vec.as_mut_slice()[16..].to_vec());
 
                                 cai_seg_cnt += 1;
-                            } else {
+                            } else if raw_vec.len() > 28 {
```
gpeacock, line 189: These magic numbers are kind of scary, maybe add some comments about where these numbers come from

gpeacock (approved):

sdk/src/asset_handlers/jpeg_io.rs
mauricefisher64: There is no magic number.  The data starts after the header.

PR #117, opened 2022-08-25
Fast XMP extraction from PNG

mauricefisher64:
Don't use PNG lib since it decodes the image.  Use our chunk parsing to handle the iTXt chunk.

## Checklist


sdk/src/asset_handlers/png_io.rs
```diff
@@ -167,6 +190,92 @@ impl CAILoader for PngIO {
...
+                        pcp.length as usize
+                            - (key.len() + _langtag.len() + _transkey.len() + 5)
+                                as usize
+                    ]; // data len - size of key - size of land - size of transkey - 3 "0" string terminators - compressed u8 - compression method u8
+                    if asset_reader.read_exact(&mut data).is_err() {
+                        return false;
+                    }
+
+                    // convert to string, decompress if needed
+                    let val = if compressed {
```
gpeacock, line 252: I read "Standard XMP is written as uncompressed iTXt"  here: https://exiftool.org/TagNames/PNG.html#TextualData  So maybe we will never encounter this case, but maybe we will. It also entirely likely that we will at some point generalize this for reading exif and other things.

gpeacock (approved):

sdk/src/asset_handlers/png_io.rs
mauricefisher64: I disabled it.  We can resurrect if needed.

PR #119, opened 2022-08-26
(IGNORE) Fail a crate release if an API breaking change is made …

scouten-adobe:
… without an appropriate version bump

This new step uses the `cargo-semver-checks` crate to compare the proposed new API surface against the previous release and will fail the publish step if the appropriate (MINOR) or (MAJOR) tag was omitted from any PR.

(You can "fix" the problem by submitting an empty PR with the appropriate tag and then re-attempting the publish.)


scouten-adobe:
Actually, NVM. I'm going to address this a different way (though PR validation).

PR #121, opened 2022-08-26
Convert status_log error val to a string so that we can return full errors

gpeacock:
Allows us to return complex errors while still adding them to logs. You can't copy an error, but you can display it.

## Checklist

### DIFF
--- a/sdk/src/validation_status.rs
+++ b/sdk/src/validation_status.rs
@@ -92,6 +92,18 @@ impl ValidationStatus {
     }
 
     // Maps errors into validation_status codes.
+    fn code_from_error_str(error: &str) -> &str {
+        match error {
+            e if e.starts_with("ClaimMissing") => CLAIM_MISSING,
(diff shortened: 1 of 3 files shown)


sdk/src/validation_status.rs
```diff
@@ -91,6 +91,18 @@ impl ValidationStatus {
         is_success(&self.code)
     }
 
+    // Maps errors into validation_status codes.
```
mauricefisher64, line 94: Will we make this list complete?  Not sure it is possible today.

mauricefisher64 (approved):
👍

sdk/src/validation_status.rs
gpeacock: Ideally we get rid of the mapping and assign validation status codes when we generate the errors. But I suspect a few cases will always fall through. FYI, there's a new 1.1 approved general.error code we should add in those cases.

PR #123, opened 2022-08-29
(MINOR) Add support for manifest removal

mauricefisher64:
New internal API to allow for the removal of a manifest.  This will be needed for future use cases 

## Checklist


sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1066,6 +1066,135 @@ impl AssetIO for BmffIO {
...
+        };
+
+        // write content before ContentProvenanceBox
+        input.seek(SeekFrom::Start(0))?;
+        let mut b = vec![0u8; start];
+        input.read_exact(&mut b)?;
+        temp_file.write_all(&b)?;
+
+        // calc offset adjustments
+        // value will be negative sibce the box is truncated
```
scouten-adobe, line 1133: Typo: sibce -> since

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1066,6 +1066,135 @@ impl AssetIO for BmffIO {
...
+                .map_err(|_err| Error::BadParam("value out of range".to_string()))?;
+
+            (start, end)
+        } else {
+            return Err(Error::BadParam("value out of range".to_string()));
+        };
+
+        // write content before ContentProvenanceBox
+        input.seek(SeekFrom::Start(0))?;
+        let mut b = vec![0u8; start];
```
scouten-adobe, line 1128: This will cause problems for especially large files, right? (I'm thinking about video.)

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1257,4 +1386,27 @@ pub mod tests {
...
+        let mut success = false;
+        if let Ok(temp_dir) = tempdir() {
+            let output = temp_dir_path(&temp_dir, "mp4_test.mp4");
+
+            if let Ok(_size) = std::fs::copy(&source, &output) {
+                let bmff = BmffIO::new("mp4");
+
+                if let Ok(()) = bmff.remove_cai_store(&output) {
+                    match bmff.read_cai_store(&output) {
+                        Ok(_) => success = false,
```
scouten-adobe, line 1403: Assigning to `success` and matching makes it a bit hard to follow the overall control flow.

Perhaps you could use `if let ... else` and have more specific assertions at the point where each failure case is detected?

sdk/src/asset_handlers/c2pa_io.rs
```diff
@@ -60,6 +60,10 @@ impl AssetIO for C2paIO {
 
         Ok(vec![hop])
     }
+
+    fn remove_cai_store(&self, _asset_path: &Path) -> Result<()> {
+        Ok(())
```
scouten-adobe, line 65: Seems like this should be some kind of error signaling a non-sensical request instead of `Ok(())`.

sdk/src/asset_handlers/png_io.rs
```diff
@@ -384,6 +384,44 @@ impl AssetIO for PngIO {
...
+                chunk crc (4 bytes in crc in format defined in PNG spec)
+        */
+
+        // erase existing
+        let empty_buf = Vec::new();
+        let mut iter = ps.into_iter();
+        if let Some(existing_cai) = iter.find(|pcp| pcp.name == CAI_CHUNK) {
+            // replace existing CAI
+            let start = usize::value_from(existing_cai.start)
+                .map_err(|_err| Error::BadParam("value out of range".to_string()))?; // get beginning of chunk which starts 4 bytes before label
```
scouten-adobe, line 411: I'm not sure `BadParam` (typically error on the part of the caller) makes sense here. This is really saying the existing file is has faulty data in it, right?

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: No, the C2PA box is right after the ftyp box.  The ftyp is box first and is usually about 25-30 bytes so we only have a few bytes to copy before we hit the C2PA box.  After that notice we do chunk the large portion of the file.

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: Fixed

sdk/src/asset_handlers/png_io.rs
mauricefisher64: Or the start cannot fit into a usize.

sdk/src/asset_handlers/c2pa_io.rs
mauricefisher64: Yeah, I am not sure about this one.  No one will ever call this so success or failure does not matter at this time.

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1066,6 +1066,135 @@ impl AssetIO for BmffIO {
         let vec: Vec<HashObjectPositions> = Vec::new();
         Ok(vec)
     }
+
+    fn remove_cai_store(&self, asset_path: &Path) -> Result<()> {
```
gpeacock, line 1070: This should really be a READ+SEEK trait but I realize there's temp_files inside so there's no point. I guess eventually we need trait abstraction for a temp file? Not to worry about now, but I want us to get there evenutally.

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1066,6 +1066,135 @@ impl AssetIO for BmffIO {
...
+            };
+
+        let mut temp_file = Builder::new()
+            .prefix("c2pa_temp")
+            .rand_bytes(5)
+            .tempfile()?;
+
+        let (start, end) = if let Some(c2pa_length) = c2pa_length {
+            let start = usize::value_from(c2pa_start)
+                .map_err(|_err| Error::BadParam("value out of range".to_string()))?; // get beginning of chunk which starts 4 bytes before label
```
gpeacock, line 1116: This is a pretty obscure error if this ever comes up. It isn't really a bad parameter, more like invalid bmff data?

sdk/src/asset_handlers/png_io.rs
gpeacock: Hey, same comment I made above. I think we should identify these as invalid bmff content.

gpeacock (changes requested):
The exp-test1.png fixture is way larger than all other image we have in the sdk combined. We should try to shrink it down (that 10mb crate limit we have to deal with).

sdk/src/asset_handlers/c2pa_io.rs
mauricefisher64: I don't want to return an error for a no op

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1257,4 +1386,25 @@ pub mod tests {
         }
         assert!(success)
     }
+
+    #[test]
+    fn test_remove_c2pa() {
+        let source = fixture_path("video1.mp4");
+
+        if let Ok(temp_dir) = tempdir() {
```
scouten-adobe, line 1394: Since you're in test code, why not use `.unwrap()` rather than the nested `if let`s?

sdk/src/asset_handlers/c2pa_io.rs
scouten-adobe: Agree with leaving this as `Ok(())`, but can we at least log that this was a non-sensical request?

sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -431,4 +450,25 @@ pub mod tests {
         let result = extract_xmp(&seg);
         assert_eq!(result, None);
     }
+
+    #[test]
+    fn test_remove_c2pa() {
+        let source = crate::utils::test::fixture_path("CA.jpg");
+
+        if let Ok(temp_dir) = tempfile::tempdir() {
```
scouten-adobe, line 458: Same comment about `if let` vs `.unwrap()`.

sdk/src/asset_handlers/png_io.rs
```diff
@@ -433,4 +469,25 @@ pub mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_remove_c2pa() {
+        let source = crate::utils::test::fixture_path("exp-test1.png");
+
+        if let Ok(temp_dir) = tempfile::tempdir() {
```
scouten-adobe, line 477: Here too …

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: Some intermediate version of Clippy made me do it.  If necessary I can see if it is still an issue.

gpeacock (approved):

PR #124, opened 2022-08-31
(IGNORE) Enforce semantic versioning requirements during PR validation and publish workflows

scouten-adobe:

This adds steps to the CI and publish workflows to ensure that the correct semantic version bumps occur when breaking API changes are proposed.

This will cause a PR to fail validation if the `(MINOR)` tag is omitted when it should be present.

Similarly, we re-check during the release process and stop the release if a minor version change should have been triggered.

All of these will need to change to `(MAJOR)` once we do 1.0 release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,7 +254,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
         with:
-          toolchain: nightly
+          toolchain: stable
           override: true
 
       - name: Get latest existing tag
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,3 +240,59 @@ jobs:
         with:
           version: latest
           args: --all-targets --all-features
+
+  version_bump:
+    name: Ensure (MINOR) tag is used when making an API breaking change
+    # Change all of these steps to (MAJOR) after 1.0 release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
+      
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
+        with:
+          toolchain: nightly
+          override: true
+
+      - name: Get latest existing tag
+        uses: actions-ecosystem/action-get-latest-tag@v1
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
+        id: get-latest-tag
+        with:
+          semver_only: true
+
+      - name: Set new version
+        uses: paulhatch/semantic-version@v4.0.2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -240,3 +240,59 @@ jobs:
...
+          tag_prefix: "v"
+          format: "${major}.${minor}.${patch}"
+          major_pattern: "(MAJORXXXX)"
+          minor_pattern: "(MINORXXXX)"
+
+      - name: "Bump crate version (NOTE: Not pushed back to repo!)"
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
+        run: |
+          sed -i "s/^version = \"[^\"]*\"$/version = \"$VERSION\"/;" sdk/Cargo.toml
+          git config user.email "nobody@example.com"
```
gpeacock, line 281: does this email get used in any way?

.github/workflows/ci.yml
```diff
@@ -240,3 +240,59 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
+      
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
+        with:
+          toolchain: nightly
```
gpeacock, line 257: I know we are already dependent on nightly, but it would be good to avoid that complication if we can.
Will this nightly requirement change at some point?

gpeacock (approved):

.github/workflows/ci.yml
jackfarzan: seconding this question

scouten-adobe: Effectively, no, but it's required by git. We're temporarily committing the change to Cargo.toml to bump the version, but not pushing that commit back to the repo (which is why I say "effectively" not used).

Git won't allow you to make a commit, even locally, without a name, e-mail address, and commit message. The e-mail address can be faked, but it has to at least look like an e-mail address.

jackfarzan (approved):

.github/workflows/ci.yml
```diff
@@ -240,3 +240,59 @@ jobs:
...
           args: --all-targets --all-features
+
+  version_bump:
+    name: Ensure (MINOR) tag is used when making an API breaking change
+    # Change all of these steps to (MAJOR) after 1.0 release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        if: ${{ !contains(github.event.pull_request.title, '(MINOR)') }}
```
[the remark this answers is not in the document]
scouten-adobe: Well, as a matter of _style_ I prefer them as a prefix, but … the tool we're using to manage the version bumps searches the entire title. And that was how I got in trouble a month or so ago.

.github/workflows/ci.yml
scouten-adobe: Thanks for the catch. Doesn't look like nightly is needed for this job, so I switched it to stable.

PR #125, opened 2022-09-06
Add sign feature for signing manifests without file i/o 

gpeacock:
Add sign feature for signing manifests without file i/o 
Only supports jpeg in so far
Manifest.embed_stream method added

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/jpeg_io.rs
+++ b/sdk/src/asset_handlers/jpeg_io.rs
@@ -13,7 +13,7 @@
 
 use std::{
     fs::{read, File},
-    io::Cursor,
+    io::{Cursor, Write},
     path::*,
 };
 
@@ -24,7 +24,7 @@ use img_parts::{
 };
 
 use crate::{
-    asset_io::{AssetIO, CAILoader, CAIRead, HashBlockObjectType, HashObjectPositions},
+    asset_io::{AssetIO, CAILoader, CAIRead, CAIWriter, HashBlockObjectType, HashObjectPositions},
     error::{wrap_io_err, Error, Result},
 };
 
@@ -141,6 +141,68 @@ fn delete_cai_segments(jpeg: &mut img_parts::jpeg::Jpeg) -> Result<()> {
     }
     Ok(())
 }
+
+//fn write_cai<W: Write>(reader: &mut dyn CAIRead , writer: W, store_bytes: &[u8]) -> Result<Vec<u8>> {
+fn write_cai<W: Write>(buf: Vec<u8>, writer: W, store_bytes: &[u8]) -> Result<()> {
+    //     let mut buf = Vec::new();
+    // // read the whole file
+    // reader.read_to_end(&mut buf).map_err(wrap_io_err);
+    let mut jpeg = Jpeg::from_bytes(buf.into()).map_err(|_err| Error::EmbeddingError)?;
+
+    // remove existing CAI segments
+    delete_cai_segments(&mut jpeg)?;
+
+    let jumbf_len = store_bytes.len();
+    let num_segments = (jumbf_len / MAX_JPEG_MARKER_SIZE) + 1;
+    let mut seg_chucks = store_bytes.chunks(MAX_JPEG_MARKER_SIZE);
+
+    for seg in 1..num_segments + 1 {
+        /*
+            If the size of the box payload is less than 2^32-8 bytes,
+            then all fields except the XLBox field, that is: Le, CI, En, Z, LBox and TBox,
+            shall be present in all JPEG XT marker segment representing this box,
+            regardless of whether the marker segments starts this box,
+            or continues a box started by a former JPEG XT Marker segment.
+        */
+        // we need to prefix the JUMBF with the JPEG XT markers (ISO 19566-5)
+        // CI: JPEG extensions marker - JP
+        // En: Box Instance Number  - 0x0001
+        //          (NOTE: can be any unique ID, so we pick one that shouldn't conflict)
+        // Z: Packet sequence number - 0x00000001...
+        let ci = vec![0x4A, 0x50];
+        let en = vec![0x02, 0x11];
+        let z = seg.to_be_bytes();
+
+        let mut seg_data = Vec::new();
+        seg_data.extend(ci);
+        seg_data.extend(en);
+        seg_data.extend(&z[4..]);
+        if seg > 1 {
+            // the LBox and TBox are already in the JUMBF
+            // but we need to duplicate them in all other segments
+            let lbox_tbox = &store_bytes[..8];
+            seg_data.extend(lbox_tbox);
+        }
+        if seg_chucks.len() > 0 {
+            // make sure we have some...
+            if let Some(next_seg) = seg_chucks.next() {
+                seg_data.extend(next_seg);
+            }
+        } else {
+            seg_data.extend(store_bytes);
+        }
+        let seg_bytes = Bytes::from(seg_data);
+        let app11_segment = JpegSegment::new_with_contents(markers::APP11, seg_bytes);
+        jpeg.segments_mut().insert(seg, app11_segment); // we put this in the beginning...
+    }
+
+    jpeg.encoder()
+        .write_to(writer)
+        //.map_err(wrap_io_err)
+        .map_err(|_err| Error::BadParam("JPEG write error".to_owned()))?;
+
+    Ok(())
+}
 pub struct JpegIO {}
 
 impl CAILoader for JpegIO {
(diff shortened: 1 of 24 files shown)


sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -80,7 +87,8 @@ fn add_required_segs(asset_path: &std::path::Path) -> Result<()> {
             let mut no_bytes: Vec<u8> = vec![0; 50]; // enough bytes to be valid
             no_bytes.splice(16..20, C2PA_MARKER); // cai UUID signature
             let aio = JpegIO {};
-            aio.save_cai_store(asset_path, &no_bytes)?;
+            aio.write_cai(stream, &no_bytes)?;
+            stream.seek(SeekFrom::Start(0))?;
```
scouten-adobe, line 91: Why is this last `.seek` necessary? Shouldn't the subsequent caller be responsible for positioning the streem where it needs it? Or is there an invariant that the stream is always at start-of-file?

sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -80,7 +87,8 @@ fn add_required_segs(asset_path: &std::path::Path) -> Result<()> {
...
         }
     } else {
         return Err(Error::BadParam(
```
scouten-adobe, line 94: Can this be `Error::UnsupportedType` as you changed it at line 408?

sdk/src/asset_io.rs
```diff
@@ -53,6 +58,16 @@ pub trait CAILoader {
     fn read_xmp(&self, asset_reader: &mut dyn CAIRead) -> Option<String>;
 }
 
+pub trait CAIWriter {
+    // Return entire CAI block as Vec<u8>
```
scouten-adobe, line 62: "Return entire block?" … but the return type is `Result<()>`

sdk/src/asset_io.rs
```diff
@@ -44,6 +44,11 @@ impl CAIRead for std::fs::File {}
 impl CAIRead for std::io::Cursor<&[u8]> {}
 impl CAIRead for std::io::Cursor<Vec<u8>> {}
 
+pub trait CAIReadWrite: CAIRead + Write {}
```
scouten-adobe, line 47: This could be in another PR later, but I'd like us to remove the `CAI` prefixes (and also not add a new `C2pa` prefix). Seems like Rust's module system should suffice where we are now using prefixes.

sdk/src/error.rs
```diff
@@ -241,7 +241,7 @@ pub enum Error {
     CborError(#[from] serde_cbor::Error),
 
     #[error(transparent)]
-    #[cfg(feature = "file_io")]
+    #[cfg(feature = "sign")]
     OpenSslError(#[from] openssl::error::ErrorStack),
```
scouten-adobe, line 245: Scoping this to `sign` feature seems odd to me. Do we not also use OpenSSL for verifying signatures?

sdk/src/manifest.rs
```diff
@@ -735,6 +735,26 @@ impl Manifest {
...
 
+    /// Embed a signed manifest into a stream using a supplied signer.
+    ///
+    #[cfg(feature = "sign")]
+    pub fn embed_stream(
+        &mut self,
+        format: &str,
+        stream: &mut dyn CAIReadWrite,
+        signer: &dyn Signer,
+    ) -> Result<Vec<u8>> {
```
scouten-adobe, line 746: What is the content of the `Vec<u8>` in the success case?

sdk/src/openssl/ec_signer.rs
```diff
@@ -45,14 +46,15 @@ pub struct EcSigner {
...
+    #[cfg(feature = "file_io")]
     fn from_files<P: AsRef<Path>>(
         signcert_path: P,
         pkey_path: P,
         alg: SigningAlg,
         tsa_url: Option<String>,
     ) -> Result<Self> {
-        let signcert = fs::read(signcert_path).map_err(wrap_io_err)?;
-        let pkey = fs::read(pkey_path).map_err(wrap_io_err)?;
+        let signcert = fs::read(signcert_path).map_err(Error::IoError)?;
```
scouten-adobe, line 56: Oh, nice simplification!

sdk/src/openssl/ed_signer.rs
```diff
@@ -35,14 +36,15 @@ pub struct EdSigner {
 }
 
 impl ConfigurableSigner for EdSigner {
+    #[cfg(feature = "file_io")]
```
scouten-adobe, line 39: Not crucial for this PR, but I wonder if we could move the `from_files` into a default implementation in `ConfigurableSigner`? Seems like a lot of boilerplate in these files.

sdk/src/validator.rs
```diff
@@ -51,8 +51,8 @@ impl CoseValidator for DummyValidator {
 // • RS512	RSASSA-PKCS1-v1_5 using SHA-512
 // • ED25519 Edwards Curve ED25519
 
-/// return validator for supported C2PA  algorthms
-#[cfg(feature = "file_io")]
+/// return validator for supported C2PA  algorithms
+#[cfg(feature = "sign")]
```
scouten-adobe, line 55: I'm not sure the feature name `sign` is appropriate given this usage.

scouten-adobe (changes requested):
Overall, love it!

In addition to the specific concerns listed below, I'm also concerned about the drop in coverage. Can we adjust so we don't leave so much code untested?

sdk/src/asset_handlers/jpeg_io.rs
gpeacock: It just seemed like a good precaution, but it isn't necessary

sdk/src/asset_io.rs
gpeacock: yeah that changed, thanks for catching it.

sdk/src/error.rs
gpeacock: Apparently we don't us OpensslError for verify, so I'd need to allow an unused reference if I don't scope it this way.

gpeacock: I'll see if feature = "openssl" will work.

sdk/src/manifest.rs
gpeacock: Need to document that, it is the embedded manifest. It could be useful for uploading to a cloud.

sdk/src/openssl/ed_signer.rs
gpeacock: good idea

sdk/src/asset_io.rs
gpeacock: Agreed, These should all be renamed, but none of them are exposed

sdk/src/error.rs
gpeacock: We were disabling openssl when not using file_io because we don't use openssl with wasm.

sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -151,6 +157,7 @@ impl CAILoader for JpegIO {
 
         // load the bytes
         let mut buf: Vec<u8> = Vec::new();
```
mauricefisher64, line 159: we could improve the performance here slightly by preallocating the buffer.

mauricefisher64 (approved):
👍 

I have already started work on a unified API that handles buffers, steams, files, etc.  I will extend it to this work when I get a chance.

sdk/src/manifest.rs
```diff
@@ -738,6 +738,23 @@ impl Manifest {
     /// Embed a signed manifest into a stream using a supplied signer.
     ///
     #[cfg(feature = "sign")]
+    pub fn embed_from_memory(
+        &mut self,
+        format: &str,
+        asset: &[u8],
+        signer: &dyn Signer,
+    ) -> Result<Vec<u8>> {
```
scouten-adobe, line 746: Still looking for documentation on what the `Vec<u8>` contains.

scouten-adobe (changes requested):
Aside from the doc request, looks great!

sdk/src/asset_handlers/jpeg_io.rs
gpeacock: done

sdk/src/openssl/ed_signer.rs
gpeacock: I made this change

sdk/src/manifest_store.rs
```diff
@@ -146,12 +146,12 @@ impl ManifestStore {
     /// Loads a ManifestStore from a file
     pub async fn from_bytes_async(
         format: &str,
-        image_bytes: Vec<u8>,
+        image_bytes: &[u8],
```
scouten-adobe, line 149: Changed title of PR to include (MINOR) since this changes an existing API. (Yes, I know, I asked for the change. :-) )

scouten-adobe (approved):

sdk/src/asset_handlers/jpeg_io.rs
gpeacock: Preallocate to what size? Do you mean by seeking to the end?

PR #126, opened 2022-09-06
(IGNORE) Disable cargo-semver-checks step

scouten-adobe:
Looks like there's a compatibility issue that needs resolving in cargo-semver-check crate.

Restore when https://github.com/obi1kenobi/cargo-semver-check/issues/122 is fixed.

PR #129, opened 2022-09-12
(IGNORE) Restore cargo-semver-checks step

scouten-adobe:
Reverts #126 now that new version of crate is available.


mauricefisher64 (approved):
👍

gpeacock (approved):
Ok as long as it works now :)

PR #130, opened 2022-09-12
Add ManifestStore::from_manifest_and_asset_bytes_async

emensch:
Adds ManifestStore::from_manifest_and_asset_bytes_async method
This is useful for validating a remote manifest (.c2pa) with the associated asset.

## Checklist


sdk/src/manifest_store.rs
```diff
@@ -169,6 +169,18 @@ impl ManifestStore {
                 .map(|_| Self::from_store(&store, &mut validation_log))
         })
     }
+
+    pub async fn from_manifest_and_asset_bytes_async(
```
scouten-adobe, line 173: This is a public API function, yes? If so, can we please add proper documentation for it?

scouten-adobe (changes requested):

sdk/src/manifest_store.rs
emensch: Yes - of course. Just wanted to get this out for preliminary review.

sdk/src/manifest_store.rs
```diff
@@ -155,6 +155,39 @@ impl ManifestStore {
...
             .map(|store| Self::from_store(&store, &mut validation_log))
     }
+
+    /// Asynchronously loads a manifest from a buffer holding a binary manifest (.c2pa) and validates against an asset buffer
+    ///
+    /// # Example: Creating a manifest store from a .c2pa manifest and validating it against an asset
+    /// ```
+    /// use c2pa::{Result, ManifestStore};
+    ///
+    /// fn main() -> Result<()> {
```
scouten-adobe, line 165: I'd encourage you to use `#` markers to hide the outer function wrapper and possibly the async block.

See https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example.

Not a blocker for PR IMHO.

scouten-adobe (approved):

sdk/src/manifest_store.rs
emensch: Cool - will update.

sdk/src/manifest_store.rs
```diff
@@ -335,7 +335,9 @@ mod tests {
         assert!(manifest.time().is_some());
     }
 
-    #[actix::test]
+    #[cfg_attr(not(target_arch = "wasm32"), actix::test)]
```
gpeacock, line 338: This is just so ugly, but it looks like what we have done in other places to make this work with wasm.

sdk/src/manifest_store.rs
```diff
@@ -297,8 +297,8 @@ mod tests {
         assert!(manifest.time().is_some());
     }
 
+    #[cfg_attr(not(target_arch = "wasm32"), actix::test)]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
-    #[ignore]
     #[allow(dead_code)]
     async fn manifest_report_image_wasm() {
```
gpeacock, line 303: If we are going to support running this test outside of wasm, we should change the name

sdk/src/manifest_store.rs
```diff
@@ -335,7 +335,9 @@ mod tests {
         assert!(manifest.time().is_some());
     }
 
-    #[actix::test]
+    #[cfg_attr(not(target_arch = "wasm32"), actix::test)]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    #[allow(dead_code)]
```
gpeacock, line 340: Do we need to allow dead code if we are supporting both alternatives?

gpeacock (changes requested):
if you want a test to run outside of wasm, it should not be named wasm

sdk/src/manifest_store.rs
emensch: Likely not. I'll try removing

sdk/src/manifest_store.rs
emensch: Agreed it isn't pretty. _Very_ open to alternatives, this is just what I saw used elsewhere

mauricefisher64: I was looking at a crate that would wrap complex macros like this to a single variable like #if_cfg(my_simple_var).  I will see if I can find it again.

dkozma: Yeah unfortunately the wasm_bindgen stuff is still pretty ugly, however we do use it in other places at the moment. IMO we should address fixing this in a future PR rather than switching everything over in this one.

sdk/src/manifest_store.rs
```diff
@@ -302,4 +333,19 @@ mod tests {
...
+        let asset_bytes = include_bytes!("../tests/fixtures/cloud.jpg");
+        let manifest_bytes = include_bytes!("../tests/fixtures/cloud_manifest.c2pa");
+
+        let manifest_store =
+            ManifestStore::from_manifest_and_asset_bytes_async(manifest_bytes, asset_bytes)
+                .await
+                .unwrap();
+        assert!(!manifest_store.manifests().is_empty());
+        assert!(manifest_store.validation_status().is_none());
+        println!("{}", manifest_store);
```
dkozma, line 349: Do we want to print the manifest store as part of this test? Not sure if it will cloud test output.

dkozma (approved):
Small comment but 👍🏼 overall.

gpeacock (approved):

PR #135, opened 2022-09-20
manifest_data was missing for remote manifests

gpeacock:
Inrgredient::from_file was not returning manifest_data for remote manifests.

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -1871,8 +1871,7 @@ impl Store {
             Ok(manifest_bytes) => Ok(manifest_bytes),
             Err(Error::JumbfNotFound) => {
                 if external_manifest.exists() {
-                    let external_manifest_bytes = std::fs::read(external_manifest)?;
-                    Ok(external_manifest_bytes)
+                    std::fs::read(external_manifest).map_err(Error::IoError)
                 } else {
                     // check for remote manifest
                     let mut asset_reader = std::fs::File::open(in_path)?;
@@ -1884,8 +1883,7 @@ impl Store {
                     .provenance
                     {
                         if cfg!(feature = "fetch_remote_manifests") {
-                            let remote_manifest_bytes = Store::fetch_remote_manifest(&ext_ref)?;
-                            Ok(remote_manifest_bytes)
+                            Store::fetch_remote_manifest(&ext_ref)
                         } else {
                             // return an error with the url that should be read
                             Err(Error::RemoteManifestUrl(ext_ref))
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -1857,26 +1857,22 @@ impl Store {
         }
     }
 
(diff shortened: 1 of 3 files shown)


sdk/src/store.rs
```diff
@@ -1889,7 +1885,7 @@ impl Store {
                     {
                         if cfg!(feature = "fetch_remote_manifests") {
                             let remote_manifest_bytes = Store::fetch_remote_manifest(&ext_ref)?;
-                            Store::from_jumbf(&remote_manifest_bytes, validation_log)
+                            Ok(remote_manifest_bytes)
```
mauricefisher64, line 1888: these two lines can just be:

Store::fetch_remote_manifest(&ext_ref)?;

sdk/src/store.rs
```diff
@@ -1857,26 +1857,22 @@ impl Store {
...
-            Ok(manifest_bytes) => {
-                // load and validate with CAI toolkit and dump if desired
-                Store::from_jumbf(&manifest_bytes, validation_log)
-            }
+            Ok(manifest_bytes) => Ok(manifest_bytes),
             Err(Error::JumbfNotFound) => {
                 if external_manifest.exists() {
                     let external_manifest_bytes = std::fs::read(external_manifest)?;
-                    Store::from_jumbf(&external_manifest_bytes, validation_log)
+                    Ok(external_manifest_bytes)
```
mauricefisher64, line 1875: The two lines can just be:

=> std::fs::read(external_manifest)?;

sdk/src/store.rs
```diff
@@ -1903,6 +1899,24 @@ impl Store {
...
+    fn load_cai_from_file(
+        in_path: &Path,
+        validation_log: &mut impl StatusTracker,
+    ) -> Result<Store> {
+        match Self::load_jumbf_from_path(in_path) {
+            Ok(manifest_bytes) => {
+                // load and validate with CAI toolkit
+                Store::from_jumbf(&manifest_bytes, validation_log)
+            }
+            Err(e) => Err(e),
```
mauricefisher64, line 1916: The error may not be meaningful if you don't pass a StatusTracker to load_jumbf_path.  You may just get LogStop.

gpeacock: This was taken from the previous code that called load_jumbf_from_File(), which also does not take a StatusTracker.
load_cai_from_file had these three error return states that do log log to a StatusTracker:
                            // return an error with the url that should be read
                            Err(Error::RemoteManifestUrl(ext_ref))
                        }
                    } else {
                        Err(Error::JumbfNotFound)
                    }
                }
            }
            Err(e) => Err(e),

gpeacock: So this should create the same result.

mauricefisher64: Maybe I am misreading this.

mauricefisher64 (approved):
👍

PR #136, opened 2022-09-22
(MINOR) Remove previously embedded manifests for remote manifests

gpeacock:
Assets with remote manifests should remove previously embedded manifests
This fixes a bug when writing remote manifests and not updating an embedded manifest. The embedded manifest will now be removed from the output asset.
## Checklist

### DIFF
--- a/sdk/src/manifest.rs
+++ b/sdk/src/manifest.rs
@@ -1158,4 +1158,30 @@ pub(crate) mod tests {
             TEST_SMALL_JPEG
         );
     }
+
+    #[cfg(all(feature = "file_io", feature = "xmp_write"))]
+    #[actix::test]
+    async fn test_embed_sidecar_with_parent_manifest() {
+        let temp_dir = tempdir().expect("temp dir");
+        let source = fixture_path("XCA.jpg");
+        let output = temp_dir.path().join("XCAplus.jpg");
+        let sidecar = output.with_extension("c2pa");
+        let fp = format!("file:/{}", sidecar.to_str().unwrap());
+        let url = url::Url::parse(&fp).unwrap();
+
+        let signer = temp_signer();
+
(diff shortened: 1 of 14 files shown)


sdk/src/store.rs
```diff
@@ -32,7 +32,7 @@ use crate::{
     cose_validator::verify_cose,
     jumbf_io::{
         get_file_extension, get_supported_file_extension, is_bmff_format, load_jumbf_from_file,
```
mauricefisher64, line 34: We should probably do one more thing in this function.  If we remove the manifest we should clear the XMP.  It will be cleared for the remote case but not the sidecar case.

mauricefisher64: I can add that in a different PR as to not hold this up.

mauricefisher64 (approved):
👍

sdk/src/manifest.rs
```diff
@@ -1158,4 +1158,30 @@ pub(crate) mod tests {
             TEST_SMALL_JPEG
         );
     }
+
+    #[cfg(all(feature = "file_io", feature = "xmp_write"))]
+    #[actix::test]
+    async fn test_embed_sidecar_with_parent_manifest() {
```
scouten-adobe, line 1164: Why is this an `async` test? I don't see any `await` or `async` inside the function.

scouten-adobe (changes requested):

sdk/src/store.rs
gpeacock: good point. We might need to check if there was something we needed to delete in the case where xmp_write is disabled, otherwise we could return an error when there's nothing wrong.

We also need to figure out how to remove c2pa jumbf from in memory buffers. But that can also wait for later.

sdk/src/manifest.rs
gpeacock: good point, I removed async from some of the other tests that didn't need it.

PR #141, opened 2022-09-30
(MINOR) User, UserCbor and Uuid assertions removed from SDK

gpeacock:
Remove User, UserCbor and Uuid assertions from the public SDK
Use add_labeled_assertion instead.
## Checklist


sdk/src/manifest.rs
```diff
@@ -699,26 +699,23 @@ impl Manifest {
     ///
     /// ```
     /// # use c2pa::Result;
-    /// use c2pa::{
-    ///     assertions::User,
-    ///     create_signer,
-    ///     Manifest,
-    ///     SigningAlg,
-    /// };
+    /// use c2pa::{create_signer, Manifest, SigningAlg,};
```
scouten-adobe, line 702: Extra comma here shouldn't be necesary.

scouten-adobe (changes requested):
Looks good. I'd like to see something in the README (breaking changes since 0.14.0 to call attention to how anybody using the now-private interface should update their code.

dkozma (approved):
LGTM after fixing @scouten-adobe 's comment

sdk/src/manifest.rs
gpeacock: removed

scouten-adobe (commented):
I think any developer relying on existing functionality could feel pretty frustrated upon taking this upgrade and not having an obvious guide to how to migrate.

I'll grant that an upgrading guide isn't particularly common, but I've seen some crates that do this (for example, https://github.com/rust-random/rand#versions) and I think it's a pattern worth adopting.

FWIW I do this in the XMP Toolkit crate: https://github.com/adobe/xmp-toolkit-rs#breaking-changes-in-0x-series

What you've added in the changelog says that there is something that is breaking, but isn't particularly helpful in guiding any developer affected by the change toward a resolution of their problem.

I won't give this a :+1: without a guide, but I'll remove my change request.

gpeacock:
Updated readme and pulled in the latest changes. The example using the User assertion has been replaced the an an example of using a custom struct and add_labeled_assertion

PR #147, opened 2022-10-25
Fix for XMP links being mistaken for remote URLs

mauricefisher64:
Add change to cose_validator so that certs without organization is not an error

Old CAI manifests may still have XMP jumbf references.  This PR check to make sure those references are valid URLs before attempting a remote manifest fetch.  Also fix bogus error when a certificate does not have an organization set.  That is not a C2PA requirement.  

## Checklist

### DIFF
--- a/sdk/src/cose_validator.rs
+++ b/sdk/src/cose_validator.rs
@@ -887,7 +887,7 @@ fn validate_with_cert(
     let pk_der = pk.raw;
 
     if validator.validate(sig, data, pk_der)? {
-        Ok(extract_subject_from_cert(&signcert)?)
+        extract_subject_from_cert(&signcert).or_else(|_| Ok("".to_string()))
     } else {
         Err(Error::CoseSignature)
     }
@@ -928,7 +928,7 @@ async fn validate_with_cert_async(
     let validator = get_validator(signing_alg);
 
     if validator.validate(sig, data, pk_der)? {
-        Ok(extract_subject_from_cert(&signcert)?)
+        extract_subject_from_cert(&signcert).or_else(|_| Ok("".to_string()))
     } else {
         Err(Error::CoseSignature)
     }
--- a/sdk/src/cose_validator.rs
+++ b/sdk/src/cose_validator.rs
@@ -887,7 +887,7 @@ fn validate_with_cert(
     let pk_der = pk.raw;
 
     if validator.validate(sig, data, pk_der)? {
-        extract_subject_from_cert(&signcert).or_else(|_| Ok("".to_string()))
+        Ok(extract_subject_from_cert(&signcert).unwrap_or_default())
     } else {
         Err(Error::CoseSignature)
     }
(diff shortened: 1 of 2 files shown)


sdk/src/cose_validator.rs
```diff
@@ -887,7 +887,7 @@ fn validate_with_cert(
     let pk_der = pk.raw;
 
     if validator.validate(sig, data, pk_der)? {
-        Ok(extract_subject_from_cert(&signcert)?)
+        extract_subject_from_cert(&signcert).or_else(|_| Ok("".to_string()))
```
scouten-adobe, line 890: This might read better as `Ok(extract_subject_from_cert(&signcert).unwrap_or_default())`.

`String::default()` is the empty string.

sdk/src/cose_validator.rs
```diff
@@ -928,7 +928,7 @@ async fn validate_with_cert_async(
     let validator = get_validator(signing_alg);
 
     if validator.validate(sig, data, pk_der)? {
-        Ok(extract_subject_from_cert(&signcert)?)
+        extract_subject_from_cert(&signcert).or_else(|_| Ok("".to_string()))
```
scouten-adobe, line 931: Same as line 890

sdk/src/store.rs
```diff
@@ -1981,6 +1984,14 @@ impl Store {
         }
     }
 
+    /// check the input url to see if it is a supported remotes URI
+    pub fn is_valid_remote_url(url: &str) -> bool {
+        match url::Url::parse(url) {
+            Ok(u) => !(u.scheme() != "http" && u.scheme() != "https"),
```
scouten-adobe, line 1990: Couldn't this be simplified to `Ok(u) => u.scheme() == "http" || u.scheme() == "https"),` ?

scouten-adobe (changes requested):
Minor changes as noted in comments; otherwise :+1:.

sdk/src/store.rs
gpeacock: or u.scheme().starts_with("http"),

gpeacock (approved):
I agree with Eric's changes, otherwise good to go

sdk/src/cose_validator.rs
mauricefisher64: implemented

sdk/src/store.rs
scouten-adobe: The latter has a risk of accepting bogus schemes such as "httpx".

sdk/src/store.rs
```diff
@@ -1981,6 +1984,14 @@ impl Store {
         }
     }
 
+    /// check the input url to see if it is a supported remotes URI
+    pub fn is_valid_remote_url(url: &str) -> bool {
+        match url::Url::parse(url) {
+            Ok(u) => u.scheme().starts_with("http"),
```
scouten-adobe, line 1990: This could misfire with a bogus URL scheme such as "httpx". Not sure how that might be abused, but I'd prefer to see the specific test for "http" and "https."

scouten-adobe (changes requested):

scouten-adobe (approved):

PR #148, opened 2022-10-26
Fix manifest.set_thumbnail when add_thumbnails is enabled

gpeacock:
Allow overriding a manifest thumbnail when the thumbnail feature is enabled.

## Checklist

### DIFF
--- a/sdk/src/manifest.rs
+++ b/sdk/src/manifest.rs
@@ -536,9 +536,12 @@ impl Manifest {
             self.set_title(ingredient.title());
         }
 
-        #[cfg(feature = "add_thumbnails")]
-        if let Ok((format, image)) = crate::utils::thumbnail::make_thumbnail(path.as_ref()) {
-            self.set_thumbnail(format, image);
+        // if a thumbnail is not already defined, create one here
+        if self.thumbnail().is_none() {
+            #[cfg(feature = "add_thumbnails")]
+            if let Ok((format, image)) = crate::utils::thumbnail::make_thumbnail(path.as_ref()) {
+                self.set_thumbnail(format, image);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


sdk/src/manifest.rs
```diff
@@ -536,9 +536,12 @@ impl Manifest {
             self.set_title(ingredient.title());
         }
 
-        #[cfg(feature = "add_thumbnails")]
-        if let Ok((format, image)) = crate::utils::thumbnail::make_thumbnail(path.as_ref()) {
-            self.set_thumbnail(format, image);
+        // if a thumbnail is not already defined, create one here
+        if self.thumbnail().is_none() {
+            #[cfg(feature = "add_thumbnails")]
+            if let Ok((format, image)) = crate::utils::thumbnail::make_thumbnail(path.as_ref()) {
```
mauricefisher64, line 542: Do you scale down the thumbnail or do we take whatever they supply?

dkozma: I'd prefer if we just passed in whatever the user sends without modifying the thumbnail in any way. This should also eliminate the need to package in any image manipulation libraries if `add_thumbnails` is turned off.

dkozma:
@gpeacock Thanks for fixing this. I noticed that `add_thumbnails` is a default feature, which is a bit unfortunate since I have to clear out the default feature set if I want to turn this auto-thumbnailing feature off. That means if we add any default features in the future, I would clear them out as well. Can we change this to be an opt-in feature instead (move it out of default features)?

sdk/src/manifest.rs
mauricefisher64: I agree with that in principle except now we have an attack vector.

dkozma: For the attack vector situation I guess I would be OK with attempting to decode and check if it is a legitimate image (similar to imagemagick's verify command), however it would be non-optimal if we had to re-encode an already-lossy JPEG thumbnail and increase compression artifacts. Additionally, there was a bug where we didn't write the EXIF orientation tag in the thumbnail, so if decided to re-encode any supplied thumbnails we would need to make sure to copy all of the pertinent headers over.

gpeacock: It was always the case that if you disable thumbnail generation, you have to set your own thumbnails. Nikon is using that feature. This just makes the feature work as intended if add_thumbnails is enabled. If we are not generating the thumbnails, then we should just take whatever they supply. Part of the point here is to not introduce the overhead of our default thumbnail generation code.
@dkozma FYI, disabling add_thumbnails will remove the image library and all the associated code. I agree that I'd like to get away from having default features, but this seems like a pretty common one to expect to work by default. I'm torn on that.

gpeacock: @mauricefisher64 At this level of the SDK there's only so much we can do to avoid misuse. We would either need to force developers to only use our thumbnail generation, or trust them to do the right thing.
FYI, I also allow overriding the thumbnail generator via options in from_file_with_options, but I may remove that since it feels redundant.

dkozma: @gpeacock I would expect most implementors to already have the libraries and tools to generate thumbnails in their codebase. Furthermore, we can't expect add_thumbnails to work with the plethora of video codecs and other documents we are planning on supporting.

mauricefisher64: Well what we will see is manifests that are not performant and present a bad experience.  Even if we don't scale we should check the size to make sure it is reasonable.

mauricefisher64: @dkozma Not sure I understand the comment.  There are only a few thumbnail formats supported in the spec.  It is not open-ended.

mauricefisher64 (approved):
👍

PR #158, opened 2022-12-02
Fixes support for instanceId on action and generate parameters.ingredient field when possible

gpeacock:
fix support for ingredientId on action and generate parameters.ingredient field when possible
Update make_test_images to add instanceId to actions

## Checklist

### DIFF
--- a/sdk/src/manifest.rs
+++ b/sdk/src/manifest.rs
@@ -591,9 +591,13 @@ impl Manifest {
             }
         }
 
+        let mut ingredient_map = HashMap::new();
         // add all ingredients to the claim
         for ingredient in &self.ingredients {
-            ingredient.add_to_claim(&mut claim, self.redactions.clone())?;
+            ingredient_map.insert(
+                ingredient.instance_id(),
+                ingredient.add_to_claim(&mut claim, self.redactions.clone())?,
+            );
         }
 
         let salt = DefaultSalt::default();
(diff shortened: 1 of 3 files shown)


mauricefisher64 (approved):
Seem fine to me, just a few optimization comments

sdk/src/manifest.rs
```diff
@@ -602,9 +606,33 @@ impl Manifest {
         for manifest_assertion in &self.assertions {
             match manifest_assertion.label() {
                 Actions::LABEL => {
-                    let actions: Actions = manifest_assertion.to_assertion()?;
-                    // todo: fixup parameters field from instance_id to ingredient uri for
-                    // c2pa.transcoded, c2pa.repackaged, and c2pa.placed action
+                    let mut actions: Actions = manifest_assertion.to_assertion()?;
+
+                    // fixup parameters field from instance_id to ingredient uri
+                    let needs_ingredient: Vec<(usize, crate::assertions::Action)> = actions
```
mauricefisher64, line 612: Could probably return a reference in the filter map to avoid the clone.

sdk/src/manifest.rs
```diff
@@ -602,9 +606,33 @@ impl Manifest {
...
+                                Some((i, a.clone()))
+                            } else {
+                                None
+                            }
+                        })
+                        .collect();
+
+                    for (index, action) in needs_ingredient {
+                        if let Some(id) = action.instance_id() {
+                            if let Some(hash_url) = ingredient_map.get(id) {
```
mauricefisher64, line 628: Could this be collapsed to: 

for (index, action) in needs_ingredient {
      if let Some(hash_url) = ingredient_map.get(action.instance_id()) {

gpeacock: I don't think that would work since instance_id() returns an option and the key is not an Option type.

sdk/src/manifest.rs
gpeacock: This way I avoid transferring partial ownership to needs_ingredient, and I need to clone any references later anyway.

PR #162, opened 2022-12-13
Create external manifests for unknown types

mauricefisher64:
This PR allows us to make and validate unknown/unsupported asset types.  When an asset extension is unknown we will make an external manifest.

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -1718,8 +1718,16 @@ impl Store {
     ) -> Result<Vec<u8>> {
         // clone the source to working copy if requested
 
-        get_supported_file_extension(asset_path).ok_or(Error::UnsupportedType)?; // verify extensions
-        let ext = get_supported_file_extension(dest_path).ok_or(Error::UnsupportedType)?;
+        // force generate external manifests for unknown types
+        let ext = match get_supported_file_extension(dest_path) {
+            Some(ext) => ext,
+            None => {
+                let pc = self.provenance_claim_mut().ok_or(Error::ClaimEncoding)?;
+                pc.set_external_manifest(); // generate external manifests for unknown types
+                MANIFEST_STORE_EXT.to_owned()
+            }
+        };
+
         if asset_path != dest_path {
             fs::copy(asset_path, dest_path).map_err(Error::IoError)?;
         }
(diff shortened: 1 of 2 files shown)


sdk/src/store.rs
```diff
@@ -2419,6 +2443,67 @@ pub mod tests {
...
+        // Create a 3rd party claim
+        let mut claim_capture = Claim::new("capture", Some("claim_capture"));
+        create_capture_claim(&mut claim_capture).unwrap();
+
+        // Do we generate JUMBF?
+        let signer = temp_signer();
+
+        // Move the claim to claims list. Note this is not real, the claims would have to be signed in between commmits
+        store.commit_claim(claim1).unwrap();
+        store.save_to_asset(&ap, &signer, &op).unwrap();
```
scouten-adobe, line 2473: I'd like to see this test verify that it actually went to an external file. I see the `save_to_asset` / `load_from_asset` pairing, but this test could just as easily pass for a file format where the manifest were embedded, right?

sdk/src/store.rs
```diff
@@ -1772,8 +1784,12 @@ impl Store {
                 crate::claim::RemoteManifest::NoRemote => dest_path.to_path_buf(),
                 crate::claim::RemoteManifest::SideCar => {
                     // remove any previous c2pa manifest from the asset
-                    remove_jumbf_from_file(dest_path)?;
-                    dest_path.with_extension(MANIFEST_STORE_EXT)
+                    match remove_jumbf_from_file(dest_path) {
```
scouten-adobe, line 1787: Duplicate logic of what's at line 1755. Could this be consolidated?

scouten-adobe (approved):
Couple minor nits; otherwise LGTM.

(Marking approve since I may not be able to re-review this afternoon.)

sdk/src/store.rs
mauricefisher64: How do you embed at manifest in a txt file?

sdk/src/store.rs
mauricefisher64: Not easily since this block is within a feature flag

mauricefisher64: I pushed a collapsed version but harder to read the cases

sdk/src/store.rs
mauricefisher64: Yes, Eric's point is valid but not for this test.  I intentionally chose a format that will never contain a manifest.

PR #163, opened 2022-12-15
Visualizations

mauricefisher64:
Improve SDK with new visualizations for tree view of manifest and the ability to dump the certificate chain for the active manifest.  Also include the new capability to remove manifests.  

## Checklist


dkozma:
@mauricefisher64 Should the tree view be part of c2patool instead of in the main SDK? Or would this be valuable for other apps as well? If so, should visualization be a feature? Just thinking about things like embedded device or web constraints where we want to keep the library as lean as possible.

scouten-adobe:
FWIW I share @dkozma's question about whether this could be done outside the c2pa crate.

When you say, "We have not exposed what I need in the SDK to do treeview in c2patool," my first instinct is to see if we could focus on that first.

I wonder what other innovation we're thwarting by not exposing those same APIs.

mauricefisher64:
That sounds like a different set of questions and not this PR.  We can decide not to add these features until then but I don't see the need for that.

scouten-adobe (approved):

PR #164, opened 2022-12-15
Address new Clippy warnings for Rust 1.66

scouten-adobe:


mauricefisher64 (approved):
Interesting how dereferencing Paths is now automatic and the auto upcasting for integers.  Thanks for fixing

gpeacock (approved):

PR #193, opened 2023-02-27
Update RemoteSigner for Wasm compatibility

dkozma:
Allows us to bypass the `Send` requirement for `async_trait` running in Wasm since that is not currently multithreaded and Rust does not have support for this in Wasm yet.

## Checklist

### DIFF
--- a/sdk/src/signer.rs
+++ b/sdk/src/signer.rs
@@ -77,7 +77,8 @@ use async_trait::async_trait;
 /// This trait exists to allow the signature mechanism to be extended.
 ///
 /// Use this when the implementation is asynchronous.
-#[async_trait]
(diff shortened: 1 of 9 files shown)


sdk/src/signer.rs
```diff
@@ -77,7 +77,8 @@ use async_trait::async_trait;
 /// This trait exists to allow the signature mechanism to be extended.
 ///
 /// Use this when the implementation is asynchronous.
-#[async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```
gpeacock, line 80: Is there any reason not to always define async_trait this way?

dkozma: I think it makes sense if there's a chance that that trait will be used in a Wasm context.

PR #198, opened 2023-03-02
(MINOR) Bump MSRV to 1.63.0

scouten-adobe:
Reason: Upstream dependency [openssl-src](https://crates.io/crates/openssl-src/versions) released a minor version upgrade (111.25.1+1.1.1t) a few days ago which _says_ that it bumps MSRV to 1.57.0, but that appears to be incorrect. Local testing shows that it no longer compiles with any version prior to 1.63.0.


mauricefisher64 (approved):
👍

PR #203, opened 2023-03-09
(MINOR) Riff support with refactored AssetIO

mauricefisher64:
This PR was meant to be RIFF support of (WAV, AVI & WEBP).  However because XMP support varies for different formats (we see this with BMFF as well) we needed a better abstraction of how we read & write assets.  So the AssetIO interface was refactored so that anything that deals with the reading and writing of support assets (including reporting type support and XMP) has been moved to the individual handlers.  Now the code in jumb_io, store, etc.  does not need specific knowledge of an asset's  capabilites.  It can be determines by querying traits.  

## Checklist


sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "c2pa"
-version = "0.18.1"
```
scouten-adobe, line 3: Please revert this change. The version number bump will happen automatically when we publish the next release. Leaving this here would cause a gap in versioning.

scouten-adobe (changes requested):
Need to flag the Cargo.toml issue now, but I'll have to come back to this later today or tomorrow. A few too many things going on today. 😬

sdk/Cargo.toml
mauricefisher64: The message I go from the error said to make it 0.19.0.  Should I change that back?

scouten-adobe: Yes, please change it back and ignore the error. More (frustrating) info in Jira under CAI-3479.

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -0,0 +1,424 @@
+// Copyright 2022 Adobe. All rights reserved.
```
scouten-adobe, line 1: 2023

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -0,0 +1,424 @@
...
+        // add c2pa manifest
+        if is_riff_chunk && !data.is_empty() && !format.contains("webp") {
+            children_contents.push(ChunkContents::Data(C2PA_CHUNK_ID, data.to_vec()));
+        }
+
+        for child in children {
+            children_contents.push(inject_c2pa(&child, file, data, format)?);
+        }
+
+        // for non webp we can place at the front
```
scouten-adobe, line 93: Comment looks like a copy/paste error. For webp, the c2pa needs to be at the end, I take it?

sdk/src/asset_io.rs
```diff
@@ -69,7 +69,24 @@ pub trait CAIWriter {
     ) -> Result<Vec<HashObjectPositions>>;
 }
 
-pub trait AssetIO {
+pub trait AssetIO: Sync + Send {
+    // create instance of AssetIO handler.  The extension type is passed in so
+    // that format specific customizations can be used during manifest embedding
+    fn new(asset_type: &str) -> Self
+    where
```
scouten-adobe, line 76: I'm curious about this interface. Is there a case where an `AssetIO` implementation would _not_ be `Sized`? I'm wondering if this trait bound should move up to `AssetIO` itself rather than being on `fn new`.

sdk/src/jumbf_io.rs
```diff
@@ -12,64 +12,79 @@
 // each license.
 
 use std::{
+    collections::HashMap,
     fs::{self, File},
     io::Cursor,
     path::{Path, PathBuf},
 };
 
+use lazy_static::lazy_static;
```
scouten-adobe, line 21: Need to pause review here. Will continue later today or tomorrow as time permits.

sdk/src/jumbf_io.rs
```diff
@@ -110,52 +125,33 @@ pub fn save_jumbf_to_memory(
...
-        "avif" | "heif" | "heic" | "mp4" | "m4a" | "application/mp4" | "audio/mp4"
-        | "image/avif" | "image/heic" | "image/heif" | "video/mp4"
-            if cfg!(feature = "bmff") && !cfg!(target_arch = "wasm32") =>
-        {
-            Some(Box::new(BmffIO::new(&asset_type)))
-        }
-        "tif" | "tiff" | "dng" => Some(Box::new(TiffIO {})),
-        _ => None,
+
+    if let Some(h) = ASSET_HANDLERS.get(&asset_type) {
```
scouten-adobe, line 141: This could probably be simplified to `ASSET_HANDLERS.get(&asset_type).map(|h| h.get_reader())`

sdk/src/jumbf_io.rs
```diff
@@ -110,52 +125,33 @@ pub fn save_jumbf_to_memory(
...
-        // "png" | "image/png" => Some(Box::new(PngIO {})),
-        // "avif" | "heif" | "heic" | "mp4" | "m4a" | "application/mp4" | "audio/mp4"
-        // | "image/avif" | "image/heic" | "image/heif" | "video/mp4"
-        //     if cfg!(feature = "bmff") && !cfg!(target_arch = "wasm32") =>
-        // {
-        //     Some(Box::new(BmffIO::new(&asset_type)))
-        // }
-        _ => None,
+
+    if let Some(h) = CAI_WRITERS.get(&asset_type) {
```
scouten-adobe, line 151: Similar to line 141.

scouten-adobe (approved):
:+1: with changes as noted

sdk/src/asset_io.rs
mauricefisher64: If you move the sized to the trait you get this error on get_handler interface:

"for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>"

PR #205, opened 2023-03-10
Adds base_path_take to ResourceStore

gpeacock:
Allows for memory writes to file based manifests
Fixes thumbnail write when reading from file based manifests.


sdk/src/manifest.rs
```diff
@@ -629,7 +629,12 @@ impl Manifest {
         if self.thumbnail().is_none() {
             #[cfg(feature = "add_thumbnails")]
             if let Ok((format, image)) = crate::utils::thumbnail::make_thumbnail(path.as_ref()) {
+                // Do not write this as a file when reading from files
```
scouten-adobe, line 632: If needed urgently, ok, but it looks to me like there should be some kind of option on `self.set_thumbnail()` instead of clearing and then restoring `self.base_path`.

gpeacock: This is special case that I don't think most users will ever need. I could make the base_path_take private if that helps. When we read a manifest in order to add it to an asset, it is a reading operation. But I discovered that we were writing a thumbnail file to the source folder. That's because internally we add a thumbnail before creating the claim. If it is a file based manifest, that was causing us to write out a file before we read it back in. 
I could to this differently by adding the thumbnail after creating the claim, but that would require significant restructuring. So in this this case, I needed a way to load one asset into memory, only when called by the internal API.

I thought about adding a parameter to set_thumbnail, but that would only have meaning if file_io is enabled, so we would need multiple versions of the function, with and without. I could go down that road, but I doubt there will be any other use cases for generating mixed in memory and from file resources in a manifest.

scouten-adobe: I decided to run that experiment in #208 and I agree that there's some mild restructuring and more code duplication than I'd like.

I don't know … it's kind of lesser-of-evils situation IMHO. Merge this or #208 as you prefer.

gpeacock: Thanks for the looking into that. I think I'll go with this approach because it provides a public way to remove the base_path, and that could be a desirable feature in some instances. 
I don't like the idea of set methods that take option parameters in order to allow setting None, but we still should have some way to set to None. A take method has the advantage of setting a value to None while also returning the current value if needed. Maybe we should follow that pattern more often.

scouten-adobe (approved):

PR #206, opened 2023-03-10
Fix hash algo warning in Wasm and hashing for RSA-PSS SHA-384/512

dkozma:

Since WebCrypto uses a hashing identifier such as `SHA-256` and the hashing utils (and I'm guessing the OpenSSL code) uses `sha256`, we were getting a warning that it couldn't find a match and that it defaulted to `sha256`. This also probably broke hash verification with RSA-PSS SHA-384 and SHA-512.

## Checklist


mauricefisher64 (approved):
I thought I had fixed this a while ago.  Maybe it is still in some experimental branch.

PR #208, opened 2023-03-10
Avoid writing thumbnail when reading a manifest as an ingredient

scouten-adobe:
Alternative approach to #205.


gpeacock:
Closing this as we took a different approach but will revisit updating the resource API later

PR #225, opened 2023-04-05
Updated several X509-related crate dependencies

scouten-adobe:
* Update x509-certificate to 0.19.0 in /sdk
* Update bcder from 0.6.0 to 0.71 in /sdk (required by the x509-cert upgrade)
* Replace asn1 code which had been copied from cryptographic-message-syntax crate with an actual dependency on cryptographic-message-syntax crate (helps with upgrade to bcder dependency)


sdk/Cargo.toml
```diff
@@ -60,6 +60,7 @@ chrono = { version = "0.4.24", default-features = false, features = ["serde", "w
 ciborium = "0.2.0"
 conv = "0.3.3"
 coset = "0.3.1"
+cryptographic-message-syntax = "0.22.0"
```
scouten-adobe, line 63: It looks like the ASN1 code has been more or less broken out from its original home in PyOxidizer into a standalone CMS crate that we can import here and it appears well-maintained (last push was ~2 weeks ago). Using this code allows us to update the bcder crate as well, which was blocking some other upgrades.

sdk/src/asn1/rfc5652.rs
```diff
@@ -652,12 +663,27 @@ impl SignedAttributes {
     /// Obtain an instance where the attributes are sorted according to DER
     /// rules. See the comment in [SignerInfo::signed_attributes_digested_content].
     pub fn as_sorted(&self) -> Result<Self, std::io::Error> {
-        // The official rules say you sort by the first element in the container.
-        // That's an Oid, which implements comparisons. So our job is easy.
-        let mut res = self.0.clone();
-        res.sort_by(|a, b| a.typ.as_ref().cmp(b.typ.as_ref()));
+        // All elements of the set have the same type. So sorting is based on encoded
```
mauricefisher64, line 666: Curious why you rewrote this section?

mauricefisher64 (approved):
Looks good.  Just one question

scouten-adobe:
BTW I've asked about having the ASN1 primitives broken out into their own crate or create feature?

https://github.com/indygreg/cryptography-rs/issues/9

PR #229, opened 2023-04-06
Upgrade x509-parser to 0.15.0

scouten-adobe:


sdk/src/cose_validator.rs
```diff
@@ -193,46 +191,45 @@ fn check_cert(
             let seq = parameters
                 .as_sequence()
                 .map_err(|_err| Error::CoseInvalidCert)?;
-            if seq.len() < 3 {
```
scouten-adobe, line 196: @mauricefisher64 when you review this, please look at this sequence extra carefully. The underlying x509-parser APIs got rewritten pretty substantially between 0.11 and 0.15 and I had to spend quite a bit of time figuring out the underlying intent and replacing it against the new APIs.

PR #199, opened 2023-03-07
Add support for ARM V8 (aarch64) on Linux

scouten-adobe:

PR #233, opened 2023-04-11
Add support for ARM on Linux

scouten-adobe:
(Adapted from similar support recently added to XMP Toolkit.)


scouten-adobe:
#199 was largely similar and was approved already. This is essentially an update to that, so taking that approval to merge.

Issue #234, opened 2023-04-11
Remove dependency on twoway crate


scouten-adobe:
Running `cargo deny` on this crate yields the following warning:

```
error[A003]: Crate `twoway` deprecated by the author
    ┌─ /.../c2pa-rs/Cargo.lock:223:1
    │
223 │ twoway 0.2.2 registry+https://github.com/rust-lang/crates.io-index
    │ ------------------------------------------------------------------ unmaintained advisory detected
    │
    = ID: RUSTSEC-2021-0146
    = Advisory: https://rustsec.org/advisories/RUSTSEC-2021-0146
    = The commit [`e99b3c7`](https://github.com/bluss/twoway/commit/e99b3c718df1117ad7f54c33f6540c8f46cc17dd) releasing version 0.2.2 explicitly deprecates `twoway` in favour of [`memchr`](https://crates.io/crates/memchr) crate.
    = Announcement: https://github.com/bluss/twoway
    = Solution: No safe upgrade is available!
    = twoway v0.2.2
      ├── c2pa v0.20.0
      │   └── make_test_images v0.20.0
      └── make_test_images v0.20.0 (*)
```

scouten-adobe:
@adobe export issue to Jira project CAI

mauricefisher64:
Fixing in PR: https://github.com/contentauth/c2pa-rs/pull/361

PR #235, opened 2023-04-11
(IGNORE) Tighten cargo deny settings

scouten-adobe:


scouten-adobe:
A few TO-DOs that are driven by external dependencies or work that we still need to do:

* Remove dependency on old/unsafe version of `time` crate: See https://github.com/indygreg/cryptography-rs/issues/10.
* Reinstate wildcard dependency warning: See https://github.com/contentauth/c2pa-rs/issues/234.
* Remove dependency on now-deprecated `twoway` crate: See https://github.com/contentauth/c2pa-rs/issues/234.

PR #240, opened 2023-04-20
Ingredient async and thumbnail support

gpeacock:
Adds Ingredient::from_memory_async, from_stream_async and from_stream_info, set_memory_thumbnail

### DIFF
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -421,25 +421,40 @@ impl Ingredient {
     /// [`ManifestStore`]: crate::ManifestStore
     #[cfg(feature = "file_io")]
     pub fn from_file_info<P: AsRef<Path>>(path: P) -> Self {
-        fn make_id(id_type: &str) -> String {
-            let uuid = Uuid::new_v4();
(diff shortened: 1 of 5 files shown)


sdk/src/ingredient.rs
```diff
@@ -670,7 +693,79 @@ impl Ingredient {
...
+    /// Thumbnail will be set only if one can be retrieved from a previous valid manifest
+    pub async fn from_memory_async(format: &str, buffer: &[u8]) -> Result<Self> {
+        let mut stream = Cursor::new(buffer);
+        Self::from_stream_async(format, &mut stream).await
+    }
+
+    /// Creates an `Ingredient` from a stream (async version).
+    ///
+    /// This does not set title or hash
+    /// Thumbnail will be set only if one can be retrieved from a previous valid manifest
```
dkozma, line 716: What does this mean?
>Thumbnail will be set only if one can be retrieved from a previous valid manifest

Do you mean automatically? I currently am setting a thumbnail after creating the `Ingredient` that was captured from the editing tool on import - just wanted to confirm that this workflow is valid.

PR #248, opened 2023-05-17
Riff streaming support

mauricefisher64:
Support for RIFF streaming write.  XMP support for RIFF.  Format compatibility changes. 

## Checklist


sdk/src/asset_handlers/riff_io.rs
```diff
@@ -50,6 +52,24 @@ const C2PA_CHUNK_ID: ChunkId = ChunkId {
...
+    value: [0x56, 0x50, 0x38, 0x58],
+}; // VP8X  chunk to hold auxiliary info
+
+const VP8_ID: ChunkId = ChunkId {
+    value: [0x56, 0x50, 0x38, 0x20],
+}; // VP8 chunk
+
+const VP8L_ID: ChunkId = ChunkId {
+    value: [0x56, 0x50, 0x38, 0x4C],
+}; // VP8 chunk
```
scouten-adobe, line 65: Copy/paste error on the label here?

sdk/src/asset_io.rs
```diff
@@ -48,13 +48,59 @@ impl CAIRead for std::io::Cursor<&mut [u8]> {}
 impl CAIRead for std::io::Cursor<Vec<u8>> {}
 impl CAIRead for NamedTempFile {}
 
+// Helper struct to create a concrete type for CAIRead when
+// that is required
+pub struct CAIReadWrapper<'a> {
```
scouten-adobe, line 53: This feels like a workaround and I'm wondering if we can limit its exposure to `pub(crate)` visibility rather than adding it to the API surface.

sdk/src/asset_io.rs
```diff
@@ -48,13 +48,59 @@ impl CAIRead for std::io::Cursor<&mut [u8]> {}
...
 pub trait CAIReadWrite: CAIRead + Write {}
 
 impl CAIReadWrite for std::fs::File {}
 impl CAIReadWrite for std::io::Cursor<&mut [u8]> {}
 impl CAIReadWrite for std::io::Cursor<Vec<u8>> {}
 impl CAIReadWrite for NamedTempFile {}
 
+// Helper struct to create a concrete type for CAIReadWrite when
+// that is required
+pub struct CAIReadWriteWrapper<'a> {
```
scouten-adobe, line 78: Same as for `CAIReadWrapper`.

scouten-adobe (commented):
I'm not famliar with the RIFF file formats, so my comments are more around API surface design.

scouten-adobe:
@dyro would you please also give this a close look since you're working in the asset I/O code these days?

sdk/src/asset_io.rs
mauricefisher64: Some APIs only accept concrete instances.  I agree this does not have to be public. This should go away in a future version of Rust.

mauricefisher64:
Understood, I agree with the comments.  I made those structs pub(crate).

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -61,54 +80,159 @@ where
...
+                // must be even
+                d.push(0);
+            }
+            children_contents.push(ChunkContents::Data(C2PA_CHUNK_ID, d));
         }
 
         Ok(ChunkContents::Children(id, chunk_type, children_contents))
     } else if id == riff::SEQT_ID {
-        let children = read_items(&mut chunk.iter_no_type(file));
+        let children = read_items(&mut chunk.iter_no_type(stream));
```
gpeacock, line 225: nice to see things moving to streams :)

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -198,46 +366,94 @@ impl AssetIO for RiffIO {
...
-            .write(true)
-            .open(asset_path)
-            .map_err(Error::IoError)?;
-        match new_contents.write(&mut output) {
-            Ok(_) => Ok(()),
-            Err(e) => Err(Error::IoError(e)),
-        }
+        new_contents
+            .write(&mut writer)
+            .map_err(|_e| Error::EmbeddingError)?;
```
gpeacock, line 440: This makes it easier to tell where the error happened, but we also lose track of the actual error for things like permissions and such. I'm not sure what our overall policy is for this kind of thing

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -198,46 +366,94 @@ impl AssetIO for RiffIO {
...
+        Ok(())
     }
 
-    fn get_object_locations(
+    fn get_object_locations_from_stream(
```
gpeacock, line 444: hurray!

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -314,38 +526,119 @@ impl RemoteRefEmbed for RiffIO {
...
+                            // save contents
+                            let mut writer = CAIReadWriteWrapper {
+                                reader_writer: output_stream,
+                            };
+                            new_contents
+                                .write(&mut writer)
+                                .map_err(|_e| Error::EmbeddingError)?;
+                            Ok(())
+                        } else {
+                            const MIN_XMP: &str = r#"<?xpacket begin="" id="W5M0MpCehiHzreSzNTczkc9d"?><x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="XMP Core 6.0.0"><rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"><rdf:Description rdf:about="" > </rdf:Description></rdf:RDF> </x:xmpmeta> "#;
```
gpeacock, line 591: Probably should have this in some common place, I'd imagine we need to do the same thing whenever there isn't already any XMP. Not important as long as this is the only place, but is it?

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -314,38 +526,119 @@ impl RemoteRefEmbed for RiffIO {
...
-                #[cfg(feature = "xmp_write")]
-                {
-                    match self.riff_format.as_ref() {
-                        "avi" | "wav" => {
-                            crate::embedded_xmp::add_manifest_uri_to_file(asset_path, &manifest_uri)
+                match self.riff_format.as_ref() {
+                    "avi" | "wav" | "webp" => {
+                        if let Some(curr_xmp) = self.read_xmp(input_stream) {
+                            let mut new_xmp = add_provenance(&curr_xmp, &manifest_uri)?;
+                            if new_xmp.len() % 2 == 1 {
```
gpeacock, line 553: This must have been fun to figure out

sdk/src/asset_io.rs
```diff
@@ -48,13 +48,59 @@ impl CAIRead for std::io::Cursor<&mut [u8]> {}
 impl CAIRead for std::io::Cursor<Vec<u8>> {}
 impl CAIRead for NamedTempFile {}
 
+// Helper struct to create a concrete type for CAIRead when
```
gpeacock, line 51: I'm a little unclear on why you need these, but they look well defined. Maybe a little more explanation about when to use them here?

gpeacock (approved):
A few minor comments but nothing requiring a change

sdk/src/asset_handlers/riff_io.rs
mauricefisher64: I won't be the only place.  I will figure it out on the next required case.

sdk/src/asset_io.rs
mauricefisher64: Comment added

sdk/src/asset_handlers/riff_io.rs
mauricefisher64: Now I have the pattern

sdk/src/asset_handlers/riff_io.rs
mauricefisher64: Moved to xmp_utils

sdk/src/asset_handlers/riff_io.rs
mauricefisher64: The mapping here is because of the return type.  Let me see what I can do about this.

sdk/src/asset_handlers/riff_io.rs
mauricefisher64: 👍

PR #249, opened 2023-05-17
Support for writing and reading manifest data from simple PDFs (without incremental updates or signatures)

dyro:
This PR introduces the `pdf_utils::Pdf` struct and the `pdf_utils::C2paPdf` trait, public (to the crate), `lopdf` abstractions with functionality to read and write manifest data from PDFs. This code works against `wasm` and `non-wasm` targets.

I intend to add read support next, then round out the work with incremental update support. I opened this PR against `origin/pdf` instead of `main` as I want to get this in front of folks before getting to far and this work doesn't support reading manifests or incremental PDF updates. 

## Checklist


sdk/Cargo.toml
```diff
@@ -66,6 +66,7 @@ fast-xml = "0.23.1"
 hex = "0.4.3"
 img-parts = "0.3.0"
 log = "0.4.8"
+lopdf = "0.30.0"
```
scouten-adobe, line 69: IMHO we should gate these dependencies on a "pdf" crate feature.

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,403 @@
+use std::io::Write;
```
scouten-adobe, line 1: Please add copyright header

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,403 @@
...
+            Reference(file_spec_ref),
+        ];
+
+        if let Ok(names) = self.document.catalog_mut()?.get_mut(NAMES_KEY) {
+            // Determine if `/Names` value is direct or indirect.
+            let names = if let Ok(names) = names.as_reference() {
+                self.document.get_object_mut(names)?.as_dict_mut()?
+            } else {
+                names.as_dict_mut()?
+            };
```
dyro, line 104: This pattern occurs a few times throughout the PR but has proved hard to abstract. `lopdf` doesn't have a great way to follow references when interacting with a mutable `lopdf` document. This is because when following references, you need a mutable reference to the document object, and a mutable reference to the indirect object you are attempting to get, which means you have two mutable references to `Document`.  I'm working on switching this code to make use of `ObjectIds` instead of the mutable reference. This is proving difficult for other reasons!

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,403 @@
...
+        first_page_dict.set(ANNOTATIONS_KEY, annotations.to_vec());
+
+        Ok(())
+    }
+
+    /// Gets the `Annotation`s for the provided Page `ObjectId`.
+    ///
+    /// This function will create an empty vector of annotations if the associated page of the PDF
+    /// doesn't have a vector of annotations already defined.
+    fn get_page_annotations(&self, page_ref: ObjectId) -> Result<Vec<Object>, Error> {
```
dyro, line 223: Working on a way to get rid of the `clones` below, but I'm running into the same issue above where the borrow checker complains of multiple mutable references to the `lopdf::Document`.

sdk/src/utils/pdf_utils.rs
dyro: Additionally, an issue with this pattern is that there's no guarantee there will only be one reference between objects. `lopdf` has implemented dereferencing helpers when using immutable references, but doesn't have anything for mutable references.

scouten-adobe (changes requested):

sdk/src/utils/pdf_utils.rs
scouten-adobe: Do you need to return owned `Vec<Object>` here? Is there any reason the success side of this `Result` couldn't be `Option<[&Object]>` (with some lifetime as needed)?

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// Saves the in-memory PDF to the provided `writer`.
+    fn save_to<W: Write>(&mut self, writer: &mut W) -> Result<(), std::io::Error> {
+        self.document.save_to(writer)
+    }
+
+    fn is_password_protected(&self) -> bool {
+        self.document.is_encrypted()
+    }
+
+    /// Determines if this PDF has a C2PA manifest embedded.
```
scouten-adobe, line 70: Thanks for the detailed comments throughout this file. 😎

sdk/Cargo.toml
dyro: That's a good idea - will do

dyro:
retest

sdk/src/claim.rs
```diff
@@ -1938,12 +1938,10 @@ impl Claim {
     // Do any assertions of this type exist?
     pub fn has_assertion_type(&self, in_label: &str) -> bool {
         let (label, _) = Claim::assertion_label_from_link(in_label);
-        let found = self
-            .assertion_store
-            .iter()
-            .find(|&x| x.assertion.label().starts_with(&label));
 
-        !matches!(found, None)
+        self.assertion_store
```
scouten-adobe, line 1942: This will likely conflict with my change for Rust 1.72 compatibility because new Clippy didn't like the `!matches` construction. Your overall change looks better, so please resolve by accepting your version.

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,507 @@
...
+    #[error(transparent)]
+    UnableToReadPdf(#[from] lopdf::Error),
+
+    #[error("Unable to add C2PA manifest as an annotation to the PDF.")]
+    AddingAnnotation,
+}
+
+const C2PA_MIME_TYPE: &str = "application/x-c2pa-manifest-store";
+
+pub(crate) trait C2paPdf: Sized {
```
scouten-adobe, line 44: I'm curious (not at all a blocker) to see if you see the `C2paPdf` trait being used anywhere other than for the `Pdf` struct you define below. If not, what's the benefit of the abstraction?

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        Ok(())
+    }
+
+    /// Gets a reference to the `C2PA` manifest bytes of the PDF.
+    ///
+    /// This method will read the bytes of the manifest, whether the manifest was added to the
+    /// PDF via an `Annotation` or an `EmbeddedFile`.
+    ///
+    /// Returns an [Ok(None)] if no manifest is present. Returns a [Ok(Some(&Vec<u8>))] when a manifest
```
scouten-adobe, line 196: I think these values should be wrapped with backticks and not braces for proper Markdown formatting.

scouten-adobe (changes requested):
One tiny documentation nit and a couple of non-blocker questions to consider.

I'd like @mauricefisher64 to review this since he knows PDF better than I do, but otherwise, this looks really solid. Easy code to follow!

sdk/src/utils/pdf_utils.rs
dyro: I added this to make unit testing easier in manifest generation code. It's hard to make PDFs that cause some functions in this trait fail. For example, `Pdf::write_manifest_as_annotation` calls `lopdf` code that can fail, but it's hard to generate a PDF that cause that to fail. Using a trait, we can instead use a mocking tool like `mockall` to do something like this: 

```rust
let mut pdf = MockC2paPdf::default();
pdf
    .expect_write_manifest_as_annotation()
    .returning(|_| Err(lopdf::Error::Blah))

// etc...
```

Happy to remove if you think it's not necessary though!

sdk/src/claim.rs
dyro: ah, I think you're right. will change!

sdk/src/utils/pdf_utils.rs
dyro: Okay, will merge!

sdk/src/utils/pdf_utils.rs
dyro: I added this to make unit testing easier in manifest generation code. It's hard to make PDFs that cause some functions in this trait fail. For example, `Pdf::write_manifest_as_annotation` calls `lopdf` code that can fail, but it's hard to generate a PDF that cause that to fail. Using a trait, we can instead use a mocking tool like `mockall` to do something like this: 

```rust
let mut pdf = MockC2paPdf::default();
pdf
    .expect_write_manifest_as_annotation()
    .returning(|_| Err(lopdf::Error::Blah))

// etc...
```

Happy to remove if you think it's not necessary though!

sdk/src/utils/pdf_utils.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        Ok(())
+    }
+
+    /// Gets a reference to the `C2PA` manifest bytes of the PDF.
+    ///
+    /// This method will read the bytes of the manifest, whether the manifest was added to the
+    /// PDF via an `Annotation` or an `EmbeddedFile`.
+    ///
+    /// Returns an `Ok(None)` if no manifest is present. Returns a `Ok(Some(&Vec<u8>))` when a manifest
```
mauricefisher64, line 196: The function need to return all the manifest stores found.  For example if someone adds another manifest, it will appear as an additional annotation.  Also we have the case where there is a file manifest then someone adds an annot manifest.

dyro: Thanks for the feedback, @mauricefisher64. I'll update the fn to return a `&[&[u8]]`.

mauricefisher64: Ok, feel free to make it Vec<Vec<8>> if you cannot return multiple &u8[] references.

mauricefisher64 (commented):
👍

dyro:
hey @scouten-adobe & @mauricefisher64 : I believe I've addressed all of your feedback. Please take a look again when you have a moment!

sdk/src/utils/pdf_utils.rs
scouten-adobe: Ahh, that makes sense. Please leave it as is.

sdk/src/utils/pdf_utils.rs
```diff
@@ -193,14 +193,14 @@ impl C2paPdf for Pdf {
...
     /// is present.
-    fn read_manifest_bytes(&self) -> Result<Option<&Vec<u8>>, Error> {
+    fn read_manifest_bytes(&self) -> Result<Option<Vec<&[u8]>>, Error> {
         if !self.has_c2pa_manifest() {
             return Ok(None);
-        }
+        };
 
-        Ok(Some(
+        Ok(Some(vec![
```
scouten-adobe, line 203: I don't understand why this API returns a vector here. Is there ever any case where there would be more than one byte slice that could be returned?

scouten-adobe (changes requested):
Please either explain the Vec of slices or remove the abstraction. Otherwise 👍

sdk/src/utils/pdf_utils.rs
dyro: I'll update the docs to reference the spec with a quick explanation, if you are okay with the `Vec<&[u8]>` return type

scouten-adobe: OK, that makes sense. Please do update the docs.

mauricefisher64 (approved):
👍

mauricefisher64:
Not holding up this PR but does the lopdf handle PDF permissions and passwords?

dyro:
@mauricefisher64: Yes, `lopdf` has support for passwords. After creating a `Document` instance, you call the its `decrypt` method with a password. Wasn't sure how to integrate that into the c2pa-rs public API

scouten-adobe (approved):

PR #251, opened 2023-05-18
Bmff v2

mauricefisher64:
This PR add Merkle tree validation to BMFF and a new streaming API to validate fragments for fragmented BMFF.

## Checklist


sdk/src/claim.rs
```diff
@@ -862,7 +868,7 @@ impl Claim {
     /// asset_bytes - reference to bytes of the asset
     pub async fn verify_claim_async<'a>(
         claim: &Claim,
-        asset_bytes: &'a [u8],
+        asset_data: &mut ClaimAssetData<'_>,
```
gpeacock, line 871: What forces the mut here? Is is that the stream needs to be mut?

sdk/src/ingredient.rs
```diff
@@ -740,13 +740,13 @@ impl Ingredient {
...
                         Ok(store) => {
                             // verify the store
-                            //todo, change this when we have a stream version of verify
-                            let mut buf: Vec<u8> = Vec::new();
-                            stream.rewind()?;
-                            stream.read_to_end(&mut buf).map_err(Error::IoError)?;
-                            Store::verify_store_async(&store, &buf, &mut validation_log)
-                                .await
-                                .map(|_| store)
+                            Store::verify_store_async(
```
gpeacock, line 743: Hurray!

sdk/src/utils/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+            index /= 2;
+        }
+        Ok(proof)
+    }
+
+    pub fn tree_dump(&self) {
+        for (i, layer) in self.layers.iter().enumerate() {
+            println!("Level: {i}");
+            for (j, mn) in layer.iter().enumerate() {
+                println!("{} (Node: {j})", Hexlify(&mn.0));
```
gpeacock, line 114: Nice! We may want to use this in other places

sdk/src/utils/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+
+// Implements Merkle tree support corresponding to the C2PA spec variant.  The Merkle tree is not reduced and
+// all leaves live at the bottom most level.  If the last layer node is an odd index (lacking a matching pair),
+// its node value is propagated to parent layer, no cloning or hashing is expected.  Null tree entries do not contribute to the hashes.
+pub struct C2PAMerkleTree {
+    pub leaves: Vec<MerkleNode>,
+    pub layers: Vec<Vec<MerkleNode>>,
+}
+
+#[allow(dead_code)]
```
gpeacock, line 30: Can we use a feature flag here?

sdk/src/utils/hash_utils.rs
```diff
@@ -350,6 +432,24 @@ pub fn verify_asset_by_alg(
...
+
+pub fn verify_stream_by_alg<R>(
+    alg: &str,
+    hash: &[u8],
+    reader: &mut R,
+    hash_range: Option<Vec<HashRange>>,
+    is_exclusion: bool,
+) -> bool
+where
+    R: Read + Seek + ?Sized,
```
gpeacock, line 444: Should we be using this instead of CAIRead, at least on any public interface?

sdk/src/store.rs
```diff
@@ -3725,6 +3801,29 @@ pub mod tests {
         let store = Store::load_from_asset(&ap, true, &mut report).expect("load_from_asset");
         println!("store = {store}");
     }
+
+    /*
+    #[test]
```
gpeacock, line 3806: Did you mean to leave it commented out. If so please add a comment saying why

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1547,6 +1556,268 @@ impl RemoteRefEmbed for BmffIO {
...
+        }
+
+        Ok(SampleToChunkBox {
+            entry_count,
+            entries,
+        })
+    }
+}
+
+#[allow(dead_code)]
```
gpeacock, line 1715: Is there a feature we should be using here?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1547,6 +1556,268 @@ impl RemoteRefEmbed for BmffIO {
...
+
+    if let Some(moov_tokens) = bmff_map.get("/moov") {
+        boxes.push(bmff_tree[moov_tokens[0]].data.clone());
+    }
+
+    Ok(boxes)
+}
+
+#[allow(dead_code)]
+struct SampleSizeBox {
```
gpeacock, line 1628: This doesn't look like it is exported anywhere, can you add a comment on why we are keeping it?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -57,6 +74,63 @@ impl ExclusionsMap {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
```
gpeacock, line 77: Probably should move this into utils at some point, but ok here as long as it is only used here.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -224,45 +325,447 @@ impl BmffHash {
...
+                }
+                current = rest;
+            }
+        } else {
+            output.insert(0, current);
+        }
+        Ok(output)
+    }
+
+    pub fn verify_hash(&self, asset_path: &Path, alg: Option<&str>) -> crate::error::Result<()> {
```
gpeacock, line 363: why don't we need a file_io feature on this?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -224,45 +325,447 @@ impl BmffHash {
...
+        let mut data = fs::File::open(asset_path)?;
+        self.verify_stream(&mut data, alg)
+    }
+
+    // Verifies BMFF hashes from a single file asset.  Both flat and fragmented (in a single file)
+    // are supported.
+    pub fn verify_stream(
```
gpeacock, line 370: Are we going to need async versions of these? If so, maybe we should start with that and just have blocking wrappers.

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1018,103 +1006,124 @@ fn get_manifest_token(
...
+
+                        // set reader to start of box contents
+                        skip_bytes_to(reader, box_info.data.offset + HEADER_SIZE + 16)?;
+
+                        // Fullbox => 8 bits for version 24 bits for flags
+                        let (_version, _flags) = read_box_header_ext(reader)?;
+                        data_len -= 4;
+
+                        // get the purpose
+                        let mut purpose = Vec::with_capacity(64);
```
gpeacock, line 1059: Is this a spec number or a good guess?

gpeacock (changes requested):
This was a lot of work! Well done. Some suggestions added, see comments.
My biggest concern is if we are going to need async support maybe we should start with it.
I wonder if we should just publish async stream APIs and make all the wrappers optional.

sdk/src/claim.rs
mauricefisher64: The stream pointers

sdk/src/ingredient.rs
```diff
@@ -1032,9 +1032,13 @@ impl Ingredient {
                 let mut buf: Vec<u8> = Vec::new();
                 stream.rewind()?;
                 stream.read_to_end(&mut buf).map_err(Error::IoError)?;
-                Store::verify_store_async(&store, &buf, &mut validation_log)
-                    .await
-                    .map(|_| store)
+                Store::verify_store_async(
```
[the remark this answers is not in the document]
mauricefisher64: I think you are right

sdk/src/store.rs
mauricefisher64: The test files are from MS.  We need to get files we can ship.

sdk/src/utils/hash_utils.rs
mauricefisher64: This used by WASM.  I will change once we have WASM stream support

sdk/src/utils/merkle.rs
mauricefisher64: Why, we have to read these and will write them shortly.

sdk/src/assertions/bmff_hash.rs
mauricefisher64: I think it already works.

sdk/src/assertions/bmff_hash.rs
mauricefisher64: We could but it is only called from store functions that are already behind that flag

sdk/src/assertions/bmff_hash.rs
```diff
@@ -274,14 +777,104 @@ impl AssertionBase for BmffHash {
...
+    let len = reader.seek(SeekFrom::End(0))?;
+
+    if old_pos != len {
+        reader.seek(SeekFrom::Start(old_pos))?;
+    }
+
+    Ok(len)
+}
+
+/* we need shippable examples
```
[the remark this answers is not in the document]
mauricefisher64: 👍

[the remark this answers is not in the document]
mauricefisher64: I don't know anyone who has examples of this other than MS.

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: purpose is in the spec

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: It can go, I switched to use MP4 crate for this

mauricefisher64: removed

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1547,6 +1556,268 @@ impl RemoteRefEmbed for BmffIO {
         Err(Error::UnsupportedType)
     }
 }
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
mauricefisher64: removed the unused code

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: Dead code removed

gpeacock (approved):

Issue #195, opened 2023-02-28
Validation and c2pa.placed


faceless2:
Cross-posted from the Discord group. I'm not quite sure where to put these, as I think they're issues with the test files as well as issues with the API.

There are two public files from https://github.com/c2pa-org/public-testfiles/tree/main/image that c2patool says are valid, but I'm not so sure:

## adobe-20220124-CICACACA.jpg

The active manifest ends in 8a8f462cb30d: the "c2pa.actions" assertion, fourth action is "c2pa.placed"but has no parameters, so no url:
```
            {
              "action": "c2pa.placed",
              "instanceId": "xmp:iid:a649c709-5c9f-47a7-aa1d-850b26789d62"
            },
```
But [18.10](https://c2pa.org/specifications/specifications/1.2/specs/C2PA_Specification.html#_actions) says
 >  c2pa.placed" action "shall" have parameters

and [15.7.3](https://c2pa.org/specifications/specifications/1.2/specs/C2PA_Specification.html#_assertion_validation) point a.iii.A.I.1  says this is an `assertion.action.ingredientMismatch`

## adobe-20220124-CAIAIIICAICIICAIICICA.jpg

The active manifest ends in 065a84099cc7: the "c2pa.actions" assertion fourth action is "c2pa.placed" and refers to "c2pa.ingredient", which has a relationship of "parentOf" 

```
            {
              "action": "c2pa.placed",
              "instanceId": "xmp.iid:813ee422-9736-4cdc-9be6-4e35ed8e41cb",
              "parameters": {
                "ingredient": {
                  "hash": "tTBD4/E0R0AjLUdJFpsVz3lE/KJUq22Vz0UGqzhEpVs=",
                  "url": "self#jumbf=c2pa.assertions/c2pa.ingredient"
                }
              }
            },

...

        "c2pa.ingredient": {
          "dc:format": "image/jpeg",
          "dc:title": "A.jpg",
          "documentID": "xmp.did:813ee422-9736-4cdc-9be6-4e35ed8e41cb",
          "instanceID": "xmp.iid:813ee422-9736-4cdc-9be6-4e35ed8e41cb",
          "relationship": "parentOf",
          "thumbnail": {
            "hash": "z55bRqFSv/HdQlFpU96AUPwDnHMWi/XVVanedKE7kxc=",
            "url": "self#jumbf=c2pa.assertions/c2pa.thumbnail.ingredient.jpeg"
          }
        },

```
[...]

PR #254, opened 2023-05-23
Fix for #195 make_test_images missing ingredient references

gpeacock:
Removed logic in Manifest to deduplicate ingredients with identical hashes
it was not comparing the ingredient fields for differences in things like instanceId
Applications are now responsible for deduplicating ingredients
Updated make_test_images to deduplicate identical ingredients
and have actions reference the same instance id when needed


dyro (approved):
LGTM

PR #260, opened 2023-06-01
Make remote manifest handling consistent across input types

dkozma:
While working on the Node.js SDK, I noticed that we were only listening to the `fetch_remote_manifests` flag when we were going through the file-based code path, not the in-memory ones. This attempts to make the remote fetching consistent whenever we try to load a store from a file, stream, or buffer.

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -42,8 +42,8 @@ use crate::{
         labels::{ASSERTIONS, CREDENTIALS, DATABOXES, SIGNATURE},
     },
     jumbf_io::{
-        get_assetio_handler, load_jumbf_from_memory, load_jumbf_from_stream,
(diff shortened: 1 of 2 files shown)


sdk/src/store.rs
```diff
@@ -2311,8 +2342,8 @@ impl Store {
                 )
                 .provenance
                 {
-                    // return an error with the url that should be read
```
mauricefisher64, line 2314: I don't think we need both a public **load_cai_from_memory** and public **load_from_memory**.  We have two paths that do the same thing.  I think we can remove public from **load_cai_from_memory** .  I think only the "load_from" functions should be public.

dkozma: Made the change - will this require a `(MINOR)` tag in the PR title?

PR #262, opened 2023-06-05
Fix timestamp Accuracy decoding

mauricefisher64:
Restore fix to rfc3161.rs to properly decode Accuracy values.  

## Checklist


scouten-adobe (approved):
Do you know if this has been fixed in the underlying cryptography-rs crate from which this originally came? I've sent a few PRs his way recently and he's been quite responsive.

mauricefisher64:
I don't think so since we fixed this earlier.   I forgot I had made the changes and the update we pulled had the bug again.  We don't see it with our images because our TSA does not fill in Accuracy.   I found it testing 3rd party images.

PR #264, opened 2023-06-09
Update Timestamp message imprint to include entire protected header

mauricefisher64:
Update Sig_structure used in Timestamp message imprint to use entire protected header.  This matches the interpretation by other implementations

## Checklist


gpeacock (approved):

gpeacock (approved):
Set parameter values was a good catch. 
The from_json_value function is an odd relic, but your extra checking is probably the right thing to do. It would also work the way it was, but not catch CBOR errors until embed.

PR #268, opened 2023-06-19
Fix for returning input stream data when using `embed_from_memory`

dkozma:
It looks like we were returning the data from the input stream in `embed_from_memory` instead of the data from the stream containing the embedded asset.

## Checklist

### DIFF
--- a/sdk/src/manifest.rs
+++ b/sdk/src/manifest.rs
@@ -1033,8 +1033,8 @@ impl Manifest {
         // todo:: see if we can pass a trait with to_vec support like we to for Strings
         let asset = asset.to_vec();
         let mut stream = std::io::Cursor::new(asset);
-        self.embed_stream(format, &mut stream, signer)?;
-        Ok(stream.into_inner())
+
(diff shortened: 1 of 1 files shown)


mauricefisher64 (approved):
👍

sdk/src/manifest.rs
```diff
@@ -1033,8 +1033,8 @@ impl Manifest {
         // todo:: see if we can pass a trait with to_vec support like we to for Strings
         let asset = asset.to_vec();
         let mut stream = std::io::Cursor::new(asset);
-        self.embed_stream(format, &mut stream, signer)?;
-        Ok(stream.into_inner())
+
+        self.embed_stream(format, &mut stream, signer)
```
gpeacock, line 1037: When I first wrote this it always replaced the input stream, but didn't update this function when when the logic changed to generate a new output stream. A good reason why every function, no matter how thin a wrapper, should have a unit test!

gpeacock (approved):

PR #271, opened 2023-06-21
Add `Send` to `CAIRead` trait so that it can be used across threads

dkozma:
I need to pass asset data across threads when working in the Node.js SDK due to how we need to offload computation off of the main thread. Since we operate off the `CAIRead` trait, we need to explicitly add the `Send` trait to allow this functionality.

## Checklist


mauricefisher64 (approved):
👍

PR #288, opened 2023-08-01
Adds a way to force no claim thumbnail generation

gpeacock:
When add_thumbnails is enabled, setting the manifest thumbnail format = "none" will force no claim thumbnail generation

### DIFF
--- a/sdk/src/manifest.rs
+++ b/sdk/src/manifest.rs
@@ -239,7 +239,7 @@ impl Manifest {
     /// Sets the thumbnail from a ResourceRef.
     pub fn set_thumbnail_ref(&mut self, thumbnail: ResourceRef) -> Result<&mut Self> {
         // verify the resource referenced exists
-        if !self.resources.exists(&thumbnail.identifier) {
+        if thumbnail.format != "none" && !self.resources.exists(&thumbnail.identifier) {
             return Err(Error::NotFound);
         };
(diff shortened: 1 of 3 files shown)


sdk/src/manifest.rs
```diff
@@ -239,7 +239,7 @@ impl Manifest {
     /// Sets the thumbnail from a ResourceRef.
     pub fn set_thumbnail_ref(&mut self, thumbnail: ResourceRef) -> Result<&mut Self> {
         // verify the resource referenced exists
-        if !self.resources.exists(&thumbnail.identifier) {
+        if thumbnail.format != "none" && !self.resources.exists(&thumbnail.identifier) {
```
dyro, line 242: Do we need to do the same thing in Ingredient.rs? https://github.com/contentauth/c2pa-rs/blob/cf041601e27ce7aed2ddc805744e60d28320c502/sdk/src/ingredient.rs#L359

dyro (approved):
Approved with one question

sdk/src/manifest.rs
gpeacock: We don't need to do this for ingredients because the api lets you override the thumbnail generation for ingredients, though I suppose it wouldn't hurt to be consistent.

gpeacock: I may add it later but it is much more complicated and not needed now.

PR #294, opened 2023-08-25
(IGNORE) Create new nightly build task

scouten-adobe:
Create nightly build branch that can be used for testing downstream builds


jackfarzan (changes requested):
The workflow itself looks good, however shall we add some info in the readme about the build options available?

PR #302, opened 2023-09-07
Lock openssl-sys version to 0.9.92

scouten-adobe:
Temporary measure while we sort out integration issues with the openssl 3 integration that's in openssl-sys 0.9.93.


dyro (approved):
Works for me locally 👍

PR #309, opened 2023-09-14
implements pdf read support

dyro:
This feature adds pdf read support to c2pa-rs. The support is limited, however. Only PDFs with a single manifest (no incremental update manifests) are currently supported. More complete support for PDF is coming shortly.

This PR also introduces a new test dependency named `mockall`. This allows unit testing the new pdf_io module without having to read entire PDF files into memory. Let me know what you think! We've been using this on the services and really enjoying how it makes unit testing easier. 

## Checklist


sdk/src/jumbf_io.rs
```diff
@@ -380,13 +385,15 @@ pub mod tests {
...
+        let pdf_supported = supported.iter().any(|s| s == "pdf");
+        assert_eq!(pdf_supported, cfg!(feature = "pdf"));
+
         assert!(supported.iter().any(|s| s == "jpg"));
         assert!(supported.iter().any(|s| s == "jpeg"));
         assert!(supported.iter().any(|s| s == "png"));
         assert!(supported.iter().any(|s| s == "mov"));
         assert!(supported.iter().any(|s| s == "mp4"));
         assert!(supported.iter().any(|s| s == "m4a"));
-        assert!(supported.iter().any(|s| s == "jpg"));
```
dyro, line 389: extra assertion for jpg here

sdk/src/utils/pdf_utils.rs
```diff
@@ -61,7 +62,10 @@ pub(crate) trait C2paPdf: Sized {
     fn write_manifest_as_annotation(&mut self, vec: Vec<u8>) -> Result<(), Error>;
 
     /// Returns a reference to the C2PA manifest bytes.
-    fn read_manifest_bytes(&self) -> Result<Option<Vec<&[u8]>>, Error>;
+    #[allow(clippy::needless_lifetimes)] // required for automock::mockall
+    fn read_manifest_bytes<'a>(&'a self) -> Result<Option<Vec<&'a [u8]>>, Error>;
```
dyro, line 66: the `mockall` macro requires the lifetime to be explicitly defined

sdk/src/utils/pdf_utils.rs
```diff
@@ -41,12 +41,13 @@ pub(crate) enum Error {
...
 const C2PA_MIME_TYPE: &str = "application/x-c2pa-manifest-store";
 
+#[cfg_attr(test, mockall::automock)]
 pub(crate) trait C2paPdf: Sized {
     /// Load a PDF from a slice of bytes.
     fn from_bytes(bytes: &[u8]) -> Result<Self, Error>;
 
     /// Save the `C2paPdf` implementation to the provided `writer`.
-    fn save_to<W: Write>(&mut self, writer: &mut W) -> Result<(), std::io::Error>;
+    fn save_to<W: Write + 'static>(&mut self, writer: &mut W) -> Result<(), std::io::Error>;
```
dyro, line 50: Same here: `mockall` requires generics items defined in traits to include `'static` lifetimes. I'm looking for a better solution, but this should work in the short term.

sdk/src/asset_handlers/pdf_io.rs
```diff
@@ -0,0 +1,256 @@
...
+};
+
+static SUPPORTED_TYPES: [&str; 2] = ["pdf", "application/pdf"];
+static WRITE_NOT_IMPLEMENTED: &str = "PDF write functionality will be added in a future release";
+
+pub struct PdfIO {}
+
+impl CAIReader for PdfIO {
+    fn read_cai(&self, asset_reader: &mut dyn CAIRead) -> crate::Result<Vec<u8>> {
+        let mut buffer: Vec<u8> = vec![];
```
scouten-adobe, line 30: Long-term, is there a way we can avoid this full copy of the asset?

sdk/src/utils/pdf_utils.rs
```diff
@@ -31,7 +31,7 @@ static SUBTYPE_KEY: &[u8] = b"Subtype";
 static NAMES_KEY: &[u8] = b"Names";
 
 #[derive(Debug, Error)]
-pub(crate) enum Error {
+pub enum Error {
```
scouten-adobe, line 34: If this is becoming part of the public API surface, it needs documentation.

sdk/src/utils/pdf_utils.rs
scouten-adobe: Appreciate the documentation about the `allow` here.

scouten-adobe (changes requested):
One minor question about docs; otherwise good

scouten-adobe (approved):

sdk/src/error.rs
```diff
@@ -51,6 +54,9 @@ pub enum Error {
     #[error("required feature missing")]
     MissingFeature(String),
 
+    #[error("feature implementation incomplete")]
+    NotImplemented(String),
```
gpeacock, line 58: We have needed this for long time

sdk/src/jumbf_io.rs
```diff
@@ -33,6 +33,8 @@ use crate::{
 lazy_static! {
     static ref ASSET_HANDLERS: HashMap<String, Box<dyn AssetIO>> = {
         let handlers: Vec<Box<dyn AssetIO>> = vec![
+            #[cfg(feature = "pdf")]
+            Box::new(crate::asset_handlers::pdf_io::PdfIO::new("")),
```
gpeacock, line 37: Probably better to put the use statement up top behind the feature, since you refer to it more than once

sdk/src/utils/pdf_utils.rs
gpeacock, line 1: This feels like it belongs under asset_handlers, utils isn't for asset specific things. We want to keep those separate so we can move them to their own crate(s) eventually.

sdk/src/error.rs
```diff
@@ -272,6 +278,10 @@ pub enum Error {
 
     #[error("prerelease content detected")]
     PrereleaseError,
+
+    #[error(transparent)]
+    #[cfg(feature = "pdf")]
+    PdfError(#[from] pdf_utils::Error),
```
gpeacock, line 284: hmmm, maybe we should have a class of AssetIO errors with subclasses for each type. I'm kind of nervous about adding top level errors for every type of supported asset structurally.

gpeacock: The current model is to use Error::InvalidAsset() and pass a string giving the particular problem

sdk/src/asset_handlers/pdf_io.rs
```diff
@@ -0,0 +1,256 @@
...
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+use std::{fs::File, path::Path};
+
+use crate::{
+    asset_io::{AssetIO, CAIReader, CAIWriter, HashObjectPositions},
+    utils::pdf_utils::{C2paPdf, Pdf},
```
gpeacock, line 18: I don't think this belongs in utils.

gpeacock (changes requested):
some questions and suggestions

sdk/src/asset_handlers/pdf_io.rs
dyro: Will move, thanks!

sdk/src/error.rs
dyro: Ah, that makes sense. I'll switch to using `InvalidAsset`. Thanks!

sdk/src/utils/pdf_utils.rs
dyro: That's good feedback- moved 👍

dyro:
Hey, @gpeacock : I've addressed you feedback in the following commits: https://github.com/contentauth/c2pa-rs/pull/309/files/a76eb8a41425415c6f2439be61693858a103ac8b..e1b2c3fba6c4f2066ca87636237bc1601e8cdab5

sdk/src/asset_handlers/pdf_io.rs
mauricefisher64: Looks like they support Document::load_from(reader: Read), this can be changed.

mauricefisher64 (approved):
👍

sdk/src/asset_handlers/pdf_io.rs
dyro: Updated - I ran into an issue with using mockall and the trait `Read`, but I've removed the constructors from the `C2paPdf` trait, as those aren't very useful for testing anyway, and got around the issue altogether!

PR #311, opened 2023-09-26
(MINOR) Signer can call timestamp authority directly

scouten-adobe:
Adds a new trait function to `Signer` and `AsyncSigner` which allows the trait implementation to call the RFC 3161 timestamp service directly. Default implementation is provided which behaves as before, so this is not (should not be) a breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
-        rust_version: [stable, 1.70.0]
+        rust_version: [stable, 1.71.0]
 
     steps:
       - name: Checkout repository
@@ -52,7 +52,7 @@ jobs:
       fail-fast: false
       matrix:
         target: [aarch64-unknown-linux-gnu]
-        rust_version: [stable, 1.70.0]
+        rust_version: [stable, 1.71.0]
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
-        rust_version: [stable, 1.71.0]
+        rust_version: [stable, 1.70.0]
 
     steps:
       - name: Checkout repository
@@ -52,7 +52,7 @@ jobs:
       fail-fast: false
       matrix:
         target: [aarch64-unknown-linux-gnu]
-        rust_version: [stable, 1.71.0]
+        rust_version: [stable, 1.70.0]
 
     steps:
       - name: Checkout repository
--- a/export_schema/Cargo.toml
+++ b/export_schema/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.26.0"
 authors = ["Dave Kozma <dkozma@adobe.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
-rust-version = "1.65.0"
+rust-version = "1.70.0"
 
 [dependencies]
 anyhow = "1.0.40"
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
-        rust_version: [stable, 1.70.0]
+        rust_version: [stable, 1.71.0]
```
gpeacock, line 19: What caused the need for the version bump?

sdk/src/signer.rs
```diff
@@ -34,6 +34,18 @@ pub trait Signer {
...
     }
 
+    /// Request RFC 3161 timestamp to be included in the manifest data
+    /// structure.
+    ///
+    /// `message` is a preliminary hash of the claim
+    ///
+    /// The default implementation will send the request to the URL
+    /// provided by [`Self::time_authority_url()`], if any.
+    fn send_rfc3161_request(&self, message: &[u8]) -> Option<Result<Vec<u8>>> {
```
gpeacock, line 44: I'm not sure I like the idea of the RFC reference being in the function name here. I'd agree it is technically correct and specific, but I think simply describing what the function does, rather than requiring them to look up an RFC is much clearer overall. It would be different if there were multiple protocols we were supporting and we needed to make clear which this one is.

gpeacock (approved):

.github/workflows/ci.yml
scouten-adobe: async-generic has a compile-time issue with 1.70.

sdk/src/signer.rs
scouten-adobe: @gpeacock good suggestion. I've changed the name to `send_timestamp_request`.

dyro (approved):
`async_generic` looks slick!

.github/workflows/ci.yml
scouten-adobe: @gpeacock I was able to resolve the issue in `async-generic`, so I've now reverted the MSRV change.

export_schema/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.26.0"
 authors = ["Dave Kozma <dkozma@adobe.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
-rust-version = "1.65.0"
+rust-version = "1.70.0"
```
scouten-adobe, line 7: FWIW when we're doing an MSRV change, we should be changing this value. Looks like the MSRV was changed in documentation and CI while I was away, but we missed this value.

sdk/src/cose_sign.rs
```diff
@@ -155,31 +106,37 @@ pub async fn cose_sign_async(
...
-    let tbs = coset::sig_structure_data(
-        coset::SignatureContext::CoseSign1,
-        sign1.protected.clone(),
-        None,
-        aad,
-        sign1.payload.as_ref().unwrap_or(&vec![]),
-    );
-    sign1.signature = signer.sign(tbs).await?;
+    if _async {
+        let tbs = coset::sig_structure_data(
```
scouten-adobe, line 131: @mauricefisher64 can this logic be coalesced with the block up at line 121 or does it need to be in this sequence?

mauricefisher64: Yes, you can eliminate 122-126 and just sign the way we do with async buy explicitly signing the tbs.

sdk/src/time_stamp.rs
```diff
@@ -85,17 +96,6 @@ pub(crate) fn cose_sigtst_to_tstinfos(
     }
 }
 
-/// Get URL to Time Authority to use
-#[allow(dead_code)] // in case we make use of this later
```
mauricefisher64, line 89: Does @gpeacock still support this override?

mauricefisher64 (approved):
👍

sdk/src/cose_sign.rs
scouten-adobe: @mauricefisher64 thanks … done!

PR #313, opened 2023-09-28
Adds embed_to_stream

gpeacock:
Adds embed_to_stream
with output stream and returning c2pa_data
replaces embed_stream
Fixes a bug in embed_stream where missing thumbnails were not reported


sdk/src/manifest.rs
```diff
@@ -2209,6 +2223,33 @@ pub(crate) mod tests {
...
+        let mut source = std::io::Cursor::new(vec![1, 2, 3]);
+        let mut dest = std::io::Cursor::new(Vec::new());
+        let signer = temp_signer();
+        let result =
+            manifest.embed_to_stream("image/jpeg", &mut source, &mut dest, signer.as_ref());
+        assert!(result.is_err());
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("resource not found: does_not_exist.jpg"));
```
dyro, line 2250: Not required for merging, but I've been using this pattern to test for errors so we don't have to test against the error strings of the c2pa-rs error. 

```suggestion
        assert!(matches!(
            result,
            Err(crate::Error::ResourceNotFound(_)
        ))
```

sdk/src/manifest.rs
```diff
@@ -1023,11 +1023,30 @@ impl Manifest {
 
     /// Embed a signed manifest into a stream using a supplied signer.
     /// returns the bytes of the new asset
+    /// We plan to deprecate this, please use embed_to_stream instead
```
dyro, line 1026: Is it worth adding the `#[deprecated]` tag now?

dyro (approved):
Approved with a question and suggestion 👍

scouten-adobe (commented):
Agree with @dyro's questions and comments.

sdk/src/manifest.rs
gpeacock: I updated this to be deprecated, which, Rust being Rust, then forced me to update any other calls we had to it in our SDK so that I didn't get warnings.

sdk/src/manifest.rs
gpeacock: I'm going to stick with this test here, because I'm trying to validate the filename inside the error as well as the error type. But if I was just validating for the type of error, I agree the other is better.

PR #315, opened 2023-10-03
Add ability to customize HTTP headers on timestamp request to Signer and AsyncSigner traits

scouten-adobe:


sdk/src/time_stamp.rs
```diff
@@ -117,8 +118,15 @@ fn time_stamp_request_http(
 
     let body_reader = std::io::Cursor::new(body);
 
-    let response = ureq::post(url)
-        .set("Content-Type", HTTP_CONTENT_TYPE_REQUEST)
+    let mut req = ureq::post(url).set("Content-Type", HTTP_CONTENT_TYPE_REQUEST);
```
dyro, line 121: Maybe we should set the “Content-Type” header after we've added the user provided headers, so we can prevent a user from breaking the spec by providing their own ContentType header?

dyro (approved):
Looks clean ⚡

sdk/src/signer.rs
```diff
@@ -42,8 +50,10 @@ pub trait Signer {
     /// The default implementation will send the request to the URL
     /// provided by [`Self::time_authority_url()`], if any.
     fn send_timestamp_request(&self, message: &[u8]) -> Option<Result<Vec<u8>>> {
```
mauricefisher64, line 52: Where is this called.   Does this function need to be in this trait or can this crate::time_stamp::default_rfc3161_request just be called where you call this function?

scouten-adobe: `send_timestamp_request` was introduced in #311 and allows the implementor of `Signer` to completely replace the timestamp call. Unless we decide to remove that API, this needs to remain as-is.

sdk/src/time_stamp.rs
mauricefisher64: The content type for timestamp servers must be "application/timestamp-query"

mauricefisher64: Similarly the response type must be: "application/timestamp-reply"

scouten-adobe: @mauricefisher64 I moved the request header to after the user-specified headers per @dyro's suggestion so that it will overwrite any user-specified `Content-Type` header. At line 111 of this file, we define ...

```
    const HTTP_CONTENT_TYPE_REQUEST: &str = "application/timestamp-query";
```

.. so the request should be formatted as you say.

PR #317, opened 2023-10-04
Support for validating JPEGs that contain MPF (multi-picture format).

mauricefisher64:
Allow box hashing to contain arbitrary data between segments if they are contained within a range.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ The SDK has been tested on the following operating systems:
  | `png`         | `image/png`                                         |
  | `svg`         | `image/svg+xml`                                     |
  | `tif`,`tiff`  | `image/tiff`                                        |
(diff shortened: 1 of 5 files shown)


scouten-adobe:
@mauricefisher64 I'll amend this PR to silence this warning for now. I've filed an issue at the `jfifdump` repository asking them to explore alternatives to the unmaintained `json` crate which is at heart here (https://github.com/vstroebel/jfifdump/issues/4). If that isn't addressed in a timely fashion, we should consider submitting a PR to `jfifdump` or finding an alternative as this is a security vulnerability.

gpeacock (approved):
We should make sure we do some significant regression testing on jpegs with this level of change right before MAX.

PR #319, opened 2023-10-04
Add ManifestStore::from_stream

gpeacock:
Adds ManifestStore.from_stream() so you can create a manifest store from a stream.


sdk/src/manifest.rs
```diff
@@ -1023,11 +1023,30 @@ impl Manifest {
 
     /// Embed a signed manifest into a stream using a supplied signer.
     /// returns the bytes of the new asset
+    /// We plan to deprecate this, please use embed_to_stream instead
```
scouten-adobe: Should we formally mark this as deprecated?

https://doc.rust-lang.org/reference/attributes/diagnostics.html#:~:text=The%20deprecated%20attribute&text=rustdoc%20will%20show%20item%20deprecation,string%20in%20the%20deprecation%20message.

sdk/src/manifest.rs
```diff
@@ -1022,12 +1022,34 @@ impl Manifest {
     }
 
     /// Embed a signed manifest into a stream using a supplied signer.
-    /// returns the bytes of the new asset
+    ///
+    /// Returns the bytes of the new asset.
+    ///
+    /// We plan to deprecate this; please use embed_to_stream instead.
```
scouten-adobe, line 1028: Should we formally mark this as deprecated?

See https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute.

scouten-adobe (approved):
I added a few proofreading/documentation edits and one question, not a blocker.

PR #322, opened 2023-10-16
(MINOR) Reuse claim thumbnail as ingredient thumbnail if the store is valid

gpeacock:
Reuses claim thumbnail as ingredient thumbnail if the store is valid.
This relies on JUMBF URLs as resource identifiers
Adds optional alg and hash to Resource references
Removes requirement that a resource exists in resources when setting Ingredient Resource ref.
The ingredient c2pa_manifest.format value is now "application/c2pa" instead of "c2pa"


scouten-adobe (approved):
@gpeacock I made a few documentation-level edits; if these are OK with you, then I'm :+1: on this PR.

PR #325, opened 2023-10-17
(MINOR) Support databox thumbnails CAI-4142

gpeacock:
Support databox thumbnails
Use jumbf uris for databox entries in manifeststore
[CAI-4141] always return application/c2pa for format instead of c2pa
Separates mime extension handling into a mime.rs utility.

### DIFF
--- a/sdk/src/utils/mime.rs
+++ b/sdk/src/utils/mime.rs
@@ -1,0 +1,72 @@
+// Copyright 2022 Adobe. All rights reserved.
+// This file is licensed to you under the Apache License,
+// Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+// or the MIT license (http://opensource.org/licenses/MIT),
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+pub fn extension_to_mime(extension: &str) -> Option<&'static str> {
+    Some(match extension {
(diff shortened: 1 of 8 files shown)


sdk/src/utils/mime.rs
```diff
@@ -0,0 +1,72 @@
...
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+pub fn extension_to_mime(extension: &str) -> Option<&'static str> {
```
dyro, line 14: Maybe some docs here since this is a new public interface

dyro (approved):
Couple suggestions for some simple docs

mauricefisher64 (approved):
👍

sdk/src/utils/mime.rs
gpeacock: I added docs but the the utils stuff is all crate private so it won't be in the public SDK, unless we want to add it there.

sdk/src/utils/mime.rs
```diff
@@ -0,0 +1,72 @@
...
+        "wav" => "audio/vnd.wav",
+        "aif" | "aifc" | "aiff" => "audio/aiff",
+        "ogg" => "audio/ogg",
+        "pdf" => "application/pdf",
+        "ai" => "application/postscript",
+        _ => return None,
+    })
+}
+
+pub fn format_to_extension(format: &str) -> Option<&'static str> {
```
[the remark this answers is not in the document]
gpeacock: Added a doc line but it isn't public

sdk/src/utils/mime.rs
dyro: ah, got it!

PR #326, opened 2023-10-17
(IGNORE) Readme Release notes update.

gpeacock:
Updates the Readme section in release notes for significant changes


scouten-adobe (changes requested):
I really don't see the point of this being in the README. I'd like to avoid manually duplicating the changelog which conveys much of the same information and is automatically generated.

Can we talk about how to limit the scope of what is manually authored to those things that are likely to cause SDK users to have to _change_ their code?

gpeacock:
I'll delete this PR and add a new one to remove the existing release notes. We should not have something there which we are not updating. It means we need to be much better at documenting things in the changelog.

PR #330, opened 2023-10-25
(IGNORE) Refactor implementation of update_data_hash and update_bmff_hash

scouten-adobe:
(This supports an upcoming PR where I will need similar functionality for another assertion type.)


mauricefisher64 (approved):
👍

PR #331, opened 2023-10-25
(IGNORE) Add new (crate private) API function Claim::add_raw_assertion

scouten-adobe:


gpeacock (commented):
How does it help to add a crate private function that is never used? Doesn't  clippy complain about that? This is to add an assertion that doesn't implement AssertionBase?
I have a hunch there is a better way to do this, but I'd like a unit test or more context about how it would be used.

scouten-adobe:
Withdrawing this PR. Found a better way through this that doesn't require this PR.

PR #333, opened 2023-11-01
(PATCH) switches af relationship for a reference to the c2pa data to an array of references, one of which is the c2pa spec

dyro:
This is a small change that changes the value of the AF relationship key in the PDF's catalog from a reference to an array of references. This change was made because a PDF can have multiple associated files. The previous version assumed the c2pa file spec reference would be the only associated file.

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/pdf.rs
+++ b/sdk/src/asset_handlers/pdf.rs
@@ -51,6 +51,11 @@ pub enum Error {
     // the manifest bytes in the PDF's embedded files.
     #[error("Unable to find C2PA manifest in the PDF's embedded files.")]
     UnableToFindEmbeddedFileManifest,
+
+    /// This error occurs when we an error was encountered trying to find the PDF's C2PA embedded
+    /// file specification in the array of Associated Files defined in the catalog.
+    #[error("Unable to find a C2PA embedded file specification in PDF's associated files array")]
+    FindingC2PAFileSpec,
 }
 
 const C2PA_MIME_TYPE: &str = "application/x-c2pa-manifest-store";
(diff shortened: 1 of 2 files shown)


sdk/src/asset_handlers/pdf.rs
```diff
@@ -361,16 +408,28 @@ impl Pdf {
...
-    /// Sets the Associated File (/AF) key of the PDF to the provided embedded file spec reference.
-    fn set_af_relationship(&mut self, embedded_file_spec_ref: ObjectId) -> Result<(), Error> {
-        self.document
-            .catalog_mut()?
-            .set(ASSOCIATED_FILE_KEY, embedded_file_spec_ref);
+    /// Creates, or appends to, the Associated File (`AF`) array the embedded file spec reference of the
+    /// C2PA data.
+    fn push_associated_file(&mut self, embedded_file_spec_ref: ObjectId) -> Result<(), Error> {
+        let catalog = self.document.catalog_mut()?;
+        if catalog.get_mut(ASSOCIATED_FILE_KEY).is_err() {
```
mauricefisher64, line 415: No big deal but there is a Hashmap function for this case of if entry not present create or else return the object.  It is called Vacanct and looks like this in practice:

if let std::collections::hash_map::Entry::Vacant(e) = self.claims_map.entry(label) {
                let index = self.claims.push_get_index(ingredient_claim);
                e.insert(index);
            }
Where e is the new entry or the existing entry you want to modify.

mauricefisher64 (approved):
👍

sdk/src/asset_handlers/pdf.rs
dyro: nice, will take a look at that!

PR #334, opened 2023-11-02
(IGNORE) Removed Release Notes section from Readme

gpeacock:
Removed Release Notes section from Readme.
It was not being updated. Use Changelog instead.


scouten-adobe (approved):

crandmck:
The release notes go up through 0.29, which is quite recent.  It might make sense to keep it for now.

PR #341, opened 2023-11-14
Update image crate dependency and limit features

scouten-adobe:
This should help resolve recent build failures on Windows.


scouten-adobe:
I believe we should merge this despite the Windows build failure as it reduces the dependency service.

See https://github.com/rust-lang/cargo/issues/12979 for explanation.

mauricefisher64 (approved):
👍

PR #352, opened 2023-12-01
CAI-5041 Clear Windows temp attribute

gpeacock:
NamedTempFile sets the windows Temp attribute, so clear it before renaming


mauricefisher64 (commented):
👍 

Thanks, I was looking make a similar change

scouten-adobe (approved):

PR #369, opened 2024-01-23
Expose authoring support in WASM

duggaraju:
Expose async version of 'embed_from_memory' which can be called from wasm.
## Checklist
Since signing methods in the JavaScript land are async need to use AsyncSigner for signing.
So created an embed_from_memory_async API 
combine sign_claim and sign_claim_async into a single function using async_generic macro.
Created a WebCryptoSigner type that implements AsyncSigner and used that for the test case.


sdk/src/utils/test.rs
```diff
@@ -371,6 +409,102 @@ impl crate::signer::RemoteSigner for TempRemoteSigner {
...
+        Reflect::set(&alg, &"hash".into(), &"SHA-256".into()).unwrap();
+        let promise = crypto
+            .sign_with_object_and_u8_array(&alg, &key, &mut digest)
+            .unwrap();
+        let result = JsFuture::from(promise).await.unwrap();
+        Ok(Uint8Array::new(&result).to_vec())
+    }
+
+    fn reserve_size(&self) -> usize {
+        10000
```
mauricefisher64, line 504: How do you know this size is sufficient.  The size need to cover the size of certificate chain + size of timestamp response + size of OCSP response + size of structure.  If the size if not sufficient the code will throw an error.

sdk/src/utils/test.rs
```diff
@@ -371,6 +409,102 @@ impl crate::signer::RemoteSigner for TempRemoteSigner {
     }
 }
 
+#[cfg(target_arch = "wasm32")]
```
mauricefisher64, line 412: Is this signer meant to be real or stay as a test?  If it is real can you move it to the wasm folder

sdk/src/utils/test.rs
```diff
@@ -371,6 +409,102 @@ impl crate::signer::RemoteSigner for TempRemoteSigner {
...
+            .unwrap();
+        let result = JsFuture::from(promise).await.unwrap();
+        let mut digest = Uint8Array::new(&result).to_vec();
+
+        let key = Uint8Array::new_with_length(self.key.len() as u32);
+        key.copy_from(&self.key);
+        let usages = Array::new();
+        usages.push(&"sign".into());
+        let alg = Object::new();
+        Reflect::set(&alg, &"name".into(), &"ECDSA".into()).unwrap();
```
mauricefisher64, line 485: Do you plan to support all the signature types?  I know Ed25519 is not supported win WebCrypto.

sdk/src/lib.rs
```diff
@@ -82,6 +82,7 @@
 //! # }
 //! ```
 
+pub use asn1::rfc3161::{MessageImprint, TimeStampReq};
```
mauricefisher64, line 85: Why did this need to be exposed?

sdk/src/utils/test.rs
mauricefisher64: If this is just test code then the fixed size if fine.

sdk/src/store.rs
```diff
@@ -1892,6 +1879,13 @@ impl Store {
     /// When called, the stream should contain an asset matching format.
     /// on return, the stream will contain the new manifest signed with signer
     /// This directly modifies the asset in stream, backup stream first if you need to preserve it.
+    #[async_generic(async_signature(
```
mauricefisher64, line 1882: good consolidation

sdk/src/store.rs
```diff
@@ -388,61 +389,47 @@ impl Store {
     }
 
     /// Sign the claim and return signature.
-    pub fn sign_claim(
+    #[async_generic(async_signature(
```
mauricefisher64, line 392: good consolidation

sdk/src/utils/test.rs
duggaraju: This is just test code.

sdk/src/lib.rs
duggaraju: Right now the default implementation of send_time_stamp_request in AsyncSigner trait does the required marshalling for the timestamp body (instead of just making the HTTP call). If I need to do the same outside in an implementation (e.g a fetch call in JavaScript) I will have to redo all the ASN.1 serialization of the timestamp request. I was trying to reuse the same by exposing those helpers. If we could refactor the AsyncSigner trait to separate the HTTP request part and the timestmap request marshalling part then no need to expose the same.

sdk/src/utils/test.rs
duggaraju: The implementation can be moved to wasm folder. But I just had the logic there for a ES256 certificate.  Eventually we will have to add logic  for other certificate types (or fail).  Also, it is only used in test code now, not sure if we want to expose this for taget_arch="wasm32"

sdk/src/store.rs
duggaraju: @mauricefisher64  I want to keep all the signing methods under a separate feature flag so web assembly size doesn't increase  for verification only.  Right now there is a feature flag called openssl_sign and that assumes openssl.  I would like a generic feature flag called sign for signing in general. Is that, ok? or can you think of a better name?

sdk/src/utils/test.rs
duggaraju: Right now it only supports es256 and that is why I kept it in the test code. A generic web crypt signer would need to support all types of keys. But I was also wondering if c2pa-js is the right place for such code.

mauricefisher64: Maybe, since we have users builded to other bindings the might not have access to web crypto

sdk/src/store.rs
mauricefisher64: No sure how much that would save you since the same functions libraries that sign are the same as those needed for validation.  The other way around might yield some savings (have validation as an option).  We are working on that change now.

sdk/src/utils/test.rs
mauricefisher64: Ok, leave this as a test signer until we figure this out.

sdk/src/lib.rs
mauricefisher64: What do you mean default implementation for AsyncSigner.   We don't have such a thing except for unit tests of the async interfaces.  The AsyncSigner trait does not have references to TimeStampReq or MessageImprint.  Those are implementation detail of those test signers.  If you were to implement your own AsyncSigner outside of the SDK you can do anything you want as long as you meet the trait requirements.

sdk/src/store.rs
duggaraju: Having a separate feature for verfication works too. Will leave this as is for now

sdk/src/lib.rs
duggaraju: I mean the default implementation of the function send_timestamp_request in AsyncSigner trait.  The default_rfc3161_request implementation uses those. I exposed the same to use outside the crate without having to re-implement it.
```rust
    async fn send_timestamp_request(&self, message: &[u8]) -> Option<Result<Vec<u8>>> {
        let headers: Option<Vec<(String, String)>> = self.timestamp_request_headers();

        self.time_authority_url()
            .map(|url| crate::time_stamp::default_rfc3161_request(&url, headers, message))
    }
```

mauricefisher64: I see, there was a change to the Signer API to require time stamping.  Let me do a quick check on if we want to keep this behavior.

mauricefisher64: We do no plan to re-export internal structures.  I am not opposed to a creating a function that returns the TimeStampRequest bytes that you can use to call the TSA url.

duggaraju: Sure. I can make that change.

duggaraju: Modified the AsyncSigner interface to have a separate method for the message body.  and have a default implementation. Refactor timestamp.rs to accommodate the same.

sdk/src/signer.rs
```diff
@@ -129,6 +129,10 @@ pub trait AsyncSigner: Sync {
         None
     }
 
+    fn timestamp_request_body(&self, message: &[u8]) -> Result<Vec<u8>> {
```
mauricefisher64, line 132: Changes look good.  Where is this used?

duggaraju: Ideally it would be used in the send_timestamp_request() implemenation as below. Howerver the current implemenation has a check to compare the nonce from the request and response as an extra step of validation.  With current API that would require parsing the body back again. Is that OK @mauricefisher64 or can we ignore the check?
```rust
    async fn send_timestamp_request(&self, message: &[u8]) -> Option<Result<Vec<u8>>> {
        let headers: Option<Vec<(String, String)>> = self.timestamp_request_headers();

        if let (Some(url), Some(body)) == (self.time_authority_url(), self.timestamp_request_body(message)) {
           let body = body?;
           // send http request. with headers and body.
       }
    }

```

duggaraju: Actually fixed this by parsing the header bytes back to TimeStampReq and validating the nonce.

mauricefisher64:
I have not forgotten about this PR.  We are working through a few PRs before this that have similar changes.

mauricefisher64:
Can you bring this up to date with main.  Thanks.  Things are changing fast internally so need to make sure this is still accurate.

duggaraju:
rebased to latest on the main branch and resolved conflicts.

duggaraju:
Looks like wasm-tests are failing post rebase. Let me fix those. Not sure the other code coverage related failures.

duggaraju:
Fix WASM tests and issues after rebasing to latest on main. Fix clippy/fmt issues.

duggaraju:
@mauricefisher64  can you approve the workflow again.

duggaraju:
I understood the test failure. The change to AsyncSigner broke the default implementation of SyncSigner. Changed both to be identical now. @mauricefisher64 please review change to signer.rs and timestamp.rs and then you can enable the workflow again.

mauricefisher64 (approved):
Looks good

gpeacock (approved):

duggaraju:
Thanks @gpeacock for approving the changes. Let me know if anything is needed before this can be merged.

PR #404, opened 2024-02-09
Remove verify after signing when compiling without openssl

gpeacock:
This allows for a sign only build without openssl. It removes the Bogus verifier in that workflow.
Also removes the default Signer traits implementation for send_timestamp_request on wasm builds.


sdk/src/store.rs
```diff
@@ -418,70 +402,6 @@ impl Store {
         }
     }
 
-    /// Sign the claim and return signature.
```
mauricefisher64, line 421: There was actually a reason this was in store.rs and not claim.rs.  claims don't sign themselves.  The are signed by outside actors.  They do not have the global knowledge that a store does.

sdk/src/cose_sign.rs
```diff
@@ -56,33 +54,11 @@ use crate::{
     box_size: usize
 ))]
 pub fn sign_claim(claim_bytes: &[u8], signer: &dyn Signer, box_size: usize) -> Result<Vec<u8>> {
-    // Must be a valid claim.
```
mauricefisher64, line 59: This is the same refactoring that MS made in their PR.  Good simplification.

sdk/src/store.rs
gpeacock: I've reverted all the reworking of signing

mauricefisher64 (approved):
👍

PR #415, opened 2024-02-28
Trust support

mauricefisher64:
This PR brings the trust support features to the SDK.   It also ties in configuration settings to control the behavior.  Default behavior mimics the current SDK, i.e. 

- no trust checking by default
- no ocsp fetching by default
- check signature after signing by default

## Checklist


sdk/src/claim.rs
```diff
@@ -1018,10 +1019,11 @@ impl Claim {
     /// Verify claim signature, assertion store and asset hashes
     /// claim - claim to be verified
     /// asset_bytes - reference to bytes of the asset
-    pub async fn verify_claim_async<'a>(
+    pub(crate) async fn verify_claim_async<'a>(
```
gpeacock, line 1022: Since the whole Claim structure isn't public, we don't need to declare this crate private, but it doesn't hurt.

sdk/src/lib.rs
```diff
@@ -140,17 +140,18 @@ pub(crate) mod hashed_uri;
...
 pub(crate) mod time_stamp;
+pub(crate) mod trust_handler;
 pub(crate) mod utils;
 pub mod validation_status;
 pub use hash_utils::HashRange;
 pub(crate) use utils::{cbor_types, hash_utils};
 pub use utils::{cbor_types::DateT, hash_utils::hash_stream_by_alg};
 pub(crate) mod validator;
-#[cfg(target_arch = "wasm32")]
+//#[cfg(target_arch = "wasm32")]
```
gpeacock, line 154: why do we want to include wasm always? Can we remove this or add a comment?

sdk/src/claim.rs
mauricefisher64: I have started to make this explicit.  Sometimes things get exposed that we did not intend.

sdk/src/lib.rs
mauricefisher64: I don't remember making this change.  Let me look into it.

mauricefisher64: I reverted back

sdk/src/claim.rs
gpeacock: sounds good

sdk/src/cose_validator.rs
```diff
@@ -25,40 +26,43 @@ use x509_parser::{
     prelude::*,
 };
 
+#[cfg(feature = "openssl_sign")]
```
gpeacock, line 29: We should declare this under "openssl" instead of "openssl_sign" since it is for verifying, not just signing. I've been making that change in some other places.

sdk/src/cose_validator.rs
```diff
@@ -814,6 +828,76 @@ fn get_timestamp_info(sign1: &coset::CoseSign1, data: &[u8]) -> Result<TstInfo>
...
+fn check_trust(
+    th: &dyn TrustHandlerConfig,
+    chain_der: &[Vec<u8>],
+    cert_der: &[u8],
+    validation_log: &mut impl StatusTracker,
+) -> Result<bool> {
+    // is the certificate trusted
+
+    let verify_result: Result<bool> = if _sync {
+        #[cfg(not(feature = "openssl_sign"))]
```
gpeacock, line 842: change these to "openssl" so we will still have them when openssl_sign isn't required.

sdk/src/cose_validator.rs
```diff
@@ -814,6 +828,76 @@ fn get_timestamp_info(sign1: &coset::CoseSign1, data: &[u8]) -> Result<TstInfo>
...
+            if trusted {
+                let log_item =
+                    log_item!("Cose_Sign1", "signing certificate trusted", "verify_cose")
+                        .validation_status(validation_status::SIGNING_CREDENTIAL_TRUSTED);
+                validation_log.log_silent(log_item);
+                Ok(true)
+            } else {
+                let log_item =
+                    log_item!("Cose_Sign1", "signing certificate untrusted", "verify_cose")
+                        .error(Error::CoseCertUntrusted)
```
gpeacock, line 883: It seems weird having errors we don't actually ever return as errors, but only in validation. At some point we may want to put these in a separate category.

sdk/src/manifest_store.rs
```diff
@@ -11,23 +11,75 @@
...
     claim::ClaimAssetData,
+    settings,
     status_tracker::{DetailedStatusTracker, StatusTracker},
     store::Store,
     utils::base64,
     validation_status::{status_for_store, ValidationStatus},
     CAIRead, Manifest, Result,
 };
 
+fn configure_trust(store: &mut Store, options: &ManifestStoreOptions<'_>) -> Result<()> {
```
gpeacock, line 30: So will all this be replaced with settings, when we have it?

sdk/src/store.rs
```diff
@@ -113,20 +116,89 @@ impl Store {
...
+            #[cfg(all(not(feature = "openssl_sign"), target_arch = "wasm32"))]
+            trust_handler: Box::new(crate::wasm::WebTrustHandlerConfig::new()),
+            #[cfg(all(not(feature = "openssl_sign"), not(target_arch = "wasm32")))]
+            trust_handler: Box::new(crate::trust_handler::TrustPassThrough::new()),
             provenance_path: None,
-        }
+        };
+
+        // load the trust handler settings, don't worry about status as these are checked during setting generation
+        let _ = get_settings_value::<Option<String>>("trust.trust_anchors").map(|ta_opt| {
```
gpeacock, line 134: There has to be a better way to do this. This looks really odd.

sdk/tests/integration.rs
```diff
@@ -107,6 +122,10 @@ mod integration_1 {
     #[test]
     #[cfg(feature = "file_io")]
     fn test_embed_json_manifest() -> Result<()> {
+        let _protect = PROTECT.lock().unwrap();
```
gpeacock, line 125: What is this for?

gpeacock (approved):
I have some suggested changes and some questions.
I'd really rather not publish the extended ManifestStoreOptions interface instead of Settings. I don't understand why we need this. Is Settings not ready enough?

sdk/src/manifest_store.rs
```diff
@@ -228,17 +255,93 @@ impl ManifestStore {
...
+    /// let options = ManifestStoreOptions {
+    ///     data_dir: Some(Path::new("../target/tmp/manifest_store")),
+    ///     ..Default::default()
+    /// };
+    /// let manifest_store = ManifestStore::from_file_with_options("tests/fixtures/C.jpg", &options)?;
+    /// println!("{}", manifest_store);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn from_file_with_options<P: AsRef<Path>>(
```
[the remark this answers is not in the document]
mauricefisher64: We can.  Just need Verify to update.  I will back out these changes since they were for a temporary build.

sdk/src/store.rs
mauricefisher64: Do you want to pass it in?

sdk/src/cose_validator.rs
mauricefisher64: I will force this one to return upon further reflection.

PR #417, opened 2024-03-06
Fix use of deprecated method `chrono::NaiveDateTime::timestamp`

scouten-adobe:
Use `.and_utc().timestamp()` instead.


mauricefisher64 (approved):
I just made the same changes.

PR #419, opened 2024-03-07
add a thumb resource when referencing an embedded uri

gpeacock:
this gives a client a way to view the thumbnail, while still reusing the embedded one.
When using data_dir it will be written as a file


dyro (approved):
Looks clean!

PR #423, opened 2024-03-11
Update minimum dependency of rasn-* crates

scouten-adobe:
Should fix the build issue with the minimum-direct-dependencies build.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -92,9 +92,9 @@ png_pong = "0.8.2"
 rand = "0.8.5"
 rand_chacha = "0.3.1"
 range-set = "0.0.9"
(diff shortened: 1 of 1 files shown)


mauricefisher64:
Interesting.  I tried updating the RASN versions but it did not fix it for me.

vancejc-mt:
Looks like the `0.12.6` version of `rasn` got yanked (tried to build off `main` and ran into this off a fresh repo clone) - would `0.12.5` be suitable?

vancejc-mt:
> Looks like the `0.12.6` version of `rasn` got yanked (tried to build off `main` and ran into this off a fresh repo clone) - would `0.12.5` be suitable?

Just saw the new PR to roll it back, thanks guys

PR #427, opened 2024-03-14
When saving, create box hash assertion if asset handler supports box hashing

vancejc-mt:

This PR makes box hashing the default hashing assertion if an asset supports it.

In order to make box hashing the default, we modified the core `store.rs` code in the following ways:
- If an asset handler supports box hashing (`handler.asset_box_hash_ref()` is `Some`) we'll always generate a box hash assertion (otherwise we'll fall back to a normal data hash assertion)
- We removed a comparison (at the end of `start_save` and `start_save_stream`) which compared the preliminary JUMBF store size to the actual JUMBF store size.  We believe this comparison was invalid - the preliminary JUMBF store won't include a `c2pa` box if the asset has never been signed previously.  When generating the real JUMBF store the `c2pa` box has been added in, which then causes a size difference between the preliminary and actual stores.
- Bug fix - `start_save` in was incorrectly calling `pc.box_hash_assertions().is_empty()` to check for any existing hash assertion; we'll now call `pc.hash_assertions().is_empty()`

We also had a separate bug fix in `manifest.rs` unit tests (which was now exposed as jpegs default to box hashing):
- Bug fix - the `c2pa::manifest::tests::test_redaction` was incorrectly using `load_from_asset()` to create a store from the previously returned JUMBF store data (from the `.embed()`) method.  The UT requests that we do validation, which will never work when creating the store from the store data (he doesn't have the original asset to hash); this only worked when doing data hashes because he's ignoring the error which gets written to the log.  This change was required for box hashing because any error in validation for box hashes is actually returned with an `Err` result.

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -2401,7 +2401,7 @@ impl Store {
                         box_hash_handler,
                         false,
                     )?;
-                    pc.update_box_hash(box_hash)?;
+                    pc.replace_box_hash(box_hash)?;
                 }
                 // Otherwise, fall back to data hashing.
                 else {
@@ -2643,7 +2643,7 @@ impl Store {
                             box_hash_handler,
                             false,
                         )?;
-                        pc.update_box_hash(box_hash)?;
+                        pc.replace_box_hash(box_hash)?;
                     }
                     // Otherwise, fall back to data hashing.
                     else {
@@ -3198,7 +3198,7 @@ pub mod tests {
 
     use super::*;
     use crate::{
-        assertions::{labels::BOX_HASH, Action, Actions, BoxHash, Uuid},
+        assertions::{labels::BOX_HASH, Action, Actions, Uuid},
         claim::AssertionStoreJsonFormat,
         jumbf_io::{get_assetio_handler_from_path, update_file_jumbf},
         status_tracker::*,
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -26,7 +26,7 @@ use crate::{
     },
     assertions::{
         labels::{self, CLAIM},
-        DataBox, DataHash, Ingredient, Relationship,
+        BoxHash, DataBox, DataHash, Ingredient, Relationship,
     },
     asset_io::{
         CAIRead, CAIReadWrite, HashBlockObjectType, HashObjectPositions, RemoteRefEmbedType,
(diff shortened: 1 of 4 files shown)


sdk/src/store.rs
```diff
@@ -2332,38 +2332,51 @@ impl Store {
...
+            if let Some(handler) = get_assetio_handler(format) {
+                // If our asset supports box hashing, generate and add a box
+                // hash assertion.
+                if let Some(box_hash_handler) = handler.asset_box_hash_ref() {
+                    let mut box_hash = BoxHash::new();
+                    box_hash.generate_box_hash_from_stream(
+                        &mut intermediate_stream,
+                        pc.alg(),
+                        box_hash_handler,
+                        false,
```
mauricefisher64, line 2349: If you set this parameter as true it will calculate the box hash.  Since the point of box hashes is that they can be pre-calculated you should not need the replace_box_hash function.  That mean you would not code below that does the fix up.

sdk/src/store.rs
```diff
@@ -2372,34 +2385,44 @@ impl Store {
...
-                )?
-            };
 
-            // patch existing claim hash with updated data
-            for hash in updated_hashes {
-                pc.update_data_hash(hash)?;
+            if let Some(handler) = get_assetio_handler(format) {
+                // If our asset supports box hashing, generate and update
+                // the existing box hash assertion.
+                if let Some(box_hash_handler) = handler.asset_box_hash_ref() {
```
mauricefisher64, line 2396: The replace block should not be needed for the box hash case if you calculate it as stated above.

sdk/src/store.rs
```diff
@@ -2372,34 +2385,44 @@ impl Store {
...
+
+                    // patch existing claim hash with updated data
+                    for hash in updated_hashes {
+                        pc.update_data_hash(hash)?;
+                    }
+                }
             }
 
             // regenerate the jumbf because the cbor changed
             data = self.to_jumbf_internal(reserve_size)?;
```
mauricefisher64, line 2425: This check will indeed be true for the non-box hash case or else we have a problem.

sdk/src/store.rs
```diff
@@ -2558,33 +2579,51 @@ impl Store {
...
             let mut needs_hashing = false;
-            if pc.box_hash_assertions().is_empty() {
-                // 2) Get hash ranges if needed, do not generate for update manifests
-                let mut hash_ranges = object_locations(&output_path)?;
-                let hashes: Vec<DataHash> = if pc.update_manifest() {
-                    Vec::new()
-                } else {
-                    Store::generate_data_hashes(dest_path, pc.alg(), &mut hash_ranges, false)?
-                };
+            let ext = get_file_extension(&output_path).ok_or(Error::UnsupportedType)?;
```
mauricefisher64, line 2583: Same as the comments above

sdk/src/store.rs
vancejc-mt: So - we actually have some questions about the `minimal_form` functionality; if we're understanding the specification correctly, box hashes which have ranges _should_ include the data in between boxes, but otherwise data which is in between boxes shouldn't actually be included in the hash.  From the looks of the code, the `minimal_form` path actually concatenates all boxes, which would mean that data which was previously excluded (because it wasn't included in a range) would get calculated into the hash.  If this is the case, we think that the `minimal_form` behavior might be breaking the spec?

Do you know if this is not the right reasoning?

vancejc-mt: Aside from that; even if we don't send `minimal_form` the `generate_box_hash_from_stream` call still does calculate the hashes for each box, but the preliminary hash may not be correct.   For instance, if the asset wasn't signed previously, the C2PA box may not exist (and adding it may modify offsets throughout the asset) - in these cases, I think you would still need to recalculate the hashes afterward?

This behavior is what caused us to remove the size comparison at the end `start_save` and `start_save_stream` because the preliminary store and actual store had different sizes (due to the addition of the C2PA box)

sdk/src/store.rs
vancejc-mt: Okay, should we add the check back in for the normal data hash scenario?

sdk/src/store.rs
mauricefisher64: We developed box hash to cover case where the caller wants to hash independent of where the c2pa box is but the order of the content is known.  For example a camera has hardware hasher that runs while they are saving the file.  They wanted to be able to specify the hashes during their writing of the info.  We expect the use of box hash to have added the box hash assertion.  Our SDK is designed for that use case.  We have not enabled writing of box hashes yet because of the issues you highlight and more.  

minimal_form was for the case when we generate box_hashes, which we don't at this time.  It looks like there is a bug in would affect writing (if it were enabled) PNG/JPEG in our code.  The get_box_map implementations are supposed to also insert a C2PA box into the map when not present to indicate where we would write the C2PA box.  It think if you returned a C2PA box from that function then what I described above should work since that is how we intended the box hash to work, i.e. user supplies where the manifest goes among the boxes.

vancejc-mt: That makes sense - we had originally been returning a dummy C2PA box from `get_box_map` for our font implementation (as it made the JUMBF store check validate) but noticed that JPEG/PNG didn't have that behavior, and the comments didn't suggest it was expected.

So - we have a question on how we should move forward; we're actively trying to finish off the box hash assertion code in our font asset handler, and we believe this is the last issue we need to get sorted out.  So we're wondering:
  - Did you want us to modify the JPEG/PNG `get_box_map` calls to create dummy C2PA boxes as part of this PR?  Then we could put the JUMBF store check back in (theoretically) and get everything cleared out at once
  - Or.. does it make sense for us to modify the JPEG/PNG `get_box_map` code and stick it up in a separate PR?
  - Or.. do we need to contact someone else to get the JPEG/PNG code modifications, instead of us doing the modifications ourselves?

If it's easier, we can have a discussion over Discord to figure out what the best direction is, just let us know :)

I think down the road we should maybe talk more about the `minimal_form` behavior; but for now (as I think it's just an optimization) we can probably table that and just get the normal mode working as expected.

mauricefisher64: Please file a bug for fixing JPEG/PNG.  We have to make sure we actually write out the C2PA box where we said we would place it.  Let's pick up the conversation on Discord on how to move forward.

vancejc-mt: Sounds good; I'll write up the bug now and then contact you once I'm free from our morning meetings.

vancejc-mt: Created https://github.com/contentauth/c2pa-rs/issues/429 - let me know if you need more detail or information on the issue and I'd be happy to add as much as needed

mauricefisher64: Looks good.

PR #431, opened 2024-03-23
Adds Action changes field as option vec of serde_json value

gpeacock:
There are no Rust APIS for this yet.
But it allows the values to pass through via actions.
This is a minimal change to allow the new actions. changes field to be set via JSON and read back that way.
Adding detailed Rust APIs should wait until the spec for this is completed

### DIFF
--- a/sdk/src/assertions/actions.rs
+++ b/sdk/src/assertions/actions.rs
@@ -158,7 +158,8 @@ impl Action {
         matches!(
             self.software_agent,
             Some(SoftwareAgent::ClaimGeneratorInfo(_))
-        ) || self.changed.is_some() // was a String in v1 but never used so presume v2 if see it
+        ) ||
+        self.changes.is_some() // only defined for v2
     }
 
     /// Returns the label for this action.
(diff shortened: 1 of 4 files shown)


sdk/src/assertions/actions.rs
```diff
@@ -105,11 +105,17 @@ pub struct Action {
 
     /// A semicolon-delimited list of the parts of the resource that were changed since the previous event history.
     ///
+    /// This is deprecated in favor of the `changes` field.
```
dkozma, line 108: Looking at the CDDL, it looks like `changed` is still valid as an Actions V1 assertion - would that still count as "deprecated"?

dkozma (approved):
Question about a comment, but LGTM!

mauricefisher64 (approved):
👍

sdk/src/assertions/actions.rs
gpeacock: I had deprecated it, and I don't think people should use it, but I can remove this comment.

PR #435, opened 2024-03-28
Streaming write support for BMFF

mauricefisher64:
Adds streaming write support for BMFF.  Support XMP in BMFF formats.

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/bmff_io.rs
+++ b/sdk/src/asset_handlers/bmff_io.rs
@@ -1778,7 +1778,7 @@ impl RemoteRefEmbed for BmffIO {
                     (end, end)
                 };
 
-                // write content before ContentProvenanceBox
+                // write content before XMP box
                 input_stream.rewind()?;
                 let mut before_manifest = input_stream.take(start as u64);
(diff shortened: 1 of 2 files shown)


gpeacock (commented):
Thank you so much for doing this, it looks great!

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -251,7 +257,7 @@ pub(crate) struct BoxInfoLite {
     pub size: u64,
 }
 
-fn read_box_header_ext(reader: &mut dyn CAIRead) -> Result<(u8, u32)> {
+fn read_box_header_ext<R: Read + Seek + ?Sized>(reader: &mut R) -> Result<(u8, u32)> {
```
gpeacock, line 260: if only we could do this everywhere!

gpeacock (approved):
Have a great sabbatical!

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: It may be worth trying again.  I think the issue I had was was with Sync and Send.

PR #437, opened 2024-03-29
(MINOR) Initial V2 API work

gpeacock:
This adds a new Builder and Reader model based entirely on streams. 
The Builder is based on the Manifest API and the Reader on ManifestStore.
The builder is used to create new manifests to be signed. The reader is always read only.
Ingredients are added directly to the builder.
Resources such as thumbnails and icons can be accessed via Uris in resource references.
Resources in a reader are always accessed via jumbf uris. 
The state of the builder can be captured into a zip stream and reconstructed at any time.
A new CallBackSigner interface has been added which ensures the client code does not need access to private keys.

The new APIs are intended to replace the older ones. The older APIs will eventually be deprecated.
The Builder and Reader APIs will be supported in language bindings for C/C++, Python and others.


sdk/examples/v2api.rs
```diff
@@ -0,0 +1,196 @@
...
+/// This example demonstrates how to use the new v2 API to create a manifest store
+/// It uses only streaming apis, showing how to avoid file i/o
+/// This example uses the `ed25519` signing algorithm
+fn main() -> Result<()> {
+    let title = "v2_edited.jpg";
+    let format = "image/jpeg";
+    let parent_name = "CA.jpg";
+    let mut source = Cursor::new(TEST_IMAGE);
+
+    let modified_core = r#"{
```
mauricefisher64, line 76: Are you making use of the settings?

sdk/src/assertions/box_hash.rs
```diff
@@ -301,18 +301,20 @@ impl AssertionBase for BoxHash {
...
     use super::*;
+    #[cfg(test)]
     use crate::{jumbf_io::get_assetio_handler_from_path, utils::test::fixture_path};
 
     #[test]
     fn test_hash_verify_jpg() {
         let ap = fixture_path("CA.jpg");
 
-        let bhp = get_assetio_handler_from_path(&ap)
+        let bhp = crate::jumbf_io::get_assetio_handler_from_path(&ap)
```
mauricefisher64, line 317: Do you plan to expose the assert handlers or is this just a test/work-in-progress?

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -1301,7 +1302,7 @@ fn tiff_clone_with_tags<R: Read + Seek + ?Sized, W: Read + Write + Seek + ?Sized
     }
 
     tc.clone_tiff(&mut tiff_tree, page_0, asset_reader)?;
-
+    bo.flush()?;
```
mauricefisher64, line 1305: I would just the flush to the clone_tiff function so that all calls are covered.

sdk/src/store.rs
```diff
@@ -2075,6 +2077,46 @@ impl Store {
         }
     }
 
+    pub async fn save_to_stream_remote_signed(
```
mauricefisher64, line 2080: Is this an API we want to support?

sdk/src/utils/mime.rs
```diff
@@ -18,12 +18,12 @@ pub fn extension_to_mime(extension: &str) -> Option<&'static str> {
         "png" => "image/png",
         "gif" => "image/gif",
         "psd" => "image/vnd.adobe.photoshop",
-        "tiff" => "image/tiff",
+        "tiff" | "tif" => "image/tiff",
```
mauricefisher64, line 21: There are formats missing like the avi

mauricefisher64:
Looks OK to merge behind the flags.  The only things I would double check is which functions are being exposed.  There are some thing I don't think we want to support.

sdk/src/assertions/box_hash.rs
gpeacock: It is referenced this way because it isn't exposed. My plan was not even expose jumbf_io. But I may allow a couple of functions under a feature flag.

sdk/src/utils/mime.rs
gpeacock: I'll take a look

sdk/examples/v2api.rs
gpeacock: yes, I'm starting to use settings

sdk/src/store.rs
gpeacock: Not really, but I don't know how we can support adobe_c2pa without it yet.

sdk/src/utils/mime.rs
gpeacock: I added avi , video/avi mappings. Eventually I'd like to generate these mappings from the asset_io drivers somehow.

sdk/src/callback_signer.rs
```diff
@@ -0,0 +1,147 @@
...
+///
+/// The callback should return a signature for the given data.
+/// The signature should be in the format expected by the `Signer`.
+/// The callback should return an error if the data cannot be signed.
+pub type CallbackFunc = dyn Fn(*const (), &[u8]) -> std::result::Result<Vec<u8>, Error>;
+
+/// Defines a signer that uses a callback to sign data.
+///
+/// The private key should only be known by the callback.
+/// This structure is private to this module.
```
scouten-adobe, line 34: If this is true, you could remove this comment and use `pub(crate)` or `pub(super)` so that the compiler will enforce that for you.

scouten-adobe (changes requested):

sdk/src/assertions/metadata.rs
```diff
@@ -248,7 +248,7 @@ pub enum ReviewCode {
     Other(String),
 }
 
-/// A rating on an [`Assertion`].
+/// A rating on an Assertion.
```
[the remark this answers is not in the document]
gpeacock: Because this references the low level internal Assertion which the new API no longer exposes. We expose ManifestAssertion, so I could reference that here. Ideally we would rename ManifestAssertion to Assertion in the public interface, but that would have backward compatibility issues.

sdk/src/asset_handlers/c2pa_io.rs
```diff
@@ -43,6 +44,36 @@ impl CAIReader for C2paIO {
     }
 }
 
+impl CAIWriter for C2paIO {
```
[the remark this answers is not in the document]
gpeacock: Yes, I want to start by removing it from the public interface and then filter through the code.

sdk/src/callback_signer.rs
gpeacock: The comment is wrong, this is now part of the public interface. I'll remove the comment.

gpeacock: Those comments are out of date. The intention is to have this be a standard part of the API. The various bindings depend on this, so it makes sense to share it here.

sdk/src/v2_api/builder.rs
```diff
@@ -0,0 +1,1150 @@
...
+
+    //#[allow(clippy::unwrap_used)]
+    /// Unzip a stream into a Builder
+    /// # Arguments
+    /// * `stream` - A stream to read the zip from
+    /// # Returns
+    /// * A new Builder
+    /// # Errors
+    /// * If the zip cannot be read
+    pub fn unzip(stream: impl Read + Seek) -> Result<Self> {
```
dkozma, line 334: Do you think these functions would be clearer/more consistent if they were called something like `to_archive` and `from_archive`?

sdk/src/v2_api/reader.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// * `format` - The format of the stream
+    /// * `stream` - The stream to verify the store against
+    /// # Returns
+    /// A reader for the manifest store
+    /// # Errors
+    /// If the c2pa_data is not valid, or severe errors occur in validation
+    /// validation status should be checked for non severe errors
+    #[async_generic(async_signature(
+        c2pa_data: &[u8],
+        format: &str,
```
dkozma, line 154: I noticed that `from_manifest_data_and_stream` is `(c2pa_data, format)` but `from_bytes` is `(format, bytes)` - IMO it would be good to always have `format` first or always have it after the input data argument to stay consistent.

sdk/src/v2_api/reader.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// # Example
+    /// ```no_run
+    /// use c2pa::Reader;
+    /// let stream = std::io::Cursor::new(include_bytes!("../../tests/fixtures/CA.jpg"));
+    /// let reader = Reader::from_stream("image/jpeg", stream).unwrap();
+    /// let status = reader.validation_status();
+    /// ```
+    /// # Note
+    /// The validation status should be checked for validation errors`
+    pub fn validation_status(&self) -> Option<&[ValidationStatus]> {
```
dkozma, line 204: Should we mention that we only report error codes here? The C2PA spec considers validation status to be made up of both success or error codes.

Related to this, I had to fix a bug in Verify where someone was dragging an asset that had an ingredient with a `validationStatus` showing both error and success codes (which is valid I believe), and since Verify is used to that just being error codes, reported an error since the array wasn't empty. This is all to say people may not realize they will only get failure codes here.

sdk/src/v2_api/reader.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub fn active_label(&self) -> Option<&str> {
+        self.manifest_store.active_label()
+    }
+
+    /// Return a Manifest for a given label
+    /// # Arguments
+    /// * `label` - The label of the manifest to return
+    /// # Returns
+    /// The manifest if it exists
+    pub fn get(&self, label: &str) -> Option<&Manifest> {
```
dkozma, line 225: Since this is on an `impl` called `Reader` should this be `get_manifest`, `manifest` (maybe more Rusty), or `manifest_by_label`?

From the [_Rust API Guidelines_](https://rust-lang.github.io/api-guidelines/naming.html#c-getter):
>The `get` naming is used only when there is a single and obvious thing that could reasonably be gotten by a getter. For example [`Cell::get`](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.get) accesses the content of a `Cell`.

I'm not sure if it's totally obvious that calling `get` on a `Reader` returns a `Manifest` (could it be a Manifest Store?), but if it is, I'm ok with leaving as is.

sdk/tests/common/compare_readers.rs
```diff
@@ -0,0 +1,185 @@
+// Copyright 2024 Adobe. All rights reserved.
```
dkozma, line 1: Is this a utility? I wonder if this should go in `sdk/tests/utils` to be more clear.

sdk/tests/common/test_signer.rs
```diff
@@ -0,0 +1,26 @@
+use c2pa::{CallbackSigner, SigningAlg};
```
dkozma, line 1: This file looks like a great example to call out for how to make your own `CallbackSigner`.

dkozma (commented):
This is looking really good - I feel like I'll have more thoughts once I start using `Reader` and `Builder`.

sdk/src/v2_api/reader.rs
gpeacock: That's because the format in this case applies to the stream, not the c2pa_data. So format still comes first. I suppose I could make c2pa_data the last parameter if that would help.

sdk/src/assertions/box_hash.rs
```diff
@@ -301,18 +301,20 @@ impl AssertionBase for BoxHash {
     }
 }
 
+#[cfg(feature = "file_io")]
 #[cfg(test)]
 mod tests {
     #![allow(clippy::unwrap_used)]
 
     use super::*;
+    #[cfg(test)]
```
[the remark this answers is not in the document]
gpeacock: Not something I changed, but I can fix it.

sdk/src/v2_api/builder.rs
gpeacock: yes I probably should not use the term "zip" for these. I'll change that.

sdk/src/callback_signer.rs
```diff
@@ -0,0 +1,147 @@
...
+        (self.callback)(self.context, data)
+    }
+
+    fn alg(&self) -> SigningAlg {
+        self.alg
+    }
+
+    fn certs(&self) -> Result<Vec<Vec<u8>>> {
+        let mut pems = pem::parse_many(&self.certs).map_err(|e| Error::OtherError(Box::new(e)))?;
+        Ok(pems.drain(..).map(|p| p.into_contents()).collect())
```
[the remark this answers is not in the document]
gpeacock: No particular reason, into_iter makes more sense in this context, I'll change it.

sdk/src/callback_signer.rs
```diff
@@ -0,0 +1,147 @@
...
+pub type CallbackFunc = dyn Fn(*const (), &[u8]) -> std::result::Result<Vec<u8>, Error>;
+
+/// Defines a signer that uses a callback to sign data.
+///
+/// The private key should only be known by the callback.
+/// This structure is private to this module.
+/// Should only be created using the `create_callback_signer` function.
+pub struct CallbackSigner {
+    /// An opaque context for the signer, used to store any necessary state
+    pub context: *const (),
```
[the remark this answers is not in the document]
gpeacock: Yes, but a pointer to something potentially not in the Rust memory space. It could be allocated in C or Python.

sdk/src/v2_api/reader.rs
gpeacock: Yes it should explain that in the comments. Note that this validation_status is not on an ingredient, but on a manifest_store. Maybe we should have a configuration option to keep the valid status information, but it will tend to bloat the ingredients.

sdk/src/v2_api/reader.rs
gpeacock: Well it used to be called ManifestStore, and I suppose it made more sense that way. But the Reader is a ManifestStore reader.

sdk/src/v2_api/V2_API_NOTES.md
```diff
@@ -0,0 +1,70 @@
+## V2 API Notes
```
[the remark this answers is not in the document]
gpeacock: Everything we do here is open source, including these comments. I think these notes are generally useful, and I'm not really sure where they would go other than in the repo.

sdk/src/manifest_store.rs
```diff
@@ -76,6 +82,30 @@ impl ManifestStore {
         self.manifests.get(label)
     }
 
+    // writes a resource identified uri to the given stream
+    pub fn get_resource(&self, uri: &str, stream: impl Write + Read + Seek + Send) -> Result<u64> {
```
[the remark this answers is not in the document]
gpeacock: This would only work with a jumbf uri from the manifest store. It is meant for retrieving resources from the store. I could add some error checking for that. As it is I think you would just get a Not Found result.

sdk/tests/common/compare_readers.rs
gpeacock: Integration tests don't have access to unit test code. So I either need a separate copy or we would need to make some test code public.

sdk/Cargo.toml
```diff
@@ -51,7 +52,8 @@ name = "show"
 required-features = ["file_io"]
 
 [[example]]
-name = "custom_assertion"
+name = "data_hash"
```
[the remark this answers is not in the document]
gpeacock: Just a placeholder in case we need extra features added for this example later. I could remove it.

sdk/Cargo.toml
```diff
@@ -107,16 +110,18 @@ serde = { version = "1.0.197", features = ["derive"] }
 serde_bytes = "0.11.5"
 serde_cbor = "0.11.1"
 serde_derive = "1.0.197"
-serde_json = "1.0.114"
+serde_json = { version = "1.0.114", features = ["preserve_order"] }
+serde_with = "3.4.0"
 serde-transcode = "1.1.1"
 sha2 = "0.10.2"
-tempfile = "3.1.0"
+tempfile = "3.10.1"
```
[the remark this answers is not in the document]
gpeacock: That's a good callout. I looked into it, but it's a PR on it's own. So I can follow up with that.

sdk/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "c2pa"
+name="c2pa"
```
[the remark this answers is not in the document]
gpeacock: I have that installed as well. The better answer would be to catch it in CI

sdk/examples/v2api.rs
```diff
@@ -0,0 +1,195 @@
...
```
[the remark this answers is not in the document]
gpeacock: done

Signing Scenarios.txt
```diff
@@ -0,0 +1,55 @@
+Signing Scenarios
```
dkozma, line 1: Did you mean to put this in the root of the repo?

sdk/src/callback_signer.rs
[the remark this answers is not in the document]
dkozma: I see this is the same, but I don't feel strongly about it, so no worries - this was more for my knowledge.

dkozma (approved):
This looks great, awesome work! The only issue I have that would be good to be addressed is if you meant to put `Signing Scenarios.txt` in the root folder, since that seems like it might have been a mistake?

Signing Scenarios.txt
gpeacock: Thanks, I didn't mean to check that in,.

sdk/examples/custom_assertion.rs
```diff
@@ -1,89 +0,0 @@
-// Copyright 2022 Adobe. All rights reserved.
```
scouten-adobe, line 1: Not a blocker for this PR, but let's make sure that there's an example for somebody building something like the [identity assertion](https://github.com/scouten-adobe/cawg-identity-core/) to build upon. (Maybe that will be a canonical example?)

sdk/examples/data_hash.rs
```diff
@@ -148,7 +148,7 @@ fn user_data_hash_with_sdk_hashing() {
     output_file.write_all(&final_manifest).unwrap();
 
     // make sure the output file is correct
-    let manifest_store = ManifestStore::from_file(&dest).unwrap();
```
scouten-adobe, line 151: This shift from `ManifestStore` to `Reader` seems odd. What is the semantic intent here?

sdk/src/builder.rs
```diff
@@ -0,0 +1,1105 @@
+// Copyright 2023 Adobe. All rights reserved.
```
scouten-adobe, line 1: 2024

sdk/src/builder.rs
```diff
@@ -0,0 +1,1105 @@
...
+    /// Convert the Builder into a archive formatted stream.
+    ///
+    /// The archive is a zip formatted stream containing the manifest.json, resources, and ingredients.
+    /// # Arguments
+    /// * `stream` - A stream to write the zip into.
+    /// # Errors
+    /// * If the archive cannot be written.
+    pub fn to_archive(&mut self, stream: impl Write + Seek) -> Result<()> {
+        drop(
+            // this drop seems to be required to force a flush before reading back.
```
scouten-adobe, line 358: That seems odd.

sdk/src/builder.rs
```diff
@@ -0,0 +1,1105 @@
...
+                }
+                builder.definition.ingredients[index]
+                    .resources_mut()
+                    .add(id, data)?;
+            }
+        }
+        Ok(builder)
+    }
+
+    // Convert a Manifest into a Claim
```
scouten-adobe, line 449: --- continuing review from here ---

sdk/examples/data_hash.rs
gpeacock: At first I was going to update the whole code base to use Reader instead of ManifestStore and remove ManifestStore entirely. I changed course and decided to leave the old APIs in for now. This is one place I had updated to Reader. I can change it back for now if you prefer, but it will become Reader eventually.

sdk/examples/custom_assertion.rs
gpeacock: Yes, I'm working on updated examples for CBOR and JSON assertions, and it would be a good idea to add an identity example!

sdk/src/builder.rs
```diff
@@ -0,0 +1,1105 @@
...
+    /// # Errors
+    /// * If the manifest cannot be signed.
+    #[async_generic(async_signature(
+        &mut self,
+        format: &str,
+        source: &mut R,
+        dest: &mut W,
+        signer: &dyn AsyncSigner
+    ))]
+    pub fn sign<R, W>(
```
scouten-adobe, line 694: Perhaps this method name should be `sign_and_embed`? The name `sign` by itself doesn't convey the "embed" part of what you wrote in the doc comment above.

sdk/src/builder.rs
```diff
@@ -0,0 +1,1105 @@
...
+                .resources
+                .get(&builder.definition.thumbnail.unwrap().identifier)
+                .unwrap()
+                .into_owned(),
+            b"12345"
+        );
+    }
+
+    #[test]
+    fn test_from_json() {
```
scouten-adobe, line 889: Continuing from here in a little while …

sdk/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// # Example
+    /// ```no_run
+    /// use std::io::Cursor;
+    ///
+    /// use c2pa::Reader;
+    /// let mut stream = Cursor::new(include_bytes!("../tests/fixtures/CA.jpg"));
+    /// let reader = Reader::from_stream("image/jpeg", stream).unwrap();
+    /// println!("{}", reader.json());
+    /// ```
+    #[async_generic(async_signature(
```
scouten-adobe, line 56: I don't think you need `async_signature` here since the sync and async signatures are the same.l

sdk/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// # Arguments
+    /// * `c2pa_data` - The c2pa data (a manifest store in JUMBF format)
+    /// * `format` - The format of the stream
+    /// * `stream` - The stream to verify the store against
+    /// # Returns
+    /// A [`Reader`] for the manifest store
+    /// # Errors
+    /// If the c2pa_data is not valid, or severe errors occur in validation
+    /// validation status should be checked for non severe errors
+    #[async_generic(async_signature(
```
scouten-adobe, line 129: Same here about `async_signature`.

sdk/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+    validation_status::ValidationStatus, Manifest,
+};
+
+/// A reader for the manifest store.
+#[derive(Debug)]
+pub struct Reader {
+    pub(crate) manifest_store: ManifestStore,
+}
+
+impl Reader {
```
scouten-adobe, line 37: Not sure _what_ to suggest, but I'm not sure `Reader` by itself is a sufficient name. Perhaps `ManifestStoreReader`?

sdk/tests/common/mod.rs
```diff
@@ -0,0 +1,99 @@
+mod compare_readers;
```
scouten-adobe, line 1: Missing copyright headers

sdk/tests/common/test_signer.rs
scouten-adobe, line 1: Missing copyright headers

sdk/tests/test_builder.rs
```diff
@@ -0,0 +1,26 @@
+use std::io::Cursor;
```
scouten-adobe, line 1: Missing copyright headers

scouten-adobe: (also in remaining files in this PR)

scouten-adobe (approved):
:+1: after considering changes in latest review

sdk/tests/common/mod.rs
gpeacock: good catch, thanks

sdk/src/builder.rs
gpeacock: That's why I commented about it :)

sdk/examples/data_hash.rs
gpeacock: I reverted this for now to be more consistent as an example.

gpeacock:
Forcing merge after all reviews passed due to MINOR VERSION check being stuck.

PR #460, opened 2024-04-30
Gpeacock/embed_remote_settings

gpeacock:
Adds a way to set the remote_url and no_embed option for Builder.
Enables BMFF format for streams.
Adds io and streaming tests.

### DIFF
--- a/V2_API_NOTES.md
+++ b/V2_API_NOTES.md
@@ -50,6 +50,19 @@ The source asset is the asset that we will hash and sign. It can the output from
 
 If there is no parent ingredient defined, and the source has a manifest store, the sdk will generate a parent ingredient from the parent.
 
+### Remote URLs and embedding
+The default operation of c2pa signing is to embed a c2pa manifest store into an asset.
+We also return the c2pa manifest store so that it can be written to a sidecar or uploaded to a remote service.
+- The API supports embedding a a remote url reference into the asset. 
+- The remote URL is stored in different ways depending on the asset, but is often stored in XMP data.
+- The remote URL must be added to the asset before signing so that it can be hashed along with the asset.
+- Not all file formats support embedding remote URLs or embedding manifests stores.
+- If you embed a manifest or a remote URL, a new asset will be created with the new data embedded.
+- If you don't embed, then the original asset is unmodified and there is no need to write one out.
+- The remote url can be set with builder.remote_url.
+- If embedding is not needed, set the builder.no_embed flag to true.
+
+
 ## Testing
 We need a more comprehensive set of tests for the rust codebase.
 
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -17,7 +17,7 @@ use std::{
     fmt, fs,
     io::{BufReader, Cursor, SeekFrom},
     ops::Deref,
-    path::{Path, PathBuf},
+    path::Path,
 };
 
 use mp4::*;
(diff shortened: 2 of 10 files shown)


V2_API_NOTES.md
```diff
@@ -50,6 +50,19 @@ The source asset is the asset that we will hash and sign. It can the output from
 
 If there is no parent ingredient defined, and the source has a manifest store, the sdk will generate a parent ingredient from the parent.
 
+### Remote URLs and embedding
+The default operation of c2pa signing is to embed a c2pa manifest store into an asset.
+We also return the c2pa manifest store so that it can be written to a sidecar or uploaded to a remote service.
+- The API supports embedding a a remote url reference into the asset.
```
dyro, line 56: extra 'a' here

sdk/src/assertions/bmff_hash.rs
```diff
@@ -356,11 +354,10 @@ impl BmffHash {
         let bmff_exclusions = &self.exclusions;
 
         // convert BMFF exclusion map to flat exclusion list
-        let mut data = fs::File::open(asset_path)?;
-        let exclusions =
-            bmff_to_jumbf_exclusions(&mut data, bmff_exclusions, self.bmff_version > 1)?;
+        //let mut data = fs::File::open(asset_path)?;
```
dyro, line 357: Should this be removed?

sdk/src/builder.rs
```diff
@@ -193,6 +193,12 @@ pub struct Builder {
     #[serde(flatten)]
     pub definition: ManifestDefinition,
 
+    /// Optional remote URL for the manifest
+    pub remote_url: Option<String>,
+
+    // If true, the manifest store will not be embedded in the asset on sign
+    pub no_embed: bool,
+
```
dyro, line 201: Seems like we'd eventually get to the `DestOption` bitmask version of this. Should we do that now? I know we don't have the sidecar APIs figured out yet, so maybe this is a problem to solve later on?

sdk/src/claim.rs
```diff
@@ -938,15 +938,13 @@ impl Claim {
     }
 
     // Crate private function to allow for patching a BMFF hash with final contents.
-    #[cfg(feature = "file_io")]
```
dyro, line 941: :zap:

sdk/src/claim.rs
```diff
@@ -938,15 +938,13 @@ impl Claim {
...
     pub(crate) fn update_bmff_hash(&mut self, bmff_hash: BmffHash) -> Result<()> {
         self.replace_assertion(bmff_hash.to_assertion()?)
     }
 
     // Patch an existing assertion with new contents.
     //
     // `replace_with` should match in name and size of an existing assertion.
-    #[cfg(feature = "file_io")]
```
dyro, line 949: :zap:

sdk/src/jumbf_io.rs
```diff
@@ -406,4 +407,46 @@ pub mod tests {
...
+            let handler = get_caiwriter_handler(asset_type).unwrap();
+            let mut removed = Cursor::new(Vec::new());
+            handler
+                .remove_cai_store_from_stream(&mut writer, &mut removed)
+                .unwrap();
+            removed.set_position(0);
+            let result = load_jumbf_from_stream(asset_type, &mut removed);
+            assert!(matches!(result.err().unwrap(), Error::JumbfNotFound));
+        }
+    }
```
dyro, line 451: 👍 Nice to have a test for this

dyro (approved):
Looks good! Couple of quick questions / suggestions

sdk/src/builder.rs
gpeacock: DestOption doesn't work here because we need to specify the remote_url. I'm not sure this API is final. Things are different with streams and I'm working on that. Basically you always get signed c2pa_manifest you can write to the cloud or a sidecar. But the output file may need to have a url, a manifest, or both, and we also have to strip out manifests and urls when needed.

PR #464, opened 2024-05-04
(Minor) Additional unit tests and fixes for injection (or previously untested) issues.

gpeacock:
Note: on some formats c2pa manifests fail to be removed.
On others XMP write is supposed to be implemented but read does not work.
These errors are bypassed for now but should be followed up with issues

### DIFF
--- a/sdk/src/error.rs
+++ b/sdk/src/error.rs
@@ -216,10 +216,10 @@ pub enum Error {
     ResourceNotFound(String),
 
     #[error("XMP read error")]
-    XmpReadError,
+    XmpReadError(String),
 
     #[error("XMP write error")]
-    XmpWriteError,
+    XmpWriteError(String),
 
     #[error("XMP is not supported")]
     XmpNotSupported,
--- a/sdk/src/jumbf_io.rs
+++ b/sdk/src/jumbf_io.rs
@@ -317,8 +317,13 @@ pub mod tests {
     #![allow(clippy::panic)]
     #![allow(clippy::unwrap_used)]
 
+    use std::io::Seek;
+
     use super::*;
-    use crate::utils::test::{create_test_store, temp_signer};
+    use crate::{
+        asset_io::RemoteRefEmbedType,
+        utils::test::{create_test_store, temp_signer},
+    };
 
     #[test]
     fn test_get_assetio() {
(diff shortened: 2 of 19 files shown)


vancejc-mt:
Question here - would it be possible to make the `add_xmp_key` function public?  We're getting off the `xmp_toolkit` with the recent pull of `main`, but these utilities don't have a way for us to add in `DocumentID`, `InstanceID`, or their related namespaces.

sdk/src/error.rs
```diff
@@ -216,10 +216,10 @@ pub enum Error {
     ResourceNotFound(String),
 
     #[error("XMP read error")]
-    XmpReadError,
+    XmpReadError(String),
```
dyro, line 219: 👍

sdk/src/jumbf_io.rs
```diff
@@ -408,45 +413,146 @@ pub mod tests {
...
+        //reader.rewind().unwrap();
+        //test_remote_ref("mp3", &mut reader); // not working
+    }
+
+    #[test]
+    fn test_streams_avif() {
+        let mut reader = std::fs::File::open("tests/fixtures/sample1.avif").unwrap();
+        test_jumbf("avif", &mut reader);
+        //reader.rewind().unwrap();
+        //test_remote_ref("avif", &mut reader);  // not working
```
dyro, line 528: One thing you can do is mark this test as _ignored_ - that way the unit test output will show that we ignored this test, instead of commenting out the unsupported assertion. Maybe separate the assertions _test_jumbf_ and _test_remote_ref_ into two test cases and ignore the unsupported part? 

```rust
#[test]
#[ignore]
fn test_streams_remote_ref() {
    let mut reader = std::fs::File::open("tests/fixtures/sample1.avif").unwrap();
    reader.rewind().unwrap();
    test_remote_ref("avif", &mut reader);  // not working
}
```

dyro (approved):
Looks clean. I left one suggestion on test organization, but you're welcome to ignore it to get a release out.

sdk/src/jumbf_io.rs
gpeacock: The better thing would be to fix the code to support it as it should. I tried that and ran out of time. So I'll continute with this as is and get back to it.

PR #471, opened 2024-05-24
fix: Don't take ownership of data when using `AsyncSigner`; use a reference instead

brogdonm:

During use of the `AsyncSigner` trait, we noticed the `sign` method has a different signature than the simple `Signer` trait. The current implementation passed ownership of the data to the signer instead of passing a reference.

## Checklist


brogdonm:
@mauricefisher64 @gpeacock The vulnerability failure appears after merging the latest `main` into this branch.

sdk/src/signer.rs
```diff
@@ -126,7 +126,7 @@ use async_trait::async_trait;
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 pub trait AsyncSigner: Sync {
     /// Returns a new byte array which is a signature over the original.
-    async fn sign(&self, data: Vec<u8>) -> Result<Vec<u8>>;
+    async fn sign(&self, data: &[u8]) -> Result<Vec<u8>>;
```
duggaraju, line 129: This would require that the data would have to implement Sync and Send potentially since the async function execution could spawn across threads.  In wasm this might complicate the implementation.  Does wasm-pack test work fine with this change?

brogdonm: As far as I can tell, all tests pass in the GitHub actions.

gpeacock: I agree this feels like an unsafe change due to the asynchronous function. The tests may be passing because lifetime can be determined by the complier, but there are likely cases where they would not.

duggaraju: At the least this function signature would require lifetime for the data.  And it might also require that data also implement Sync + Send to work correctly

brogdonm:
I'll try to take a look at this tomorrow and fix up where I can.

PR #473, opened 2024-05-30
Implement `Serialize` and custom `Debug` for `Reader`

ok-nick:
* Add owned `Store` field to `ManifestStore` to open access to all `Store` internal methods
* Manual implementation of `Debug` for `Reader`, which provides the user with detailed debug information about the manifest (effectively exposing `ManifestStoreReport` in the unstable API)
* Implement `Serialize` on `Reader` so that the user can serialize the manifest store into any `serde`-supported format (including json)

The `Reader` already implements `Display`, which provides user-friendly information about the manifest (whereas `Debug` uses the more explicit/detailed `ManifestStoreReport`).

One concern is exposing things like `ManifestStoreReport::dump_tree` and `ManifestStoreReport::cert_chain`, which provides short and concise debug information. Should this info be included in the `Debug` impl?

Additionally, should `Reader::json` provide the "simplified" json or the explicit json?

## Checklist

### DIFF
--- a/sdk/src/manifest_store.rs
+++ b/sdk/src/manifest_store.rs
@@ -45,6 +45,9 @@ pub struct ManifestStore {
     #[serde(skip_serializing_if = "Option::is_none")]
     /// ValidationStatus generated when loading the ManifestStore from an asset
     validation_status: Option<Vec<ValidationStatus>>,
+    #[serde(skip)]
+    /// The internal store representing the manifest store
+    store: Store,
 }
 
 impl ManifestStore {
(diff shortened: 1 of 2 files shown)


ok-nick:
Adding `Store` to `ManifestStore` causes a breaking change since `ManifestStore` no longer implements `UnwindSafe` nor `RefUnwindSafe`. This is because `Store` contains a trait object, which cannot guarantee unwind safety unless the trait object requires the traits mentioned or we mask it with `AssertUnwindSafe`. See https://github.com/rust-lang/rust/issues/65717 and https://github.com/rust-lang/rust/issues/65717.

ok-nick:
For now we require `TrustHandlerConfig` to be `UnwindSafe` and `RefUnwindSafe`, but ideally they'd either be (1) a generic on `Store` or (2) each impl is gated behind a feature flag for each platform.

mauricefisher64 (approved):
👍

PR #479, opened 2024-06-06
fix: No need in the `Builder::add_resource` taking ownership of the resource stream; use reference instead

brogdonm:

Previously, the `Builder::add_resource` method takes ownership of the resource stream. This is unnecessary and taking the stream as a reference allows for the stream to be used over and over again between calls (especially in long running processes).

The method also required the stream to be seekable, but this isn't necessary either as the stream is read into a memory buffer.

## Checklist


brogdonm:
@mauricefisher64 @gpeacock The vulnerability failure appears after merging the latest `main` into this branch.

PR #482, opened 2024-06-13
Add iterators over manifests and resources in unstable API

ok-nick:
While migrating `c2patool` to the unstable API, I ran across an issue where it's not possible to iterate unknown manifests and parse the format from a resource. Previously, `Manifest::from_file_with_resources` was used to extract resources, however, this has two downsides: (1) we have no control over how they are stored on the filesystem and (2) it only extracts known resources (with standardized labels, like thumbnails). 

So, we introduce two new methods:
* Added `Reader::iter_manifests`
* Added `Manifest::iter_resources`

The first one allows us to iterate all manifests (even if we don't know the label) and the second one allows us to iterate all resources (including unknown ones) into a `ResourceRef`, which provides us with a mime type (format) and uri (identifier).

## Checklist


sdk/src/manifest.rs
```diff
@@ -451,6 +451,14 @@ impl Manifest {
         self.signature_info.to_owned().and_then(|sig| sig.time)
     }
 
+    /// Returns an iterator over [`ResourceRef`][ResourceRef]s.
+    pub fn iter_resources(&self) -> impl Iterator<Item = ResourceRef> + '_ {
```
mauricefisher64, line 455: could the name here just be "resources"

sdk/src/reader.rs
```diff
@@ -187,6 +187,11 @@ impl Reader {
         self.manifest_store.active_label()
     }
 
+    /// Returns an iterator over [`Manifest`][Manifest]s.
```
mauricefisher64, line 190: Same as above, could name just be "manifests"

sdk/src/manifest.rs
ok-nick: There already exists a method named `Manifest::resources` which returns a `ResourceStore`. I think the intention with the unstable API is to phase out use of resource stores in favor of methods like  `Reader::resource_to_stream`? Resource store also has a method named resources already, so we can't move it there without a breaking change.

mauricefisher64: Ok

mauricefisher64 (approved):
👍

PR #483, opened 2024-06-13
url crate version 2.5.1 introduces new license "Unicode-3.0"

scouten-adobe:
We need legal review on this license before we can accept it.

Related to https://github.com/servo/rust-url/pull/923


mauricefisher64 (approved):
👍

PR #472, opened 2024-05-29
External placed manifest

mauricefisher64:
## Changes in this pull request
Support for external placed manifests (CAWG) work.  
Fixed implementations of BMFF streaming

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #491, opened 2024-06-25
Let's not assume that third-party assertions are using serde_cbor

scouten-adobe:
This was in an earlier version of #472 but seems to have been lost when it was merged.

CAWG identity assertion project needs this.


mauricefisher64 (approved):
👍

PR #494, opened 2024-06-28
(PATCH) ensures temp files are removed

dyro:

This PR changes the behavior of the `asset_io::rename_or_copy` function by deleting the copied file after it has been copied to the destination path.

This mainly affected our windows clients as the `std::fs::rename` fails when renaming a file from one drive to another drive on windows.

The new behavior is the following:

1. Atttempt to rename the file
2. If that fails, copy the file
3. If the copy succeeds, delete the temporary file.

## Checklist


mauricefisher64 (approved):
👍

PR #498, opened 2024-07-08
Temporarily allow unused `JsonAssertionData` to fix unused error in CI

ok-nick:
Temporarily comment `JsonAssertionData` to fix unused error in CI.

## Checklist

### DIFF
--- a/sdk/src/assertion.rs
+++ b/sdk/src/assertion.rs
@@ -471,12 +471,12 @@ impl Assertion {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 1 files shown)


sdk/src/assertion.rs
```diff
@@ -471,12 +471,12 @@ impl Assertion {
     }
 }
```
mauricefisher64, line 473: Maybe add just an unused tag so the code is still available.  I think this is used by another PR somewhere.

mauricefisher64 (approved):
👍

PR #499, opened 2024-07-08
feat: Add ZIP support (+ EPUB, Office Open XML, Open Document, and OpenXPS) and Collection Data Hash assertion

### LINKED ISSUE #349: Add support for ZIP-based formats
Feature request from [Discord](https://discord.com/channels/983153151341371422/1129423756964659201/1178772273579180042):

C2PA 1.4 has added support for [ZIP-based formats](https://c2pa.org/specifications/specifications/1.4/specs/C2PA_Specification.html#_embedding_manifests_into_zip_based_formats) - this is a request to add compatibility with this to c2pa-rs and downstream tools.

### LINKED ISSUE #406: Zip format support

### LINKED ISSUE #487: Support for Collections Assertion
Collections Assertions allow for auxiliary files to be validated as part of the manifest.  This task is support the generation and validation of collections.

ok-nick:
Adds ZIP support (and by extension EPUB, Office Open XML, Open Document, and OpenXPS support) and the corresponding Collection Data Hash assertion.

### Related Issues
* Completes #487 
* Completes #406 
* Completes #349 

## Checklist


sdk/src/assertions/collection_hash.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub size: Option<u64>,
+
+    #[serde(rename = "dc:format", skip_serializing_if = "Option::is_none")]
+    pub dc_format: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub data_types: Option<Vec<AssetType>>,
+}
+
+impl CollectionHash {
```
mauricefisher64, line 35: Collection hashes are not just for ZIPs.  This needs to be thought through on how to make this work in general.  I think there would need to be a new API to allow users to set the list of included external files or in the case of the zip something like this would happen automatically.

ok-nick: I was thinking about that, I'll probably split the collection assertion off into a separate PR. Maybe we can allow the user to specify their own "URI resolver" of some such to fetch inclusion ranges. Then for ZIP, we can offer a predefined resolver.

sdk/src/assertions/collection_hash.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn add_uri_map(&mut self, uri_map: UriHashedDataMap) {
+        self.uris.push(uri_map);
+    }
+
+    // The base path MUST be the folder of the manifest. A URI MUST NOT reference a path outside of that folder.
+    pub fn gen_hash<R>(&mut self, base_path: &Path) -> Result<()>
```
mauricefisher64, line 52: I think base_path should be part of the constructor and this function should just calc the hashes when requested.

sdk/src/asset_handlers/zip_io.rs
```diff
@@ -0,0 +1,380 @@
...
+                dc_format: None,
+                data_types: None,
+            });
+        }
+    }
+
+    Ok(uri_maps)
+}
+
+pub fn uri_inclusions<R>(stream: &mut R, uri_maps: &[UriHashedDataMap]) -> Result<Vec<HashRange>>
```
mauricefisher64, line 293: The hash is per each individual files compressed content.  I don't this this function is needed.  See my suggestions for uri_maps function.

sdk/src/asset_handlers/zip_io.rs
```diff
@@ -0,0 +1,380 @@
...
+            .by_name(&file_name)
+            .map_err(|_| Error::JumbfNotFound)?;
+
+        if !file.is_dir() {
+            uri_maps.push(UriHashedDataMap {
+                // TODO: temp unwrap
+                #[allow(clippy::unwrap_used)]
+                uri: file.enclosed_name().unwrap(),
+                hash: Vec::new(),
+                size: Some((file.data_start() + file.compressed_size()) - file.header_start()),
```
mauricefisher64, line 283: The size should only be for the compressed data.

sdk/src/asset_handlers/zip_io.rs
```diff
@@ -0,0 +1,380 @@
...
+                dc_format: None,
```
mauricefisher64, line 284: These should be filled in

sdk/src/assertions/collection_hash.rs
ok-nick: The problem is that base_path is only used when the assertion is for non-zips. We can have a constructor that takes in a path, something like with_base_path, but then we'd have to store it optionally. If the assertion is for a non-zip and constructed without a base path, then it's already invalid.

ok-nick: Thinking about it again, we should only have public methods to construct a collection hash with a base path. When using collection hash for zips, user's should never need to construct a collection hash explicitly.

sdk/src/asset_handlers/zip_io.rs
ok-nick: Pending spec revision.

ok-nick:
Temporarily using forked repo until https://github.com/zip-rs/zip2/pull/232 is merged

scouten-adobe:
Created CAI-12644 to track this.

sdk/src/store.rs
mauricefisher64: Should there be a check here for the zip care

ok-nick: Yes, good catch.

mauricefisher64 (approved):

sdk/src/store.rs
```diff
@@ -3324,6 +3314,81 @@ impl Store {
...
+                return Err(Error::BadParam(
+                    "update manifests are not supported for ZIP-based assets".to_string(),
+                ));
+            }
+
+            // hash all of the existing files and embed a placeholder central directory hash.
+            // this is okay because adding new file entries does not change the hashes of existing
+            // file entries. when we insert the real manifest we can go back and hash the central directory.
+            let mut new_collection_hash = None;
+            if pc.hash_assertions().is_empty() {
```
scouten-adobe:
```rust
if pc.hash_assertions().is_empty() {
```
`hash_assertions()` now aggregates DataHash, BmffHash, BoxHash, *and* CollectionHash. The BMFF branch above uses the type-specific `pc.bmff_hash_assertions()` for the equivalent check — should this use the equally type-specific `pc.collection_hash_assertions()` (added in this PR at `claim.rs:3919`) instead? As written, if the claim already carries any other hash-binding assertion when signing a ZIP asset, this whole block is skipped and no CollectionHash gets generated at all. Worth double-checking, since `collection_hash_assertions()` currently has no callers anywhere in the tree.

sdk/src/assertions/collection_hash.rs
```diff
@@ -0,0 +1,660 @@
...
+            )?);
+        }
+
+        Ok(())
+    }
+
+    /// Validate the hashes for the files in the collection.
+    ///
+    /// The base path is the directory that each URI is resolved against.
+    pub fn verify_hash(&self, base_path: &Path) -> Result<()> {
```
scouten-adobe, line 167: `verify_hash` only iterates `self.uris` and checks each listed file exists with a matching hash — it never enumerates `base_path`'s actual contents. The ZIP path has a built-in backstop against this (the central-directory hash changes if entries are added), but this directory-based path doesn't have an equivalent, so an extra file dropped into `base_path` that isn't in the signed `uris` list would go undetected. Is that intentional for this asset shape, or should there be an "extra/unaccounted files" check analogous to `ASSERTION_COLLECTIONHASH_INCORRECT_FILE_COUNT`?

sdk/src/assertions/collection_hash.rs
```diff
@@ -0,0 +1,660 @@
...
+
+        Ok(())
+    }
+
+    /// Generate the collection hash for a ZIP-based asset.
+    pub fn gen_hash_from_zip_stream<R>(&mut self, stream: &mut R) -> Result<()>
+    where
+        R: Read + Seek + ?Sized,
+    {
+        self.gen_zip_central_directory_hash(stream)?;
```
scouten-adobe, line 212:
```rust
self.gen_zip_central_directory_hash(stream)?;
self.gen_zip_uri_hashes(stream)?;
```
Both of these (and `zip_central_directory_range`/`zip_uri_ranges` in `verify_zip_stream_hash` below) independently construct a fresh `ZipArchive` over the stream, so the central directory gets fully parsed twice per sign and twice per verify. For large multi-entry packages (EPUB/ODS/DOCX with thousands of parts) this doubles the parse cost for no new information — could these share one parsed `ZipArchive`/entry list?

sdk/src/store.rs
```diff
@@ -3324,6 +3314,81 @@ impl Store {
...
+            }
+
+            data = self.to_jumbf_internal(reserve_size)?;
+            jumbf_size = data.len();
+
+            // we only need to compute the central directory hash here because everything else is
+            // already hashed in the previous step.
+            if let Some(mut collection_hash) = new_collection_hash {
+                if !remove_manifests {
+                    let mut scratch = io_utils::stream_with_fs_fallback(threshold, input_len)?;
```
scouten-adobe, line 3351: This writes a full copy of the asset (via `save_jumbf_to_stream` into `scratch`) just to learn the resulting central-directory bytes for hashing, then discards `scratch` — the real output comes from separate copy/embed passes later (~line 3384, and again in `finish_save_stream`). For a large ZIP-based asset this adds a full extra copy pass compared to the ~two passes other formats need for placeholder + final embed. Is there a way to derive the post-insert central-directory layout from the already-parsed original CD plus the known placeholder-entry size, without rewriting the bulk file data just to throw it away?

sdk/src/store.rs
```diff
@@ -3324,6 +3314,81 @@ impl Store {
                     pc.update_bmff_hash(bmff_hash)?;
                 }
             }
+        } else if is_zip {
```
scouten-adobe, line 3315: The BMFF and ZIP hash-assertion handling are two parallel, hand-inlined `if is_bmff {...} else if is_zip {...}` branches here (mirrored by label-string `if/else if` dispatch in `claim.rs`'s `verify_internal`), each repeating the same placeholder-generate → embed-JUMBF → recompute-hash recipe with format-specific calls. Given the PR description also calls out OOXML/ODF as in scope, would a shared trait (implemented by both `BmffHash` and `CollectionHash`) be worth it here, so a future format doesn't mean another copy-pasted branch in both files?

sdk/src/assertions/collection_hash.rs
```diff
@@ -0,0 +1,660 @@
...
+        if !path.is_file() {
+            return Err(Error::BadParam(format!(
+                "collection hashes must only contain files, got `{}`",
+                path.display()
+            )));
+        }
+
+        for component in path.components() {
+            match component {
+                Component::CurDir | Component::ParentDir => {
```
scouten-adobe, line 98: The same relative-component check is written three times — inline here in `add_file`, inline in `verify_zip_stream_hash` (line 297), and again as the `validate_uri` helper (line 326, itself only called from `gen_hash`/`verify_hash`, not from the other two). They've also already drifted: `add_file` raises `Error::BadParam`, `validate_uri` raises `Error::C2PAValidation`, for the identical condition. Worth consolidating into one helper both call sites share?

scouten-adobe (changes requested):
Ran a structured review over this PR (ZIP/EPUB/OOXML/ODF support + Collection Data Hash assertion). Left inline notes below where GitHub would let me anchor them to the diff. Requesting changes mainly for two correctness issues that look like real bugs:

- `sdk/src/store.rs:3330` — the emptiness check uses the aggregate `hash_assertions()` instead of the type-specific `collection_hash_assertions()` (added in this PR but never called), which could silently skip generating the CollectionHash assertion entirely. See inline note.
- Two more notes below couldn't be anchored inline because they're in files this PR doesn't touch, but are worth a look before merging since this PR is what makes them reachable for the first time:
  - `sdk/src/manifest.rs:602` — the match arm that hides internal hash-binding assertions from the public `Manifest` (`labels::DATA_HASH | labels::BMFF_HASH | labels::BOX_HASH => { /* do not include */ }`) wasn't updated to also exclude `labels::COLLECTION_HASH`, even though `labels::HASH_LABELS` elsewhere already treats it as a peer of the other three. As written, the internal `c2pa.hash.collection.data` assertion falls through to a later arm and gets pushed into `manifest.assertions`, leaking it into every verify/read JSON report for ZIP-based assets.
  - `c2pa_c_ffi/src/error.rs:127` — the new `c2pa::Error::ZipError` variant (`sdk/src/error.rs:360`) has no arm in `C2paError::from_c2pa_error`'s match and falls into the catch-all `_ => Self::Other(err_str)`, so C API consumers get a generic "Other" code for every ZIP-related failure instead of the granularity other error families get here.

The rest are lower-priority correctness questions plus some architecture/efficiency notes worth considering before this lands, but not necessarily blocking.

sdk/src/assertions/collection_hash.rs
ok-nick: Good catch. I'm going to keep the add_file case specifically alone for now because I prefer the more explicit error type, whereas the other case deals with valdiation.

sdk/src/store.rs
ok-nick: Yes this is a good idea but let's leave that to the asset IO refactoring.

sdk/src/store.rs
ok-nick: Yes but this problem exists in the other hard bindings as well due to how the APIs work. The goal of the new asset IO refactoring is to avoid these intermediate streams entirely. I defer to that work.

sdk/src/assertions/collection_hash.rs
ok-nick: It's unusually complicated to solve this because the `zip` crate doesn't provide all the information we need for parsing the ZIP (see `zip_uri_ranges` and the data descriptor parsing logic). Ideally that functionality can be upstreamed, but in any case we need the original stream we can't pass `ZipArchive` to both functions easily.

I'm not too worried since this function is only used in tests. In the future, following the asset IO refactoring I mentioned in the other comment, we would only need to read the ZIP archive once and cache the necessary locations.

sdk/src/assertions/collection_hash.rs
ok-nick: The spec doesn't say anything about this. I was working on writing up an issue to see how it should be handled, but as of now this is how the spec works.

sdk/src/store.rs
ok-nick: Sure, that would be a clean addition. However, note there cannot be multiple hash assertions anyways so it would fail validation.

sdk/src/assertions/collection_hash.rs
ok-nick: https://github.com/zip-rs/zip2/issues/974

scouten-adobe (changes requested):
A few things still remain:

* Path validation issue I sent privately.
* manifest.rs:602 — COLLECTION_HASH is still missing from the "hide from public Manifest" match arm; not touched.
* c2pa_c_ffi/src/error.rs:127 — ZipError still falls into _ => Self::Other; not touched.
collection_hash.rs:161 — verify_hash still has no check for extra/unaccounted files in directory-based collections.
* collection_hash.rs:206 and verify path — central directory still parsed twice per sign/verify.
* store.rs:3353 — the throwaway scratch full-file copy is unchanged.

scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) failed to backport this pull request in [workflow run 33900042913](https://github.com/contentauth/c2pa-rs/actions/runs/33900042913).

| Target | Status |
|--------|--------|
| `stable` | :x: Failed |

```bash
git fetch origin stable
git worktree add -d .worktree/backport-499-to-stable origin/stable
cd .worktree/backport-499-to-stable
git switch --create backport-499-to-stable
git cherry-pick -x b00371ad2065dab3335b9989bd9f83f206670000
```

</details>

PR #501, opened 2024-07-10
Expose `SignatureInfo` publicly

ok-nick:
Expose `SignatureInfo` publicly (already used in public API).

## Checklist


gpeacock:
I think if we are making the struct public, it makes sense to also make the fields public. We could alternatively provide getters.

ok-nick:
If user's are deserializing into the struct, I think it makes sense to have the fields public so they know what they're deserializing into.

gpeacock (approved):

PR #506, opened 2024-07-15
(MINOR) Add region of interest assertion definition

ok-nick:
Adds region of interest assertion definition.

Partially or completely solves #397 

## Checklist


gpeacock (changes requested):

ok-nick:
Some of the region of interest struct fields are floats, which cannot derive `Eq`. We remove `Eq` from existing structs, causing a breaking change.

gpeacock:
overidding minor warning since the script is still broken and does not detect it is there now.

PR #507, opened 2024-07-16
Builder Archive update

gpeacock:
Adds version and from indexed ingredients to manifests c2pa storage
Fixes Clippy changes so that make test passes


sdk/src/builder.rs
```diff
@@ -432,6 +447,29 @@ impl Builder {
...
+                let manifest_label = file
+                    .name()
+                    .split('/')
+                    .nth(1)
+                    .ok_or(Error::BadParam("Invalid manifest path".to_string()))?;
+                let manifest_label = manifest_label.replace(['_'], ":");
+                for ingredient in builder.definition.ingredients.iter_mut() {
+                    if let Some(active_manifest) = ingredient.active_manifest() {
+                        if manifest_label.starts_with(active_manifest) {
+                            ingredient.set_manifest_data(data.clone())?;
```
ok-nick, line 465: Can multiple ingredients share the same internal binary manifest? If that's the case, then we need to check if the binary manifest already exists when we write them (ZIP will error if it already exists, check out #499). Otherwise, we can avoid a clone by breaking out of the loop here.

sdk/src/builder.rs
```diff
@@ -432,6 +447,29 @@ impl Builder {
...
+            // Load the c2pa_manifests.
+            // We add the manifest data to any ingredient that has a matching active_manfiest label.
+            if file.name().starts_with("manifests/") && file.name() != "manifests/" {
+                let mut data = Vec::new();
+                file.read_to_end(&mut data)?;
+                let manifest_label = file
+                    .name()
+                    .split('/')
+                    .nth(1)
+                    .ok_or(Error::BadParam("Invalid manifest path".to_string()))?;
```
ok-nick, line 460: It's worth checking out `ZipFile::encode_name()` to get back a `Path`.

ok-nick (dismissed):

sdk/src/builder.rs
gpeacock: Yes, we can have multiple ingredients sharing the same manifest data, so we need to handle that case. It would be nice if Rust had a way to clone only when necessary here.

sdk/src/builder.rs
gpeacock: Thanks, I'll take a look at that.

gpeacock: do you mean enclosed_name()? The main purpose of that is to prevent attacks when unzipping files into folders, and we aren't doing that here. I think it is actually simpler to process this as a string since we need to convert the underscores back into colons anyway.

ok-nick (approved):

Issue #451, opened 2024-04-23
Test tool for API 2024, create and compare manifests


dkozma:

gpeacock:
Integrate with Nicks work

Issue #396, opened 2024-02-06
Testing infrastructure


gpeacock:
This makes sure that anything is included in the Rust integration tests are part of automation.

PR #513, opened 2024-07-17
chore: Upgrade testing infrastructure

ok-nick:
Upgrade testing infrastructure.
* Use [insta-rs](https://insta.rs) for snapshot testing
* Test reading old c2pa-rs version manifests are compatible with the current version
  * CI runs each publish and calls c2pa-compat which generates embedded/remote/json manifests for one of every type of asset handler (e.g. jpeg, bmff, png, gif, etc.).
    * Assets are binary diffed against original assets and compressed (only storing modified manifest blocks).
    * The json manifest used for signing should cover all possible features that c2pa-rs supports in the current version.
  * CI sends PR  with new compat snapshots for the current version.
  * Integration test runs against all compat snapshot versions and verifies reading all the old c2pa-rs version manifests are exactly the same as what's stored.
  * Ensures compatibility and tests against regressions.
* CI changes
  * Add back code coverage.
  * Insta snapshot test reporting.
  * Deny cargo doc warnings.

### Related Issues
* #451 
* #396
* #412

## Checklist


sdk/Cargo.toml
```diff
@@ -121,7 +121,7 @@ sha2 = "0.10.2"
 tempfile = "3.10.1"
 thiserror = "1.0.61"
 treeline = "0.1.0"
-url = "2.2.2, <2.5.1"  # Can't use 2.5.1 or newer until new license is reviewed.
+url = "2.2.2, <2.5.1"                                               # Can't use 2.5.1 or newer until new license is reviewed.
```
gpeacock, line 124: why the extra spaces?

gpeacock:
it feels redundant having tests/fixtures/assets/. 
We should ether put the assets folder in the tests folder or put the contents of assets into fixtures. 
It looks like it is missing the fixtures/certs folder, so I can't run tests.

sdk/Cargo.toml
ok-nick: Weird formatter mistake

ok-nick:
Blocked by issues with SVG and RIFF

sdk/Cargo.toml
```diff
@@ -165,10 +165,16 @@ web-sys = { version = "0.3.58", features = [
...
-] } # allow integration tests to use the new API
+] }
 jumbf = "0.4.0"
-
+insta = { version = "1.39.0", features = ["json", "redactions", "filters"] }
+# The deps below are used for tests/compat.rs
+tiny_http = "0.12.0"
+bsdiff = "0.2.0"
+# TODO: do we want to use a compression algo already used in the spec?
+lz4_flex = "0.11.3"
```
gpeacock, line 177: probably a good idea to use something we use elsewhere, but not critical.

sdk/Cargo.toml
```diff
@@ -177,3 +183,8 @@ wasm-bindgen-test = "0.3.31"
 actix = "0.13.1"
 ed25519-dalek = "2.1.1"
 tokio = { version = "1.36.0", features = ["full"] }
+c2pa = { path = ".", features = [
```
gpeacock, line 186: nice way to add this, I didn't know that would work!

gpeacock (changes requested):
It looks like it all works - I can make a small change and see it is caught. But I think we need more instruction about how to use and maintain it. My concern is that when the spec changes the format of a UUID, these will break, and we will need to fix the tests. So  how do we deal with expected changes going forward?

ok-nick:
As far as stabilizing values like UUIDs go, it should be a simple change to the Stabilizer. However, if the format of the JSON were to change, there can be some issues.

In the case where new fields are added to the JSON, the test currently ignores them by default. In the case where the location of the field changes or is removed, there’s currently no way to specify that. Ideally, we’d have a system built on top that tells us the known changes/additions/removals from the prior version.

That sounds simple enough, but there’s cases where a field may change that’s inside of an array of maps, we’d need a way to specify that as well. Insta-rs has a cool way of doing this with their redaction selectors, but that code isn’t decoupled. Unfortunately, we can’t leverage insta-rs here because redaction selectors must be hardcoded (and due to other limitations). If we want this type of functionality, we’d need to implement our own selector syntax, although I’m unsure of the difficulty and if we can leverage any existing crates.

For the time being and for the simplest functionality, we can ignore any unknown fields that don’t exist in both JSONs and verify that existing fields have the same exact values. However, if we can guarantee that fields will never be removed/moved, then we can definitely create a simple impl for known changes.

TLDR; I'll clarify maintenance in the docs and there is a lot of potential for making this suite stricter

sdk/Cargo.toml
ok-nick: We may want to consider using the brotli algorithm as it's used by compressed manifests in the spec.

ok-nick:
FYI, CI is failing due to a weird issue with snapshot filtering, where it isn't filtering the current version used in the PR [here](https://github.com/contentauth/c2pa-rs/blob/6c859363cbd3640c1891016a463f64bebbf25e5d/sdk/tests/common/mod.rs#L59). It doesn't happen locally, my thought is that it's using the latest c2pa-rs version only in CI somehow? I believe it can be temporarily fixed by merging main into this branch, but that shouldn't be necessary.

scouten-adobe:
Closing for lack of activity. @ok-nick feel free to re-open if you think we'll revisit this.

PR #514, opened 2024-07-17
Ensure Ingredient data_types make it to the store and back.

gpeacock:
Ensure Ingredient data_types are written to the ingredient assertion and are read back from it.
Add test cases to verity this.


dyro (approved):

mauricefisher64 (approved):
👍

PR #516, opened 2024-07-18
Introduce a mutex around the FFI calls to OpenSSL

scouten-adobe:
This mutex is currently behind a new feature flag `openssl_ffi_mutex`, which is not enabled by default.

OpenSSL code is not re-entrant; if tests run fast enough, that will cause unexpected behavior.

### DIFF
--- a/sdk/src/openssl/ffi_mutex.rs
+++ b/sdk/src/openssl/ffi_mutex.rs
@@ -1,0 +1,62 @@
+// Copyright 2024 Adobe. All rights reserved.
+// This file is licensed to you under the Apache License,
+// Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+// or the MIT license (http://opensource.org/licenses/MIT),
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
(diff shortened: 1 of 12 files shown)


mauricefisher64 (approved):
👍

sdk/src/openssl/ffi_mutex.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[cfg(feature = "openssl_ffi_mutex")]
+static FFI_MUTEX: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+pub(crate) struct OpenSslMutex<'a> {
+    // Dead code here is intentional. We don't need to read the () contents
+    // of this guard. We only need to ensure that the guard is dropped when
+    // this struct is dropped.
+    #[allow(dead_code)]
+    #[cfg(feature = "openssl_ffi_mutex")]
```
gpeacock, line 25: If we want this to be an optional feature, then why not declare it where we use it instead of having a dummy implementation and always calling it?

Alternatively we could isolate the openssl methods we call into a wrapper that does the right checks when needed.

dkozma (approved):

sdk/src/openssl/ffi_mutex.rs
gpeacock: I think the goal is to have this always enabled in all builds, so ok to continue as is, and we will remove the feature ASAP when tested.

PR #521, opened 2024-07-22
(MINOR) Updates needed for v2 JavaScript SDK

dkozma:
I had to make some edits to c2pa-rs to enable things we need on the new JavaScript SDK:

- Add a missing code path for adding ingredients asynchronously for them to be validated properly in Wasm
- Unfortunate duplication of the `AsyncSigner` code for Wasm that does not implement `Send`, since [futures/async functions in Wasm do not support `Send`](https://github.com/rustwasm/wasm-bindgen/issues/2409#issuecomment-820750943) and I don't know how/if you can define conditional supertraits in Rust (I tried to no avail).
- Added JSON schema (schemars) definitions for the `Settings` and `ManifestDefinition` APIs

## Checklist


sdk/src/ingredient.rs
```diff
@@ -841,36 +843,61 @@ impl Ingredient {
...
+                        .await
+                    {
+                        Ok(_) => Ok(store),
+                        Err(err) => Err(err),
+                    }
+                }
+            } else {
+                // This will always be Err in this situation
+                #[allow(clippy::unwrap_used)]
+                Err(jumbf_store.unwrap_err())
```
dkozma, line 883: There might be a nicer way to do this, however I wasn't sure how to handle this case without `match`, which I can't use since it creates a new function block inside the arms, which stops `async_generic` from working properly (would have to conditionally provide a sync or async function wrapper).

sdk/src/signer.rs
```diff
@@ -177,7 +176,58 @@ pub trait AsyncSigner: Sync {
...
+    /// This is useful for cases where the signer needs to handle the COSE structure directly.
+    /// Not recommended for general use.
+    fn direct_cose_handling(&self) -> bool {
+        false
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+#[async_trait(?Send)]
+pub trait AsyncSigner {
```
dkozma, line 199: Read the "Changes in this pull request" part of the PR description for info on why I had to duplicate this.

make_test_images/Cargo.toml
```diff
@@ -13,14 +13,14 @@ c2pa = { path = "../sdk", default-features = false, features = [
 	"unstable_api",
 ] }
 env_logger = "0.11"
-log = "0.4.8" 
-image = { version = "0.25.1", default-features = false, features = [
```
dkozma, line 17: I've pinned this to `0.25.1` to fix the tests - since Clippy would fail with the current version since we are now using a deprecated API (there is now an `ImageReader` that takes the place of `io::Reader`) and the minimum version tests would fail if I updated the API.

sdk/src/ingredient.rs
gpeacock: it could be cleaner but it is ok

sdk/src/signer.rs
gpeacock: I'm not sure we need to have send in the other cases, but I suppose it all depends on how it is being used.

gpeacock (approved):

PR #522, opened 2024-07-23
Export `AssertionDefinition` and `ActionTemplate` in public API

ok-nick:
Export `AssertionDefinition` + `ActionTemplate` in public API.

image

## Checklist

### DIFF
--- a/sdk/src/assertions/mod.rs
+++ b/sdk/src/assertions/mod.rs
@@ -14,7 +14,7 @@
 //! Assertion helpers to build, validate, and parse assertions.
 
 mod actions;
-pub use actions::{c2pa_action, Action, Actions, SoftwareAgent};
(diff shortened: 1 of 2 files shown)


gpeacock:
This should be ok. I'd updated this structure so that we can support CBOR based assertions better. Previously the definition at this level only supported JSON. It's a little weird because the default assertion encoding is CBOR when passing a full definition in JSON. Generally that's fine since CBOR is a superset of JSON. But if you call the APIs directly you can add an assertion with CBOR specific encoding.

gpeacock (changes requested):
Do we have an integration test that shows these can be used outside the crate? We should add something for that.

ok-nick:
@gpeacock These structs specifically or constructing a builder using Rust structs as a whole? It'd be a good idea to implement it for all structs either way, I can send a separate PR implementing that.

gpeacock (approved):

PR #524, opened 2024-07-24
Fix GIF `remove_cai_store_from_stream` behavior

ok-nick:
GIF should copy the entire input stream to the output if no c2pa block was found when calling `remove_cai_store_from_stream`.

## Checklist


ok-nick:
Even better if we can point the input_stream to output_stream with no i/o, but I'm not sure if we can provide these guarantees without side-effects.

mauricefisher64 (approved):
👍

PR #525, opened 2024-07-25
Update crate to fix bad certificate dump content

mauricefisher64:
Fix issue reported in https://github.com/contentauth/c2patool/issues/173. This was caused by bug in x509_certificate crate

## Checklist

### DIFF
--- a/make_test_images/Cargo.toml
+++ b/make_test_images/Cargo.toml
@@ -11,6 +11,7 @@ anyhow = "1.0.40"
 c2pa = { path = "../sdk", default-features = false, features = [
 	"openssl",
 	"unstable_api",
+	"file_io",
 ] }
 env_logger = "0.11"
 log = "0.4.8"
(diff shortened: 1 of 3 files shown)


make_test_images/Cargo.toml
```diff
@@ -11,6 +11,7 @@ anyhow = "1.0.40"
 c2pa = { path = "../sdk", default-features = false, features = [
 	"openssl",
 	"unstable_api",
+	"file_io",
```
gpeacock, line 14: why do we need to add file_io here? I'd updated this to not require it.

gpeacock (approved):
Approved by unclear on why we are adding file_io to make_test_images with this.

ok-nick (approved):

make_test_images/Cargo.toml
mauricefisher64: I got build errors without it.

PR #529, opened 2024-07-26
Ingredient checks

mauricefisher64:
This PR addresses  a number of hacks reported against the implementation.

Support 2.0 style recursive ingredient checks.  
Check trust for recursed ingredients
Display only signed timestamps not unprotected TstInfo structures
Fail fetching databoxes when HashURI is wrong
Fix hash generation failure for empty streams
Complete validation of wasm signing paths

## Checklist


sdk/src/manifest.rs
```diff
@@ -1966,19 +1976,19 @@ pub(crate) mod tests {
...
+            .unwrap();
 
-        /* to be enabled later
-                // try to load the manifest
-                let mut validation_log = DetailedStatusTracker::new();
-                Store::from_jumbf(&out_manifest, &mut validation_log).expect("manifest_load_error");
-        */
+        // try to load the manifest
+        let mut validation_log = DetailedStatusTracker::new();
+        Store::from_jumbf(&out_manifest, &mut validation_log).expect("manifest_load_error");
```
gpeacock, line 1991: we should be able to get this result from the above Reader::from_stream_async result, no need to call again here.

sdk/src/manifest.rs
```diff
@@ -2002,19 +2012,19 @@ pub(crate) mod tests {
...
```
gpeacock, line 2027: I don't think we need to call both Reader::from_stream.. and Store::from_jumbf, since the first one calls the second one.

sdk/src/ingredient.rs
```diff
@@ -1844,7 +1844,7 @@ mod tests_file_io {
         let ap = fixture_path("CIE-sig-CA.jpg");
         let ingredient = Ingredient::from_file(ap).expect("from_file");
         println!("ingredient = {ingredient}");
-        assert!(ingredient.validation_status().is_none());
+        assert!(ingredient.validation_status().is_some());
```
gpeacock, line 1847: This should not be reporting errors at the ingredient validation status level. what changed?

gpeacock (changes requested):
I'll build this locally and give it a try. I don't think the one ingredient should be reporting errors at that level, so I'd like to know why it is now.

sdk/src/ingredient.rs
gpeacock: It is identifying a CoseTimeStampMismatch on the nested manifest with the known error. I think that makes sense because this is a new check that would not have been performed when that nested image was added. So it is correct given that the image is now out of date. That's a useful test, but not what this test was designed to test. We should update the image that is saved in fixtures and this should not report an error anymore. 

```
  "validation_status": [
    {
      "code": "general.error",
      "url": "self#jumbf=/c2pa/contentauth:urn:uuid:5b639b8e-edc1-4e41-9c8e-c87fb1e36921/c2pa.signature",
      "explanation": "claim signature is not valid: CoseTimeStampMismatch"
    }
  ],
```

sdk/src/manifest.rs
mauricefisher64: I just uncommented your tests that were disabled.  Feel free to change.

sdk/src/manifest.rs
mauricefisher64: I will remove those.

sdk/src/manifest.rs
mauricefisher64: I removed

gpeacock (approved):

PR #532, opened 2024-07-30
(IGNORE) Clarify use of unstable_api feature flag.

crandmck:
Clarify use of unstable_api feature flag in README and examples.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -19,7 +19,7 @@
 //! This library supports reading, creating and embedding C2PA data
 //! with a variety of asset types.
 //!
-//! Some functionality requires you to enable specific crate features, 
+//! Some functionality requires you to enable specific crate features,
 //! as noted in the documentation.
 //!
 //! The library has a new experimental Builder/Reader API that will eventually replace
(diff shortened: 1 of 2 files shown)


sdk/src/lib.rs
```diff
@@ -19,21 +19,27 @@
 //! This library supports reading, creating and embedding C2PA data
 //! with a variety of asset types.
 //!
-//! We have a new experimental Builder/Reader API that will eventually replace
+//! To read with file-based methods, you must add the `file_io` dependency to your Cargo.toml.
```
gpeacock, line 22: The documentation flags code that depends on a particular feature, so I don't know that we need to call it out here at all. Maybe just a generic statement about how some functionality requires features to be enabled and will be noted in the documentation.

"Available on crate feature file_io only."

gpeacock (changes requested):

sdk/src/lib.rs
crandmck: OK, I changed it to 
> Some functionality requires you to enable specific crate features, as noted in the documentation.

gpeacock (approved):

Issue #381, opened 2024-02-06
Read only mode for specific formats


gpeacock:
We want an API that can distinguish between formats we can read-only vs read and write.

mauricefisher64:
Adding Collin who did the work.

Issue #363, opened 2024-01-05
Duplication of code in store.rs with slightly different  behaviors.


duggaraju:
There is lot of duplication across the various functions like

start_save() vs start_save_stream()
finish_save() vs finish_save_stream() finsh_save_to_memory()

and the corresponding jumbf versions .
If they can be combined or refactored to reuse the same code then that would avoid duplication and make it easy to test or even support  signing bmff files from a memory buffer etc.

gpeacock:
Thanks, yes we agree. We want to have all the core code just implemented with streams and then provide wrappers around those functions for memory or file_io. That's in the works now.

mauricefisher64:
Done

PR #533, opened 2024-07-31
feat: Decouple `asset_handlers` into `c2pa-codecs` crate

ok-nick:
Decouples asset_handler/asset_io into its own crate.

### Changes
* Separate/group encoding and decoding
  * Introduces `Encode`/`Decode` trait with the following methods:
    * Encode: `write_c2pa`, `remove_c2pa`, `patch_c2pa`, `write_xmp`, `write_xmp_provenance`, `remove_xmp`, `remove_xmp_provenance`
    * Decode: `read_c2pa`, `read_xmp`, `read_xmp_provenance`
  * Asset handlers (codecs) now take a read-only stream on construction
    * When signing, the typical workflow is read->write->read->write, there's a lot we can cache
  * Streams are no longer trait objects and instead generics (less overhead)
* Decouple hashers to the parsers
  * Introduces a `Hash` trait  with 5 methods: `hash`, `data_hash`, `box_hash`, `bmff_hash`, `collection_hash`
  * Parsers can choose their default hasher (usually defined by spec) with `Hasher::hash` whilst also implementing any other supported hashes
  * Data hash now returns a list of byte spans corresponding to the manifest rather than explicitly defining byte ranges over the entire file
* Add file signature inference
  * Introduces a `Supporter` trait with 3 methods: `supports_signature`, `supports_extension`, `supports_mime`.
* Add `Embed` trait (composed manifests)
  * Construct an `Embeddable` with `embeddable`, read it from a stream with `read_embeddable` and write it to a stream with `write_embeddable`
* `Codec` is the new entry-point for asset handlers
  * It implements all traits and forwards it to the corresponding codec based on the file type
  * It is an enum over all codecs, removing the need for boxing trait objects
  * For instance, `Codec::from_stream(&mut stream).read_c2pa()`
* Granular parsing errors
  * It's now clear specifically where parsing went wrong, rather than "not found"
* Codecs are individually locked behind feature flags
  * If you only want png, enable the png feature, same with gif, jpeg, etc.
* External codecs
  * The `Codec` struct can be created with an external codec that implements the desired traits (user-defined codecs)
* Updated dependencies
  * Many parser dependencies have become unmaintained or deprecated

### Testing
* Test the codecs separately from c2pa-rs
  * Use verified test image suites provided online, things like pngsuite, imagetestsuite, etc.
* Test all codecs simulatenously
  * Construct Codec and test read_xmp for all file types (no longer needs to be implemented for each asset handler)
* Fuzz the codecs
  * Individually fuzz each method of the codecs using afl or proptest

### Related issues
* #398
* #381
* #363

## Checklist


scouten-adobe:
Closing for lack of activity. @ok-nick feel free to re-open if you think we'll revisit this.

PR #551, opened 2024-08-13
Fix remote embedding RIFF when specifying mime type

ok-nick:
When signing a remote manifest into a RIFF file and specifying a mime type (such as one derived from `crate::format_from_path`), RIFF will error with `UnsupportedType`. Currently, RIFF explicitly checks if the format passed was an extension (not mime type as well). We can either add the mime types to this check or remove the check completely (since all supported formats are handled the same anyways).

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/riff_io.rs
+++ b/sdk/src/asset_handlers/riff_io.rs
@@ -541,90 +541,85 @@ impl RemoteRefEmbed for RiffIO {
     ) -> Result<()> {
         match embed_ref {
             crate::asset_io::RemoteRefEmbedType::Xmp(manifest_uri) => {
-                match self.riff_format.as_ref() {
-                    "avi" | "wav" | "webp" => {
-                        if let Some(curr_xmp) = self.read_xmp(input_stream) {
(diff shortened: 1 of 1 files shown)


sdk/src/asset_handlers/riff_io.rs
```diff
@@ -541,90 +541,85 @@ impl RemoteRefEmbed for RiffIO {
...
-                                let mut reader = CAIReadWrapper {
-                                    reader: input_stream,
-                                };
-                                Chunk::read(&mut reader, 0)?
-                            };
-
-                            if top_level_chunks.id() != RIFF_ID {
-                                return Err(Error::InvalidAsset("Invalid RIFF format".to_string()));
-                            }
+                if let Some(curr_xmp) = self.read_xmp(input_stream) {
```
mauricefisher64, line 544: The reason for the old code is to make sure we confirmed that those formats did indeed support XMP (not all formats do).  I think this change is OK since those are currently the only RIFF based asset types we support.

mauricefisher64 (approved):
👍

PR #564, opened 2024-08-16
Inline certs for wasm test signer

mauricefisher64:
WASM cannot read the fixtures folder so inline the test certs.

## Checklist


sdk/src/wasm/rsa_wasm_signer.rs
```diff
@@ -120,15 +120,144 @@ impl Signer for RsaWasmSigner {
     }
 }
 
+// If "../../tests/fixtures/certs/rs256.pub" or "../../tests/fixtures/certs/rs256.pem") change
+// CHAIN_BYTES and KEY_BYTES should be updated here.
+// We embed the certs here because WASM cannot read the fixtures.
```
dkozma, line 125: Technically, this is because we don't package the fixtures in the published crate due to [this line](https://github.com/contentauth/c2pa-rs/blob/ace4dd5480e1fdabd720dde98d30b2d8c6b8508d/sdk/Cargo.toml#L21), which makes sense since we don't generally want to add all of those files and bloat the package. I believe this is only appearing in Wasm since it is the only place where we are trying to access these fixtures during compile time and not during testing, when we have the entire repo checked out.

dkozma (approved):

PR #566, opened 2024-08-20
feat: Introduce `DynamicAssertion` trait

scouten-adobe:


sdk/src/dynamic_assertion.rs
```diff
@@ -0,0 +1,85 @@
...
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+use std::{fmt::Debug, slice::Iter};
+
+//use ciborium::de;
```
gpeacock, line 16: nix this until we use it, unless you want this as a todo comment.

sdk/src/dynamic_assertion.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl PreliminaryClaim {
+    pub fn new() -> Self {
+        Self {
+            assertion_uris: Vec::new(),
+        }
+    }
+
+    pub fn add_assertion_uri(&mut self, uri: HashedUri) {
```
gpeacock, line 77: We need comments on all public APIs. Add 
#![deny(missing_docs)] and build docs. Otherwise integration tests won't work.

gpeacock (commented):
Looks good other than needing some comments. I'll start the rework to move things into the Signer.

sdk/src/signer.rs
```diff
@@ -85,6 +85,11 @@ pub trait Signer {
     fn direct_cose_handling(&self) -> bool {
         false
     }
+
+    /// Returns a list of dynamic assertions that should be included in the manifest.
+    fn dynamic_assertions(&self) -> Vec<Box<dyn DynamicAssertion>> {
```
scouten-adobe, line 90: I'm wondering if it makes sense to have this fn on the non-async `Signer` trait. Do we anticipate a use case where they would be supported / needed?

scouten-adobe (commented):
I have a question about supporting `DynamicAssertion` in the non-async `Signer` case; otherwise, this looks spot on.

sdk/src/signer.rs
gpeacock: We don't plan on it, but I don't see the harm in having it here.

gpeacock (approved):

PR #568, opened 2024-08-20
(IGNORE) Re-upgrade async-generic to 1.1

scouten-adobe:
This was (inadvertently?) reverted to an older version in #529.


mauricefisher64 (approved):
👍

PR #579, opened 2024-09-11
upgrades to riff@2.0.0, preventing panic on invalid riff files

dyro:
Some RIFF files, specifically files with an incorrect length defined in the riff chunk, were causing the library RIFF to panic. Riff recently implemented a fix for this behavior, by having iterators return a `Result<Chunk>` instead of a `Chunk`, which was being `unwrap`ed in Riff's `IntoIter` trait implementation. This PR does the following:

1. Introduces a fixture with the invalid `wav` file. I modified the `wav` file to have the length of the root chunk be in incorrect. 
2. Adds a test which caught the panic caused by Riff.
3. Upgrades the version of Riff to `2.0.0`
4. Modifies the `RiffIo` implementation of `AssetIo` to handle invalid chunks from Riff, if they are present in the iterator. 
5. Removes a few unnecessary path prefixes.

## Checklist


sdk/src/asset_handlers/riff_io.rs
```diff
@@ -298,12 +297,10 @@ impl CAIReader for RiffIO {
             reader: input_stream,
         };
 
-        for c in top_level_chunks.iter(&mut chunk_reader) {
-            if c.id() == XMP_CHUNK_ID {
-                let output = c.read_contents(&mut chunk_reader).ok()?;
-                let output_string = String::from_utf8_lossy(&output);
-
-                return Some(output_string.to_string());
+        for chunk in top_level_chunks.iter(&mut chunk_reader).flatten() {
```
dyro, line 300: This implementation is skipping chunks that have an error and continuing with the iterator. We could change this to return a None after first finding an invalid chunk. Any thoughts?

mauricefisher64: Yeah, I would fail on invalid content, that is the issue with **flatten**.  You will skip problems.  I think there was another flatten in the changes.

dyro: Will change, thanks!

mauricefisher64: Is it common for RIFF to have invalid content that we can skip and the file remain useful?

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -670,6 +662,21 @@ pub mod tests {
         assert!(success)
     }
 
+    #[test]
+    fn test_read_cai_with_incorrect_header_size_does_not_panic() {
+        let riff_io = RiffIO::new("wav");
+
+        let panic_result = panic::catch_unwind(|| {
```
gpeacock, line 669: I think this is our first "don't panic!" test.

gpeacock (approved):
Nice cleanup while you were at it.

mauricefisher64 (approved):
👍

sdk/src/asset_handlers/riff_io.rs
dyro: It appears to depend on the implementation. Apple Music and Premiere Pro were fine with the incorrect length header in the DATA chunk. Not sure how they are reading the audio data. The data chunk was the last chunk in this example. Maybe they are checking for the end of the file as a safe guard?

PR #580, opened 2024-09-11
feat: Add `get_provenance_cert_chain` to `Reader`

iameli:
I found myself needing to inspect the cert chain present in a C2PA asset over in [c2pa-go](https://github.com/aquareum-tv/c2pa-go). This reader interface seemed to be the only piece exposed in the UniFFI bindings, so I added a helper function here. I have no idea if I'm doing it the right way, but it does work.

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -168,6 +168,11 @@ impl Reader {
         self.manifest_store.to_string()
     }
 
+    /// Return certificate chain for the provenance claim
+    pub fn get_provenance_cert_chain(&self) -> Result<String> {
+        self.manifest_store.store().get_provenance_cert_chain()
+    }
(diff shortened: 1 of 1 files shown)


gpeacock:
I'm trying to keep the API space to a minimum, so let me think about this. I'm planning to expose the detailed manifest store report to this API (with something like json-detailed() and that could include the certs from every manifest in the store.

scouten-adobe:
@iameli would you please enable maintainers' contributions? This would allow us to update to base branch. Otherwise, we'll keep asking you to update and hope that we notice at the right time. 😐

scouten-adobe:
Closing for lack of activity. Please re-open if this can be updated to current. This may be supported now by crJson support.

PR #589, opened 2024-09-19
fix: requires "StatusTracker" to implement "Send"

dyro:
This PR requires `StatusTracker` implementations to implement the `Send` trait. This is required because c2pa-node is calling the function 
`Ingredient::from_manifest_and_asset_bytes_async`, which eventually calls `Store::ingredient_checks_async` in c2pa-rs. 

Marking the `Store::ingredient_checks_async` as not `Sync` with the `#[async_recursion(?Send)]` macro is preventing calls to `Store::ingredient_checks_async` with Tokio via the `tokio::spawn` function call. 

Another approach to this may be to only mark the `Store::ingredient_checks_async` 's status tracker argument, and all of its callers in c2pa-rs as `Send`. Something like this: `&mut (impl StatusTracker + Send)`. 

## Checklist

### DIFF
--- a/sdk/src/status_tracker.rs
+++ b/sdk/src/status_tracker.rs
@@ -69,7 +69,7 @@ impl LogItem {
     }
 }
 
-pub trait StatusTracker {
+pub trait StatusTracker: Send {
     // should we stop on the first error
     fn stop_on_error(&self) -> bool;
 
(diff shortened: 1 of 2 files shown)


mauricefisher64 (approved):
👍

PR #593, opened 2024-09-24
Fixes issue where nested assertion-uri-hash errors were being reported at the active manifest level.

gpeacock:
Validation status now generates absolute URIs for these cases.
Fixed long standing bug in not deduping deeply nested manifest ingredient errors.
Added make_test_images script to generate a deeply nested assertion_uri_hash error case.
Added unit test to Reader for this.
Fixes CAI-6669.


mauricefisher64 (approved):
👍

PR #603, opened 2024-09-30
fix: Write absolute urls to manifest store resource references.

gpeacock:
Write absolute urls to manifest store resource references
Also match on either an absolute or relative reference. This allows Reader.resource_to_stream to read resources from nested manifests. Updates tests to ignore relative/absolute changes in urls. Adds from_json_with_stream_full_resources() Builder unit test.

### DIFF
--- a/make_test_images/src/compare_manifests.rs
+++ b/make_test_images/src/compare_manifests.rs
@@ -183,7 +183,10 @@ fn compare_json_values(
                 || path.ends_with(".instanceId")
                 || path.ends_with(".time")
                 || path.contains(".hash")
-                || val1.is_string() && val2.is_string() && val1.to_string().contains(":urn:uuid:"))
+                || val1.is_string()
+                    && val2.is_string()
+                    && (val1.to_string().contains(":urn:uuid:")
+                        || val2.to_string().contains(":urn:uuid:")))
             {
                 if val2.is_null() {
                     issues.push(format!("Missing {}: {}", path, val1));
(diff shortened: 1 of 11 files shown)


make_test_images/src/compare_manifests.rs
```diff
@@ -183,7 +183,10 @@ fn compare_json_values(
                 || path.ends_with(".instanceId")
                 || path.ends_with(".time")
                 || path.contains(".hash")
-                || val1.is_string() && val2.is_string() && val1.to_string().contains(":urn:uuid:"))
+                || val1.is_string()
+                    && val2.is_string()
```
mauricefisher64, line 187: Is this something we want to test against since it will not be available in 2.1?

sdk/src/manifest_store.rs
```diff
@@ -89,22 +89,41 @@ impl ManifestStore {
...
-            None => self.get_active(),
+        // add logic to search for local or absolute uri identifiers
+        let (manifest, label) = match manifest_label_from_uri(uri) {
+            Some(label) => (self.get(&label), label),
+            None => (
+                self.get_active(),
+                self.active_label().unwrap_or_default().to_string(),
+            ),
         };
+        let relative_uri = to_relative_uri(uri);
```
mauricefisher64, line 100: This seems to have become a common pattern.  Ensure it is relative then make absolute.

let absolute_uri = if uri.isrelative() { to_absolute_uri(&label, uri) } else { uri.to_owned());

We could move this logic to inside of the to_absolute_uri.

mauricefisher64 (approved):
👍

make_test_images/src/compare_manifests.rs
gpeacock: We will need to update this to handle 2.1 uris.

PR #606, opened 2024-10-02
chore: Add m4a test

tmathern:
- Add test to verify m4a support for signing (m4a is a BMFF format. The added test file uses AAC as audio codec).

## Checklist


gpeacock (approved):
Good to have this. Not a blocker, but is there a smaller sample we could use? We try to keep the fixtures as small as possible so that our build server time doesn't increase too much.

tmathern:
> Good to have this. Not a blocker, but is there a smaller sample we could use? We try to keep the fixtures as small as possible so that our build server time doesn't increase too much.

The Music app won't let me tune the conversion settings (I used the mp3 we have as sample to convert it tho m4a using the Music app on mac). If we have shorter mp3s or we can cut them to be shorter, that would also cut the file size.

PR #621, opened 2024-10-08
doc: Update API documentation

crandmck:
Review doc comments for Reader.

Edits to doc comments in Rust code for clarity and completeness.

TO DO:
- Change all references to errors to be to the [`Error`] enum

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -79,8 +79,8 @@ pub struct ManifestDefinition {
     #[serde(default = "default_instance_id")]
     pub instance_id: String,
 
-    /// An optional ResourceRef to an image.  Must be available when the manifest is signed.
-    /// Represents the state of the asset that was signed.
+    /// An optional ResourceRef to a thumbnail image that represents the asset that was signed.
+    /// Must be available when the manifest is signed.
     pub thumbnail: Option<ResourceRef>,
 
     /// A List of ingredients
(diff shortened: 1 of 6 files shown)


sdk/src/builder.rs
```diff
@@ -112,6 +115,8 @@ fn default_vec<T>() -> Vec<T> {
     Vec::new()
 }
 
+/// Assertion data is almost always a JSON structure, but this allows for
```
gpeacock, line 118: This comment applies to the Cbor variant of this enum. I think all we need to say here is that.
This allows the assertion to be expressed as CBOR or JSON. 
The default is CBOR unless you specify that an assertion should be JSON.

sdk/src/builder.rs
```diff
@@ -80,6 +79,8 @@ pub struct ManifestDefinition {
     #[serde(default = "default_instance_id")]
     pub instance_id: String,
 
+    /// An optional ResourceRef to an image.  Must be available when the manifest is signed.
```
gpeacock, line 82: This needs to say something about being a thumbnail image that represents the asset.

sdk/src/ingredient.rs
```diff
@@ -544,15 +544,15 @@ impl Ingredient {
...
+    /// Generates an `Ingredient` from a stream, including XMP info.
     pub fn from_stream_info<F, S>(stream: &mut dyn CAIRead, format: F, title: S) -> Self
     where
         F: Into<String>,
         S: Into<String>,
     {
         let format = format.into();
 
-        // try to get xmp info, if this fails all XmpInfo fields will be None
+        // try to get xmp info, if this fails all XmpInfo fields will be None.
```
gpeacock, line 555: Capitalize the T at the beginning.

gpeacock (changes requested):
I requested some changes but otherwise good to go.

sdk/src/builder.rs
crandmck: Addressed.

sdk/src/builder.rs
crandmck: Addressed.

sdk/src/ingredient.rs
crandmck: Addressed, although this is not a doc comment.

crandmck:
Addressed your comments.

gpeacock (approved):

Issue #622, opened 2024-10-08
Add support for multiple data hash exclusions


brogdonm:
From our understanding, the data hashes, should allow for multiple exclusions. Searching through the "Issues", I found:

https://github.com/contentauth/c2pa-rs/issues/560

Which may address this at some point, since the current hash block object types are:

```rust
pub enum HashBlockObjectType {
    Cai,
    Xmp,
    Other,
}
```

With that said, the current code simply clumps all `HashBlockObjectType::Cai` types together to have one range. This could be an issue where an asset type has the "Cai" block split in two different areas of the file (e.g., the beginning in a lookup table and the end).

In our work we have added the ability to treat each `HashBlockObjectType::Cai` as an exclusion range, de-fragmenting any areas that are adjacent to each other in memory.
> Note: I plan to subsequently raise a PR for this

brogdonm:
@llangsner Oh, I see the github label of `wontfix`. Should I close #623 then?

PR #623, opened 2024-10-09
622: C2PA-542: Allow for multiple data exclusions, minimyzing where appropriate

brogdonm:

The PR changes how the `HashBlockObjectType::Cai` is treated when building up the data exclusions for data hashes. This PR takes all of the `HashBlockObjectType::Cai` blocks, de-fragments any adjacent blocks and builds multiple exclusions.

In our work with fonts, we had the C2PA table directory entry in the beginning of the file, which pointed to the actual C2PA table at the end of the file. With the previous code, this resulted in nearly the entire file being excluded (this was not good and not desirable).

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -1580,39 +1580,47 @@ impl Store {
         // sort blocks by offset
         block_locations.sort_by(|a, b| a.offset.cmp(&b.offset));
 
-        // generate default data hash that excludes jumbf block
-        // find the first jumbf block (ours are always in order)
-        // find the first block after the jumbf blocks
-        let mut block_start: usize = 0;
-        let mut block_end: usize = 0;
-        let mut found_jumbf = false;
+        // Setup to assume fragmented CAI blocks
+        let mut exclusions = Vec::<(usize, usize)>::new();
         for item in block_locations {
             // find start of jumbf
-            if !found_jumbf && item.htype == HashBlockObjectType::Cai {
-                block_start = item.offset;
-                found_jumbf = true;
-            }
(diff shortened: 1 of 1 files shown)


brogdonm:
@scouten-adobe @mauricefisher64 I wanted to check in and see if there is any interest in this PR? I've been keeping it up to date but noticed @llangsner closed #622 a couple weeks ago.

mauricefisher64:
I think it is a worthy inclusion.  I will try to address this next week.  Thanks.

sdk/src/store.rs
```diff
@@ -1578,50 +1578,54 @@ impl Store {
...
             // find start of jumbf
-            if !found_jumbf && item.htype == HashBlockObjectType::Cai {
-                block_start = item.offset;
-                found_jumbf = true;
-            }
-
-            // find start of block after jumbf blocks
-            if found_jumbf && item.htype == HashBlockObjectType::Cai {
-                block_end = item.offset + item.length;
+            if item.htype == HashBlockObjectType::Cai {
```
mauricefisher64, line 1585: Curious how you would make this case in the wild.  None of our asset handlers will do this.  They do not by design require multiple exclusions.

brogdonm: When we implemented for fonts, this was a requirement. As we had multiple sections that needed to be excluded. We have since moved to general box hash support, but since that is not the main line of logic in the SDK we still have this in place when using just a data hash.

brogdonm:
After talking with @mauricefisher64 deciding to close this.

PR #627, opened 2024-10-10
feat: Associate ingredients with actions using parameters.org.cai.ingredientsIds array

gpeacock:
The previous method of using an action.instanceId is now deprecated and will be removed in the future.
 This fixes a non compliance issue with the spec.
```
 {
       "action": "c2pa.opened",
       "parameters": {
             "org.cai.ingredientIds": ["xmp.iid:813ee422-9736-4cdc-9be6-4e35ed8e41cb" ]
       }
}

```


mauricefisher64 (approved):
👍

dyro (approved):

PR #628, opened 2024-10-15
fix: Harden SDK against attempting buffers that are too large

mauricefisher64:
vec! will panic if the requested size is beyond the available memory limits.  This PR makes sure that user supplied values use a safer way to allocate than vec!.  It also adds a safe single line implementation of the common pattern:

let mut data = vec![0u8; data_len];
reader.read_exact(&mut data)?;

changes to:

let data = reader.read_to_vec(data_len)?;

## Checklist

### DIFF
--- a/sdk/src/error.rs
+++ b/sdk/src/error.rs
@@ -293,6 +293,12 @@ pub enum Error {
 
     #[error("prerelease content detected")]
     PrereleaseError,
+
+    #[error("insufficient memory space for operation")]
+    InsufficientMemory,
+
+    #[error("parameters our of range")]
+    OutOfRange,
 }
 
 /// A specialized `Result` type for C2PA toolkit operations.
--- a/sdk/src/error.rs
+++ b/sdk/src/error.rs
@@ -297,7 +297,7 @@ pub enum Error {
     #[error("insufficient memory space for operation")]
     InsufficientMemory,
 
-    #[error("parameters our of range")]
+    #[error("parameters out of range")]
     OutOfRange,
 }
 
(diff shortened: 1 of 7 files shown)


sdk/src/error.rs
```diff
@@ -293,6 +293,12 @@ pub enum Error {
 
     #[error("prerelease content detected")]
     PrereleaseError,
+
+    #[error("insufficient memory space for operation")]
+    InsufficientMemory,
+
+    #[error("parameters our of range")]
```
dyro, line 300: s/our/out

dyro (approved):
Looks clean! Quick question and found a spelling mistake.

sdk/src/utils/io_utils.rs
```diff
@@ -53,3 +53,55 @@ pub(crate) fn stream_len<R: Read + Seek + ?Sized>(reader: &mut R) -> Result<u64>
...
+
+    // make sure we can allocate vec
+    let mut output: Vec<T> = Vec::new();
+    output
+        .try_reserve_exact(num_items)
+        .map_err(|_e| Error::InsufficientMemory)?;
+
+    // fill if requested
+    if let Some(i) = init_with {
+        output.resize(num_items, i);
```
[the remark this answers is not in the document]
mauricefisher64: Fill will fill to the "len" of the object.  The len is currently 0 (capacity is num_items).  resize only changes the size if needed filling in the passed value to capacity.

sdk/src/error.rs
mauricefisher64: Fixed

PR #631, opened 2024-10-17
fix: Adds identified RangeType to region of interest

gpeacock:
fix: Adds identified RangeType to region of interest.
RangeType was missing the "identified' type now defined in the 2.1 spec, so this adds support for that type.
Also updated so that None fields in RegionOfInterest types do not serialize.

### DIFF
--- a/sdk/src/assertions/metadata.rs
+++ b/sdk/src/assertions/metadata.rs
@@ -326,21 +326,14 @@ pub mod tests {
                 .set_region_of_interest(RegionOfInterest {
                     region: vec![Range {
                         range_type: RangeType::Temporal,
-                        shape: None,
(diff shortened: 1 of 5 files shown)


sdk/src/assertions/metadata.rs
```diff
@@ -326,21 +326,14 @@ pub mod tests {
...
-                        text: None,
+                        ..Default::default()
                     }],
-                    name: None,
-                    identifier: None,
-                    region_type: None,
-                    role: None,
-                    description: None,
-                    metadata: None,
+                    ..Default::default()
```
dyro, line 336: Nice clean up!

dyro (approved):

PR #649, opened 2024-10-25
feat: Add fragmented mp4 Builder and Reader support

gpeacock:
Adds builder.sign_from_fragmented_files()
Adds Reader::from_fragmented_files()
Adds Reader::from_fragment and from_fragment_async
Adds Reader::to_folder -- to extract all resources
Deprecates all manifest_store constructors. Use Reader instead.
Fixes an issue where we were getting duplicate c2pa_rs ClaimGeneratorInfo values. 
Now we no longer add a c2pa_rs ClaimGeneratorInfo but insert a org.cai.c2pa_rs field into the first claim_generator to track the version. This will be more compatible with 2.1 in the future.

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -12,7 +12,7 @@
 // each license.
 
 #[cfg(feature = "file_io")]
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::{
     collections::HashMap,
     io::{Read, Seek, Write},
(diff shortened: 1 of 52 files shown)


scouten-adobe:
@gpeacock FYI I re-ran the PR title check and it has passed now.

sdk/src/builder.rs
```diff
@@ -972,6 +976,65 @@ impl Builder {
...
+        self.definition.instance_id = format!("xmp:iid:{}", Uuid::new_v4());
+        if self.definition.title.is_none() {
+            if let Some(title) = path.file_name() {
+                self.definition.title = Some(title.to_string_lossy().to_string());
+            }
+        }
+        Ok(())
+    }
+
+    /// Sign a set of fragmented BMFF files for DASH streaming.
```
mauricefisher64, line 1007: You can remove the data comment.  It should be BMFF based.  So HLS for example should work as well.

mauricefisher64: We should add a note here that the code does not work if the files already have C2PA data.  That is a known limitation.  I will add a story to support that.

dyro (approved):
Caught a grammer mistake and there's a question about deprecated version in the deprecated attributes you added.

PR #588, opened 2024-09-19
feat(sdk): Add support for MP3 without ID3 header

shc261392:
## Changes in this pull request

MP3 files does not necessarily have ID3 header and the current version will always raise "unsupported type" error when an file is MP3 format but has no ID3 header. Getting the ID3 header in MP3 file handler only serves the purpose of skipping the size of the header, hence when the MP3 file does not have an ID3 header, we could check for the MP3 header sync word (11 bits of 1) to make sure it is a valid MP3 file instead and skip 0 bytes as there is no ID3 header to skip.

This change would enable c2pa-rs to work with MP3 files that lack ID3 headers.

## Checklist
- [v] This PR represents a single feature, fix, or change.
- [v] All applicable changes have been documented.
- [v] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #652, opened 2024-10-29
fix: Add support for MP3 without ID3 header

shc261392:

MP3 files does not necessarily have ID3 header and the current version will always raise "unsupported type" error when an file is MP3 format but has no ID3 header. Getting the ID3 header in MP3 file handler only serves the purpose of skipping the size of the header, hence when the MP3 file does not have an ID3 header, we could check for the MP3 header sync word (11 bits of 1) to make sure it is a valid MP3 file instead and skip 0 bytes as there is no ID3 header to skip.

This change would enable c2pa-rs to work with MP3 files that lack ID3 headers.

## Checklist
- [v] This PR represents a single feature, fix, or change.
- [v] All applicable changes have been documented.
- [v] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.


shc261392:
@mauricefisher64 Hi could you check on this pull request? This is identical to https://github.com/contentauth/c2pa-rs/pull/588 except I used a personal fork to enable maintainer contribution so that you may edit it if needed.

mauricefisher64:
Can you bring the branch up to date with main.  I will try to merge it then.

mauricefisher64 (approved):
👍

PR #674, opened 2024-11-14
fix: For Issue 672, Callback is unsound

gpeacock:


andrewhalle (approved):

dyro (approved):
The first commit looks good, but it looks like one of the merges removed your changes.

PR #625, opened 2024-10-09
fix: Fix #624 (edge cases when combining the box hashes)

brogdonm:
## Changes in this pull request

This addresses the issue #624 by only combining boxes that were found.

## Checklist
- [X] This PR represents a single feature, fix, or change.
- [X] All applicable changes have been documented.
- [X] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #678, opened 2024-11-15
chore: Delete "Preflight crate publish" step

scouten-adobe:
This step is no longer viable in the new monorepo configuration. Changes in one crate that are required by another cause this step to fail and that's more noise than signal.


scouten-adobe:
Should address issue raised in https://github.com/contentauth/c2pa-rs/pull/625#issuecomment-2477764221.

PR #683, opened 2024-11-16
chore: Move `c2pa::CoseValidator` to `c2pa_crypto::RawSignatureValidator`

scouten-adobe:


internal/crypto/src/openssl/validators/ecdsa_validator.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        let sig_der = if is_p1363 {
+            // Convert P1363 signature to DER signature.
+            let sig_len = sig.len() / 2;
+
+            let r = BigNum::from_slice(&sig[0..sig_len])?;
+            let s = BigNum::from_slice(&sig[sig_len..])?;
+            EcdsaSig::from_private_components(r, s)?.to_der()?
+        } else {
+            sig.to_vec()
```
scouten-adobe, line 70: @mauricefisher64 do you know of a case where the signature is _not_ encoded as P1363?

internal/crypto/src/openssl/validators/rsa_validator.rs
```diff
@@ -0,0 +1,110 @@
...
+    alg: SigningAlg,
+}
+
+impl RsaValidator {
+    pub fn new(alg: SigningAlg) -> Self {
+        RsaValidator { alg }
+    }
+}
+
+impl CoseValidator for RsaValidator {
```
scouten-adobe, line 29: Ignore this for now … I haven't retrofitted RSA to the new pattern yet.

internal/crypto/src/raw_signature/validator.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// Which validators are available may vary depending on the platform and
+/// which crate features were enabled.
+pub fn validator_for_signing_alg(alg: SigningAlg) -> Option<Box<dyn RawSignatureValidator>> {
+    #[cfg(feature = "openssl")]
+    if let Some(validator) = crate::openssl::validators::validator_for_signing_alg(alg) {
+        return Some(validator);
+    }
+
+    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
```
scouten-adobe, line 46: @cdmurph32 when we have a chance, I'd like to learn how to test on the WASI-not-WASM environments.

scouten-adobe: Also, I've been trying to achieve code coverage on WASM, but it looks just out of reach. This (https://github.com/taiki-e/cargo-llvm-cov/pull/338) looks tantalizingly close.

internal/crypto/src/tests/openssl/validators/rsa_validator.rs
```diff
@@ -0,0 +1,107 @@
...
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+#[test]
```
scouten-adobe, line 14: Ignore … not yet retrofitted.

internal/crypto/src/webcrypto/validators/ecdsa_validator.rs
```diff
@@ -0,0 +1,74 @@
...
+/// An `EcdsaValidator` can validate raw signatures with one of the ECDSA
+/// signature algorithms.
+pub enum EcdsaValidator {
+    /// ECDSA with SHA-256
+    Es256,
+
+    /// ECDSA with SHA-384
+    Es384,
+    // ECDSA with SHA-512
+    // Es512, // not yet implemented (check with Colin)
```
scouten-adobe, line 30: @cdmurph32 why was Es512 not implemented?

internal/crypto/src/openssl/validators/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::{raw_signature::RawSignatureValidator, SigningAlg};
+
+mod ecdsa_validator;
+pub use ecdsa_validator::EcdsaValidator;
+
+mod ed25519_validator;
+pub use ed25519_validator::Ed25519Validator;
+
+// mod rsa_validator;
+// pub use rsa_validator::{RsaLegacyValidator, RsaValidator}; // ???
```
tmathern, line 26: Why the `???`?

internal/crypto/src/raw_signature/validator.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use crate::SigningAlg;
+
+/// A `RawSignatureValidator` implementation checks a signature encoded using a
+/// specific signature algorithm and a private/public key pair.
+///
+/// IMPORTANT: This signature is typically embedded in a wrapper provided by
+/// another signature mechanism. In the C2PA ecosystem, this wrapper is
+/// typically COSE, but `RawSignatureValidator` does not implement COSE.
+pub trait RawSignatureValidator {
```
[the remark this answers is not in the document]
tmathern: Will these changes be breaking changes then?

internal/crypto/src/tests/raw_signature/validator.rs
```diff
@@ -0,0 +1,121 @@
...
+
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen_test::wasm_bindgen_test;
+
+use crate::{
+    raw_signature::{validator_for_signing_alg, RawSignatureValidationError},
+    SigningAlg,
+};
+
+const SAMPLE_DATA: &[u8] = b"some sample content to sign";
```
tmathern, line 22: Could we had tests with different strings? Or vary string used for testing?
I would consider even to go as far as picking strings from those: https://github.com/minimaxir/big-list-of-naughty-strings

internal/crypto/src/raw_signature/validator.rs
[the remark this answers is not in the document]
scouten-adobe: No. `CoseValidator` was not public in c2pa-rs.

internal/crypto/src/openssl/validators/mod.rs
scouten-adobe: I'm not sure if `RsaLegacyValidator` needs to be made public or brought over.

internal/crypto/src/tests/raw_signature/validator.rs
scouten-adobe: Mostly because I'm trying to avoid re-creating test assets.

internal/crypto/src/tests/openssl/validators/rsa_validator.rs
scouten-adobe: Now retrofitted.

sdk/src/time_stamp.rs
```diff
@@ -627,8 +583,19 @@ pub(crate) fn verify_timestamp(ts: &[u8], data: &[u8]) -> Result<TstInfo> {
             let validated_res: Result<bool> = if _sync {
                 #[cfg(feature = "openssl")]
                 {
-                    let validator = get_local_validator(sig_alg, hash_alg)?;
-                    validator.validate(&sig_val.to_bytes(), &tbs, &signing_key_der)
+                    let Some(validator) =
```
scouten-adobe, line 586: IMHO this can all boil down to the synchronous version since we now have equivalent code for WASM. @cdmurph32 @mauricefisher64 any objection to removing the async path?

sdk/src/validator.rs
```diff
@@ -29,53 +25,3 @@ pub struct ValidationInfo {
     pub cert_chain: Vec<u8>, // certificate chain used to validate signature
     pub revocation_status: Option<bool>,
 }
-
-/// Trait to support validating a signature against the provided data
-pub(crate) trait CoseValidator {
```
scouten-adobe, line 34: Replaced by `c2pa_crypto::raw_signature::RawSignatureValidator`.

sdk/src/validator.rs
```diff
@@ -29,53 +25,3 @@ pub struct ValidationInfo {
...
-// • PS384 (RSASSA-PSS using SHA-384 and MGF1 with SHA-384)
-// • PS512 (RSASSA-PSS using SHA-512 and MGF1 with SHA-512)
-// • RS256	RSASSA-PKCS1-v1_5 using SHA-256
-// • RS384	RSASSA-PKCS1-v1_5 using SHA-384
-// • RS512	RSASSA-PKCS1-v1_5 using SHA-512
-// • ED25519 Edwards Curve ED25519
-
-/// return validator for supported C2PA  algorithms
-#[cfg(feature = "openssl")]
-pub(crate) fn get_validator(alg: SigningAlg) -> Box<dyn CoseValidator> {
```
scouten-adobe, line 62: Replaced by `c2pa_crypto::raw_signature::validator::validator_for_signing_alg`.

sdk/src/openssl/mod.rs
```diff
@@ -16,33 +16,16 @@ mod rsa_signer;
 #[cfg(feature = "openssl_sign")]
 pub(crate) use rsa_signer::RsaSigner;
 
-#[cfg(feature = "openssl")]
```
scouten-adobe, line 19: These all moved into `c2pa_crypto::openssl::validators`.

sdk/src/time_stamp.rs
```diff
@@ -448,273 +404,277 @@ fn get_validator_type(sig_alg: &bcder::Oid, hash_alg: &bcder::Oid) -> Option<Str
...
+        validate_timestamp_sig(sig_alg, hash_alg, sig_val, &tbs, &signing_key_der)?;
+
+        #[cfg(not(feature = "openssl"))]
+        {
+            #[cfg(target_arch = "wasm32")]
+            {
+                if _sync {
+                    validate_timestamp_sig(sig_alg, hash_alg, sig_val, &tbs, &signing_key_der)?;
+                } else {
+                    // TO REVIEW: Worth keeping this WASM-specific async path alive, or can we fully switch over to the synchronous path? (I'd prefer the latter.)
```
scouten-adobe, line 597: Reviewers: Please let me know if the WASM-specific async path is worth preserving. I'd love to simplify down to the new sync-only path.

internal/crypto/src/raw_signature/validator.rs
```diff
@@ -0,0 +1,143 @@
...
+    #[error("the signature does not match the provided data or public key")]
+    SignatureMismatch,
+
+    /// An error was reported by the OpenSSL native code.
+    ///
+    /// NOTE: We do not directly capture the OpenSSL error itself because it
+    /// lacks an Eq implementation. Instead we capture the error description.
+    #[cfg(feature = "openssl")]
+    #[error("an error was reported by OpenSSL native code: {0}")]
+    OpenSslError(String),
```
gpeacock, line 122: I think this is a very good choice.

gpeacock (approved):

internal/crypto/src/raw_signature/validator.rs
```diff
@@ -0,0 +1,143 @@
...
+        public_key: &[u8],
+    ) -> Result<(), RawSignatureValidationError>;
+}
+
+/// Return a built-in signature validator for the requested signature
+/// algorithm.
+///
+/// Which validators are available may vary depending on the platform and
+/// which crate features were enabled.
+pub fn validator_for_signing_alg(alg: SigningAlg) -> Option<Box<dyn RawSignatureValidator>> {
```
mauricefisher64, line 42: I have to think about this.  It may be too restrictive.  There are cases where we validate patterns that don't match our predefined signature types.  For example You can have an EC signature that uses an arbitrary hash length.  They are not limited to for example ES256 where the sig is on curve prime256v1 but the hash is Sha256.  You can use that curve with other hash values.  We see this when validating certs.

internal/crypto/Cargo.toml
```diff
@@ -63,9 +63,26 @@ default-features = false
 features = ["now", "wasmbind"]
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
+ecdsa = "0.16.9"
+ed25519-dalek = "2.1.1"
+p256 = "0.13.2"
+p384 = "0.13.0"
+rsa = { version = "0.9.6", features = ["sha2"] }
+spki = "0.7.3"
 wasm-bindgen = "0.2.83"
```
cdmurph32, line 72: wasm-bindgen, web-sys, web-time can not be included in the wasi target. I know this PR isn't really about adding WASI support, but there are other wasi targets in the Cargo.toml

internal/crypto/src/raw_signature/validator.rs
cdmurph32: We still don't have test automation, but we can at least set up manual testing by creating interfaces and run cli commands against them.
https://github.com/cdmurph32/c2pa-component

I'd love to create a test framework for WASI, it will take a lot of work though. It could be written in any supported language.

internal/crypto/src/openssl/validators/ecdsa_validator.rs
mauricefisher64: So when we sign with OpenSSL or we are checking certificates the values are DER.

internal/crypto/src/tests/raw_signature/validator.rs
```diff
@@ -0,0 +1,289 @@
...
+use crate::{
+    raw_signature::{
+        validator_for_sig_and_hash_algs, validator_for_signing_alg, RawSignatureValidationError,
+    },
+    SigningAlg,
+};
+
+const SAMPLE_DATA: &[u8] = b"some sample content to sign";
+
+#[test]
```
mauricefisher64, line 28: With these changes I thought these unit tests would work for both WASM and non-WASM cases?

internal/crypto/src/webcrypto/validators/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+mod rsa_validator;
+pub use rsa_validator::RsaValidator;
+
+/// Return a validator for the given signing algorithm.
+pub fn validator_for_signing_alg(alg: SigningAlg) -> Option<Box<dyn RawSignatureValidator>> {
+    match alg {
+        SigningAlg::Es256 => Some(Box::new(EcdsaValidator::Es256)),
+        SigningAlg::Es384 => Some(Box::new(EcdsaValidator::Es384)),
+        SigningAlg::Es512 => None, /* why is this unimplemented? */
```
mauricefisher64, line 41: Interesting?  This is implemented in the current c2pa-rs

internal/crypto/src/webcrypto/validators/ecdsa_validator.rs
mauricefisher64: Looks like the switch to Rustls here is a regression.  We currently support ES512 with SubtleCrypto.

internal/crypto/src/raw_signature/validator.rs
mauricefisher64: Should we allow the Error to return LogItem to get the detailed error info?

internal/crypto/src/raw_signature/validator.rs
scouten-adobe: OK, without test automation I think we have to call it unofficial support. I'll carry over things where I see them but I think it will be hard to maintain effectively.

internal/crypto/src/webcrypto/validators/ecdsa_validator.rs
scouten-adobe: OK, that's the reason to keep the async path open. SubtleCrypto requires async.

sdk/src/time_stamp.rs
scouten-adobe: Until we can support Es512 and others, then we need to keep the async path open. I'll leave the code as is.

internal/crypto/src/tests/raw_signature/validator.rs
scouten-adobe: Not sure what you're saying here. Next line flags it was a WASM-compatible test.

Unfortunately, the WASM test framework doesn't respect the standard `#[test]` annotation. 😢

internal/crypto/src/webcrypto/validators/ecdsa_validator.rs
scouten-adobe: This code path is synchronous-only.

internal/crypto/src/raw_signature/validator.rs
scouten-adobe: @mauricefisher64 I saw this error case a few times while debugging the new code and I was pretty pleased with the description that was provided. It included the OpenSSL file name and line number.

internal/crypto/src/openssl/validators/ecdsa_validator.rs
mauricefisher64: Is you sign something with straight OpenSSL it will be a DER signature.  You can't use a signature that came from Cose because it will be in the other format.

internal/crypto/src/raw_signature/validator.rs
mauricefisher64: I see.  We will have to do some testing against those odd cert cases.

scouten-adobe: @mauricefisher64 would love to have new test cases for anything else you're familiar with.

PR #691, opened 2024-11-20
Report OCSP revocation and update status in SignatureInfo

gpeacock:
Add functionality to report OCSP revocation status and update information in the `SignatureInfo` structure. Introduce a new method `Reader.detailed_json()` for enhanced JSON output. Remove the obsolete `revocation_status` field and replace it with `revocation_date` and `ocsp_next_update`.

### DIFF
--- a/sdk/src/manifest_store_report.rs
+++ b/sdk/src/manifest_store_report.rs
@@ -45,7 +45,10 @@ impl ManifestStoreReport {
     pub(crate) fn from_store(store: &Store) -> Result<Self> {
         let mut manifests = HashMap::<String, ManifestReport>::new();
         for claim in store.claims() {
-            manifests.insert(claim.label().to_owned(), ManifestReport::from_claim(claim)?);
+            manifests.insert(
+                claim.label().to_owned(),
+                ManifestReport::from_claim(claim, store)?,
+            );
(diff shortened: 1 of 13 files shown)


mauricefisher64 (approved):
👍

sdk/src/manifest_store_report.rs
```diff
@@ -105,8 +108,18 @@ impl ManifestStoreReport {
...
         println!("{cert_str}\n\n");
 
-        if let Some(ocsp_info) = store.get_ocsp_status() {
-            println!("{ocsp_info}");
+        let claim_label = store
+            .provenance_path()
+            .ok_or(crate::Error::ProvenanceMissing)?;
+        if let Some(ocsp_info) = store.get_ocsp_status(&claim_label) {
+            if let Some(revoked_at) = &ocsp_info.revoked_at {
+                println!("Certificate Status: Revoked, revoked at: {}", revoked_at);
```
mauricefisher64, line 116: If there is a update time and it has expired without error then may we should just display a "Not Revoked" or "Cert Good" message.

gpeacock:
I'm going to close this since we have taken other approaches to this. If it comes up again I can resurrect it.

PR #694, opened 2024-11-21
chore: Move WASM asynchronous raw signature mechanism into c2pa-crypto

scouten-adobe:


internal/crypto/src/webcrypto/async_validators/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        public_key: &[u8],
+    ) -> Result<(), RawSignatureValidationError>;
+}
+
+/// Return an async validator for the given signing and hash algorithm.
+pub fn async_validator_for_signing_alg(
+    alg: SigningAlg,
+) -> Option<Box<dyn AsyncRawSignatureValidator>> {
+    match alg {
+        // SigningAlg::Es256 => Some(Box::new(EcdsaValidator::Es256)),
```
scouten-adobe, line 52: Will fill in the other cases in a subsequent draft of this PR.

scouten-adobe:
Review notes: This completes the move of raw-signature validation into c2pa-crypto. There are known failures in the WASM code path which I will investigate in the morning. This should give you a sense of the general direction.

internal/crypto/src/raw_signature/validator.rs
```diff
@@ -133,6 +133,15 @@ pub enum RawSignatureValidationError {
...
     InvalidSignature,
+
+    /// The time stamp uses an unsupported signing or hash algorithm.
+    #[error("signature uses an unsupported algorithm")]
+    UnsupportedAlgorithm,
+
+    /// An unexpected internal error occured while requesting the time stamp
+    /// response.
+    #[error("internal error ({0})")]
+    InternalError(&'static str),
```
tmathern, line 144: If it's a timestamp error, why not give it a better name?

internal/crypto/src/webcrypto/async_validators/ecdsa_validator.rs
```diff
@@ -0,0 +1,150 @@
...
+    Es256,
+
+    /// ECDSA with SHA-384
+    Es384,
+
+    // ECDSA with SHA-512
+    Es512,
+}
+
+#[async_trait(?Send)]
```
tmathern, line 38: Why `?Send` (why the question mark?))

internal/crypto/src/raw_signature/validator.rs
tmathern: Sounds good, thanks for clarifying (because while reading the command I thought it's specifically for a single case).

PR #696, opened 2024-11-22
feat: Move time stamp code into c2pa-crypto

scouten-adobe:


scouten-adobe:
@mauricefisher64 @mikeklein13 do we have any sample time stamp data that we could use to achieve independent code coverage for this?

scouten-adobe:
Per @mauricefisher64, look at the `claim_v2` branch to catch up on some recent changes to time stamp work.

internal/crypto/src/time_stamp/verify.rs
```diff
@@ -0,0 +1,368 @@
...
+    // Look for any valid signer.
+    for signer_info in sd.signer_infos.iter() {
+        // Find signer's cert.
+        let cert = match certs.iter().find_map(|cc| {
+            let c = match cc {
+                Certificate(c) => c,
+                _ => return None,
+            };
+
+            match &signer_info.sid {
```
mklein13, line 64: Move these to separate functions that clearly define their purpose

internal/crypto/src/time_stamp/verify.rs
```diff
@@ -0,0 +1,368 @@
...
+                        && sn.serial_number == c.tbs_certificate.serial_number
+                    {
+                        Some(c.clone())
+                    } else {
+                        None
+                    }
+                }
+
+                SignerIdentifier::SubjectKeyIdentifier(ski) => {
+                    const SKI_OID: ConstOid = bcder::Oid(&[2, 5, 29, 14]);
```
mklein13, line 76: Consider using Rust crypto Constants for OIDs: https://github.com/RustCrypto/formats/tree/master/const-oid

internal/crypto/src/time_stamp/verify.rs
```diff
@@ -0,0 +1,368 @@
...
+                if _sync {
+                    // IMPORTANT: The synchronous implementation of validate_timestamp_sync
+                    // on WASM is unable to support _some_ signature algorithms. The async path
+                    // should be used whenever possible.
+                    validate_timestamp_sig(sig_alg, hash_alg, sig_val, &tbs, &signing_key_der)?;
+                } else {
+                    // NOTE: We're keeping the WASM-specific async path alive for now because it
+                    // supports more signature algorithms. Look for future WASM platform to provide
+                    // the opportunity to unify.
+                    validate_timestamp_sig_async(
```
mklein13, line 256: Does validation need an async call path?

internal/crypto/src/time_stamp/provider.rs
```diff
@@ -0,0 +1,165 @@
...
+use x509_certificate::DigestAlgorithm;
+
+use crate::{asn1::rfc3161::TimeStampReq, time_stamp::TimeStampError};
+
+/// A `TimeStampProvider` implementation can contact a [RFC 3161] time stamp
+/// service and generate a corresponding time stamp for a specific piece of
+/// data.
+///
+/// [RFC 3161]: https://datatracker.ietf.org/doc/html/rfc3161
+pub trait TimeStampProvider {
```
mklein13, line 26: Provide a builder interface over a trait.

mklein13 (changes requested):
Thank you for your hard work on this! I have a couple of suggestions that I think could help improve the library’s structure and usability:

Decouple HTTP I/O:
Consider focusing the library on just building HTTP requests, parsing responses, and verifying them. This would simplify the library’s structure and make the code easier to test. For users needing more convenience, you might offer a way to pass in a function that takes an HTTP request and returns an HTTP response. This approach would let users choose any HTTP client they prefer while shifting connection handling to the user rather than the library. _(Won't implement in this PR. Tracking as CAI-7305.)_

Leverage standardized types:
It might be worth using the standardized types from the http crate for request building and parsing responses. This could enhance compatibility and consistency while making the library easier to integrate into other projects. _(Won't implement in this PR. Tracking as CAI-7306.)_

Decouple verifying logic from response parsing:
Combining both responsablities leads to complicated code that's difficult to fully test.

internal/crypto/src/time_stamp/http_request.rs
```diff
@@ -0,0 +1,126 @@
...
+    message: &[u8],
+) -> Result<Vec<u8>, TimeStampError> {
+    let request = Constructed::decode(
+        bcder::decode::SliceSource::new(data),
+        bcder::Mode::Der,
+        TimeStampReq::take_from,
+    )
+    .map_err(|_err| TimeStampError::InternalError("failure to decode Constructed TimeStampReq"))?;
+
+    let ts = time_stamp_request_http(url, headers, &request)?;
```
andrewhalle, line 40: Shouldn't `time_stamp_request_http()` be async-generic? As-is, I think `default_rfc3161_request_async()` would block the executor because `time_stamp_request_http()` uses `ureq` which uses blocking I/O.

mauricefisher64: It has to work in the sync path too

internal/crypto/src/time_stamp/http_request.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    const HTTP_CONTENT_TYPE_REQUEST: &str = "application/timestamp-query";
+    const HTTP_CONTENT_TYPE_RESPONSE: &str = "application/timestamp-reply";
+
+    let mut body = Vec::<u8>::new();
+    request
+        .encode_ref()
+        .write_encoded(bcder::Mode::Der, &mut body)?;
+
+    let mut req = ureq::post(url);
```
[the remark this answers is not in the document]
mauricefisher64: We do no have state in the SDK.   If you want to do that you do it in your custom signer.

internal/crypto/src/time_stamp/verify.rs
mauricefisher64: No, why.  This is not separate functionality.  It is doing a specific process.

internal/crypto/src/time_stamp/verify.rs
mauricefisher64: I don't want to be dependent on RustCrypto either

mauricefisher64: I would use the Oid crate instead.  The problem you will hit is every crate has a different Oid library with different functionality.  So what happened is you are constantly doing conversions since you cannot simply compare Oid for different libraries.  They have different memory configurations.  Before we just start changing stuff make sure you know the implications.

internal/crypto/src/time_stamp/verify.rs
mauricefisher64: It has an async path for wasm.

internal/crypto/src/time_stamp/http_request.rs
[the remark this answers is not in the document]
scouten-adobe: Won't implement in this PR. Tracking as CAI-7308.

internal/crypto/src/tests/time_stamp/verify.rs
```diff
@@ -0,0 +1,82 @@
...
+    let ts = include_bytes!("../fixtures/time_stamp/video1.ts.bin");
+    let mut data = include_bytes!("../fixtures/time_stamp/video1.data.bin").to_vec();
+    assert_ne!(data[4], 4);
+    data[4] = 4;
+
+    let err = verify_time_stamp(ts, &data).unwrap_err();
+    assert_eq!(err, TimeStampError::InvalidData);
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), actix::test)]
```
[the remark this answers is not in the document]
scouten-adobe: Won't implement in this PR. Tracking as CAI-7309.

internal/crypto/src/time_stamp/verify.rs
```diff
@@ -0,0 +1,368 @@
...
+    },
+    raw_signature::validator_for_sig_and_hash_algs,
+    time_stamp::{TimeStampError, TimeStampResponse},
+};
+
+/// Decode the TimeStampToken info and verify it against the supplied data.
+///
+/// TEMPORARILY PUBLIC while refactoring
+#[async_generic]
+pub fn verify_time_stamp(ts: &[u8], data: &[u8]) -> Result<TstInfo, TimeStampError> {
```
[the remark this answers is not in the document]
scouten-adobe: Won't implement in this PR. Tracking as CAI-7310.

internal/crypto/src/time_stamp/verify.rs
scouten-adobe: Won't implement in this PR. Tracking as part of CAI-7310.

internal/crypto/src/time_stamp/verify.rs
scouten-adobe: Won't implement in this PR. Tracking as CAI-7311.

internal/crypto/src/time_stamp/provider.rs
scouten-adobe: Won't implement in this PR. I don't really agree with the feedback, so not tracking as an issue. If there's a larger pattern change, let's discuss in architecture/API review.

internal/crypto/src/time_stamp/http_request.rs
scouten-adobe: Won't implement in this PR. Tracking as CAI-7312.

scouten-adobe:
@mauricefisher64 Almost there: Still need to bring over `claim_v2`'s version of `get_tst_info_from_signed_data`. Out of gas for tonight; will look at it when I spin up tomorrow.

mklein13 (approved):
Discussed offline. Suggested updates will be addressed after this is merged.

internal/crypto/src/time_stamp/provider.rs
mauricefisher64: Yeah, not sure what a  build would build.  Also, timestamp in in the Signer interface and only used when we are doing the fetching of the response.  Signers can implement their own implementation and return the time stamp response.

internal/crypto/src/time_stamp/response.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    /// Return `true` if the request was successful.
+    pub fn is_success(&self) -> bool {
+        matches!(
+            self.0.status.status,
+            PkiStatus::Granted | PkiStatus::GrantedWithMods
+        )
+    }
+
+    /// TO REVIEW: Does this need to be public?
```
mauricefisher64, line 59: I don't think any of these need to be public.  They may be needed by verify.rs however.

internal/crypto/src/time_stamp/response.rs
```diff
@@ -0,0 +1,192 @@
...
+
+#[derive(AsnType, Clone, Debug, Decode, Encode, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct ContentInfo {
+    content_type: rasn::types::ObjectIdentifier,
+
+    #[rasn(tag(explicit(0)))]
+    content: rasn::types::Any,
+}
+
+/// TO REVIEW: Does this need to be public after refactoring?
```
mauricefisher64, line 135: This is used by verify_timestamp functions.

internal/crypto/src/time_stamp/verify.rs
```diff
@@ -0,0 +1,369 @@
...
+    },
+};
+
+/// Decode the TimeStampToken info and verify it against the supplied data.
+///
+/// TEMPORARILY PUBLIC while refactoring
+#[async_generic]
+pub fn verify_time_stamp(ts: &[u8], data: &[u8]) -> Result<TstInfo, TimeStampError> {
+    // Did the time stamp expire between issuance and verification?
+    let Some(sd) = signed_data_from_time_stamp_response(ts)? else {
```
mauricefisher64, line 44: this sigTst2 compatibility this should be:  

 let Ok(Some(sd)) = get_timestamp_signed_data(ts) else {
        return Err(Error::CoseInvalidTimeStamp);
    };

internal/crypto/src/time_stamp/response.rs
scouten-adobe: Good catch. Downgraded to `pub(crate)` and dead-stripped.

internal/crypto/src/time_stamp/response.rs
scouten-adobe: Still used, but downgraded to `pub(crate)`

internal/crypto/src/time_stamp/verify.rs
scouten-adobe: I updated `signed_data_from_timestamp_response` to include the changes you had made in `get_timestamp_signed_data`.

PR #702, opened 2024-11-26
fix: Correct usage of `#[cfg]` directives

kornelski:
The `make_test_images` crate enables the `file_io` and `openssl`, which means that when `cargo test --all-targets` is used, these features will always be enabled. That made mistakes in applying `cfg(feature = "openssl")` undetectable in tests, and the flags indeed weren't quite correct.

In need to run the SDK in a sandbox that can't have any `file_io`, and I'll probably need to disable `openssl` too, so I need these combinations of features to work.

`v1_api` is not possible to disable — too much code relies on `CAIRead` and `ManifestStore`. I haven't tried fixing that feature flag, since you'll probably going to remove it soon anyway. Once `v1_api` is fixed, `cargo hack --each-feature --no-dev-deps check` works.

I haven't tried to remove unused `use` imports, since they're harmless, and hiding them exactly when needed would require a lot of extra `#[cfg]`s.

### DIFF
--- a/export_schema/Cargo.toml
+++ b/export_schema/Cargo.toml
@@ -8,6 +8,6 @@ rust-version = "1.76.0"
 
 [dependencies]
 anyhow = "1.0.40"
-c2pa = { path = "../sdk", features = ["json_schema", "unstable_api"] }
+c2pa = { path = "../sdk", default-features = false, features = ["json_schema", "unstable_api"] }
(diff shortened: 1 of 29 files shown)


scouten-adobe:
@kornelski I picked up the changes you made here, merged them with current `main`, and incorporated them into the release via PR #783. Thank you for taking the time to improve our project!

PR #704, opened 2024-11-29
fix: Verbose assertions for `is_none()`

kornelski:
This prints the `Some(…)` value on failure, which makes failures easier to diagnose.


gpeacock (approved):
Yes, I like this pattern and have been moving in that direction already. Thanks!

PR #705, opened 2024-12-03
feat: Move time stamp code into c2pa-crypto

scouten-adobe:
~~This will eventually be merged into #696, but I'm tracking the changes I previously made and ensuring they build incrementally.~~

This replaces #696.


scouten-adobe:
Closed in favor of #696. I pushed those changes there so as to retain PR feedback from @mklein13 and @andrewhalle.

PR #716, opened 2024-12-06
feat: Add `RawSigner` trait to `c2pa-crypto` (derived from `c2pa::Signer`)

scouten-adobe:


scouten-adobe:
Work in progress: I plan, as part of this PR, to move the signer implementations in to c2pa_crypto. This part moves most of the `Signer` trait into c2pa_crypto (here known as `RawSigner`, since it only implements the core signature algorithm).

More to come tomorrow …

dyro (approved):

mauricefisher64 (approved):
Approved.

PR #775, opened 2024-12-12
feat: Move `TrustHandlerConfig` trait into `c2pa-crypto` crate

scouten-adobe:


scouten-adobe:
Withdrawn in favor of #779.

PR #779, opened 2024-12-13
feat: Introduce `c2pa_crypto::CertificateAcceptancePolicy`

scouten-adobe:
This replaces the `TrustHandler` trait and its implementations in the core `c2pa` crate. Turns out most of what was there is not platform-specific and can be consolidated down into a single platform-generic struct.


scouten-adobe:
There are a couple of unit test failures in this build which I will investigate as I have time over the weekend.

internal/crypto/src/cose/certificate_acceptance_policy.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// > the validator’s trust anchor store for C2PA signers.
+    ///
+    /// This function reads zero or more X.509 root certificates in PEM format
+    /// and configures the trust handler to accept certificates that chain up to
+    /// these trust anchors.
+    ///
+    /// [§14.4.1, C2PA Signers]: https://c2pa.org/specifications/specifications/2.1/specs/C2PA_Specification.html#_c2pa_signers
+    pub fn add_trust_anchors(
+        &mut self,
+        trust_anchor_pems: &[u8],
```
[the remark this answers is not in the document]
scouten-adobe: Probably mostly because it's more easily compatible with `std::fs::read` and `include_bytes!` but I could b persuaded to change it to `&str`.

internal/crypto/src/cose/certificate_acceptance_policy.rs
```diff
@@ -0,0 +1,258 @@
...
+
+impl Default for CertificateAcceptancePolicy {
+    fn default() -> Self {
+        let mut this = CertificateAcceptancePolicy {
+            trust_anchor_ders: vec![],
+            end_entity_cert_ders: vec![],
+            additional_ekus: HashSet::default(),
+        };
+
+        this.add_valid_ekus(include_bytes!("./valid_eku_oids.cfg"));
```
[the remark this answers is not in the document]
scouten-adobe: You can use `new` instead of `default` to bypass the pre-configured EKU list.

internal/crypto/src/cose/certificate_acceptance_policy.rs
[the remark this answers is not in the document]
mklein13: What about using the Pem type vs the raw string/bytes?

internal/crypto/src/cose/certificate_acceptance_policy.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        // In testing configs, also add debug/trust anchors.
+        #[cfg(test)]
```
mauricefisher64, line 46: Will this work for unit tests in c2pa-rs.  We plan to turn on trust checks by default for all unit tests since trust is required for 2.x.

internal/crypto/src/cose/certificate_acceptance_policy.rs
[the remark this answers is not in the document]
mauricefisher64: String in many languages have limits.  Also you avoid string processing which is easy to mess up.  Let the string processing happen in the PEM libraries.

internal/crypto/src/cose/certificate_acceptance_policy.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        if eku.ocsp_signing {
+            return Some(OCSP_SIGNING_OID.clone());
+        }
+
+        // TO REVIEW: Earlier implementation used the last match; this one uses the
+        // first. Does that make a difference?
+
+        for extra_oid in eku.other.iter().as_ref() {
+            let extra_oid_str = extra_oid.to_string();
```
mauricefisher64, line 235: I don't think this behavior is correct.  Each "." in the pattern means something.  Matching a subset is incorrect.  Look at the old code.

internal/crypto/src/cose/certificate_acceptance_policy.rs
scouten-adobe: The root certs that are added at line 49 are only the test certs. When we have the C2PA root bundle, we could add another `add_trust_anchors` call with those roots and not gate that call on `#[cfg(test)]`.

internal/crypto/src/cose/certificate_acceptance_policy.rs
scouten-adobe: `self.additional_ekus` is defined as `HashSet<String>`. Unless I've misinterpreted that data type, `contains` should only return `true` if the **entire** string presented is an exact match for one of the strings in the hash set. I don't think it's prone to the subset match that you're concerned about.

scouten-adobe: The one thing that is known to be different from the old code is that this will return the _first_ OID in the cert's OID list that matches rather than the last. I couldn't find a case where that was important, but if it is, I can adapt.

internal/crypto/src/cose/certificate_acceptance_policy.rs
scouten-adobe: @mauricefisher64 see https://github.com/contentauth/c2pa-rs/pull/779/files#diff-3867f533b55dbd88258138ed6ea7d7127ee605b3b68750c6b0174baa6efd76ebR222-R229 (just added)

Issue #663, opened 2024-11-11
Replace `instant` crate with `web_time`


scouten-adobe:
The `instant` crate is no longer maintained and is the subject of a new RUSTSEC advisory. Authors recommend replacing it with [`web_time`](https://crates.io/crates/web-time).

I've added a temporary exemption for the security alert, but let's look into replacing it ASAP.

PR #784, opened 2024-12-17
Move `verify_trust` into `c2pa_crypto`

scouten-adobe:


scouten-adobe:
DRAFT. TO DO:

- [ ] Move WASM version of `verify_trust` over
- [ ] Remove `c2pa-rs` versions of functions and rewire code in `cose_validator` to use this

internal/crypto/src/cose/certificate_profile.rs
```diff
@@ -0,0 +1,518 @@
...
+    x509::{AlgorithmIdentifier, X509Version},
+};
+
+use crate::{asn1::rfc3161::TstInfo, cose::CertificateTrustPolicy};
+
+/// Verify that an X.509 certificate meets the requirements stated in [§14.5.1,
+/// Certificate Profiles].
+///
+/// [§14.5.1, Certificate Profiles]: https://c2pa.org/specifications/specifications/2.1/specs/C2PA_Specification.html#_certificate_profiles
+pub fn check_certificate_profile(
```
scouten-adobe, line 34: This is the function formerly known as `check_cert` in `cose_validator.rs`.

internal/crypto/src/tests/fixtures/raw_signature/ps256.pub
```diff
@@ -1,36 +1,76 @@
 -----BEGIN CERTIFICATE-----
```
scouten-adobe, line 1: These test fixtures were regenerated to match what was in the `c2pa-rs` test fixture directory.

In a prior PR, I had generated new `ps***` certs and private keys, but that turned out to be unnecessary.

internal/crypto/src/webcrypto/check_certificate_trust.rs
```diff
@@ -0,0 +1,286 @@
...
+    SigningAlg,
+};
+
+pub(crate) async fn check_certificate_trust(
+    ctp: &CertificateTrustPolicy,
+    chain_der: &[Vec<u8>],
+    cert_der: &[u8],
+    _signing_time_epoch: Option<i64>,
+) -> Result<(), CertificateTrustError> {
+    // TO DO: Add verification of signing time.
```
scouten-adobe, line 34: May not be relevant since we do this in `check_certificate_profile` above,

sdk/Cargo.toml
```diff
@@ -136,7 +136,6 @@ image = { version = "0.24.7", default-features = false, features = [
     "png",
 ], optional = true }
 instant = "0.1.12"
-openssl = { version = "0.10.61", features = ["vendored"], optional = true }
```
scouten-adobe, line 139: ‼️

sdk/src/cose_validator.rs
```diff
@@ -106,403 +79,6 @@ fn get_cose_sign1(
     }
 }
 
-pub(crate) fn check_cert(
```
scouten-adobe, line 109: Moved into `c2pa_crypto` crate above.

sdk/src/error.rs
```diff
@@ -288,10 +288,6 @@ pub enum Error {
     #[error("could not acquire OpenSSL FFI mutex")]
     OpenSslMutexError,
```
scouten-adobe, line 290: There are probably more error codes I could remove. Will do that in a subsequent PR.

sdk/src/salt.rs
```diff
@@ -55,20 +55,11 @@ impl Default for DefaultSalt {
 
 impl SaltGenerator for DefaultSalt {
     fn generate_salt(&self) -> Option<Vec<u8>> {
-        #[cfg(feature = "openssl_sign")]
+        #[cfg(target_arch = "wasm32")]
         {
-            let mut salt = vec![0u8; self.salt_len];
```
scouten-adobe, line 60: @mauricefisher64 check this one especially. Is there something special about `openssl`'s implementation of `rand_bytes` that we'd want to preserve? LMK and I'll adapt if so.

scouten-adobe:
Fixes #663

internal/crypto/src/cose/certificate_trust_policy.rs
```diff
@@ -54,21 +57,67 @@ impl Default for CertificateAcceptancePolicy {
...
+    /// period of validity.
+    #[allow(unused)] // parameters may be unused in some cases
+    #[async_generic]
+    pub fn check_certificate_trust(
+        &self,
+        chain_der: &[Vec<u8>],
+        end_entity_cert_der: &[u8],
+        signing_time_epoch: Option<i64>,
+    ) -> Result<(), CertificateTrustError> {
+        if _async {
```
mauricefisher64, line 93: Do we support the sync wasm path now?

scouten-adobe: No. The non `_async` case will fall through to the error response at line 116.

internal/crypto/src/openssl/check_certificate_trust.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub(crate) fn check_certificate_trust(
+    ctp: &CertificateTrustPolicy,
+    chain_der: &[Vec<u8>],
+    cert_der: &[u8],
+    signing_time_epoch: Option<i64>,
+) -> Result<(), CertificateTrustError> {
+    // First check to see if the certificate appears on the allowed list of
+    // end-entity certificates.
+    if ctp.end_entity_cert_ders().any(|der| der == cert_der) {
```
mauricefisher64, line 32: Looks like you removed this behavior:

// check the cert against the allowed list first
    let cert_sha256 = hash_sha256(ee_der);
    let cert_hash_base64 = base64::encode(&cert_sha256);
    if th.get_allowed_list().contains(&cert_hash_base64) {
        return Ok(true);
    }
This was an optimization we use for the Verify site.  Verify only stores the hashes of the der encoded pems.  That greatly reduced the size of the payload and speed of lookup that Verify performs.

mauricefisher64: Same thing in the wasm case.

internal/crypto/src/webcrypto/check_certificate_trust.rs
mauricefisher64: I think we can remove this comment since we check signing cert in cose_sign

internal/crypto/src/openssl/check_certificate_trust.rs
scouten-adobe: Ah, I wasn't aware of that use case, but I'll add it back.

sdk/src/salt.rs
mauricefisher64: The only requirement is it should be a cryptographically random number generator.

internal/crypto/src/cose/certificate_trust_policy.rs
mauricefisher64: Ok, a todo for Colin's case.

internal/crypto/src/openssl/check_certificate_trust.rs
mauricefisher64: If you look at load_allowed_list in the old versions you will see we allow you to use either PEM or DER hashes.  Verify uses DER hashes.

mauricefisher64: Note this code should go away one day when we stop supporting end-entity certs.

internal/crypto/src/cose/certificate_trust_policy.rs
```diff
@@ -28,8 +30,9 @@ pub struct CertificateTrustPolicy {
     /// Trust anchors (root X.509 certificates) in DER format.
     trust_anchor_ders: Vec<Vec<u8>>,
 
-    /// End-entity certificaters (root X.509 certificates) in DER format.
-    end_entity_cert_ders: Vec<Vec<u8>>,
+    /// Base-64 encoded SHA-256 has of end-entity certificaters (root X.509
```
scouten-adobe, line 33: Typo: `has` should be `hash`

scouten-adobe: Will fix after meeting

mauricefisher64 (approved):
Looks good

internal/crypto/src/openssl/check_certificate_trust.rs
scouten-adobe: Actually, if I read this right, we were already using DERs under the hood here. Which means Verify should be fine. And makes sense since the underlying API only accepts DER-encoded end-entity certs.

So I think this is OK as is. (Happy to change after merging if that's wrong.)

PR #801, opened 2024-12-24
feat: Move COSE signature verification into `c2pa_crypto`

scouten-adobe:


scouten-adobe:
There are some known test failures here; I'll clean those up later. (End of my flight-delay-induced working session. 🎅 )

PR #707, opened 2024-12-04
feat: Claim v2

mauricefisher64:
## Changes in this pull request
Support V2 Claim processing

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [ \x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #811, opened 2025-01-02
fix: Bring `claim_v2` changes from #707 into `c2pa_crypto`

scouten-adobe:
Review #707 for changes that should be incorporated into `c2pa_crypto` crate.

The biggest change is the introduction of the `TimeStampStorage` enum, which can select V1 (now deprecated) or V2 (new as of C2PA 2.x) handling of `sigTst` or `sigTst2` time stamp headers.

### DIFF
--- a/sdk/src/cose_sign.rs
+++ b/sdk/src/cose_sign.rs
@@ -16,7 +16,9 @@
 #![deny(missing_docs)]
 
 use async_generic::async_generic;
-use c2pa_crypto::cose::{check_certificate_profile, sign, sign_async, CertificateTrustPolicy};
+use c2pa_crypto::cose::{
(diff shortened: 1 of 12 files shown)


sdk/src/cose_sign.rs
```diff
@@ -90,9 +93,15 @@ pub fn sign_claim(claim_bytes: &[u8], signer: &dyn Signer, box_size: usize) -> R
 #[async_generic(async_signature(
     signer: &dyn AsyncSigner,
     data: &[u8],
-    box_size: usize
+    box_size: usize,
+    time_stamp_storage: TimeStampStorage,
```
scouten-adobe, line 97: @mauricefisher64 this will match up with changes you've made over in #707.

PR #813, opened 2025-01-02
feat: Review `c2pa-crypto` crate API

scouten-adobe:


internal/crypto/src/asn1/rfc3161.rs
```diff
@@ -24,11 +24,6 @@ use crate::asn1::{rfc4210::PkiFreeText, rfc5652::ContentInfo};
 /// 1.2.840.113549.1.9.16.1.4
 pub const OID_CONTENT_TYPE_TST_INFO: ConstOid = Oid(&[42, 134, 72, 134, 247, 13, 1, 9, 16, 1, 4]);
 
-/// id-aa-timeStampToken
```
scouten-adobe, line 27: I made `mod asn1` private, which led to a lot of dead-stripping such as this.

PR #819, opened 2025-01-07
chore: Install `cargo udeps` directly rather than using action

scouten-adobe:
The `cargo-udeps-action` has not been updated in more than a year and no longer operates on GitHub's Ubuntu runner.


scouten-adobe:
This PR is a workaround for https://github.com/aig787/cargo-udeps-action/issues/6.

PR #835, opened 2025-01-10
chore(crypto): Remove `openssl` feature flag

scouten-adobe:
Use OpenSSL whenever not building for WASM architecture.


sdk/Cargo.toml
```diff
@@ -32,8 +32,8 @@ file_io = ["openssl_sign"]
 serialize_thumbnails = []
 no_interleaved_io = ["file_io"]
 fetch_remote_manifests = []
-openssl = ["c2pa-crypto/openssl"]
-openssl_sign = ["openssl", "c2pa-crypto/openssl"]
+openssl = []
```
scouten-adobe, line 35: Should we remove the feature here, too?

PR #853, opened 2025-01-15
feat(crypto): Add `rsa` crate support to `rust_native_crypto` feature

scouten-adobe:
Also makes RSA-PSS signatures available in WASM.


internal/crypto/src/raw_signature/rust_native/signers/rsa_signer.rs
```diff
@@ -0,0 +1,181 @@
...
+            RsaSigningAlg::Ps512 => {
+                let s = SigningKey::<Sha512>::new(self.private_key.clone());
+                let sig = s.sign_with_rng(&mut rng, data);
+                Ok(sig.to_bytes().to_vec())
+            }
+        }
+    }
+
+    fn reserve_size(&self) -> usize {
+        1024 + self.cert_chain_len + self.time_stamp_size
```
mauricefisher64, line 161: This should probably include OCSP size if part of this signer

mauricefisher64 (approved):
Approved but I had some comments/ questions

internal/crypto/src/raw_signature/webcrypto/async_validators/mod.rs
```diff
@@ -62,21 +45,10 @@ pub(crate) fn async_validator_for_sig_and_hash_algs(
         } else if hash_alg.as_ref() == SHA512_OID.as_bytes() {
             return async_validator_for_signing_alg(SigningAlg::Es512);
         }
-    } else if sig_alg.as_ref() == ED25519_OID.as_bytes() {
```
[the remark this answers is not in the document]
scouten-adobe: Yes, it now goes through the `rust_native` path. The scope of `webcrypto` is now limited to what isn't (yet?) implemented in `rust_native`.

internal/crypto/src/raw_signature/rust_native/signers/rsa_signer.rs
scouten-adobe: @mauricefisher64 do we have a way to estimate OCSP size ahead of time?

PR #854, opened 2025-01-16
chore: Additional exports to enable easy unit tests writing

tmathern:
Additional exports to enable easy unit tests writing, based on code from https://github.com/contentauth/c2pa-rs/blob/main/cawg_identity/src/tests/claim_aggregation/interop.rs.

## Checklist


scouten-adobe (changes requested):
This PR is not viable as proposed because it introduces a circular dependency between `cawg_identity` and `c2pa-rs`. Circular crate dependencies are not allowed in Rust.

tmathern:
I moved to something else for the unit tests.

PR #856, opened 2025-01-16
fix(c2patool): Upload a distinct SBOM per platform

scouten-adobe:
(No-op change to c2patool's Cargo.toml to trigger new builds when we release again.)

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,9 +1,7 @@
 [package]
 name = "c2patool"
 default-run = "c2patool"
-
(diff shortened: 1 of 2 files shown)


cli/Cargo.toml
```diff
@@ -1,9 +1,7 @@
 [package]
 name = "c2patool"
 default-run = "c2patool"
-
```
scouten-adobe, line 4: No-op change to cause c2patool to be updated next time we do any release.

PR #864, opened 2025-01-21
fix: Use correct byte label for GIF Plain Text Extension

ok-nick:
Noticed that the plain text extension is being matched under `0x21` (the extension introducer) instead of `0x01` (the plain text extension label).

2025-01-20 at 20 50 35

## Checklist


gpeacock (approved):
Thanks Nick!

PR #869, opened 2025-01-22
feat: Change the definition of `Signer.raw_signer()` to return an `Option` defaulting to `None`

scouten-adobe:
Removes the need for each `Signer` implementation to also provide `RawSigner` and `TimeStampProvider` implementations.


scouten-adobe:
Will do same for `AsyncSigner` before completing this PR.

PR #873, opened 2025-01-23
Fix panic in decoding of GIF chunks

kornelski:
Empty data causes capacity to become negative, which wraps around and causes "capacity overflow" panic.

This changes the GIF parser to correctly read chunk lengths.


sdk/src/asset_handlers/gif_io.rs
```diff
@@ -1118,19 +1118,26 @@ impl DataSubBlocks {
     }
 
     fn to_decoded_bytes(&self) -> Vec<u8> {
-        // Amount of bytes - (length markers + terminator).
-        let mut bytes = Vec::with_capacity(self.bytes.len() - (self.bytes.len().div_ceil(255) + 1));
-        for chunk in self.bytes.chunks(256) {
-            bytes.extend_from_slice(&chunk[1..]);
+        let mut bytes = Vec::with_capacity(gif_chunks(&self.bytes).map(|c| c.len()).sum());
```
gpeacock, line 1121: If you call gif_chunks once and save the value you can avoid iterating twice.

gpeacock (changes requested):

sdk/src/asset_handlers/gif_io.rs
gpeacock: I'll fix this up later. Thanks for the fix!

gpeacock (approved):

sdk/src/asset_handlers/gif_io.rs
kornelski: The iterator is super cheap. It's intentionally run twice, because it avoids reallocation that's more expensive.

PR #879, opened 2025-01-23
feat(crypto): Make `box_size` parameter on `c2pa_crypto::cose::sign` an `Option`

scouten-adobe:
Signing via CAWG identity SDK for X.509 credentials does the padding at a different stage so I need to disable it here.

### DIFF
--- a/internal/crypto/src/cose/sign.rs
+++ b/internal/crypto/src/cose/sign.rs
@@ -76,13 +76,13 @@ use crate::{
 #[async_generic(async_signature(
     signer: &dyn AsyncRawSigner,
     data: &[u8],
-    box_size: usize,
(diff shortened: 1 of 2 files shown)


gpeacock (approved):
I thought you weren't a fan of Option parameters ;) I believe we still have copies of the cose signers in the c2pa-rs sdk. Perhaps it is time to just remove those, since a user could call the crypto versions directly. I'm not sure how that would impact our signer though.

scouten-adobe:
@gpeacock I've stripped away some of those signers for common test cases, but some remain because the `c2pa::Signer` interface has things that aren't in (and shouldn't be in) the `RawSigner` interface.

PR #881, opened 2025-01-24
fix: Remove dependency on SubtleCrypto

mauricefisher64:
Complete rust native support
Fix broken certificate dump
Support arbitrary curves for EC signatures

## Checklist


internal/crypto/src/tests/raw_signature/rust_native/validators.rs
```diff
@@ -33,7 +30,7 @@ fn es256() {
 }
 
 #[test]
-#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+#[cfg(any(target_arch = "wasm32", feature = "rust_native_crypto", test))]
```
gpeacock, line 33: Could rust_native_crypto be used outside a wasm32 architecture? I'm wondering if we need to specify wasm32 or just use the rust_native_crypto feature here?

gpeacock (approved):

internal/crypto/src/tests/raw_signature/rust_native/validators.rs
mauricefisher64: I did not know who might be depending on that feature being present, like Verify.

sdk/src/asset_handlers/mp3_io.rs
```diff
@@ -94,7 +94,7 @@ impl ID3V2Header {
     }
 
     fn decode_tag_size(n: u32) -> u32 {
-        n & 0xff | (n & 0xff00) >> 1 | (n & 0xff0000) >> 2 | (n & 0xff000000) >> 3
+        (n & 0xff) | ((n & 0xff00) >> 1) | ((n & 0xff0000) >> 2) | ((n & 0xff000000) >> 3)
```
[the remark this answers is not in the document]
mauricefisher64: Clippy's warning was to add parens because most people don't know the precedence rules.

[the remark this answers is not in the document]
tmathern: Parenthesis makes also parsing easier while reading :)

sdk/src/utils/test.rs
```diff
@@ -438,6 +438,7 @@ impl crate::signer::RemoteSigner for TempRemoteSigner {
     }
 }
 
+/* This test should be replaced by a rust_native signer if desired to sign from wasm
```
tmathern, line 441: Should we even keep it then?
At least flag it with a TO-DO so it's easy to find?

internal/crypto/src/tests/raw_signature/rust_native/validators.rs
```diff
@@ -52,7 +49,7 @@ fn es256_bad_signature() {
 }
 
 #[test]
-#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
```
scouten-adobe, line 55: We need the `cfg_attr` to remain present so that we keep these test running for WASM.

Sadly, the `#[test]` attribute is ignored by the WASM test runner.

I'll add them back. Marking changes requested, but I'll make the changes.

scouten-adobe (changes requested):
Please do not merge until I've had a chance to make edits and to restore the WASM platform tests. ETA 20-30 minutes.

scouten-adobe (approved):

scouten-adobe (changes requested):
Shoot. I just forgot that I was going to update the README. Please hold until I do that.

scouten-adobe (approved):

scouten-adobe (approved):

tmathern (approved):

sdk/src/assertions/actions.rs
```diff
@@ -119,8 +122,9 @@ pub struct Action {
 
     /// This is NOT the instanceID in the spec
     /// It is now deprecated but was previously used to map the action to an ingredient
-    #[serde(rename = "instanceId", skip_serializing)] // this should never be written to CBOR
     #[deprecated(since = "0.37.0", note = "Use `org.cai.ingredientIds` instead")]
+    #[serde(skip_serializing)]
+    #[serde(alias = "instanceId", alias = "instanceID")]
```
gpeacock, line 127: There's one possible side effect to this that could be an issue. We also use serializing when I save a Builder archive. In that case the instanceId field would get lost. But it is deprecated now in favor of using an ingredient_ids field. So maybe that is ok.

sdk/src/assertions/actions.rs
```diff
@@ -109,6 +109,9 @@ pub struct Action {
     #[serde(skip_serializing_if = "Option::is_none")]
     changed: Option<String>,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    description: Option<String>,
```
gpeacock, line 113: I think whatever test case we have that has this field in it should be removed. We should never have supported a this field here. It seems wrong to add this just to fix an incorrect test case.

gpeacock (changes requested):
I don't think we should add a description field to Action. We never had it there before. The test case was wrong. I'm not sure how we ever generated it.

gpeacock (approved):

scouten-adobe (approved):

PR #888, opened 2025-01-29
feat: Wasm32 wasi 0.41.0

cdmurph32:
feat: Add wasi compatibility.

- wasip2 will require +nightly until https://github.com/rust-lang/rust/issues/130323 is resolved and/or std::os::wasip2 is available in stable.
- Support was added to rustix for version 0.38.39 https://github.com/bytecodealliance/rustix/pull/1205
- Support was added to tempfile for version 3.14 https://github.com/Stebalien/tempfile/pull/305

Related issues:
- https://github.com/WebAssembly/wasi-cli/pull/54
- https://github.com/bytecodealliance/wasm-tools/issues/2047

Tests passing with WASI.
CARGO_TARGET_WASM32_WASIP2_RUNNER="wasmtime -S common -S http --dir ."

## Checklist


sdk/src/utils/io_utils.rs
```diff
@@ -141,6 +147,31 @@ impl<R: Read + Seek> ReaderUtils for R {
     }
 }
 
+pub(crate) fn tempfile_builder<T: AsRef<OsStr> + Sized>(prefix: T) -> Result<NamedTempFile> {
+    #[cfg(target_os = "wasi")]
```
cdmurph32, line 151: This is not available in p1, only p2 and later.

mauricefisher64:
How does this work for crypto?  Does it use the new rust native path that we have for wasm?

cdmurph32:
> How does this work for crypto? Does it use the new rust native path that we have for wasm?

Yes. Removing WebCrypto made that very straightforward.

sdk/src/utils/io_utils.rs
cdmurph32: Done

internal/crypto/Cargo.toml
```diff
@@ -108,6 +108,11 @@ num-bigint-dig = "0.8.4"
 pkcs1 = "0.7.5"
 rsa = { version = "0.9.7", features = ["pem", "sha2"] }
 spki = "0.7.3"
+
+[target.'cfg(all(target_arch = "wasm32", not(target_os = "wasi")))'.dependencies]
```
mauricefisher64, line 112: I think this whole dependency can go.

mauricefisher64:
Does this code rely on the fact that the wasm target will bring in rust_native crypto?  If so would it be better to just have a wasi target instead of a side effect.  Otherwise this could be easily broken without someone knowing they are affecting wasi.

internal/crypto/Cargo.toml
cdmurph32: I can do a wasm32-unknown-unknown cleanup later. Some of these dependencies are definitely still required.

cdmurph32:
> Does this code rely on the fact that the wasm target will bring in rust_native crypto? If so would it be better to just have a wasi target instead of a side effect. Otherwise this could be easily broken without someone knowing they are affecting wasi.

The wasm32 target_arch covers all targets of that architecture, including wasm32-wasi. That target_arch is used intentionally .

sdk/src/utils/test.rs
```diff
@@ -428,7 +431,7 @@ impl WebCryptoSigner {
     }
 }
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
```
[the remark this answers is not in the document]
cdmurph32: Within the sdk, the signer code path is the same for both wasm and native code.






We don't need special tests for wasm anymore with the signer.

Looks like the test code in that util is used by the builder unit tests. They are covered in the new CI job this PR adds.
```
test builder::tests::test_builder_remote_sign ... ok
test builder::tests::test_builder_remote_url ... ok
test builder::tests::test_builder_sign ... ok
test builder::tests::test_builder_sign_assets ... ok
```

mauricefisher64 (approved):
Approved with comments

PR #905, opened 2025-02-04
fix:  P1363 fix to handle incorrect signers

mauricefisher64:
This PR adds support to handle incorrect EC signers with ability to handle arbitrary curves.

## Checklist

### DIFF
--- a/internal/crypto/src/ec_utils.rs
+++ b/internal/crypto/src/ec_utils.rs
@@ -127,6 +127,7 @@ pub(crate) fn der_to_p1363(data: &[u8], sig_len: usize) -> Result<Vec<u8>, RawSi
 }
 
 // Returns supported EcdsaCurve for given public key.
+#[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


internal/crypto/src/ec_utils.rs
```diff
@@ -127,6 +127,7 @@ pub(crate) fn der_to_p1363(data: &[u8], sig_len: usize) -> Result<Vec<u8>, RawSi
 }
 
 // Returns supported EcdsaCurve for given public key.
+#[allow(dead_code)]
```
gpeacock, line 130: Do we need either of these utilities now? Will they serve a purpose in the future or should we just remove them?

gpeacock (approved):

internal/crypto/src/ec_utils.rs
mauricefisher64: Yes, it used by Verify for EC certs that have different curves.  The dead code is because of the feature flags.

PR #906, opened 2025-02-06
fix: Update error reporting

mauricefisher64:
Remove redundant wrapped error codes.
Initial support for new error and status code
Better reporting with correct URIs for Ingredients

## Checklist


sdk/src/claim.rs
```diff
@@ -1781,8 +1818,45 @@ impl Claim {
...
+                        log_item!(
+                            claim.uri(),
+                            "claim signature OCSP value good",
+                            "verify_internal"
+                        )
+                        .validation_status(validation_status::SIGNING_CREDENTIAL_NOT_REVOKED)
+                        .success(validation_log);
+                    } else {
+                        // adjust the error info
+                        if let Some(li) = validation_log.logged_items_mut().last_mut() {
```
scouten-adobe, line 1844: Is the assumption here that `check_ocsp_status` issued a `LogItem` that now needs to be patched with the claim's URI?

scouten-adobe (approved):
One quick clarification; otherwise :+1:.

sdk/src/claim.rs
mauricefisher64: Yes

PR #910, opened 2025-02-07
refactor: use `AsRef<Path>` in `load_jumbf_from_file()`

andrewhalle:
While using this function, I noticed that it takes an explicit `&Path`, which isn't as ergonomic as it could be. All of the functions in [`std::fs`][0] use `AsRef<Path>`, which means that this function could easily take strings as well.

[0]: https://doc.rust-lang.org/stable/std/fs/index.html

I see most of the `from_file()` functions already use `AsRef<Path>`, so this is more of a consistency PR than anything. I think there are still a few occurrences of `&Path` I didn't change; I'm happy to do so, just figured I would start the discussion with this PR.

## Checklist

### DIFF
--- a/sdk/src/jumbf_io.rs
+++ b/sdk/src/jumbf_io.rs
@@ -269,11 +269,11 @@ pub(crate) fn update_file_jumbf(
 
 #[cfg(feature = "file_io")]
 /// load the JUMBF block from an asset if available
-pub fn load_jumbf_from_file(in_path: &Path) -> Result<Vec<u8>> {
-    let ext = get_file_extension(in_path).ok_or(Error::UnsupportedType)?;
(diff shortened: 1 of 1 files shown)


gpeacock (approved):
This is a pattern I've generally tried to use in the public apis. It is easy to forget which APIs are actually public.

The jumbf_io logic should eventually go into its own crate so that it would be public from there an no longer directly part of the c2pa_rs API.

andrewhalle:
My way of figuring out what's public is usually to run `cargo doc`. I think it's actually just the `jumbf_io` functions that need updating—everything else already uses this. I'll go ahead and do that now.

scouten-adobe (approved):

PR #911, opened 2025-02-07
feat: Explore with `Reader` `TryInto` `Value` and JSON formatting

gpeacock:
This adds a TryInto serde_json::Value method for Reader.

It also has a demo unit test that reformats the JSON output assertions. I'm not sure if that function belongs in the Reader or not.

This is not ready to merge, but I'm doing a PR for easy comparision.


gpeacock:
I've integrated this work elsewhere, closing this PR

PR #915, opened 2025-02-10
fix: Restore support for claim_generator_hints.

mauricefisher64:
Support for 1.0 Claim claim_generator_hints field that was inadvertently removed.  

## Checklist


andrewhalle (approved):

gpeacock (commented):
Why support this? No one ever implemented it and it wasn't written to any claims. It was briefly in the spec, but removed quickly. Was there some test case that caused this to show up as an error?

mauricefisher64:
We had a sample from a customer and we support other 1.0 features

gpeacock (approved):
Why support this? No one ever implemented it and it wasn't written to any claims. It was briefly in the spec, but removed quickly. Was there some test case that caused this to show up as an error?

I guess it makes sense at this level to not call it an error, but we won't let you produce the field at the upper later or show it in our reports.

mauricefisher64:
Makes sense

PR #922, opened 2025-02-12
docs: Update versions and feature flags

crandmck:

Update crate version in docs.  Update feature flag docs.

Per @mauricefisher64:

- Removed doc for these feature flags: `openssl` and `openssl_sign` 
- Added description of  `pdf`

## Checklist


mauricefisher64:
With the refactoring of our crypto code, openssll, and openssl_sign don't make much sense.  The default behavior is always to use OpenSSL unless we do some wasm/wasi.  These feature flags could be removed if someone goes through and cleans up the code.  I think they are on in unit tests and samples at the moment.  No real functionality is using these flags at the moment.  
"pdf" feature enable PDF read support

crandmck:
I removed the `openssl` and `openssl_sign` flags from the doc, since it sounds like they don't actually do anything even though they are technically supported.

gpeacock (approved):

mauricefisher64:
Why do we have to tell them the what crypto is used.   If you build  c2pa-rs it will use OpenSSL.  If you build c2pa-rs for wasm/wasi it will use RustCrypto.   There is currently nothing for users to do.

PR #924, opened 2025-02-14
feat: Add support for DynamicAssertions in JSON format

gpeacock:
This allows DynamicAssertions to create JSON formatted assertions. Previously only Cbor formatted assertions were supported.

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -2097,7 +2097,7 @@ impl Store {
                     final_assertions.push(User::new(&label, &data).to_assertion()?);
                 }
                 DynamicAssertionContent::Binary(format, data) => {
-                    // todo:: support for pushing binary assertions here
+                    todo!("Binary dynamic assertions not yet supported");
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


sdk/src/store.rs
```diff
@@ -2080,16 +2082,24 @@ impl Store {
...
-            final_assertions.push(UserCbor::new(&label, da_data).to_assertion()?);
+            match da_data {
+                DynamicAssertionContent::Cbor(data) => {
+                    final_assertions.push(UserCbor::new(&label, data).to_assertion()?);
+                }
+                DynamicAssertionContent::Json(data) => {
+                    final_assertions.push(User::new(&label, &data).to_assertion()?);
+                }
+                DynamicAssertionContent::Binary(format, data) => {
+                    // todo:: support for pushing binary assertions here
```
scouten-adobe, line 2100: Suggestion: Add a `todo!(...)` macro here so we don't silently fail in the binary case.

scouten-adobe (approved):
One minor change requested (see earlier comment); otherwise :+1:

sdk/src/store.rs
gpeacock: done!

PR #936, opened 2025-02-20
feat: Configurable `Reader` type

scouten-adobe:


sdk/src/reader.rs
```diff
@@ -86,9 +86,44 @@ impl Reader {
     /// println!("{}", reader.json());
     /// ```
     #[async_generic()]
-    pub fn from_stream(format: &str, mut stream: impl Read + Seek + Send) -> Result<Reader> {
+    pub fn from_stream(format: &str, stream: impl Read + Seek + Send) -> Result<Reader> {
+        let config = ReaderConfig::default();
+
+        if _sync {
```
scouten-adobe, line 92: I only did it for this one, but I would anticipate doing this same default vs configurable approach for the other entry points.

PR #937, opened 2025-02-21
feat: Simplify `StatusTracker` interface

scouten-adobe:
Having `StatusTracker` as a non-object-safe trait made it impossible to use in further trait APIs.

See early draft of #936 for an example of this problem.

This PR changes `StatusTracker` to be a `struct`. It uses a new `enum ErrorBehavior` to describe the only two implementations of `StatusTracker` that have ever existed.


internal/status-tracker/src/status_tracker.rs
```diff
@@ -0,0 +1,169 @@
...
+        for log_item in other.logged_items() {
+            self.add_non_error(log_item.clone());
+        }
+    }
+
+    /// Adds a non-error [`LogItem`] to this status tracker.
+    ///
+    /// Primarily intended for use by [`LogItem::success()`]
+    /// or [`LogItem::informational()`].
+    pub fn add_non_error(&mut self, mut log_item: LogItem) {
```
gpeacock, line 60: Given that LogItem now contains a kind field, there's no need for separate add_non_error and add_error. We should have an add() method that does the right thing. If we aren't ready to update all the calling code, the add_non_error and add_error should simply call add().

internal/status-tracker/src/status_tracker.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Removes the current ingredient URI, if any.
+    pub fn pop_ingredient_uri(&mut self) -> Option<String> {
+        self.ingredient_uris.pop()
+    }
+
+    /// Returns the [`LogItem`]s that have error conditions (`err_val` is
+    /// populated).
+    ///
+    /// Removes matching items from the list of log items.
+    pub fn take_errors(&mut self) -> Vec<LogItem> {
```
gpeacock, line 136: How is this used? it seems really odd to have something that just extracts the errors

mauricefisher64 (approved):
Good simplification

internal/status-tracker/src/status_tracker.rs
scouten-adobe: I find 23 hits in the repo, all in test cases. Since it's been part of the public API for a while (https://docs.rs/c2pa-status-tracker/latest/c2pa_status_tracker/struct.DetailedStatusTracker.html#method.take_errors), I don't know who else is using it.

internal/status-tracker/src/status_tracker.rs
scouten-adobe: This is carried over from the existing APIs. The primary difference is that `add_error` _can_ trigger an `Err` result if the status tracker is configured to stop on first error and in that case, we need to have an `Error` type ready to use.

The `add_non_error` case has no return value, so we don't need to pass in an `Error`.

internal/status-tracker/src/tests/status_tracker.rs
```diff
@@ -41,9 +41,9 @@ mod detailed {
         assert!(!tracker.has_error("Something Else"));
 
         // Verify that one item with error was found.
-        let errors = tracker.take_errors();
+        let errors: Vec<&crate::LogItem> = tracker.filter_errors().collect();
         assert_eq!(errors.len(), 1);
-        assert_eq!(tracker.logged_items().len(), 1);
+        assert_eq!(tracker.logged_items().len(), 2);
```
scouten-adobe, line 46: Since we're not _taking_ errors any more, the expected status of the tracker itself is now different.

PR #940, opened 2025-02-24
refactor: Remove `openssl` feature flag

cdmurph32:

Moving cryptography to c2pa-crypto made the `openssl` and `openssl_sign` flags defunct.

## Checklist


gpeacock (commented):
Thanks, this looks like a very nice code clean up.
I'm curious about how an application will be able to not build with openssl now.
The openssl_sign feature was added for [c2pa_min](https://github.com/contentauth/c2pa-min) which shows how to build c2pa for signing only using no SSL.

cdmurph32:
> Thanks, this looks like a very nice code clean up. I'm curious about how an application will be able to not build with openssl now. The openssl_sign feature was added for [c2pa_min](https://github.com/contentauth/c2pa-min) which shows how to build c2pa for signing only using no SSL.

I think they would add `c2pa-crypto` as an explicit dependency with the feature `rust_native_crypto`. I think that has been true since @scouten-adobe moved the crypto to its own crate.

scouten-adobe (approved):

PR #943, opened 2025-02-25
feat: Add `StatusTracker` to `IdentityAssertion` parsing and validation APIs

scouten-adobe:


cawg_identity/src/identity_assertion/assertion.rs
```diff
@@ -63,9 +64,10 @@ impl IdentityAssertion {
     /// Iterator returns a [`Result`] because each assertion may fail to parse.
     ///
     /// Aside from CBOR parsing, no further validation is performed.
-    pub fn from_manifest(
-        manifest: &Manifest,
-    ) -> impl Iterator<Item = Result<Self, c2pa::Error>> + use<'_> {
+    pub fn from_manifest<'a>(
```
tmathern, line 67: I am always skeptical when I see lifetimes like this appear in a change.
(Remember our tooling for other languages bindings may not take those that well, and to avoid future changes I would just do the minimal amount of lifetimes specifications possible).

cawg_identity/src/tests/builder/simple_case.rs
```diff
@@ -66,17 +67,17 @@ async fn simple_case() {
...
-    let mut ia_iter = IdentityAssertion::from_manifest(manifest);
+    let mut st = StatusTracker::default();
+    let mut ia_iter = IdentityAssertion::from_manifest(manifest, &mut st);
 
     // Should find exactly one identity assertion.
     let ia = ia_iter.next().unwrap().unwrap();
-    dbg!(&ia);
-
     assert!(ia_iter.next().is_none());
+    drop(ia_iter);
```
tmathern, line 76: Why do we need an explicit drop here?

cawg_identity/src/tests/builder/simple_case.rs
```diff
@@ -66,17 +67,17 @@ async fn simple_case() {
     assert_eq!(manifest_store.validation_status(), None);
 
     let manifest = manifest_store.active_manifest().unwrap();
-    let mut ia_iter = IdentityAssertion::from_manifest(manifest);
+    let mut st = StatusTracker::default();
+    let mut ia_iter = IdentityAssertion::from_manifest(manifest, &mut st);
```
tmathern, line 71: Ack - StatusTracker becomes a parameter we carry around in most calls. Correct?

cawg_identity/src/identity_assertion/assertion.rs
scouten-adobe: Can we preflight this somehow? It will be at best difficult and quite likely impossible to make this change without the new lifetime requirement.

cawg_identity/src/tests/builder/simple_case.rs
scouten-adobe: Yes, this is in line with the general approach to handling errors in C2PA parsing and validation.

cawg_identity/src/tests/builder/simple_case.rs
scouten-adobe: This ends the lifetime of the iterator, which makes the status tracker available for mutable use later.

cawg_identity/src/identity_assertion/assertion.rs
```diff
@@ -63,14 +64,31 @@ impl IdentityAssertion {
...
+    ) -> impl Iterator<Item = Result<Self, c2pa::Error>> + use<'a> {
         manifest
             .assertions()
             .iter()
             .filter(|a| a.label().starts_with("cawg.identity"))
-            .map(|a| a.to_assertion())
+            .map(|a| (a.label().to_owned(), a.to_assertion()))
+            .inspect(|(label, r)| {
+                if let Err(err) = r {
+                    // TO DO: a.label() is probably wrong (not a full JUMBF URI)
```
scouten-adobe, line 78: @gpeacock is there a way I can get from `Manifest.assertions()` iterator to the JUMBF URI for the assertion returned?

cawg_identity/src/identity_assertion/assertion.rs
```diff
@@ -222,6 +254,7 @@ impl IdentityAssertion {
     pub async fn validate<SV: SignatureVerifier>(
         &self,
         manifest: &Manifest,
+        _status_tracker: &mut StatusTracker,
```
scouten-adobe, line 257: This placeholder will be used later, not to worry … 🙂

PR #957, opened 2025-03-10
update: Bump x509-certificate to 0.24.0

scouten-adobe:


internal/crypto/Cargo.toml
```diff
@@ -98,9 +99,6 @@ version = "0.4.39"
 default-features = false
 features = ["now"]
 
-[target.'cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))'.dependencies]
```
scouten-adobe, line 101: Per @cdmurph32's suggestion, this reverts the dependency to what was there prior to #934.

scouten-adobe: … but now uses 0.24.0.

sdk/Cargo.toml
```diff
@@ -134,7 +134,6 @@ tempfile = { version = "3.15", features = ["nightly"] }
 ureq = "2.4.0"
 
 [target.'cfg(any(target_os = "wasi", not(target_arch = "wasm32")))'.dependencies]
-x509-certificate = "0.23.1"
```
scouten-adobe, line 137: x509-certificate is no longer referenced by c2pa-rs; that moved entirely to c2pa-crypto.

scouten-adobe:
@cdmurph32 would you please add instructions on building using upstream llvm vs Xcode on Mac?

gpeacock (approved):
Thanks for this!

cdmurph32:
We might need to document the need inability to use XCode clang for wasm. This was already true for WASI, but that was more recently added.

gpeacock:
We will need to update the build servers to have llvm installed, but maybe they already do for wasi??

cdmurph32:
> We will need to update the build servers to have llvm installed, but maybe they already do for wasi??

No need. The build servers use ubuntu for wasm builds.

PR #958, opened 2025-03-11
docs: WASM building and testing

cdmurph32:
## Checklist

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -30,9 +30,12 @@ Add the `add_thumbnails` dependency to generate thumbnails for JPEG and PNG file
 c2pa = { version = "0.45.2", features = ["file_io", "add_thumbnails"] }
 ```
 
+## WebAssembly
+This crate supports compilation to both the `wasm32-unknown-unknown` and `wasm32-wasi*` family of LLVM targets.
+
 ### Building wasm
 
-Building wasm requires LLVM clang due to the `ring` crate. The version of clang that comes with XCode does not support wasm targets.
+Building wasm requires clang due to the `ring` crate. The version of clang that comes with XCode does not support wasm targets.
 
 1. On macOS, install clang with homebrew `brew install llvm`.
 2. Set the `CC` environment variable to the clang binary. Cargo provides a mechanism to do this with .cargo/config.toml in the project root.
@@ -47,7 +50,7 @@ NOTE: As of March 2025, `wasm32-wasip2` still requires the nightly toolchain as 
 - Testing WASI wasm (wasm32-wasip1 and later) requires [wasmtime](https://github.com/bytecodealliance/wasmtime).
 
 ### Example `.cargo/config.toml`
-The following is valid for all macOS targets, not only wasm.
+This config specifies homebrew clang which can build all targets on macOS, not only wasm. The test runner is set for wasm32-wasip2.
 ```
 [env]
 CC = "/opt/homebrew/opt/llvm/bin/clang"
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -30,7 +30,31 @@ Add the `add_thumbnails` dependency to generate thumbnails for JPEG and PNG file
 c2pa = { version = "0.45.2", features = ["file_io", "add_thumbnails"] }
 ```
 
-NOTE: If you are building for WASM, omit the `file_io` dependency.
+### Building wasm
+
+Building wasm requires LLVM clang due to the `ring` crate. The version of clang that comes with XCode does not support wasm targets.
+
+1. On macOS, install clang with homebrew `brew install llvm`.
```
tmathern, line 37: Can you add the minimum llvm version to install (you can pick the one you currently have). So we are sure we know at least one working setup for sure.

docs/usage.md
```diff
@@ -30,7 +30,31 @@ Add the `add_thumbnails` dependency to generate thumbnails for JPEG and PNG file
...
+2. Set the `CC` environment variable to the clang binary. Cargo provides a mechanism to do this with .cargo/config.toml in the project root.
```
tmathern, line 38: Can you add an example? So first-time readers can easily do it.

crandmck (changes requested):

docs/usage.md
cdmurph32: It's difficult to pin down a version, there has been wasm support in LLVM for a very long time. WASI is continuing to evolve, so newer LLVM versions will have better support.

@crandmck I don't understand the spirit of the question. There are many supported LLVM targets for this crate and we don't have any language indicating why you would compile for them. The user will necessarily have a use case outside the scope of this crate and know which target to compile for. No one has to compile for any target if they don't want to.

docs/usage.md
```diff
@@ -30,7 +30,31 @@ Add the `add_thumbnails` dependency to generate thumbnails for JPEG and PNG file
 c2pa = { version = "0.45.2", features = ["file_io", "add_thumbnails"] }
 ```
 
-NOTE: If you are building for WASM, omit the `file_io` dependency.
+### Building wasm
```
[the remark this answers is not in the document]
cdmurph32: Done

docs/usage.md
crandmck: IIUC, one can use c2pa-rs directly (i.e. not for another language binding).  For example, for c2patool.  Maybe I'm misunderstanding.

crandmck:
@scouten-adobe Nit: The correct spelling of the abbreviation for WebAssembly is Wasm, not WASM. Cf. https://webassembly.org/,

docs/usage.md
gpeacock: Note that make test has always tried to build wasm, and that will no longer work on macs without installing llvm. We may want to reconfigure this and make wasm/wasi builds as separate step.

crandmck (approved):

cdmurph32:
> @scouten-adobe Nit: The correct spelling of the abbreviation for WebAssembly is Wasm, not WASM. Cf. https://webassembly.org/.
> 
> > WebAssembly (abbreviated Wasm) is a binary instruction format ...

Yes. We've been a bit inconsistent about it. https://rustwasm.github.io/book/what-is-webassembly.html

crandmck:
> > @scouten-adobe Nit: The correct spelling of the abbreviation for WebAssembly is Wasm, not WASM. Cf. https://webassembly.org/.
> > > WebAssembly (abbreviated Wasm) is a binary instruction format ...
> 
> Yes. We've been a bit inconsistent about it. https://rustwasm.github.io/book/what-is-webassembly.html

As discussed in Slack, we don't need to spend any more time on this.  But I would note that [the WebAssembly spec](https://webassembly.github.io/spec/) uses "Wasm" and specifically calls out the spelling (capitalization).

PR #960, opened 2025-03-11
fix: Fix broken c2patool fragment feature

mauricefisher64:
The output path in this use case is a folder.  So skip the normal file checks.

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -1055,9 +1055,20 @@ impl Builder {
             std::fs::create_dir_all(output_path.as_ref())?;
         } else {
             // if the file exists, we need to remove it
-            return Err(crate::Error::BadParam(
-                "Destination file already exists".to_string(),
-            ));
+            if output_path.as_ref().is_file() {
+                return Err(crate::Error::BadParam(
+                    "output_path must be a folder".to_string(),
+                ));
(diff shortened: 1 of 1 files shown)


sdk/src/builder.rs
```diff
@@ -1050,7 +1050,15 @@ impl Builder {
         fragment_paths: &Vec<std::path::PathBuf>,
         output_path: P,
     ) -> Result<()> {
-        self.set_asset_from_dest(output_path.as_ref())?;
+        if !output_path.as_ref().exists() {
+            // ensure the path exists
+            std::fs::create_dir_all(output_path.as_ref())?;
+        } else {
+            // if the file exists, we need to remove it
```
gpeacock, line 1057: technically this should be "if the folder exists". I'm not positive that's a concern in this case, since we could still write to the folder, and I don't think we have -f/force option wired up here either.

gpeacock (approved):

sdk/src/builder.rs
mauricefisher64: I didn't change the behavior because all of the -f behavior is in c2patool.  It is the case here as well.  If the asset file is found at the output location an error is returned just like other assets.  I will fix the comment.

Issue #967, opened 2025-03-12
JPEG: RST0..RST7 markers should be treated as standalone markers (without length specifications)


brogdonm:
I noticed in the code for box hashing, JPEG asset handler has a method, `has_length`, which is called by the `get_seg_size` method. I read through parts of the specification for JPEG, and I believe the ReSTart markers are standalone segment markers, or in other words do not have a 2 byte length specifier following it.

For the following line: 

https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/asset_handlers/jpeg_io.rs#L681

I believe the `RST0..=RST7` check should be removed.

PR #968, opened 2025-03-12
fix: ISSUE-967: Remove the `RST0..=RST7` check from the `has_length` method for the JPEG asset handler.

brogdonm:

As #967 states, this removes the ReSTart 0 through 7 segment markers from the `has_length` check. As these markers are standalone markers and do not have a 2-byte length specified after them.

## Checklist


mauricefisher64 (approved):
Thanks for the catch

PR #982, opened 2025-03-18
fix: Remove circular dependency between C2PA and CAWG crates

scouten-adobe:


gpeacock (approved):
Note to me to add the cawg.rs demo back into a different location

PR #985, opened 2025-03-19
feat: Generate gathered assertions for v2 claims

gpeacock:
Includes validation of hard bindings when added.
There may be others to add (SoftBindings?) but this is a good start.


tmathern (approved):

sdk/src/assertions/data_hash.rs
```diff
@@ -273,11 +273,6 @@ impl AssertionBase for DataHash {
     const VERSION: Option<usize> = Some(ASSERTION_CREATION_VERSION);
 
     fn to_assertion(&self) -> Result<Assertion> {
-        if self.hash.is_empty() {
```
mauricefisher64, line 276: Is this removal for CAWG?

PR #997, opened 2025-03-26
fix: Make sure manifests are signed with end-entity certs

mauricefisher64:
Adds check to make sure end-entity certs are used where we expect them.

## Checklist

### DIFF
--- a/internal/crypto/src/cose/certificate_profile.rs
+++ b/internal/crypto/src/cose/certificate_profile.rs
@@ -33,6 +33,7 @@ use crate::{asn1::rfc3161::TstInfo, cose::CertificateTrustPolicy};
 /// [§14.5.1, Certificate Profiles]: https://c2pa.org/specifications/specifications/2.1/specs/C2PA_Specification.html#_certificate_profiles
 pub fn check_certificate_profile(
     certificate_der: &[u8],
+    is_end_entity: bool,
     ctp: &CertificateTrustPolicy,
     validation_log: &mut StatusTracker,
     _tst_info_opt: Option<&TstInfo>,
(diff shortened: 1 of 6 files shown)


internal/crypto/src/cose/certificate_profile.rs
```diff
@@ -33,6 +33,7 @@ use crate::{asn1::rfc3161::TstInfo, cose::CertificateTrustPolicy};
 /// [§14.5.1, Certificate Profiles]: https://c2pa.org/specifications/specifications/2.1/specs/C2PA_Specification.html#_certificate_profiles
 pub fn check_certificate_profile(
```
andrewhalle, line 34: Instead of a boolean `is_end_entity`, might I suggest a new function `check_end_entity_certificate_profile()`? That would call this function and then validate that it's an end-entity certificate after.

I find keeping the number of boolean parameters to functions to be beneficial.

mauricefisher64: I would have to parse the cert twice.  I will consider it.

gpeacock: I agree that the boolean argument is harder to follow, but given that this is an internal only API, performance may outweigh readability in this case.

gpeacock (approved):

internal/crypto/src/cose/certificate_profile.rs
andrewhalle: That makes sense! I'm good with it either way; I like the current version, but I'm also fine if you decide to revert to the previous version.

andrewhalle (approved):

PR #998, opened 2025-03-26
chore: Add a CAWG X.509 signing example

gpeacock:
We lost this example due to circular dependencies, so this puts it back in. I set it up in a higher level examples folder so that it can be executed with cargo run --example cawg source.ext dest.ext

### DIFF
--- a/cawg_identity/examples/cawg.rs
+++ b/cawg_identity/examples/cawg.rs
@@ -1,0 +1,137 @@
+// Copyright 2025 Adobe. All rights reserved.
+// This file is licensed to you under the Apache License,
+// Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+// or the MIT license (http://opensource.org/licenses/MIT),
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


cawg_identity/examples/cawg.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Creates a CAWG signer from a certificate chains and private keys.
+    fn async_cawg_signer() -> Result<impl AsyncSigner> {
+        let c2pa_raw_signer = raw_signature::async_signer_from_cert_chain_and_private_key(
+            CERTS,
+            PRIVATE_KEY,
+            SigningAlg::Ed25519,
+            None,
+        )?;
+
+        let cawg_raw_signer = raw_signature::async_signer_from_cert_chain_and_private_key(
```
scouten-adobe, line 84: I'd recommend using different certs for the two cases so we can demonstrate that the right cert goes in the right place.

scouten-adobe (approved):

scouten-adobe:
Looks like Clippy and WASM build are grumpy and I have one small suggestion; otherwise, LGTM.

cawg_identity/examples/cawg.rs
gpeacock: I'll do some more work on this later. I'd like to have a method that accepts two raw signers.

PR #1023, opened 2025-04-04
fix: Manually trigger release of c2patool

scouten-adobe:
No-op change as a workaround to https://github.com/release-plz/release-plz/issues/2164.


scouten-adobe:
Closing in favor of #1024, which incorporates this no-op change.

PR #1024, opened 2025-04-04
fix: Update openssl to address a recently-announced vulnerability

scouten-adobe:
Contains trivial changes to each downstream project as a hedge against https://github.com/release-plz/release-plz/issues/2164.

### DIFF
--- a/cawg_identity/src/validator.rs
+++ b/cawg_identity/src/validator.rs
@@ -37,7 +37,7 @@ impl AsyncPostValidator for CawgValidator {
         partial_claim: &PartialClaim,
         tracker: &mut StatusTracker,
     ) -> c2pa::Result<Option<Value>> {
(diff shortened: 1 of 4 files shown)


tmathern (approved):

cawg_identity/src/validator.rs
```diff
@@ -37,7 +37,7 @@ impl AsyncPostValidator for CawgValidator {
         partial_claim: &PartialClaim,
         tracker: &mut StatusTracker,
     ) -> c2pa::Result<Option<Value>> {
-        if label == "cawg.identity" {
+        if label.starts_with("cawg.identity") {
```
gpeacock, line 40: This is generally the better practice now that assertions are correctly labeled with instances.

gpeacock (approved):

PR #1031, opened 2025-04-07
fix: Adjust dependencies to avoid security warnings and yanked versions

scouten-adobe:


deny.toml
```diff
@@ -28,13 +28,13 @@ allow = [
   "Apache-2.0",
   "BSD-2-Clause",
   "BSD-3-Clause",
-  "CC0-1.0",
+  # "CC0-1.0",
```
gpeacock, line 31: Are these due to a change in policy or to work around build issues?

brogdonm: From what I saw, they are not encountered anymore with the updates to dependencies.

PR #1035, opened 2025-04-09
fix: dup check with validation kind

gpeacock:
Timestamp errors are now informational
maps error & informational codes from from v1 claim ingredients Added timestamp.malformed informational status.
signature errors have correct urls
chore :  update yanked crate
updated timestamp unit tests for informational status


mauricefisher64 (approved):
👍

PR #1047, opened 2025-04-23
feat: Adds c_api for dynamic library releases

gpeacock:
Includes automation tools:
cbindgen c2pa.h header file generation,
building platform specific libraries
Builds arm, x86 and universal libraries 
packages releases into zip file artifacts for github release
library_release.yml script added.
use shared workplace version for sdk, c2pa_c and make_test_images


.github/workflows/library-release.yml
```diff
@@ -0,0 +1,77 @@
+name: Build and Release
+
+on:
+  push:
+    tags:
+      - 'v*' # Trigger on version tags (e.g., v1.0.0)
```
scouten-adobe, line 6: I believe this change is required to match the tags that release-plz generates for the core c2pa-rs crate.

```suggestion
      - 'c2pa-v*' # Trigger on version tags (e.g., v1.0.0)
```

.github/workflows/library-release.yml
```diff
@@ -0,0 +1,77 @@
...
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: 
+          - ubuntu-latest
+          - ubuntu-latest
+          #- macos-latest
```
scouten-adobe, line 17: Why have the duplicated `macos-latest` instance commented out?

Cargo.toml
```diff
@@ -11,6 +12,9 @@ members = [
     "sdk",
 ]
 
+[workspace.package]
```
scouten-adobe, line 15: I'm not yet 100% convinced that release-plz is ready for workspace versioning. Could we please do that in a separate PR? (FYI @cdmurph32 is submitting a PR to release-plz that I think will fix the underlying issue.)

c2pa_c/build.rs
```diff
@@ -0,0 +1,58 @@
...
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+//! This creates the c2pa.h header file and the c2pa_version.txt file
+//! in the target directory. It is intended to be run as part of the build process.
+// The crate version is added to the header file.
```
scouten-adobe, line 16: Was line 16 intended to be part of the doc comment?

scouten-adobe:
Broader question while I'm working on the detailed review: Is the commit history from the original c2pa-c history worth preserving? If so, LMK and I'd be happy to help make that migration happen.

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 resolver = "2"
 
 members = [
+    "c2pa_c",
```
scouten-adobe, line 5: I feel a little weird about using the `c2pa_` prefix for this directory since we don't elsewhere in the project. Perhaps rename the directory to `c_api` or something like that?

scouten-adobe (dismissed):
Overall, :+1:. Sharing a review in progress with some changes requested. Will continue from `c_api.rs` probably tomorrow when I have more time.

.github/workflows/library-release.yml
gpeacock: For the mac I build arm, x86 and universal libraries. In order to build the universal one I need both builds on the same machine. So I only need one mac where I can cross compile to the other build and then combine them into a universal. I distribute all three libraries.

c2pa_c/src/signer_info.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use c2pa::{create_signer, Signer, SigningAlg};
+use serde::Deserialize;
+
+use crate::{Error, Result};
+
+/// SignerInfo provides the information needed to create a signer
+/// and sign a manifest.
+///
+/// The signer is created from the signcert and pkey fields.
```
scouten-adobe, line 21: Why document these in prose here rather than adjacent to the fields?

Also I notice a disconnect between `tsa_url` in the doc comment at `ta_url` in the actual struct.

c_api/build.rs
```diff
@@ -0,0 +1,60 @@
...
+            header.push_str(&format!("\n// Version: {}\n", version));
+            Some(header.clone())
+        }
+        None => Some(format!("\n// Version: {}\n", version)),
+    };
+
+    // Generate the header file.
+    cbindgen::generate_with_config(&crate_dir, config).map_or_else(
+        |error| match error {
+            cbindgen::Error::ParseSyntaxError { .. } => {}
```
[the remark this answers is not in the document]
gpeacock: I added the warning output as suggested

.github/workflows/library-release.yml
```diff
@@ -1,23 +1,116 @@
...
-  CARGO_TERM_COLOR: always
-
 jobs:
   build:
+    name: Publish c2pa-library binaries
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
```
tmathern, line 16: For nightlies, I am all-in for `ubuntu-latest` and `windows-latest`. But for "normal" release builds, I always wonder if we should pin the versions, and update should be a conscious decision.

.github/workflows/library-release.yml
```diff
@@ -1,23 +1,116 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
 
+      - name: Setup Strawberry Perl
+        if: matrix.os == 'windows-latest'
```
tmathern, line 37: Opinion: The fact that windows needs Perl, one of the only write-only languages, is quite a thing.

.github/workflows/library-release.yml
```diff
@@ -1,23 +1,116 @@
...
+    needs: build
+    runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: cargo test --verbose
+      - name: Download build artifacts
+        uses: actions/download-artifact@v3
```
tmathern, line 76: Why not v4? The Python release steps already use v4 => https://github.com/contentauth/c2pa-python/blob/b41be67cd6bc9a23f879e554c7fa8dee070e75d2/.github/workflows/build.yml#L371

.github/workflows/library-release.yml
```diff
@@ -1,23 +1,116 @@
...
+            ./windows-latest-x86_64-pc-windows-msvc-release/windows-latest-x86_64-pc-windows-msvc-release.zip
+            ./macos-latest-x86_64-apple-darwin-release/macos-latest-x86_64-apple-darwin-release.tar.gz
+          draft: false
+          prerelease: false
+          generate_release_notes: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create GitHub Release old
+        id: create_release_old
```
tmathern, line 98: Should we rename all `*old` to `*legacy`? I imagine they'll be retired at some point?

Cargo.toml
```diff
@@ -11,6 +12,8 @@ members = [
     "sdk",
 ]
 
+
+
```
tmathern, line 16: Nitpick: One empty line likely is enough?

c_api/Makefile
```diff
@@ -0,0 +1,107 @@
...
+    echo "Creating zip file: $${ZIP_NAME}"; \
+    mkdir -p $(TARGET_DIR)/artifacts $(1)/include $(1)/lib; \
+    cp $(1)/c2pa.h $(1)/include/; \
+    cp $(1)/libc2pa_c.* $(1)/lib/; \
+    cd $(1) && zip -r $${ZIP_NAME} include lib; \
+    echo "Zip file created: $${ZIP_NAME}"
+endef
+
+check-format:
+	cargo +nightly fmt -- --check
```
tmathern, line 39: Do we want to keep the `+nightly` in the check everywhere?

c_api/Makefile
```diff
@@ -0,0 +1,107 @@
...
+
+test-local:
+	cargo test
+
+# Full local validation, build and test all features including wasm
+# Run this before pushing a PR to pre-validate
+test: check-format clippy test-local
+
+fmt:
+	cargo +nightly fmt
```
tmathern, line 52: Do we want to keep the `+nightly` in the check everywhere?

c_api/build.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright 2023 Adobe. All rights reserved.
```
tmathern, line 1: Is the copyright year off by a few years? File seems new?

c_api/cbindgen.toml
```diff
@@ -0,0 +1,59 @@
+cpp_compat = true
+
+header = """// Copyright 2024 Adobe. All rights reserved.
```
tmathern, line 3: I am never sure about the year dates in those auto-generation cases.

c_api/src/json_api.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use c2pa::{Ingredient, Manifest, Reader};
+use cawg_identity::validator::CawgValidator;
+use tokio::runtime::Runtime;
+
+use crate::{Error, Result, SignerInfo};
+
+/// Returns the version of the c2pa SDK used in this library
+pub fn sdk_version() -> String {
+    String::from(c2pa::VERSION)
```
tmathern, line 21: Note: No covered by tests here, but extensively covered by the languages that bind to the related C native code.

c_api/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+/// This module exports a C2PA library
+mod c2pa_stream;
+mod c_api;
+mod error;
+#[cfg(feature = "json_api")]
```
tmathern, line 17: I like the JSON API. Many other languages now natively or quite well use JSON. Should we enable it by default? (Especially since we offer to create SignerInfo from the json. It increases API surface, but it also makes it feel more... complete?)

tmathern (approved):

.github/workflows/library-release.yml
gpeacock: That's a good idea, I'm just not sure what those versions should be. For testing (not in this script) we try nightly and   a minimal version for compatibility. There are lots of questions about what version of what these libraries should support.
I wanted to go with only MUSL builds for linux, but they don't support dynamic libraries, only binary apps.

.github/workflows/library-release.yml
gpeacock: I share your opinion of Perl ;)

c_api/Makefile
gpeacock: We use nightly only for fmt because the nightly version includes some extra checking no available elsewhere.

c_api/build.rs
gpeacock: Yes, it is, but clippy has decided that every file must have the same copyright date or it will complain bitterly and break our builds. I think that's very wrong policy, but I haven't been fighting it for now.

c_api/src/lib.rs
gpeacock: This json_api feature is for a set of Rust only functions I provided to mirror the exposed C API. I did it so that the python implementation could share some code. Now it serves no purpose and it only includes the older v1 style api functions.

gpeacock:
> Broader question while I'm working on the detailed review: Is the commit history from the original c2pa-c history worth preserving? If so, LMK and I'd be happy to help make that migration happen.

I don't think it is necessary for this. The c2pa_c repo will still exist if we need older history.

gpeacock:
The library_release github action is not working on windows yet, and needs work. but it is only manually triggered at this point. Make release on mac and linux are operational and generated zip file artifacts.
Code converage will get better when I remove the json_api soon.

PR #1048, opened 2025-04-24
chore: clean up temp directory usage in tests

cdmurph32:
A few of the tests were interfering with TempDir's automatic cleanup.

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -6562,7 +6562,7 @@ pub mod tests {
         // test adding to actual image
 
         let tempdir = tempdirectory().expect("temp dir");
-        let output_path = tempdir.into_path();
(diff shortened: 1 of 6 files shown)


sdk/src/store.rs
```diff
@@ -6594,7 +6594,10 @@ pub mod tests {
                     let signer = test_signer(SigningAlg::Ps256);
 
                     // add manifest based on
```
scouten-adobe, line 6596:
```suggestion
-                    // add manifest based on
```

scouten-adobe, line 6596:
```suggestion
```

scouten-adobe (approved):
One comment clean up to suggest; otherwise :+1:.

PR #1050, opened 2025-04-24
fix: Fix missing Action fields

mauricefisher64:
Fix missing Action fields
Fix typo in validation codes

## Checklist


sdk/src/assertions/actions.rs
```diff
@@ -194,7 +197,8 @@ impl Action {
         matches!(
             self.software_agent,
             Some(SoftwareAgent::ClaimGeneratorInfo(_))
-        ) || self.changes.is_some() // only defined for v2
+        ) || self.changes.is_some()
+            || self.description.is_some() // only defined for v2
```
gpeacock, line 201: It is probably time to always generate v2 assertions and not bother with this logic, but I'll update that later

sdk/src/assertions/actions.rs
```diff
@@ -486,6 +490,10 @@ pub struct Actions {
     /// A list of [`Action`]s.
     pub actions: Vec<Action>,
 
+    /// A list of of the software/hardware that did the action.
+    #[serde(rename = "softwareAgents", skip_serializing_if = "Option::is_none")]
+    pub software_agents: Option<Vec<ClaimGeneratorInfo>>,
```
gpeacock, line 495: If you are adding sofwareAgents we should also add sofwareAgentIndex to action, the two require each other.

mauricefisher64: The index is in the Action

gpeacock: yes, it sure is, and I guess I added it :)

gpeacock (approved):

PR #1063, opened 2025-05-06
docs: Replace old c2patool release notes with CHANGELOG

crandmck:

The C2PA Tool release notes have not been updated for a long time (since v 0.9), and the CHANGELOG really contains the most recent release notes.  This is already reflected in the [open source doc site](https://opensource.contentauthenticity.org/docs/c2patool-rns).

This PR changes links to point to the CHANGELOG instead of the old release notes, and adds a small caveat to the top of the (old) release notes,  along with a few other minor tweaks.


gpeacock (approved):
I think we could just delete the release notes file. We have it in the git history if needed.

crandmck:
> I think we could just delete the release notes file. We have it in the git history if needed.

I agree, but for the time being let's keep it around just in case a user needs to reference it (if using an old release).  I'll delete it in a few weeks and remove all links to it.

PR #1083, opened 2025-05-09
feat!: Rust native crypto by default

cdmurph32:
* Remove rust_native_crypto feature
* Add openssl feature
* Add support for es512 for rust native crypto
* Add ci test for all features except openssl

## Checklist


internal/crypto/src/tests/raw_signature/signers.rs
```diff
@@ -19,8 +19,8 @@ use crate::raw_signature::{
 };
 
 #[test]
-// #[cfg_attr(all(target_arch = "wasm32", not(target_os = "wasi")), wasm_bindgen_test)]
-#[cfg(not(target_arch = "wasm32"))]
+// #[cfg_attr(all(target_arch = "wasm32", not(target_os = "wasi")),
```
gpeacock, line 22: did you mean to leave all these in but commented out? Is there a reason to keep them?

internal/crypto/Cargo.toml
```diff
@@ -31,15 +31,7 @@ rustdoc-args = ["--cfg", "docsrs"]
...
-rust_native_crypto = [
-    "dep:const-oid",
-    "dep:der",
-    "dep:ed25519-dalek",
-    "dep:num-bigint-dig",
-    "dep:pkcs1",
-    "dep:rsa",
-    "dep:spki",
-]
+openssl = ["dep:openssl"]
```
gpeacock, line 34: it looks like this flips things around so that we always build with rust native crypto and so openssl builds would get larger than they are now. Do we know how much larger?
I'd still like to have a no-crypto option for embedded builds that don't need to validate, but we can work on that later.

gpeacock (approved):

internal/crypto/src/tests/raw_signature/signers.rs
cdmurph32: I left it as is but `cargo fmt` changed it. I'm not sure why they are here. 

I did remove the exclusion of the tests from WASI, since they run just fine.

internal/crypto/Cargo.toml
cdmurph32: Since I added additional crates to handle PEM parsing for es512 it's only about 100KB

scouten-adobe (changes requested):
Holding off on further review until we sort out which feature flags we're doing.

cdmurph32:
Closing in favor of https://github.com/contentauth/c2pa-rs/pull/1118

PR #1089, opened 2025-05-10
feat!: Merge CAWG identity SDK into main C2PA crate

scouten-adobe:


tmathern (approved):
- All those are breaking changes for other SDKs who depend on it. We'll need to plan the time accordingly to fix the dependency trees to not use cawg-identity anymore where it was used. The deprecation warnings and changes done here should help to make that easier.

gpeacock (approved):
It's up to you, but I'm not sure if it is worth maintaining all the deprecated notices in the older set of files. Could you just delete them all?
There are a lot of test images, I'm not sure if there are duplicates since they all look the same. It seems like these should be under src/tests/fixtures so we know where all the fixtures are.

scouten-adobe:
@gpeacock my plan is to publish one version of the crate with these deprecation notices and then completely delete the source.

PR #1091, opened 2025-05-12
feat(c2pa_rs): Support setting the Ingredient manifest_data field for remote manifests using Builder.

gpeacock:
When the fetch_remote_manifests feature is disabled, as in wasm builds, we need a way with the builder API to set the remote manifest_data that has been fetched in some other way. This can now be done by setting the manifest_data field in the ingredient definition. 
manifest_data {
    "format": "application/c2pa",
    "identifier": "my_identifier"
 }
 
 Then use builder.add_resource_stream("my_identifier", c2pa_data)  to add the binary resource before signing.

### DIFF
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -879,10 +879,6 @@ impl Ingredient {
         let (result, manifest_bytes) = if let Ok(manifest_bytes) = jumbf_result {
             let jumbf_store = Store::from_jumbf(&manifest_bytes, &mut validation_log);
             let result = if let Ok(mut store) = jumbf_store {
-                // if we don't have an active manifest, set it to the one we just loaded
-                // if self.active_manifest.is_none() {
-                //     self.active_manifest = store.provenance_label();
-                // };
                 if _sync {
                     match store.verify_from_stream(stream, format, &mut validation_log) {
                         Ok(_) => Ok(store),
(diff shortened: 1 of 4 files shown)


sdk/src/ingredient.rs
```diff
@@ -895,18 +899,11 @@ impl Ingredient {
...
-                result.inspect_err(|e| {
-                    // add a log entry for the error so we act like verify
-                    log_item!("asset", "error loading file", "Ingredient::from_file")
-                        .failure_no_throw(&mut validation_log, e);
-                }),
-                Some(manifest_bytes),
-            )
+            (result, Some(manifest_bytes))
         } else {
             // This will always be Err in this situation
```
scouten-adobe, line 904: You could rewrite the `if` at line 879 as a `match` statement and that would eliminate the need for the `unwrap` here.

sdk/src/manifest.rs
```diff
@@ -1870,6 +1910,7 @@ pub(crate) mod tests {
     #[test]
     #[cfg(feature = "file_io")]
     #[allow(deprecated)]
+    /// Actiions asssertions cannot be redacted, event though the redaction reference is valid
```
scouten-adobe, line 1913: Typo: Actiions -> Actions

scouten-adobe (approved):

sdk/src/ingredient.rs
gpeacock: Thanks, I this is much cleaner now.

emensch (approved):

PR #1100, opened 2025-05-14
Update c2patool CHANGELOG.md

crandmck:
To fix issues that were breaking the doc site build.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-All notable changes to this project will be documented in this file.
+All changes to C2PA Tool are documented in this file.  For additional information on versions 0.9.x and earlier, see the [Archived release motes](../docs/release-notes.md).
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html), except that – as is typical in the Rust community – the minimum supported Rust version may be increased without a major version increase.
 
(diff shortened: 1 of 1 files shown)


cli/CHANGELOG.md
```diff
@@ -23,16 +23,6 @@ _14 May 2025_
 ### Documented
 
 * Replace old c2patool release notes with CHANGELOG ([#1063](https://github.com/contentauth/c2pa-rs/pull/1063))
-<div class="github-only" >
```
scouten-adobe, line 26: Gah, I didn't see _that_ part.

cli/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-All notable changes to this project will be documented in this file.
+All changes to C2PA Tool are documented in this file.  For additional information on versions 0.9.x and earlier, see the [Archived release motes](../docs/release-notes.md).
```
scouten-adobe, line 3: Why segregate like this?

crandmck: It was simply because the changelog was added in addition to the existing manually-written release notes.  Do you want me to add the content of that file to the end of this one?  I suspect very few people will care about that info... so I don't have a strong opinion.

cli/CHANGELOG.md
crandmck: I use that in various places in markdown files to have content that we don't want to display in the doc site, but DO want in GitHub.  You don't see the tag itself in either place (only if you edit the md file).  Anyway, I removed it for now.  The issue was the lack of a new line before it, which apparently breaks the build due to linting rules.

tmathern (approved):

PR #1105, opened 2025-05-15
feat: Adds CAWG validation support to the c_api reader creation methods

gpeacock:
fix: allow c2pa_builder_sign_data_hashed_embeddable to pass a null asset pointer because that is an allowed feature

### DIFF
--- a/c_api/src/c_api.rs
+++ b/c_api/src/c_api.rs
@@ -18,10 +18,11 @@ use std::{
 
 // C has no namespace so we prefix things with C2PA to make them unique
 use c2pa::{
-    assertions::DataHash, settings::load_settings_from_str, Builder as C2paBuilder, CallbackSigner,
-    Reader as C2paReader, SigningAlg,
+    assertions::DataHash, identity::validator::CawgValidator, settings::load_settings_from_str,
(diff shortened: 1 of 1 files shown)


c_api/src/c_api.rs
```diff
@@ -428,6 +429,22 @@ pub unsafe extern "C" fn c2pa_string_free(s: *mut c_char) {
...
 
+// Run CAWG post-validation - this is async and requires a runtime.
+fn post_validate(result: Result<C2paReader, c2pa::Error>) -> Result<C2paReader, c2pa::Error> {
+    match result {
+        Ok(mut reader) => {
+            let runtime = match Runtime::new() {
+                Ok(runtime) => runtime,
+                Err(err) => return Err(c2pa::Error::OtherError(Box::new(err))),
+            };
+            match runtime.block_on(reader.post_validate_async(&CawgValidator {})) {
```
tmathern, line 440: Does the CAWG validation pass by default if there is no CAWG data?

scouten-adobe: Yes.

scouten-adobe (approved):

PR #1115, opened 2025-05-19
feat!: Merge `c2pa-status-tracker` crate into `c2pa`

scouten-adobe:


internal/status-tracker/src/lib.rs
```diff
@@ -11,21 +11,7 @@
...
-mod status_tracker;
-pub use status_tracker::{ErrorBehavior, StatusTracker};
-
-pub mod validation_codes;
-
-#[cfg(test)]
-pub(crate) mod tests;
+#[allow(unused)]
+pub(crate) fn goodbye() {
+    eprintln!("This crate is being discontinued.");
```
gpeacock, line 16: that feels kinda sad, but I don't think anyone is going to be mourning this ;)

gpeacock (approved):
Thanks for trudging though all of this!

mauricefisher64 (approved):
I did not look at every file but if it compiles it is likely correct.

PR #1118, opened 2025-05-19
feat: Make OpenSSL a default feature

cdmurph32:
Openssl will not be included when building rust native crypto
CI tests for both openssl and rust_native_crypto features

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,31 @@ on:
     - cron: "0 18 * * 1,4,6" # 1800 UTC every Monday, Thursday, Saturday
 
 jobs:
+  get-features:
+    name: Get features
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+
+      - name: Get all features
+        run: |
+          cargo metadata --format-version=1 | jq -r '[.packages[] | select(.name=="c2pa" or .name=="c2pa-c") | .features | keys | map(select(. != "default")) | .[]] | unique | join(" ")' > features.txt
+
+      - name: Upload features
+        uses: actions/upload-artifact@v4
+        with:
+          name: features
+          path: features.txt
+
   tests:
     name: Unit tests
+    needs: get-features
     if: |
       github.event_name != 'pull_request' ||
       github.event.pull_request.author_association == 'COLLABORATOR' ||
@@ -36,6 +59,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Download features
+        uses: actions/download-artifact@v4
+        with:
+          name: features
+
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -45,13 +73,40 @@ jobs:
...
+        if: runner.os == 'Windows'
+        run: |
+          $features = Get-Content features.txt
+          $features = $features -replace 'openssl', ''
+          cargo test --no-default-features --features "$features"
+
+      - name: Run tests with rust_native_crypto
+        if: runner.os != 'Windows'
+        run: |
+          FEATURES=$(cat features.txt | sed 's/openssl//g')
```
scouten-adobe, line 87: Nice!

.github/workflows/ci.yml
```diff
@@ -15,8 +15,31 @@ on:
...
+          toolchain: stable
+
+      - name: Get all features
+        run: |
+          cargo metadata --format-version=1 | jq -r '[.packages[] | select(.name=="c2pa" or .name=="c2pa-c") | .features | keys | map(select(. != "default")) | .[]] | unique | join(" ")' > features.txt
+
+      - name: Upload features
+        uses: actions/upload-artifact@v4
+        with:
+          name: features
```
scouten-adobe, line 37: I don't think this should be a release artifact. Please try to rework this using action outputs instead: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs

sdk/Cargo.toml
```diff
@@ -117,9 +123,9 @@ nom = "7.1.3"
 non-empty-string = { version = "=0.2.4", features = ["serde"] }
 nonempty-collections = { version = "0.2.9", features = ["serde"] }
 num-bigint-dig = { version = "0.8.4", optional = true }
-p256 = "0.13.2"
-p384 = "0.13.0"
-p521 = { version = "0.13.3", features = ["pkcs8", "digest", "ecdsa"] }
+p256 = { version = "0.13.2", optional = true }
+p384 = {version = "0.13.0", optional = true }
```
scouten-adobe, line 127: Can we be consistent with spacing here? The other deps have a space inside the {} pair.

scouten-adobe (changes requested):
Mostly :+1: but I have a concern about how values are passed between the GitHub Action steps.

.github/workflows/ci.yml
```diff
@@ -45,13 +73,32 @@ jobs:
...
         uses: Swatinem/rust-cache@v2
 
+      - name: Run tests with rust_native_crypto
+        shell: pwsh
+        if: runner.os == 'Windows'
+        env:
+          RUST_BACKTRACE: "1"
+          FEATURES: ${{needs.get-features.outputs.rust-native-features}}
+        run: |
+          cargo test --no-default-features --features "$env:FEATURES"
```
scouten-adobe, line 83: Would you please try to achieve code coverage for the rust-native-crypto path as well?

.github/workflows/ci.yml
```diff
@@ -270,12 +331,13 @@ jobs:
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 
-      - name: Run Wasm tests
-        run: wasm-pack test --chrome --headless
+      - name: Run Wasm tests (c2pa-rs)
```
scouten-adobe, line 334: FWIW I removed the `(c2pa-rs)` label in my PR #1115 because there were no longer any other crates running Wasm or WASI.

.github/workflows/ci.yml
```diff
@@ -289,8 +351,6 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-        # nightly required for testing until this issue is resolved:
```
scouten-adobe, line 292: Please leave this comment in unless you expect that the underlying issue will never be resolved.

In general, we want to avoid using nightly Rust for build processes because they are unpredictable. (See diffs at line 391.)

scouten-adobe (changes requested):
A couple minor things; otherwise :+1:.

.github/workflows/ci.yml
cdmurph32: Sorry I did not mean to delete this.

scouten-adobe (approved):

PR #1125, opened 2025-05-20
chore: Repair `Makefile` after build changes

gpeacock:
Everything but test-wasi works again now.


scouten-adobe (changes requested):
What's the step you use to verify that the makefile works? IMHO we should add that to CI so we identify such failures before the PRs get merged.

scouten-adobe (approved):

PR #1131, opened 2025-05-23
build: Windows PowerShell make_release

gpeacock:
Builds on x86 or arm
this is a rust_native_crypto build


scouten-adobe (approved):
Approved with the caveat that I'm not a PowerShell expert, but it looks reasonable to me.

PR #1136, opened 2025-05-27
fix: Remove use of workspace versioning for the moment

scouten-adobe:
crates.io doesn't like it. 😢

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -22,7 +22,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
 [dependencies]
 anyhow = "1.0"
 atree = "0.5.2"
-c2pa = { workspace = true , features = [
+c2pa = { version = "0.54.0", features = [
 	"fetch_remote_manifests",
 	"file_io",
 	"add_thumbnails",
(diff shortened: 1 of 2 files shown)


tmathern (approved):

cli/Cargo.toml
```diff
@@ -22,7 +22,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
 [dependencies]
 anyhow = "1.0"
 atree = "0.5.2"
-c2pa = { workspace = true , features = [
```
gpeacock, line 25: Actually I didn't intend to sync c2patool with this change yet.

gpeacock: I suppose it doesn't hurt but it could be confusing if we don't keep everything in sync. Won't the other crates have the same issue?

scouten-adobe: It looks like the cargo team doesn't plan to support this syntax for `cargo publish`: https://github.com/rust-lang/cargo/issues/11133

gpeacock: Someone else on that thread resolved the issue by moving to workspace packaging
https://github.com/boa-dev/boa/pull/2297

PR #1138, opened 2025-05-28
feat(sdk): introduces get_supported_types api

dyro:

This allows clients to determine which mime types that they can sign and which mime types that may read manifests from.

### Rust Changes

This PR introduces new associated functions named `supported_mime_types` on the `Builder` and `Reader` structs. 

### C Changes
We are exporting the above associated functions to our "C" clients as well. This PR adds the `c2pa_reader_supported_mime_types` and `c2pa_builder_supported_mime_types` functions. These are C functions that return an array of supported types. Additionally, the `c2pa_free_string_array` function was introduced to handle freeing the supported types string array. 

## Checklist


c_api/src/c_api.rs
```diff
@@ -604,6 +629,20 @@ pub unsafe extern "C" fn c2pa_reader_resource_to_stream(
...
+/// The caller is responsible for freeing the array.
+///
+/// # Parameters
+/// * count: pointer to a usize to return the number of mime types.
+///
+/// # Safety
+/// The returned value MUST be released by calling [c2pa_free_string_array].
+/// The array and its contents are no longer valid after that call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_reader_supported_mime_types(count: *mut usize) -> *mut *mut c_char {
```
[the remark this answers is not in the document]
dyro: Done, thanks!

sdk/src/identity/claim_aggregation/w3c_vc/did.rs
```diff
@@ -122,6 +122,12 @@ impl DidBuf {
         }
     }
 
+    // mismatched_lifetime_syntaxes reproting an error we are creating a
+    // [Did] from a [&str] reference.
+    //
+    // We are allowing "unknown_lints" because "mismatched_lifetime_syntaxes"
+    // is only in nightly.
+    #[allow(unknown_lints, mismatched_lifetime_syntaxes)]
```
dyro: We may want to consider changing the `Did::new_unchecked` API to take a `String` instead?

c_api/src/c_api.rs
```diff
@@ -430,6 +430,33 @@ pub unsafe extern "C" fn c2pa_string_free(s: *mut c_char) {
...
+    }
+
+    let mut_ptr = ptr as *mut *mut c_char;
+    // Free each string directly using the pointer.
+    for i in 0..count {
+        c2pa_string_free(*mut_ptr.add(i));
+    }
+
+    // Free the array.
+    Vec::from_raw_parts(mut_ptr, count, count);
```
mauricefisher64, line 457: Is there any way to protect this for bad input?

c_api/src/c_api.rs
```diff
@@ -1183,9 +1242,40 @@ pub unsafe extern "C" fn c2pa_signature_free(signature_ptr: *const u8) {
...
+unsafe fn c2pa_mime_types_to_c_array(strs: Vec<String>, count: *mut usize) -> *const *const c_char {
+    // Even if the array is exposed as a `*const *const c_char` for read-only access,
+    // the underlying memory must be allocated as `*mut *mut c_char` because freeing
+    // or deallocating memory requires a mutable pointer. This ensures the caller can
+    // safely release ownership of both the array and its strings.
+    let mime_ptrs: Vec<*mut c_char> = strs.into_iter().map(|s| to_c_string(s)).collect();
+    *count = mime_ptrs.len();
+
+    let ptr = mime_ptrs.as_ptr();
+    std::mem::forget(mime_ptrs);
```
mauricefisher64, line 1277: How does this work when the targets are compiled for different ptr size.  For example I build SDK on a 32 bit arm environment but my windows app is 64 bits?   May only be an edge error but perhaps we should note that in the function documentation.

c_api/src/c_api.rs
dyro: No, I don't think so. We have to manually ensure the invariants of the from_raw_parts are upheld. Namely, that we: 

1. Actually allocated the memory for the vector we are constructing using the Rust allocator
2. No one else is referencing the vector
3. The memory hasn't already been de-allocated.

We check if the ptr is null before we start the de-allocation, and we check if each of the strs are null before de-allocating them.

An additional check we could potentially do is to ensure the length isn't longer than the capacity, but in this case, both the length and capacity are the same. This is because the way we create the vector ensures that the capacity and length are the same.

c_api/src/c_api.rs
mauricefisher64: Perhaps our c bindings should use fixed sized units based on the target rather than usize.  Maybe u32 for 32 bit builds and u64  for 64 bit and u128 for 128bit pointers.  I don't think we need to hold up the PR for this but it would make interop safer.

sdk/src/identity/claim_aggregation/w3c_vc/did.rs
```diff
@@ -122,12 +122,6 @@ impl DidBuf {
         }
     }
 
-    // mismatched_lifetime_syntaxes reproting an error we are creating a
-    // [Did] from a [&str] reference.
-    //
-    // We are allowing "unknown_lints" because "mismatched_lifetime_syntaxes"
-    // is only in nightly.
-    #[allow(unknown_lints, mismatched_lifetime_syntaxes)]
```
dyro: Addressed by scouten via another PR.

PR #1142, opened 2025-05-29
EPUB_Frame

Kingsleyyy21:
Registeration of EPUB 

• added a mapping for the EPUB MIME type (application/epub+zip) in the 
mime utility (sdk/src/utils/mime.rs) so that the tool can recognize the “.epub” 
extension 
• registered our EPUB handler (EpubIo) in the global asset handler table (in 
sdk/src/jumbf_io.rs) 
• added “use crate::asset_handlers::epub_io::EpubIo;” at the top. 
With that registration, the c2patool now recognizes EPUB files (means no 
longer returns “Unsupported file type” when you run “c2patool sample.epub.”


ok-nick:
FYI there is existing work for this format in https://github.com/contentauth/c2pa-rs/pull/499

PR #1144, opened 2025-05-30
feat: Update Validation for 2.2 spec compliance

gpeacock:
We have a new setting - strict_v1_validation that defaults to false. This allows us to validate v1 claims under the rules that the v1.4 spec required. Setting it to true will allow you to write new v1 claims that are compliant with the 2.2 spec.
There are new cases where errors are returned where previously only validation_results were added.
Some minor logging support has been added. and there is a utils::test::setup_logging() function that can be used in unit tests.


sdk/src/store.rs
```diff
@@ -3821,30 +4136,366 @@ impl Store {
...
+
+        // if all mismatches are redactions we are good
+        if redact_matches == differences.len() {
+            return Some(redactions_to_remove);
+        }
+
+        None
+    }
+
+    // build ingredient lists for the Clain in the specified Store
```
gpeacock: change to claim not Clain

PR #1145, opened 2025-06-03
feat: Add Android platform release targets with integrated JNI support

n8fr8:
This PR integrates support for building the C API library targeting Android ARM64 and X86_64 with the local Makefile build process.

It also integrates the RUST-JNI crate to provide a first pass of an integrated Java Native Interface (JNI) which, unlike with Swift iOS, is required for calling into the library from Kotlin or Java in the Android environment. This interface will be improved in future PRs to be more in platform style.

The Makefile changes also integrate support for installing the required Android Native Development Kit (NDK) tools for completing the build.

## Checklist


gpeacock:
Is there a reason why this needs to be done in the library instead of in external JNI connected to the C bindings?

scouten-adobe (approved):
Looks OK to me with the caveat that I'd like review on some JNI specifics.

c_api/src/android.rs
```diff
@@ -0,0 +1,477 @@
+// Copyright 2024 Adobe. All rights reserved.
```
tmathern, line 1: Could we move this out, even to its own folder (crate)? Because since it's JNI, this could maybe be reused in the future for Java bindings which are not specific to Android (or refactored to share some code with that, if we ever have a Java SDK)?

c_api/src/android.rs
```diff
@@ -0,0 +1,477 @@
...
+// each license.
+
+use jni::objects::{JClass, JString};
+use jni::sys::{jint, jlong, jstring};
+use jni::JNIEnv;
+use std::ffi::{CStr, CString};
+use std::os::raw::c_char;
+use std::ptr;
+
+use crate::{
```
tmathern, line 21: The JNI bindings could depend on those even if it graduates to its own crate

c_api/src/android.rs
```diff
@@ -0,0 +1,477 @@
...
+    mut env: JNIEnv,
+    _class: JClass,
+) -> jstring {
+    let version = unsafe { c2pa_version() };
+    c_string_to_java_string(&mut env, version)
+}
+
+/// Returns the last error message
+#[no_mangle]
+pub extern "system" fn Java_org_c2pa_C2PA_getError(
```
tmathern, line 145: Will this be used in the Java code to **throw** errors, so it's more idiomatic to Java developers?

c_api/src/android.rs
```diff
@@ -0,0 +1,477 @@
...
+) -> jint {
+    let settings_cstr = java_string_to_cstring_jint!(env, settings);
+    let format_cstr = java_string_to_cstring_jint!(env, format);
+    
+    unsafe { c2pa_load_settings(settings_cstr.as_ptr(), format_cstr.as_ptr()) }
+}
+
+/// Returns a ManifestStore JSON string from a file path
+#[no_mangle]
+pub extern "system" fn Java_org_c2pa_C2PA_readFile(
```
tmathern, line 169: Ack - but we should consider using readers to do that, as they are an important piece of the API surface (and it reduces the exposed API surface too, so people use the API's main concepts).

c_api/src/android.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+    
+    let remote_url_cstr = java_string_to_cstring_jint!(env, remote_url);
+    
+    unsafe { c2pa_builder_set_remote_url(builder_ptr as *mut C2paBuilder, remote_url_cstr.as_ptr()) }
+}
+
+/// Creates and verifies a C2paReader from an asset stream with the given format
+#[no_mangle]
+pub extern "system" fn Java_org_c2pa_C2PA_readerFromStream(
```
tmathern, line 256: Here we have the reader - that is similar in spirit to `Java_org_c2pa_C2PA_readFile`, if we open the file as stream and pass it to a Reader, to finally get the manifest store as JSON.

c_api/src/android.rs
tmathern: An other argument to move this out is to add tests (unit tests, on this file), but also some Java tests. Especially to verify the stream behavior when we cross languages.
@gpeacock @n8fr8 Wdyt?

n8fr8: Happy to put it where you think it is best. I do not believe there is anything Android specific here that wouldn't also work in a generica Java/Kotlin environment.

c_api/src/android.rs
n8fr8: Yes. There will be a more idiomatic wrapper around this. In addition, we also intend to evolve this interface as much as it makes sense. Trying to minimize doing too much custom code inside the rust piece.

c_api/src/android.rs
n8fr8: We are still fully getting our heads around the best practices of the 2.x API ourselves. Readers do make sense, and again, will be more inline with the Android/Kotlin/Java idiom than just raw JSON. 

Also, we do plan to continue to have a "SimpleC2PA" library as well, on top of the "Full API", that even further simplifies integration for the most common and basic functions in a mobile app.

c_api/src/android.rs
tmathern: Let's start with that, we can move later if the need ever arises.

PR #1146, opened 2025-06-05
feat: Build C library for mobile platforms

redaranj:
Adds support for building the C library for Android and iOS and fixes an issue with the Linux ARM build exposed by the mobile changes.

Previous versions of the Makefile relied on platform detection to determine which architecture to build for, but we now must specify the architecture via the TARGET argument. I left the platform detection code in place for local development.

I have tested the workflow and am able to successfully build for all platforms and end up with a Github release containing the expected artifacts.

I’m not sure where/if I should document these changes.

## Checklist


scouten-adobe (approved):

.github/workflows/library-release.yml
```diff
@@ -15,9 +15,7 @@ jobs:
       matrix:
         include:
           - os: macos-latest
-            target: aarch64-apple-darwin
-          - os: macos-latest
```
tmathern, line 19: (Sidenote: Some people like to separate those out explicitly, and pull in only the compiled code for their specific arch)

c_api/Makefile
```diff
@@ -104,11 +110,92 @@ release-linux-gnu-x86:
 
 release-linux-gnu-arm:
 	rustup target add aarch64-unknown-linux-gnu
-	cargo build --target=aarch64-unknown-linux-gnu  $(CARGO_BUILD_FLAGS)
+	sudo apt-get update && sudo apt-get install -y gcc-aarch64-linux-gnu libc6-dev-arm64-cross
+	CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc cargo build --target=aarch64-unknown-linux-gnu  $(CARGO_BUILD_FLAGS)
 	@$(call make_zip,$(TARGET_DIR)/aarch64-unknown-linux-gnu/release,aarch64-unknown-linux-gnu)
 
+release-ios-arm64:
```
tmathern, line 120: Could you add a comment for what each target compiles? It looks like some are specific to the simulator?

redaranj:
> What are some of these builds for? i.e. x86_64-apple-ios

 x86_64-apple-ios is for iOS simulators running on Intel Macs (which are admittedly getting pretty rare).

I added comments to the Makefile about what each target is for. 

If we’d like to pare these back a bit, I think we could drop the iOS Intel simulator and both 32-bit Android targets (device and emulator) and focus our efforts on modern, 64-bit architectures.

.github/workflows/library-release.yml
redaranj: I re-added the individual Mac targets, so in the release we’ll have all three: ARM, Intel and Universal.

c_api/Makefile
redaranj: I added comments for each of the mobile targets in the Makefile

c_api/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MIT OR Apache-2.0"
 
 [lib]
 name = "c2pa_c"
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "staticlib"]
```
tmathern, line 10: @scouten-adobe @gpeacock Remind me why we had only `cdylib` (if there was a reason) ? Any concern changing that?

redaranj: We need the static library for iOS, but I don’t see a way to specify that we should only build the static lib for iOS-related targets. Instead I’m removing it from the other targets before zipping/uploading the artifacts.

scouten-adobe: I measured clean build times with the `staticlib` setting and without and there's no appreciable difference. I'm :+1: on this change.

scouten-adobe (approved):

PR #1147, opened 2025-06-05
feat: impl Soft Binding assertion definition

ok-nick:
Implements the Soft Binding assertion definition. It exposes all public fields so it's clear in the docs how to deserialize into it and it provides no functions since everything is already accessible by field. Getters/setters can be added on if wanted, but there is already a lot of variation in the implementation of assertions.

## Checklist

### DIFF
--- a/sdk/src/assertions/mod.rs
+++ b/sdk/src/assertions/mod.rs
@@ -49,6 +49,9 @@ pub use ingredient::Relationship;
 
 pub mod labels;
 
+mod metadata;
+pub use metadata::Metadata;
+
 mod assertion_metadata;
 pub use assertion_metadata::{
     c2pa_source, Actor, AssertionMetadata, AssetType, DataBox, DataSource, ReviewCode, ReviewRating,
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -1132,8 +1132,9 @@ impl BmffHash {
             crate::asset_handlers::bmff_io::write_c2pa_box(
                 &mut uuid_box_data,
                 &[],
-                false,
+                crate::asset_handlers::bmff_io::MERKLE,
                 &mm_cbor,
+                0,
             )?;
 
             let first_moof = box_infos
@@ -1226,8 +1227,9 @@ impl BmffHash {
                 crate::asset_handlers::bmff_io::write_c2pa_box(
                     &mut uuid_box_data,
                     &[],
-                    false,
+                    crate::asset_handlers::bmff_io::MERKLE,
                     &mm_cbor,
+                    0,
                 )?;
 
                 // replace temp C2PA Merkle box
@@ -1413,8 +1415,9 @@ impl BmffHash {
                 crate::asset_handlers::bmff_io::write_c2pa_box(
                     &mut uuid_box_data,
                     &[],
-                    false,
+                    crate::asset_handlers::bmff_io::MERKLE,
                     cbor_data,
+                    0,
                 )
                 .unwrap_or_default(); // this cannot fail unless the vec allocation above fails
                 uuid_box_data
@@ -1522,7 +1525,7 @@ impl BmffHash {
                 // the number of MM entries should match the number of ranges
                 if ranges.len() != mm.count {
                     return Err(Error::C2PAValidation(
-                        "number of leaves does not match the Merkle".to_string(),
+                        "number of leaves does not match the MerkleMap count".to_string(),
                     ));
                 }
 
@@ -1551,20 +1554,22 @@ impl BmffHash {
                 .collect::<Vec<(&Vec<BmffMerkleMap>, &Vec<HashRange>)>>();
 
             for (mm_index, (bmff_mm, ranges)) in merkle_maps_with_ranges.into_iter().enumerate() {
+                let current_mm = &mm_vec[mm_index];
+
+                let alg = match &current_mm.alg {
+                    Some(a) => a,
+                    None => self
+                        .alg()
+                        .ok_or(Error::HashMismatch("alg is required".to_string()))?,
+                };
+
                 // check all the ranges in this for this mdat
                 for (range_index, range) in ranges.iter().enumerate() {
-                    let alg = match &mm_vec[mm_index].alg {
-                        Some(a) => a,
-                        None => self
-                            .alg()
-                            .ok_or(Error::HashMismatch("alg is required".to_string()))?,
-                    };
-
                     // hash the entire fragment minus exclusions
                     let hash = hash_stream_by_alg(alg, reader, Some(vec![range.clone()]), false)?;
 
                     // check MerkleMap for the hash
-                    if !mm_vec[mm_index].check_merkle_tree(
+                    if !current_mm.check_merkle_tree(
                         alg,
                         &hash,
                         bmff_mm[range_index].location,
--- a/sdk/src/assertions/labels.rs
+++ b/sdk/src/assertions/labels.rs
@@ -18,6 +18,9 @@
 //! These constants do not include version suffixes.
 //!
 //! See <https://c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_c2pa_standard_assertions>.
+use std::sync::LazyLock;
+
+use regex::Regex;
 
 /// Label prefix for a claim assertion.
 ///
@@ -184,8 +187,23 @@ pub(crate) const DATABOX_STORE: &str = "c2pa.databoxes";
 
 /// Label prefix for asset reference assertion.
 ///
-/// See <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_asset_reference>
+/// See <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_asset_reference>.
 pub const ASSET_REFERENCE: &str = "c2pa.asset-ref";
+
+/// Label prefix for a metadata assertion.
+///
+/// See <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_metadata>.
+pub const METADATA: &str = "c2pa.metadata";
+
+/// Must have a label that ends in '.metadata' and is preceded by an entity-specific namespace.
+/// For example, a 'com.litware.metadata' assertion would be valid.
+pub static METADATA_LABEL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    #[allow(clippy::unwrap_used)]
+    {
+        Regex::new(r"^(?:[a-zA-Z0-9][a-zA-Z0-9_-]*)(?:\.(?:[a-zA-Z0-9][a-zA-Z0-9_-]*))*\.metadata$")
+            .unwrap()
+    }
+});
 
 /// Return the version suffix from an assertion label if it exists.
 ///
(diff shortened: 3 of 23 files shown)


sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use super::labels;
+use crate::{
+    assertion::{Assertion, AssertionBase, AssertionCbor, AssertionJson},
+    assertions::region_of_interest::RegionOfInterest,
+    cbor_types::UriT,
+    Result,
+};
+
+const ASSERTION_CREATION_VERSION: usize = 1;
```
tmathern: @ok-nick @gpeacock @mauricefisher64 Do we need to use usize for the version numbers? It will work for Rust, but cause us pain if we ever need to have other languages and tools bind to that code (because of the type). It's also a very hypothetical case, but I want to point it out.

mauricefisher64: @tmathern is it usize specifically that is causing the issue or just that the size is typed?

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Region of interest in regard to the soft binding.
+    pub region: RegionOfInterest,
+}
+
+/// Soft binding timespan for temporal assets.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SoftBindingTimespanMap {
+    /// Start of the time range (as milliseconds from media start) over which the soft binding value has been computed.
+    pub start: u32,
```
mauricefisher64: The spec does not specify these to be u32.  Make them usize.  CBOR chooses the optimal storage based on what is actually written.

sdk/src/assertions/soft_binding.rs
tmathern: The `usize` type specifically (the family of types whose size actually depend on the arch you're running on, as they don't have a generic type that fits all). Not an issue yet, but could become one - so if we could get away with an `int` from the beginning, it would be nice.

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// this structure is invalid; there is no default.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub alg: Option<String>,
+
+    /// A list of details about the soft binding.
+    pub blocks: Vec<SoftBindingBlockMap>,
+
+    /// Zero-filled bytes used for filling up space.
+    #[serde(with = "serde_bytes")]
+    pub pad: Vec<u8>,
```
gpeacock: I think pad and pad2 need to be at the end of structure, or at least they make more sense there.

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub value: String,
+}
+
+/// Soft binding scope, specifying specifically where in an asset the soft binding is applicable.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SoftBindingScopeMap {
+    /// In algorithm specific format, the part of the digital content over which the soft binding value has been computed.
+    #[deprecated]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub extent: Option<String>,
```
gpeacock: I wonder if we need to make already deprecated fields public

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Region of interest in regard to the soft binding.
+    pub region: RegionOfInterest,
+}
+
+/// Soft binding timespan for temporal assets.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SoftBindingTimespanMap {
+    /// Start of the time range (as milliseconds from media start) over which the soft binding value has been computed.
+    pub start: usize,
```
gpeacock: usize tends to be problematic in bindings, we may want to use u64 or something  more specific here.

sdk/src/assertions/soft_binding.rs
ok-nick: I think it makes sense to keep public, if we support deserializing to it, we should support accessing it directly via code. If an older manifest has this field it should still be accessible.

sdk/src/assertions/soft_binding.rs
ok-nick: It currently matches the field order in the spec, although there's no harm in moving it to the bottom of the struct.

sdk/src/assertions/soft_binding.rs
ok-nick: Originally I had them as `u32`, but Maurice suggested `usize`:
> The spec does not specify these to be u32. Make them usize. CBOR chooses the optimal storage based on what is actually written.

@mauricefisher64 Any thoughts on this?

mauricefisher64: We have some flexibility here.  We discussed this in the C2PA working group and since is was not defined in earlier versions of the spec it was decided not to codify it now.  The logical thing to do is to make sure what ever the field may contain can be represented.  In this case is it milliseconds.  For most time specs represent this as a 64 bit value.  So I would go with u64.

sdk/src/assertions/soft_binding.rs
mauricefisher64: I am with Gavin on this.  Making an accessor enforces that it cannot be written.

mauricefisher64: we should also not even serialize it when writing.

ok-nick: That's a good idea, I've implemented it in the latest commit.

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,167 @@
...
+
+use super::labels;
+use crate::{
+    assertion::{Assertion, AssertionBase, AssertionCbor},
+    assertions::region_of_interest::RegionOfInterest,
+    cbor_types::UriT,
+    Result,
+};
+
+const ASSERTION_CREATION_VERSION: usize = 1;
```
gpeacock: We don't need to specify this at all unless we have more than one version, it will default to the right thing.

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,167 @@
...
+    url: Option<UriT>,
+}
+
+impl SoftBinding {
+    /// A file or http(s) URL to where the bytes that are being hashed lived.
+    ///
+    /// This is useful for cases where the data lives in a different file chunk or side-car
+    /// than the claim.
+    #[deprecated]
+    pub fn url(&self) -> Option<&UriT> {
```
gpeacock, line 52: It seems very bizarre to have a deprecated function on a brand new assertion. What is the logic here?
Also when using the deprecated macro, we should specify when it was deprecated and what to use instead.

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,167 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub region: Option<RegionOfInterest>,
+
+    #[serde(skip_serializing)]
+    extent: Option<String>,
+}
+
+impl SoftBindingScopeMap {
+    /// In algorithm specific format, the part of the digital content over which the soft binding value has been computed.
+    #[deprecated]
```
gpeacock: why?

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// than the claim.
+    #[deprecated]
+    pub fn url(&self) -> Option<&UriT> {
+        self.url.as_ref()
+    }
+}
+
+/// Details about the soft binding, including the referenced value and scope.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SoftBindingBlockMap {
```
gpeacock: If we are making these sub structures public, we need unit tests for them. Otherwise make them private and we don't have to worry about testing the rust apis. I'm not sure about the code coverage impact.

sdk/src/assertions/soft_binding.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// A human-readable description of what this hash covers.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
+
+    /// A string describing parameters of the soft binding algorithm.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub alg_params: Option<String>,
+
+    /// Zero-filled bytes used for filling up space.
```
gpeacock: We need to work on this. If we expect a user to add these fields correctly we need to know how do that. Or we can document that the SDK does this magically, but then the SDK needs to do that.

gpeacock (changes requested):

sdk/src/assertions/soft_binding.rs
ok-nick: The spec deprecated this in 1.3 and says to use the asset reference assertion instead. I marked it as deprecated to signify to users that it shouldn't be used. We do support reading older manifests which is why this field is here although it's not recommended hence the deprecation.

sdk/src/assertions/soft_binding.rs
ok-nick: Read above.

sdk/src/assertions/soft_binding.rs
ok-nick: From my understanding this field has no purpose or use outside of the lower level APIs. I'm inclined to private the fields, add a getter, and skip it from serialization although I'd like to hear your thoughts about that. These padding fields exist in a few assertions so we may want to do the same there.

sdk/src/assertions/soft_binding.rs
ok-nick: I don't think struct definitions themselves are included in the code coverage. I added a unit test at the bottom that handles round trip serializing->deserializing these structs/fields already as well (taken from the example in the spec). I think we should keep them public since the info is relevant for the assertion.

sdk/src/assertions/mod.rs
```diff
@@ -77,3 +77,8 @@ mod embedded_data;
 pub use embedded_data::EmbeddedData;
 
 pub mod region_of_interest;
+
+mod soft_binding;
+pub use soft_binding::{
+    SoftBinding, SoftBindingBlockMap, SoftBindingScopeMap, SoftBindingTimespanMap,
```
gpeacock: Why are we adding "Map" to the end of all these structures? What is that intended to convey. In Rust the practice is to avoid redundant information in naming. We are not supposed to add Enum or Enum names or Error or error names. Why not just call these SoftBindingBlock, SoftBindingScope and SoftBindingTimespan?

gpeacock (changes requested):
I think this is ready to go but I'd like to not add the practice of appending Map to field names unless our plan is to do that everywhere, and I'm not sure what it adds.

sdk/src/assertions/mod.rs
ok-nick: I initially did this to match the spec definitions, although I agree with your sentiment. I'll change it to remove the suffix.

sdk/src/assertions/soft_binding.rs
ok-nick: I moved the padding fields to methods since they aren't used in the SDK at a lower level. I've added additional documentation although I'm waiting to confirm with @mauricefisher64 if they are correct.

sdk/src/assertions/soft_binding.rs
gpeacock: ok, that makes sense, I didn't realize that this was a 1.3 assertion we had not implemented yet.

PR #1150, opened 2025-06-09
feat: add `Reader::remote_url` and `Reader::is_embedded`

ok-nick:
Implements:
* `Reader::remote_url` (high-level)
* `Reader::is_embedded` (high-level)

Note that `Reader::remote_url` will return the remote url ONLY IF the manifest was obtained remotely. There may still be a remote manifest embedded, although we do not parse far enough to tell.

These APIs are also implemented for the C bindings.

## Checklist


sdk/src/reader.rs
```diff
@@ -844,15 +851,72 @@ impl std::fmt::Debug for Reader {
     }
 }
 
+/// Returns the locations a C2PA manifest was found within the stream.
+///
+/// This function DOES NOT validate or fetch the C2PA manifest and will never
+/// return [`ManifestLocation::Sidecar`], detection of sidecars depends on
```
tmathern:
```suggestion
/// return [`ManifestLocation::Sidecar`], as detection of sidecars depends on
```

c_api/src/c_api.rs
```diff
@@ -284,6 +308,45 @@ pub unsafe extern "C" fn c2pa_load_settings(
...
+/// c2pa_error.
+///
+/// # Safety
+/// Reads from NULL-terminated C strings.
+#[no_mangle]
+pub unsafe extern "C" fn manifest_locations_from_stream(
+    format: *const c_char,
+    stream: *mut C2paStream,
+) -> c_int {
+    let format = CStr::from_ptr(format).to_string_lossy().into_owned();
```
tmathern: Why not use the macro `from_cstr_or_return_null`, so you don't need to import the type just for that call? Like here, seems to be a very similar case: https://github.com/contentauth/c2pa-rs/blob/e20691aed676c7b827ed177a91cddf66f50cf78e/c_api/src/c_api.rs#L306?

ok-nick: Originally the function wasn’t returning a c_int, although now that would be a good idea.

mauricefisher64:
Please do not proceed with these changes in the Rust SDK until I've had a chance to think about it.  Right now I would not approve.

c_api/src/c_api.rs
```diff
@@ -41,6 +41,28 @@ mod cbindgen_fix {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum C2paManifestLocation {
+    /// The manifest is embedded in the asset.
+    Embedded,
+    /// The manifest is stored remotely.
+    Remote,
+    /// The manifest is stored separately in a sidecar.
+    Sidecar,
+}
+
```
mauricefisher64: The only comment I have is can we change this from ManifestLocation to something else like ManifestDiscovery.  The reason is that in the internal code manifest location refer to its position in the asset

c_api/src/c_api.rs
```diff
@@ -41,6 +41,28 @@ mod cbindgen_fix {
...
+/// The location in which the manifest was read.
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum C2paManifestLocation {
+    /// The manifest is embedded in the asset.
+    Embedded,
+    /// The manifest is stored remotely.
+    Remote,
+    /// The manifest is stored separately in a sidecar.
+    Sidecar,
```
mauricefisher64: Where is Sidecar used?

ok-nick: I should clarify the docs here, `Sidecar` is returned when the manifest is passed explicitly via a method such as `from_manifest_data_and_stream`. If there's a better name to describe this variant let me know.

c_api/src/c_api.rs
ok-nick: How about `ManifestSource`?

c_api/src/c_api.rs
```diff
@@ -41,6 +41,28 @@ mod cbindgen_fix {
...
+/// The source in which the manifest was read.
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum C2paManifestSource {
+    /// The manifest is embedded in the asset.
+    Embedded,
+    /// The manifest is stored remotely.
+    Remote,
+    /// The manifest is stored separately in a sidecar.
+    Sidecar,
```
gpeacock: Maybe rename this to **External** since we don't really know if it is a sidecar. For instance, the same method is used when JS has retrieved a manifest from the cloud via it's own.
Since the caller will know if they have supplied the external manifest, perhaps the only information we really need here is if the SDK has retrieved the manifest remotely or not.

sdk/src/store.rs
```diff
@@ -3741,25 +3748,30 @@ impl Store {
...
     ///
     /// This handles, embedded and remote manifests
     ///
     /// asset_type -  mime type of the stream
     /// stream - a readable stream of an asset
-    pub fn load_jumbf_from_stream(asset_type: &str, stream: &mut dyn CAIRead) -> Result<Vec<u8>> {
+    pub fn load_jumbf_from_stream(
+        asset_type: &str,
+        stream: &mut dyn CAIRead,
+    ) -> Result<ManifestDataSource> {
```
gpeacock: We could return two values here as a tuple instead of wrapping one in the other. That way we can use the same enum on the external API.
`Result<(Vec<u8>, ManifestSource)>`

gpeacock: Another thing I've been wanting is a way to know what the remote url was after retrieving the data.
So perhaps the second parameter could return the URL as an Option. 
If it is None, and you get data, then that data was embedded. If you get a URL back then we fetched the data from the URL. Another option is that we did not fetch the data from the URL but we returned an error with the URL.
`Result<(Vec<u8>, Option<String>)>`

sdk/src/reader.rs
```diff
@@ -417,6 +417,11 @@ impl Reader {
         }
     }
 
+    /// Get the source of the manifest loaded into this [`Reader`].
+    pub fn manifest_source(&self) -> ManifestSource {
```
gpeacock: It feels like we could reuse the RemoteManifest enum here:

- NoRemote - embedded
- SideCar - maybe we can add the path here so we know where we read the sidecar from.
- Remote(String) - this could return the remote URL that we read the manifest from
- EmbedWithRemote(String) Indicates that we read from the embedded but there is also a remote URL
-

sdk/src/store.rs
```diff
@@ -4504,6 +4521,49 @@ impl std::fmt::Display for Store {
...
+    Remote,
+    /// The manifest is stored separately from the asset.
+    ///
+    /// This variant occurs when the manifest bytes are passed explicitly such
+    /// as in the case of [Reader::from_manifest_data_and_stream].
+    Sidecar,
+}
+
+#[derive(Debug, Clone)]
+pub enum ManifestDataSource {
```
gpeacock: This might be better handled with just returning a tuple with an option for the URL.

c_api/src/c_api.rs
ok-nick: Yes, I like `ManifestSource::External` better as well.

That's a really good point too, maybe we would be better off with just a `Reader::is_remote` or `Reader::is_remote_source`  method?

Edit: Or we can have a `Reader::remote_url` method that returns an optional url.

sdk/src/reader.rs
ok-nick: That's a good idea. I wonder if we can get by with pulling that enum out of the claim without all of the other refactoring.

Edit: Unfortunately the enum only set in the claim when writing, not reading.

sdk/src/store.rs
ok-nick: One concern with changing the error for this case, is that a lot people may depend on reading `Error::JumbfNotFound` to detect that there is no manifest.

sdk/src/reader.rs
ok-nick: Looking over it, I think we'd be better served with a new enum. The name is a little bit misleading and the fields don't necessarily align.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -595,6 +596,39 @@ pub unsafe extern "C" fn c2pa_reader_json(reader_ptr: *mut C2paReader) -> *mut c
...
+    check_or_return_null!(reader_ptr);
+    let c2pa_reader = guard_boxed!(reader_ptr);
+
+    match c2pa_reader.remote_url() {
+        Some(url) => to_c_string(url.to_string()),
+        None => ptr::null(),
+    }
+}
+
+/// Returns the if the reader was created from an embedded manifest.
```
mauricefisher64: typo

sdk/src/reader.rs
```diff
@@ -260,7 +260,7 @@ impl Reader {
         mut fragment: impl Read + Seek + Send,
     ) -> Result<Self> {
         let mut validation_log = StatusTracker::default();
-        let manifest_bytes = Store::load_jumbf_from_stream(format, &mut stream)?;
+        let manifest_bytes = Store::load_jumbf_from_stream(format, &mut stream)?.0;
```
mauricefisher64: We should also return the type  in Store::from_stream.  That way we can collapse the line 263/264 into a single function call.  Even if you don't these two lines can be collapsed.

sdk/src/ingredient.rs
```diff
@@ -930,7 +931,7 @@ impl Ingredient {
 
         // retrieve the manifest bytes from embedded, sidecar or remote and convert to store if found
         let (result, manifest_bytes) = match Store::load_jumbf_from_stream(format, stream) {
-            Ok(manifest_bytes) => {
+            Ok((manifest_bytes, _)) => {
```
mauricefisher64, line 935: Can be collapsed to Store::from_stream

ok-nick: The problem here is that we need the manifest bytes for this call:
```rust
ingredient.update_validation_status(result, manifest_bytes, &validation_log)?;
```

mauricefisher64: Ok, it may be needed for some Ingredient uses.

sdk/src/reader.rs
```diff
@@ -862,10 +875,32 @@ pub mod tests {
...
+        let reader = Reader::from_stream("image/jpeg", Cursor::new(IMAGE_WITH_MANIFEST))?;
+        assert_eq!(reader.remote_url(), None);
+        assert!(reader.is_embedded());
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_reader_remote_url() -> Result<()> {
+        let reader = Reader::from_stream("image/jpeg", Cursor::new(IMAGE_WITH_REMOTE_MANIFEST))?;
```
gpeacock, line 897: Note that this will fail if we run it when offline. It isn't the only case of that, so it's ok for now, but I'd like to avoid the  online dependency for tests since it make them less reliable. We could work around this with a mock service or using File:// urls.

sdk/src/store.rs
```diff
@@ -3728,16 +3742,22 @@ impl Store {
     ///
     /// asset_type -  mime type of the stream
     /// stream - a readable stream of an asset
-    pub fn load_jumbf_from_stream(asset_type: &str, stream: &mut dyn CAIRead) -> Result<Vec<u8>> {
```
gpeacock, line 3731: This is a published API so it will be a breaking change. I'm ok with that since it is probably little used.

gpeacock (approved):

PR #1151, opened 2025-06-09
fix: to_archive does not store resources associated with ingredients

alextrnnn:
Scenario (detected by unit test):
- Create a builder from a manifest definition
- Add an ingredient with a stream to that builder
- Create an archive from that builder
- Create a builder from that archive
- Attempt to sign with the builder

This scenario would result in a ResourceNotFound error.

Modifications to_archive/from_archive to add/read the ingredient's resource files to/from the zip have been added to resolve this error.

## Checklist


tmathern (approved):
LGTM, but please wait for a review from @gpeacock or @mauricefisher64 regarding the validity of us supporting this scenario. Thanks.

sdk/src/builder.rs
```diff
@@ -561,6 +569,30 @@ impl Builder {
...
+                    .ok_or(Error::BadParam("Invalid resource path".to_string()))?;
+                let format = format_from_path(id)
+                    .ok_or(Error::BadParam("Invalid resource path".to_string()))?;
+                let id = id.replacen(['-'], ":", 1);
+                for ingredient in builder.definition.ingredients.iter_mut() {
+                    let base_id = ingredient.instance_id().to_string();
+                    if id.starts_with(&base_id) {
+                        ingredient
+                            .resources_mut()
+                            .add_with(&base_id, &format, data.clone())?;
```
ok-nick: After this line you can add a `break;` and avoid the `.clone()`. Instance IDs are globally unique so the current iterated resource should only match one ingredient thus making the loop safe to break.

c_api/src/c_api.rs
```diff
@@ -1317,6 +1318,80 @@ mod tests {
...
+                &mut manifest_bytes_ptr,
+            )
+        };
+
+        assert_ne!(res, -1);
+
+        let reader = Reader::from_stream("image/jpeg", &mut dest_stream).unwrap();
+        let reader_json = reader.json();
+        assert!(reader_json.contains("Test Ingredient"));
+        assert!(reader_json.contains("thumbnail"));
```
ok-nick: Tests are usually ran with all features including `add_thumbnails`, so I think there are 2 thumbnails, one for the main stream `test_image2` and one for the ingredient stream `test_image3`. May be better to check for "thumbnail.ingredient" instead (although correct me if I'm wrong).

ok-nick (changes requested):
Nice work.

sdk/src/builder.rs
```diff
@@ -567,6 +574,31 @@ impl Builder {
                 builder.resources.add(id, data)?;
             }
 
+            if file.name().starts_with("ingredient-resources/")
+                && file.name() != "ingredient-resources/"
+            {
+                let mut data = Vec::new();
+                file.read_to_end(&mut data)?;
+
+                let id = file
```
gpeacock: we need to factor out  the name munging logic since it is being repeated.

c_api/src/c_api.rs
```diff
@@ -1413,6 +1414,79 @@ mod tests {
         unsafe { c2pa_builder_free(builder) };
     }
 
+    #[test]
+    fn test_to_archive_and_from_archive_with_ingredient_thumbnail() {
```
gpeacock: This test should be in reader.rs. It would be a lot cleaner there and I don't think there's anything unique about calling it through the C wrapper.

sdk/src/builder.rs
```diff
@@ -514,6 +514,13 @@ impl Builder {
                 zip.start_file("manifests/", options)
                     .map_err(|e| Error::OtherError(Box::new(e)))?;
                 for ingredient in self.definition.ingredients.iter() {
+                    // add ingredient resource files to a ingredient folder
```
gpeacock: This is actually a little inside out. This will work, and it may be ok for now, but the better answer would be to move all the ingredient resources into the builder before we archive. That way we don't need a separate set of ingredient resources to manage.

PR #1156, opened 2025-06-16
Fix xmp jpeg write

cdmurph32:
Adding the manifest url to large XMP segments could cause the XMP data
to exceed the limit of 65503 bytes.

There were 3 causes of this issue:
1. Extra whitespace from formatting. The quick_xml writer shiftwidth
   could be much larger than that of the original. Potentially adding
   several KB of whitespace.
2. Manifest url added multiple times. The manifest url was added to
   multiple keys within the XMP data.
3. Lack of padding awareness. According to the spec, XMP segments should
   have 2KB to 4KB of padding. The padding should be adjusted to keep
   the XMP data length the same with the addition of the remote manifest
   url. This is important for data hashes as it keeps the offsets of the
   JPEG metadata the same.

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/jpeg_io.rs
+++ b/sdk/src/asset_handlers/jpeg_io.rs
@@ -620,7 +620,7 @@ impl RemoteRefEmbed for JpegIO {
                     .unwrap_or((None, MIN_XMP));
 
                 // add provenance and JPEG XMP prefix
-                let xmp = format!("{XMP_SIGNATURE}\0 {}", add_provenance(xmp, &manifest_uri)?);
+                let xmp = format!("{XMP_SIGNATURE}\0{}", add_provenance(xmp, &manifest_uri)?);
                 let segment = JpegSegment::new_with_contents(markers::APP1, Bytes::from(xmp));
                 // insert or add the segment
                 match xmp_index {
--- a/sdk/src/utils/xmp_inmemory_utils.rs
+++ b/sdk/src/utils/xmp_inmemory_utils.rs
@@ -93,7 +93,6 @@ fn extract_xmp_key(xmp: &str, key: &str) -> Option<String> {
 /// Add a value to XMP using a key, replaces the value if the key exists, or adds it only once
 fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
     let orig_length = xmp.len();
-    println!("original length: {orig_length}");
 
     // Minimal padding should be 2 KB to 4 KB. This is used if no XMP packet is found.
     let mut target_length = 4096;
@@ -109,7 +108,6 @@ fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
     } else {
         xmp
     };
-    println!("target length: {target_length}");
 
     let mut reader = Reader::from_str(xmp_body);
     // Do not trim text to preserve whitespace
@@ -201,12 +199,7 @@ fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
         }
     }
     let padding_length = target_length.saturating_sub(writer.get_ref().get_ref().len());
-    println!("padding length {padding_length}");
     write_xmp_padding(&mut writer.get_mut(), padding_length)?;
-    println!("final length: {}", writer.get_ref().get_ref().len());
-    if padding_length > 0 {
-        assert_eq!(writer.get_ref().get_ref().len(), orig_length);
-    }
     let result = writer.into_inner().into_inner();
     String::from_utf8(result).map_err(|e| Error::XmpWriteError(e.to_string()))
 }
--- a/sdk/src/utils/xmp_inmemory_utils.rs
+++ b/sdk/src/utils/xmp_inmemory_utils.rs
@@ -90,11 +90,32 @@ fn extract_xmp_key(xmp: &str, key: &str) -> Option<String> {
 }
 
 // writes the event to the writer)
-/// Add a value to XMP using a key, replaces the value if the key exists
+/// Add a value to XMP using a key, replaces the value if the key exists, or adds it only once
 fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
-    let mut reader = Reader::from_str(xmp);
-    reader.config_mut().trim_text(true);
-    let mut writer = Writer::new_with_indent(Cursor::new(Vec::new()), b' ', 2);
+    let orig_length = xmp.len();
+    println!("original length: {orig_length}");
+
+    // Minimal padding should be 2 KB to 4 KB. This is used if no XMP packet is found.
+    let mut target_length = 4096;
+    // Remove trailing xpacket if present for easier manipulation
+    let xpacket_end_single = "<?xpacket end='w'?>";
+    let xpacket_end_double = "<?xpacket end=\"w\"?>";
+    let xmp_body = if let Some(pos) = xmp.rfind(xpacket_end_single) {
+        target_length = orig_length - xpacket_end_single.len();
+        xmp[..pos].trim_end()
+    } else if let Some(pos) = xmp.rfind(xpacket_end_double) {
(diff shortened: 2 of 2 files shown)


sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -620,7 +620,7 @@ impl RemoteRefEmbed for JpegIO {
                     .unwrap_or((None, MIN_XMP));
 
                 // add provenance and JPEG XMP prefix
-                let xmp = format!("{XMP_SIGNATURE}\0 {}", add_provenance(xmp, &manifest_uri)?);
+                let xmp = format!("{XMP_SIGNATURE}\0{}", add_provenance(xmp, &manifest_uri)?);
```
cdmurph32, line 623: This extra space caused there to be a leading space in the XMP data

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -90,19 +90,38 @@ fn extract_xmp_key(xmp: &str, key: &str) -> Option<String> {
...
 fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
-    let mut reader = Reader::from_str(xmp);
-    reader.config_mut().trim_text(true);
-    let mut writer = Writer::new_with_indent(Cursor::new(Vec::new()), b' ', 2);
+    let orig_length = xmp.len();
+
+    // Minimal padding should be 2 KB to 4 KB. This is used if no XMP packet is found.
+    let mut target_length = 4096;
+    // Remove trailing xpacket if present for easier manipulation
+    let xpacket_end_single = "<?xpacket end='w'?>";
```
cdmurph32: Single quotes are what the spec has, but we specifically test for double quotes. The test images from cameras also use double quotes.
`<?xpacket end="w"?>`

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -130,13 +149,14 @@ fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
                 if !added {
                     // didn't exist, so add it
                     elem.push_attribute((key, value));
+                    added = true;
```
cdmurph32, line 152: The lack of `added = true` is why the remote manifests url was being added multiple times.

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -261,4 +317,54 @@ mod tests {
...
     }
+
+    #[cfg_attr(not(target_arch = "wasm32"), actix::test)]
+    #[cfg_attr(
+        all(target_arch = "wasm32", not(target_os = "wasi")),
+        wasm_bindgen_test
+    )]
+    #[cfg_attr(target_os = "wasi", wstd::test)]
+    async fn test_broken_xmp_read_write_stream() {
+        let source_bytes = include_bytes!("../../tests/fixtures/broken.jpg");
```
cdmurph32: Should we rename the file?

sdk/src/utils/xmp_inmemory_utils.rs
mauricefisher64: I think this needs to be more flexible.  You may also encounter end='r'.

mauricefisher64: Or we could be a good citizen and respect the 'r' which means read only.

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -90,19 +90,38 @@ fn extract_xmp_key(xmp: &str, key: &str) -> Option<String> {
...
+    let orig_length = xmp.len();
+
+    // Minimal padding should be 2 KB to 4 KB. This is used if no XMP packet is found.
+    let mut target_length = 4096;
+    // Remove trailing xpacket if present for easier manipulation
+    let xpacket_end_single = "<?xpacket end='w'?>";
+    let xpacket_end_double = "<?xpacket end=\"w\"?>";
+    let xmp_body = if let Some(pos) = xmp.rfind(xpacket_end_single) {
+        target_length = orig_length - xpacket_end_single.len();
+        xmp[..pos].trim_end()
```
mauricefisher64: As I commented above, I would just search for "<?xpacket" since the rest of the tag may vary.

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -90,19 +90,38 @@ fn extract_xmp_key(xmp: &str, key: &str) -> Option<String> {
...
+    } else if let Some(pos) = xmp.rfind(xpacket_end_double) {
+        target_length = orig_length - xpacket_end_double.len();
+        xmp[..pos].trim_end()
```
mauricefisher64, line 107: The change I suggested about removes the else if case.

sdk/src/utils/xmp_inmemory_utils.rs
mauricefisher64: This is not our file so I don't think we can include it in a unit test.

cdmurph32: That's too bad. It would be cool to have a picture from summit in the repo. :-).

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -177,6 +198,8 @@ fn add_xmp_key(xmp: &str, key: &str, value: &str) -> Result<String> {
             }
         }
     }
+    let padding_length = target_length.saturating_sub(writer.get_ref().get_ref().len());
```
mauricefisher64, line 202: Shouldn't this take into account the existing padding to keep the length the same if possible?  This seems like it would not take into account the existing padding.  Is there somewhere you account to the existing padding?

sdk/src/utils/xmp_inmemory_utils.rs
mauricefisher64: You can ask Jen if we can use it.

sdk/src/utils/xmp_inmemory_utils.rs
gpeacock: We definitely need to be flexible reading here. I've also found cases where the end just wasn't there and other tools think it is fine.

sdk/src/utils/xmp_inmemory_utils.rs
gpeacock: Thanks for tracking this down!

sdk/src/utils/xmp_inmemory_utils.rs
gpeacock: We try to keep our repo images small, this one is over 6mb.

sdk/src/utils/xmp_inmemory_utils.rs
cdmurph32: If the end is completely missing then I can't just look for `<?xpacket` since that could be the beginning. The test `add_xmp` does this.

sdk/src/utils/xmp_inmemory_utils.rs
cdmurph32: Yes. Line 107
 
`target_length = orig_length - xpacket_end_length;`

sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -23,6 +23,7 @@ use quick_xml::{
 use crate::{asset_io::CAIRead, jumbf_io::get_cailoader_handler, Error, Result};
 
 const RDF_DESCRIPTION: &[u8] = b"rdf:Description";
+const XMP_END: &[u8] = b"<?xpacket end=\"w\"?>";
```
mauricefisher64, line 26: I guess you could go with the end=\'w\' single quote version since you said that is what the spec has.

mauricefisher64: Though XML does not care so nevermind

PR #1158, opened 2025-06-17
feat: Publish C2PA C ffi API (only package renamed)

tmathern:
- Prepare publication of C2PA C ffi API
- Renames c_api to c2pa-c-ffi (package only) so that it gets published under that name (lib name remains c2pa_c)

## Checklist

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -50,9 +50,7 @@ changelog_path = "./CHANGELOG.md"
 name = "c2patool"
 
 [[package]]
-name = "c_api"
-publish = false
-release = false
+name = "c2pa-c-ffi"
 
 [[package]]
 name = "export_schema"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -812,6 +812,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "c2pa-c-ffi"
+version = "0.55.0"
+dependencies = [
+ "c2pa",
+ "cbindgen",
+ "scopeguard",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "thiserror 1.0.69",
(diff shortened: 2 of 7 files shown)


release-plz.toml
```diff
@@ -50,9 +50,7 @@ changelog_path = "./CHANGELOG.md"
 name = "c2patool"
 
 [[package]]
-name = "c_api"
-publish = false
-release = false
```
gpeacock, line 55: I'm not sure we want to release this, since that will create another gitbhub release, I think. I don't really know release-plz either. We may want to check with Eric on this.

gpeacock (changes requested):
We may need to change the folder name from c_api to c2pa_c_ffi. The lib.rs file references mod c_api and the workspace Cargo.toml includes it. Maybe we don't need to change the folder name. I know Eric didn't like having the c2pa name in there so I changed it to c_api, but that was before we wanted to publish it as a crate.
I suspect release_plz will have issues with the naming,but I'm not sure.

release-plz.toml
tmathern: To be prepared, let me make those changes in another branch. Worse case we can pick the one over the other.

tmathern: Update: Looking at the renaming, the mod name comes from the file. I would keep `c_api(.rs)` in the folder, even if we decide to rename the folder. That's the change done in https://github.com/contentauth/c2pa-rs/pull/1159/. I flagged it for the workflows to run to verify if that one can build too.
(I think indeed renaming the folder makes it clearer, if having that name would be fine).

tmathern:
Closing in favor of https://github.com/contentauth/c2pa-rs/pull/1159

PR #1160, opened 2025-06-17
test: Add more test coverage to Builder/Reader

alextrnnn:
Unit tests added to verify the functionality of Builder/Reader

## Checklist


sdk/src/builder.rs
```diff
@@ -2193,4 +2194,337 @@ mod tests {
...
+    #[test]
+    fn test_add_duplicate_resource() {
+        let mut builder = Builder::new();
+        let _ = builder.add_resource("unknown", Cursor::new(b"12345"));
+        let result = builder.add_resource("unknown", Cursor::new(b"12345"));
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_sign_multiple() {
```
gpeacock, line 2219: Nice, I'm glad to know this works. There may be some cases where it doesn't since signing can modify a builder. I'll think about that.

sdk/src/builder.rs
```diff
@@ -2193,4 +2194,337 @@ mod tests {
...
+
+        let creative_work = CreativeWork::new();
+        builder
+            .add_assertion(CreativeWork::LABEL, &creative_work)
+            .unwrap();
+
+        // Add custom CBOR assertion
+        let custom_cbor = serde_cbor::Value::Map(
+            vec![(
+                serde_cbor::Value::Text("test_key".to_string()),
```
gpeacock, line 2290: It would be good to modify this to encode a value that cannot be stored in JSON and then validate that we get it back correctly. We have some types in utils/cbor_types that would good to test this way.

sdk/src/builder.rs
```diff
@@ -2193,4 +2194,337 @@ mod tests {
...
+        let empty_json = "{}";
+        let builder = Builder::from_json(empty_json).unwrap();
+        assert!(!builder.definition.claim_generator_info.is_empty()); // Should have default
+    }
+
+    #[test]
+    fn test_empty_builders() {
+        let mut builder = Builder::new();
+
+        builder.definition.ingredients = Vec::new();
```
gpeacock, line 2269: These are the same as the defaults, so I'm not sure what this is testing. You could try signing this and it should complain that there are no assertions

sdk/src/builder.rs
```diff
@@ -2193,4 +2194,337 @@ mod tests {
...
+        }
+
+        assert_eq!(builder.definition.ingredients.len(), 3);
+        for (i, ingredient) in builder.definition.ingredients.iter().enumerate() {
+            assert_eq!(ingredient.relationship(), &relationships[i]);
+        }
+    }
+
+    #[test]
+    fn test_multiple_claim_generator_info() {
```
gpeacock, line 2385: This is an interesting one since v2 claims should not allow this, or at least they will ignore the second entry. We need to work on how to handle that somehow. Currently we allow the second entry, but for v2 claims we should not allow it. And yes, that begs the question about why it is a vector still.

sdk/src/reader.rs
```diff
@@ -991,4 +991,102 @@ pub mod tests {
...
+        assert!(reader.manifests().is_empty());
+        assert_eq!(reader.iter_manifests().count(), 0);
+        assert!(reader.validation_status().is_none());
+    }
+
+    #[test]
+    #[cfg(feature = "file_io")]
+    fn test_reader_from_file_with_sidecar() {
+        // Test that image with no manifest, but has sidecar, does not return an Error
+        let result = Reader::from_file("tests/fixtures/cloud_manifest.jpg");
```
gpeacock, line 1085: This wasn't really intended as a sidecar manifest. We should have a specific sidecar manifest. A cloud manifest runs a different path since it has a url in it.

gpeacock (changes requested):

sdk/src/builder.rs
alextrnnn: It looks like there isn't any special case for empty builders. I tried to sign and no error came up either, I looked around very briefly and I don't see special handling for empty assertions. I'll remove this test

scouten-adobe:
Too much change in the underlying code since then. Closing.

Issue #1175, opened 2025-06-19
Review and edit Rust API docs


crandmck:
Update what's displayed in https://docs.rs/c2pa/latest/c2pa.

PR #1178, opened 2025-06-20
Docs: API doc clean up and improvement

crandmck:
WIP PR to clean up the docs that are published to https://docs.rs/c2pa/latest/c2pa/ per #1175.

Hide docs for these modules:
- Dynamic_assertions 
- Crypto 
- validation_status 
- status_tracker 
- settings 

Should these be hidden, too?
- Identity
- jumbf_io 

Additional questions:
 - Do we need the Re-exports section?
 
Do we need these?
 - Action
 - ActionTemplate 
 - Actions
 - Metadata

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -17,13 +17,13 @@
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg, doc_cfg_hide))]
 
 //! This library supports reading, creating, and embedding C2PA data
-//! with a variety of asset types.
+//! for a variety of asset types.
 //!
 //! Some functionality requires you to enable specific crate features,
 //! as noted in the documentation.
 //!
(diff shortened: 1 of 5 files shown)


sdk/src/lib.rs
```diff
@@ -91,31 +91,40 @@ pub const VERSION: &str = env!("CARGO_PKG_VERSION");
...
 
 /// Supports status tracking as defined in the C2PA Technical Specification.
+#[doc(hidden)]
 pub mod status_tracker;
 
 /// The validation_results module contains the definitions for the validation results that are part of the C2PA specification.
 pub mod validation_results;
+
 /// The validation_status module contains the definitions for the validation status that are part of the C2PA specification.
+#[doc(hidden)]
```
gpeacock, line 129: I may just need to deprecate this rather than remove it from the docs, I'll think about it.

PR #1180, opened 2025-06-21
feat: Add Automatic Update Manifest generation when possible

gpeacock:

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -32,14 +32,14 @@ use crate::{
         labels, Actions, BmffHash, BoxHash, CreativeWork, DataHash, Exif, Metadata, SoftwareAgent,
         Thumbnail, User, UserCbor,
     },
-    claim::Claim,
+    claim::{Claim, ALLOWED_UPDATE_MANIFEST_ACTIONS},
     error::{Error, Result},
     jumbf_io,
     resource_store::{ResourceRef, ResourceResolver, ResourceStore},
     salt::DefaultSalt,
     store::Store,
     utils::mime::format_to_mime,
-    AsyncSigner, ClaimGeneratorInfo, HashRange, Ingredient, Signer,
+    AsyncSigner, ClaimGeneratorInfo, HashRange, Ingredient, Relationship, Signer,
 };
 
 /// Version of the Builder Archive file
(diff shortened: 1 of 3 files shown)


sdk/src/builder.rs
```diff
@@ -855,10 +902,16 @@ impl Builder {
...
-    fn to_store(&self) -> Result<Store> {
-        let claim = self.to_claim()?;
-        // commit the claim
+    fn to_store(&self, is_update: bool) -> Result<Store> {
+        let mut claim = self.to_claim(is_update)?;
+
         let mut store = Store::new();
+
+        // if this can be an update manifest, then set the update_manifest flag
+        if is_update {
```
mauricefisher64: This does not seem correct to me.  You can only add an update manifest to an existing store.  The update manifest must link to an existing active manifest.

gpeacock: This is the way I always do things. The parent will get added to the store as I process the current claim. I have this check to ensure that the right kind of parent will get added.
:

```
        // must contain only one v3 ingredient with relationship ParentOf and an active manifest
        if self.definition.ingredients.len() != 1
            || *self.definition.ingredients[0].relationship() != Relationship::ParentOf
            || self.definition.ingredients[0].active_manifest().is_none()
        {
            return false;
        }
```

sdk/src/claim.rs
```diff
@@ -2497,12 +2497,17 @@ impl Claim {
...
-                                            Claim::assertion_label_from_link(&readaction_label);
+                                        // The assertion may or may not be in the assertion store.
+                                        // It can exist and be zeroed or be removed entirely
+                                        // but it must be in the claim's assertions HashUri list
                                         parent_tested = Some(
-                                            ingredient.get_assertion(&label, instance).is_some(),
+                                            ingredient
+                                                .assertions()
+                                                .iter()
+                                                .any(|a| a.url().contains(&readaction_label)),
```
mauricefisher64: Is the redaction label you are passing in the full URL?  That is what is expected.

mauricefisher64: You are correct.  A redaction will be in in the assertion list but not the assertion store.  Good catch.

mauricefisher64 (approved):

sdk/src/claim.rs
mauricefisher64: maybe make this a Claim fuction, perhaps **fn has_assertion_entry() -> bool**

PR #1181, opened 2025-06-21
fix: Clean up logs

tmathern:
- Remove debug logs
- Add callback signer test (to help narrow down possible error sources)
- Add docs on streams
- ~Rename things for clarity~

## Checklist


c2pa_c_ffi/src/c_api.rs
```diff
@@ -141,6 +129,18 @@ macro_rules! from_cstr_option {
     };
 }
 
+// Internal routine to convert a *const c_char to a rust String or return a -1 int error.
+#[macro_export]
+macro_rules! from_cstr_or_return_int {
```
tmathern: Moved around

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1104,22 +1104,25 @@ pub unsafe extern "C" fn c2pa_signer_create(
...
                 data.as_ptr(),
                 data.len(),
                 signed_bytes.as_mut_ptr(),
                 signed_len_max,
             )
         };
-        println!("c_callback: signed_size: {}", signed_size);
-        if signed_size < 0 {
+
+        if signed_size <= 0 {
```
tmathern: That `<=` is important to catch errors where signing failed, since raising errors in the callback is cumbersome. So we'll check the signature length as we already do, and raise the error from the native code here (until we can do better).

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1104,22 +1104,25 @@ pub unsafe extern "C" fn c2pa_signer_create(
     let certs = from_cstr_or_return_null!(certs);
     let tsa_url = from_cstr_option!(tsa_url);
     let context = context as *const ();
-    println!("c2pa_signer_create");
-    let c_callback = move |context: *const (), data: &[u8]| {
+
+    // Create a callback that uses the provided C callback function
+    // The callback ignores its context parameter and will use the context set on the CallbackSigner
+    let c_callback = move |callback_context: *const (), data: &[u8]| {
```
tmathern: Prevent capture of "wrong" context.

c2pa_c_ffi/src/c_api.rs
gpeacock: The convention on any C function that returns a numeric error is that a negative number is an error. I'm afraid this won't do the right thing for C++ or Swift or others. I'm not comfortable with changing the rule here. Why can't the callback return a negative number when it has an error?

c2pa_c_ffi/src/c_api.rs
gpeacock: I don't see how this changes anything. I think it is just renaming the context parameter to callback_context.

gpeacock (dismissed):
I don't think these changes are needed. You are welcome to convince me otherwise.

c2pa_c_ffi/src/c_api.rs
tmathern: `let context = context as *const ();` is left alone and not captured, I believe.

c2pa_c_ffi/src/c_api.rs
tmathern: Well, the question is if a 0 len signature is considered valid? I think it should not be considered valid, but maybe it is?

tmathern: I can get behind that though and undo that change.

c2pa_c_ffi/src/c_api.rs
tmathern: Or at least it makes it clearer in my mind that there is no capture

tmathern:
> I don't think these changes are needed. You are welcome to convince me otherwise.

In any case, I'd like to keep the renaming. At least it makes it a little bit clearer. Also, I removed debug logs.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1600,6 +1606,146 @@ mod tests {
...
+        };
+
+        // verify signer is not null (aka could be created)
+        assert!(!signer.is_null());
+
+        unsafe { c2pa_signer_free(signer) };
+    }
+
+    #[test]
+    fn test_sign_with_callback_signer() {
```
tmathern, line 1641: Full-circle test:
- Creates a callback signer
- Uses the callback to sign
- Verfies signed data can be read

PR #1184, opened 2025-06-23
feat: Loading ingredients from folder through Builder

alextrnnn:
- Photoshop's current workflow is the following:
  - When you place an image, they will call Ingredient::from file, saving all assets to some given base_path folder
  - The JSON result is then pasted into their manifest definition
- Gavin's plan is to change to using a builder for this workflow. Since they will have existing ingredients saved in folders, we still need to read for compatibility.
- This PR is provides a helper method of reading an ingredient folder, given some base path, using the Builder workflow. An example is also provided.

## Checklist


sdk/examples/load_ingredient/main.rs
```diff
@@ -0,0 +1,20 @@
...
+// specific language governing permissions and limitations under
+// each license.
+use anyhow::Result;
+mod load_ingredient;
+
+#[cfg(feature = "file_io")]
+fn main() -> Result<()> {
+    load_ingredient::main()?;
+    Ok(())
+}
```
ok-nick: The `load_ingredient::main` function can be inlined here.

sdk/src/builder.rs
```diff
@@ -1204,6 +1205,54 @@ impl Builder {
...
         Store::get_composed_manifest(manifest_bytes, format)
     }
+
+    /// Loads an ingredient folder into the builder.
+    /// # Arguments
+    /// * `base_path` - The base path of ingredient folder to read.
+    /// # Errors
+    /// * Returns an [`Error`] if the ingredient folder could not be loaded.
+    #[cfg(feature = "file_io")]
+    pub fn load_ingredient_from_folder(&mut self, base_path: &Path) -> Result<()> {
```
ok-nick: Changing this to `add_ingredient_from_folder` and returning `&mut Ingredient` would stay more consistent with the other methods.

sdk/src/builder.rs
```diff
@@ -2231,4 +2280,35 @@ mod tests {
         assert!(reader_json.contains("Test Ingredient"));
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
+
+    #[test]
+    #[cfg(feature = "file_io")]
+    fn test_load_ingredient_from_folder() {
+        let mut source = Cursor::new(TEST_IMAGE_CLEAN);
+        let format: &'static str = "image/jpeg";
```
ok-nick: nit: can remove type definition `: &'static str`

sdk/src/builder.rs
```diff
@@ -1204,6 +1205,54 @@ impl Builder {
...
+    /// * Returns an [`Error`] if the ingredient folder could not be loaded.
+    #[cfg(feature = "file_io")]
+    pub fn load_ingredient_from_folder(&mut self, base_path: &Path) -> Result<()> {
+        let ingredient_path = PathBuf::from(base_path).join("ingredient.json");
+        let ingredient_path = ingredient_path.as_path();
+        let json = std::fs::read_to_string(ingredient_path)?;
+        let mut ingredient = Ingredient::from_json(&json)?;
+
+        // Specify ingredient is parent
+        ingredient.set_is_parent();
```
ok-nick: The `relationship` of an ingredient is defined in its `ingredient.json`, it's not always assumed to be the parent.

sdk/src/builder.rs
```diff
@@ -1204,6 +1205,54 @@ impl Builder {
...
+
+        // Make sure we will have access to thumbnail
+        if let Some(thumbnail_ref) = ingredient.thumbnail_ref() {
+            self.add_resource_ref_to_resources(base_path, thumbnail_ref)?;
+        }
+
+        // Make sure we will have access to manifest
+        if let Some(manifest_data_ref) = ingredient.manifest_data_ref() {
+            self.add_resource_ref_to_resources(base_path, manifest_data_ref)?;
+        }
```
ok-nick, line 1229: When signing, I believe resources are extracted from the ingredients added to the `Builder`, so this step may not be necessary, although you may have to set [`Ingredient::with_base_path`](https://docs.rs/c2pa/latest/c2pa/struct.Ingredient.html#method.with_base_path).

sdk/examples/load_ingredient/load_ingredient.rs
```diff
@@ -0,0 +1,53 @@
...
+    let mut builder = Builder::new();
+
+    builder.load_ingredient_from_folder(Path::new("sdk/examples/load_ingredient/ingredient"))?;
+
+    // Write the manifest builder to a zipped stream
+    let mut zipped = Cursor::new(Vec::new());
+    builder.to_archive(&mut zipped)?;
+
+    // Unzip the manifest builder from the zipped stream
+    let _ = zipped.rewind();
```
ok-nick: Old code from test?

sdk/examples/load_ingredient/load_ingredient.rs
```diff
@@ -0,0 +1,53 @@
...
+
+const CERTS: &[u8] = include_bytes!("../../tests/fixtures/certs/ed25519.pub");
+const PRIVATE_KEY: &[u8] = include_bytes!("../../tests/fixtures/certs/ed25519.pem");
+const TEST_IMAGE: &[u8] = include_bytes!("../../tests/fixtures/C.jpg");
+
+#[cfg(feature = "file_io")]
+#[allow(clippy::incompatible_msrv)]
+pub fn main() -> Result<()> {
+    let mut source = Cursor::new(TEST_IMAGE);
+    let format: &'static str = "image/jpeg";
```
ok-nick: nit: `: &'static str` can be removed here as well.

sdk/examples/load_ingredient/load_ingredient.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use anyhow::Result;
+use c2pa::{crypto::raw_signature::SigningAlg, Builder, CallbackSigner, Reader};
+
+const CERTS: &[u8] = include_bytes!("../../tests/fixtures/certs/ed25519.pub");
+const PRIVATE_KEY: &[u8] = include_bytes!("../../tests/fixtures/certs/ed25519.pem");
+const TEST_IMAGE: &[u8] = include_bytes!("../../tests/fixtures/C.jpg");
+
+#[cfg(feature = "file_io")]
+#[allow(clippy::incompatible_msrv)]
```
ok-nick: How come this lint needs to be allowed?

ok-nick (commented):
Nice work on this, I left a few comments with some suggestions for potential improvements. One concern I have is you should be able to get the same behavior by doing `Builder::add_ingredient(Ingredient::from_json(&json)?.with_base_path(path))`, should we be adding another method to the Builder for this?

sdk/examples/add_ingredient/main.rs
```diff
@@ -0,0 +1,19 @@
...
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+use anyhow::Result;
+mod add_ingredient;
+
+#[cfg(feature = "file_io")]
+fn main() -> Result<()> {
+    add_ingredient::main()
+}
```
gpeacock, line 19: I think you may need a noop main() in order for this to build without file_io. 
Have you run make test on this?

sdk/examples/add_ingredient/add_ingredient.rs
```diff
@@ -0,0 +1,52 @@
...
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+use std::{
+    io::{Cursor, Seek},
+    path::Path,
+};
+
+use anyhow::Result;
+use c2pa::{crypto::raw_signature::SigningAlg, Builder, CallbackSigner, Reader};
```
gpeacock, line 19: SigningAlg should be exported at the top level, so you should not need the crypto crate reference here.

sdk/examples/add_ingredient/add_ingredient.rs
```diff
@@ -0,0 +1,52 @@
...
+// or the MIT license (http://opensource.org/licenses/MIT),
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+use std::{
```
gpeacock, line 13: We could use a top level description of why this example exists: i.e:

This shows how to convert a saved ingredient folder into a Builder ingredient that can be archived.

sdk/src/builder.rs
```diff
@@ -1204,6 +1205,53 @@ impl Builder {
...
         Store::get_composed_manifest(manifest_bytes, format)
     }
+
+    /// Adds an ingredient folder into the builder.
+    /// # Arguments
+    /// * `base_path` - The base path of ingredient folder to read.
+    /// # Errors
+    /// * Returns an [`Error`] if the ingredient folder could not be loaded.
+    #[cfg(feature = "file_io")]
+    pub fn add_ingredient_from_folder(&mut self, base_path: &Path) -> Result<&mut Ingredient> {
```
gpeacock, line 1215: My intention was to show how to add an ingredient using the existing Builder logic. I wanted to know if we really needed to add this method or not, because I'm trying to keep the SDK as minimal as possible.

So the question is, can I set up an ingredient in a manifest definition, that does what we need here.

I.e. can I write a method in the example that, given a folder, can construct a manifest definition with an ingredient read from the ingredient.json in that file, and also add the needed resources from that file to the builder and have it build. I don't want to add this method to the SDK if we can create an example that shows those who need it how to do this without this method.

PR #1186, opened 2025-06-23
refactor: Move uri_to_path into utils

alextrnnn:
Move helper method that converts uri to path into utils

## Checklist


sdk/src/utils/io_utils.rs
```diff
@@ -234,6 +234,22 @@ pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
...
+        if path.starts_with("/c2pa/") {
+            path = path.replacen("/c2pa/", "", 1);
+        } else {
+            path = format!("{}/{path}", manifest_label);
+        }
+        path = path.replace([':'], "_");
+    }
+    path
+}
+
```
ok-nick: We can reduce allocations here by using `String::strip_prefix` or perhaps even better,`PathBuf::strip_prefix` although these changes wouldn't block this PR.

sdk/src/utils/io_utils.rs
```diff
@@ -234,16 +235,52 @@ pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
     ))?
 }
 
+/// Convert a URI to a file path using PathBuf for better path handling.
+pub fn uri_to_path(uri: &str, manifest_label: &str) -> String {
+    let Some(path_str) = uri.strip_prefix("self#jumbf=") else {
+        return uri.to_string();
+    };
+
+    let mut path = PathBuf::from(path_str);
```
ok-nick, line 249: I believe this line can be removed and the below code can work with the existing string.

sdk/src/utils/io_utils.rs
```diff
@@ -234,16 +235,52 @@ pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
...
+        path = stripped.to_path_buf();
+    } else {
+        let mut new_path = PathBuf::from(manifest_label);
+        new_path.push(path);
+        path = new_path;
+    }
+
+    let path_str = path.to_string_lossy();
+
+    path_str.replace(':', "_")
```
ok-nick: Maybe the `.replace` can be done before converting to a path and that way we can return a `PathBuf` directly from this function (if code that uses it permits).

sdk/src/utils/io_utils.rs
```diff
@@ -234,16 +235,52 @@ pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
...
+        assert_eq!(uri_to_path(uri, "unknown"), expected_path);
+        assert_eq!(uri_to_path(expected_path, "unknown"), expected_path);
+
+        let uri = "self#jumbf=c2pa.assertions/c2pa.thumbnail.claim";
+        let manifest_label = "test";
+        let expected_path = format!("{manifest_label}/c2pa.assertions/c2pa.thumbnail.claim");
+
+        assert_eq!(uri_to_path(uri, manifest_label), expected_path);
+        assert_eq!(uri_to_path(&expected_path, manifest_label), expected_path);
+    }
```
ok-nick: Nice this is very helpful to see.

sdk/src/utils/io_utils.rs
alextrnnn: Without this line it would have slightly different behavior than the old one, if the path is `"contentauth_urn_uuid_b3386820-9994-4b58-926f-1c47b82504c4/c2pa.assertions/c2pa.thumbnail.claim.jpeg"` it will prepend the manifest label even if `self#jumbf=` isn't present. In the previous code, it will only prepend the manifest label if `self#jumbf=` exists in the string. Is that okay? If it is I can remove those lines!

sdk/src/utils/io_utils.rs
```diff
@@ -234,16 +236,64 @@ pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
...
 mod tests {
     #![allow(clippy::expect_used)]
     #![allow(clippy::unwrap_used)]
 
     use std::io::Cursor;
 
+    #[cfg(feature = "file_io")]
+    #[test]
+    fn test_uri_to_path() {
+        let uri = "self#jumbf=/c2pa/contentauth:urn:uuid:b3386820-9994-4b58-926f-1c47b82504c4/c2pa.assertions/c2pa.thumbnail.claim.jpeg";
```
gpeacock: We should flip this around to be the new v2 manifest label format - the contentauth part is at the end of the label now.

sdk/src/reader.rs
```diff
@@ -626,7 +628,7 @@ impl Reader {
         for manifest in self.manifests.values() {
             let resources = manifest.resources();
             for (uri, data) in resources.resources() {
-                let id_path = Self::uri_to_path(uri, manifest.label().unwrap_or("unknown"));
```
gpeacock, line 642: If you are moving this here, we should remove the Reader::uri_to_path since it won't be referenced anymore.  The same logic is repeated in ResourceStore, so you should update it to use the utility.

I suspect we need something like this in builder to_archive and from_archive.

gpeacock (changes requested):

sdk/src/reader.rs
alextrnnn: https://github.com/contentauth/c2pa-rs/blob/ac4371eec9d52fd88ddbafa31cdd6ec1e6c1156b/sdk/src/builder.rs#L524-L531

Is this what you mean? I believe the only thing that is shared between the code is the replacement of the ":" to a "_"

sdk/src/resource_store.rs
```diff
@@ -244,20 +244,15 @@ impl ResourceStore {
...
             if self.base_path.is_some() {
-                // convert to a file path always including the manifest label
-                id = id.replace("self#jumbf=", "");
-                if id.starts_with("/c2pa/") {
-                    id = id.replacen("/c2pa/", "", 1);
-                } else if let Some(label) = self.label.as_ref() {
-                    id = format!("{label}/{id}");
-                }
-                id = id.replace([':'], "_");
+                use crate::utils::io_utils::uri_to_path;
```
gpeacock: move this to the the top of the file

sdk/src/utils/io_utils.rs
```diff
@@ -234,16 +236,64 @@ pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
...
+    } else {
+        return PathBuf::from(path_str);
+    }
+
+    let mut path = PathBuf::from(path_str);
+
+    if let Ok(stripped) = path.strip_prefix("/c2pa/") {
+        path = stripped.to_path_buf();
+    } else if let Some(manifest_label) = manifest_label {
+        let mut new_path = PathBuf::from(manifest_label);
```
gpeacock: The original logic replaced colons and underscores at the end, so that the manifest label was also converted. This logic will not convert the manifest label.

gpeacock (changes requested):
It looks like the manifest label is not converted here.

gpeacock (approved):

PR #1187, opened 2025-06-23
feat: Remove file type from thumbnail names

gpeacock:
This is a c2pa 2.2 spec requirement

### DIFF
--- a/sdk/src/assertions/thumbnail.rs
+++ b/sdk/src/assertions/thumbnail.rs
@@ -53,6 +53,16 @@ impl Thumbnail {
             content_type,
         }
     }
+
+    /// Create a new Thumbnail with a specific content type
+    /// The label should not contain the image type postfix
+    pub fn new_with_format(label: &str, data: Vec<u8>, content_type: &str) -> Self {
+        Thumbnail {
+            data,
+            label: label.to_owned(),
+            content_type: content_type.to_owned(),
+        }
+    }
 }
 
 impl AssertionBase for Thumbnail {
@@ -128,5 +138,17 @@ pub mod tests {
         let assertion = Assertion::new(labels::JPEG_CLAIM_THUMBNAIL, None, data);
         let result = Thumbnail::from_assertion(&assertion);
         assert!(result.is_err())
+    }
+
+    #[test]
+    fn assertion_thumbnail_with_format() {
+        let original = Thumbnail::new_with_format("foo", some_binary_data(), "image/png");
+        let assertion = original.to_assertion().expect("build_assertion");
+        assert_eq!(assertion.content_type(), "image/png");
+        assert_eq!(assertion.label(), "foo");
+        let result = Thumbnail::from_assertion(&assertion).expect("from_assertion");
+        assert_eq!(original.label, result.label);
+        assert_eq!(original.content_type, result.content_type);
+        assert_eq!(original.data, result.data);
     }
 }
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -26,7 +26,7 @@ use uuid::Uuid;
 #[cfg(doc)]
 use crate::Manifest;
 use crate::{
-    assertion::{get_thumbnail_image_type, Assertion, AssertionBase},
+    assertion::{Assertion, AssertionBase},
     assertions::{self, labels, AssetType, Metadata, Relationship, Thumbnail},
     asset_io::CAIRead,
     claim::{Claim, ClaimAssetData},
@@ -41,7 +41,10 @@ use crate::{
     resource_store::{skip_serializing_resources, ResourceRef, ResourceStore},
     status_tracker::StatusTracker,
     store::Store,
-    utils::{mime::extension_to_mime, xmp_inmemory_utils::XmpInfo},
+    utils::{
+        mime::{extension_to_mime, format_to_mime},
+        xmp_inmemory_utils::XmpInfo,
+    },
     validation_results::ValidationResults,
     validation_status::{self, ValidationStatus},
 };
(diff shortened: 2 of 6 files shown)


tmathern:
Sidenote - you will need a version of https://github.com/contentauth/c2pa-rs/pull/1193 too. Whichever PR pickes it up first and merged first "wins" the right to have more passing checks again.

sdk/src/ingredient.rs
```diff
@@ -1613,7 +1619,7 @@ mod tests {
         let ingredient = Ingredient::from_memory_async(format, image_bytes)
             .await
             .expect("from_memory_async");
-        // println!("ingredient = {ingredient}");
+        println!("ingredient = {ingredient}");
```
tmathern: Keep this debug statement as-is?

sdk/src/ingredient.rs
```diff
@@ -705,13 +708,7 @@ impl Ingredient {
...
 
     fn thumbnail_from_assertion(assertion: &Assertion) -> (String, Vec<u8>) {
-        let thumbnail_format = extension_to_mime(
-            &get_thumbnail_image_type(&assertion.label_root()).unwrap_or("".into()),
-        );
-        (
-            thumbnail_format.unwrap_or("image/none").to_string(),
-            assertion.data().to_vec(),
-        )
+        (assertion.content_type(), assertion.data().to_vec())
```
tmathern: Do we want this data even if no thumbnail is added before/later? Will the call be skipped if there is no thumbnail data?

sdk/src/settings.rs
```diff
@@ -683,4 +683,44 @@ pub mod tests {
 
         reset_default_settings().unwrap();
     }
+
+    #[test]
+    fn test_all_setting() {
```
tmathern, line 688: That test is not related to that change, is it? It's a bonus test?

sdk/src/utils/test.rs
```diff
@@ -102,7 +102,7 @@ pub(crate) fn static_test_v1_uuid() -> &'static str {
 /// Creates a claim for testing (v2)
 pub fn create_test_claim() -> Result<Claim> {
     // First create and add a claim thumbnail (we don't need to reference this anywhere)
-    let mut claim = Claim::new("adobe unit test", Some("adobe"), 2);
+    let mut claim = Claim::new("adobe unit test", Some("contentauth"), 2);
```
tmathern:
```suggestion
    let mut claim = Claim::new("contentauth unit test", Some("contentauth"), 2);
```

Do you want to update the other string too maybe, so it's consistent?

sdk/src/assertions/thumbnail.rs
```diff
@@ -53,6 +53,16 @@ impl Thumbnail {
...
+    /// Create a new Thumbnail with a specific content type
+    /// The label should not contain the image type postfix
+    pub fn new_with_format(label: &str, data: Vec<u8>, content_type: &str) -> Self {
+        Thumbnail {
+            data,
+            label: label.to_owned(),
+            content_type: content_type.to_owned(),
+        }
+    }
 }
```
ok-nick: Can we take a `String` for these parameters? If we are going to do an allocation anyways, best to leave it up to the caller.

sdk/src/ingredient.rs
ok-nick: Can we return `(&str, &[u8])` to avoid the allocations?

sdk/src/assertions/thumbnail.rs
```diff
@@ -53,6 +53,16 @@ impl Thumbnail {
             content_type,
         }
     }
+
+    /// Create a new Thumbnail with a specific content type
+    /// The label should not contain the image type postfix
+    pub fn new_with_format(label: &str, data: Vec<u8>, content_type: &str) -> Self {
```
ok-nick: Name this `with_format`?

Also, is this format or mime type? It looks like in other parts of the code we explicitly set it to the mime type. If that's the case it may help to rename this function to something like `with_mime`.

sdk/src/assertions/thumbnail.rs
```diff
@@ -53,6 +53,16 @@ impl Thumbnail {
             content_type,
         }
     }
+
+    /// Create a new Thumbnail with a specific content type
+    /// The label should not contain the image type postfix
```
ok-nick: I think we should also deprecate the `::new()` function if we are no longer creating v1 manifests.

sdk/src/settings.rs
gpeacock: yes this is bonus test. It was a handy way to document the current collection of settings in one place.

sdk/src/ingredient.rs
gpeacock: this is only an internal function used in two places. I updated it to only return references to the data.

sdk/src/assertions/thumbnail.rs
gpeacock: I added the EmbeddedData Assertion and we can use that for all future thumbnails and icons. That means that this Thumbnail Assertion only exists for generating v1 claims and we can remove it entirely when we remove support for generating v1 claims.

sdk/src/ingredient.rs
gpeacock: yes I updated this to return references, thanks for the suggestion

sdk/src/assertions/thumbnail.rs
gpeacock: I've removed this in favor of using the new EmbeddedData assertion.

sdk/src/assertions/thumbnail.rs
gpeacock: I removed new_with_format and now we just use EmbeddedData::new()

ok-nick (approved):

PR #1191, opened 2025-06-25
fix: change timestamp validation to return new required info fields

mauricefisher64:
Log info for every timestamp error
Timestamp must be validated against Timestamp trust before using
Support intermediate certs trust anchors for OpenSSL path 
Support separate trust list reporting
Fix for upcoming spec change to disallow empty algorithm value in a Claim

## Checklist


tmathern:
Short note and link - The checks on the PR for clippy are failing due to new enforced clippy rules after the Rust update.
Fixes are in the PR https://github.com/contentauth/c2pa-rs/pull/1193 (targets that branch).

sdk/src/crypto/asn1/mod.rs
```diff
@@ -14,7 +14,11 @@
 // crate contain dependencies on blocking calls in reqwest. Those calls are not
 // available in WASM environment.
 
+#[allow(dead_code)]
```
gpeacock, line 17: There's probably a better way to deal with this, but we can get to that later.

sdk/src/store.rs
```diff
@@ -203,6 +196,11 @@ impl Store {
         Ok(self.ctp.add_end_entity_credentials(allowed_vec)?)
     }
 
+    // Replace default trust policy
```
gpeacock: Did you mean to leave this in?

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -11,127 +11,204 @@
...
             "unable to find signed data".to_string(),
         ));
     };
 
-    let certs = sd.certificates.clone().ok_or(TimeStampError::DecodeError(
-        "time stamp contains no certificates".to_string(),
-    ))?;
+    // Grab the list of certs used in signing this timestamp
+    let Some(certs) = &sd.certificates else {
+        log_item!("", "count not parse timestamp data", "verify_time_stamp")
```
gpeacock, line 82: These seem like they could be very hard to track down without the URL/label. We won't know what manifest we are in. The StatusTracker can keep track of the current_uri like this
log_item(validation_log.current_uri().unwrap_or_default()),

mauricefisher64: Right, if I don't supply the first variable to log_item it fills it in

sdk/src/crypto/asn1/mod.rs
mauricefisher64: Ideally we could remove this if we do the rest in RASN.  I think I am down to TstInfo to implement.

sdk/src/store.rs
mauricefisher64: No I don't want this yet but we may need something like it in the future.

PR #1192, opened 2025-06-25
feat: expand settings API

### LINKED ISSUE #586: Implement Mandatory Created and Opened Actions
Ensure that we only create created or opened manifests.

With v2 claims, every manifest must either be a created manifest with a known digitalSourceType or a opened manifest with a valid parent ingredient. The API should prevent you from creating a manifest that is not one of these. It should be an error.

My proposal is that users can still manually create a manifest with the right components if they really know what they are doing. But, when signing, if we don't recognize a valid created manifest we will attempt to use the source asset as the parent and create the associated parent ingredient and opened assertion. The default generated ingredient and assertion will have limited information, but enough to be valid. If a user desires to add more information, they must either add the parent themselves or I may provide some default action fields (another topic).

We will always assume an opened manifest unless the user has provided us with acknown digitalSourceType. Rather than force them to predefine the created action correctly, we could allow a new creation_type field in the manifest defintion. If specified, it would need to have one of a specific enumerated set of values. A similar option would be to have a new builder constructor such as Builder::created(SourceType)

If an opened manifest needs to supply a different parent than the source image, they will need to set the parent ingredient before signing. This is often the case when an image has been modified so we need to capture original image as the parent and the source image will be the modified one. I may provide a Builder constructor for this case as well:
Builder::opened(source, format)

I'm not sure, with all the new changes, if we still need to provide a manifest definition when creating a builder. Many of the fields in the definition can now be provided in the setting - such as the claim-generator-info. If the builder allows for add_assertion and add_ingredient, there isn't much else to set. I've been toying around with a add_action method that would simplify a lot of the work with actions. Of course something like c2patool would still need a way to pass a full manifest.

### LINKED ISSUE #747: Convert c2patool to use settings config

### LINKED ISSUE #755: Signing Image File always creates thumbnail in C2PA-Manifest and results in big file size
Signing Image File always create thumbnail and results in big file size for signed image.

tested with c2patool 0.9.12 

Creation of thumbnail should to be optional and if enabled, you should define size (width/height) by your own

**Source JPG 653kBytes
Signed  JPG 838kBytes and thumbnail 165kBytes** 

e.g.

```
./c2patool source.jpg -m test.json  -o ext_dest.jpg
```

(test.json same as https://github.com/contentauth/c2patool/blob/main/sample/test.json and self signed certs)

source
ext_dest
c2pa thumbnail claim

### LINKED ISSUE #1006: Ingredients don't have thumbnails if image format is other than JPG and PNG
If signed image is other than JPG and PNG, ingredients don't have thumbnails

Image

Is there a particular reason why ingredient thumbnails aren't generated for images that aren't JPG or PNG? It seems like only minimal code modifications are needed to extend this functionality to other image formats as well.

By applying the following changes, I managed to get c2patool to generate an ingredient thumbnail from a WebP image:

Image

Image

### LINKED ISSUE #1215: AllActionsIncluded default to true
The spec did a flip/flop on this and now requires every actions to state explicitly that it includes all actions or else it will be assumed that it does not (and a warning would be displayed). Since this was never required before in our SDK, I think the only safe way to handle this is to default this field to true, and let users set it to false if they know or suspect that actions are missing. 
I'd suggest that we have a setting that would define if the claim generator wants this to default to true or false.

This could be part of a larger effort to set default values for any action.

### LINKED ISSUE #1216: Signer from Settings
With the new Settings configuration effort, we can new define all the values needed for a signer in the settings, so there is no longer a need to pass them separately in Signer::from_info or other methods. We could also handle CAWG signing in the same way. If you have specified CAWG configuration info for than we could automatically create a cawg signer from those settings. This vastly simplifies the API and examples. We would just need to call something like signer = Signer::from_settings(), and then pass that signer to our Builder signing method.

The settings can also allow for a remote signer URL. I'm thinking this could replace the current callback signer  function, or at least leverage it. The idea is that there would be a standard protocol for calling a simple signer similar to what we have done for the callback API. But it would call a service. That service could be local or remote. 

It would also be possible to directly support an HSA this way, but that might be better wrapped in a Rest API.

ok-nick:
Expands the interface of the `Settings` API, adding more ways to configure it and more ways to use it.

As part of this change, it standardizes on using the toml format for settings. Narrowing down supported formats on the config crate is estimated to reduce around 40-50 dependencies in the latest version.

### New Options
* `signer`
  * `local`
    * `alg` - algorithm used for signing
    * `sign_cert` - cert used for signing
    * `private_key` - private key used for signing
    * `tsa_url` - timestamp authority url used for signing
  * `remote`
    * `url` - URL to remote signer (simple POST sent for now)
    * `alg` - algorithm used for signing
    * `sign_cert` - cert used for signing
    * `tsa_url` - timestamp authority url used for signing
* `builder`
  * `claim_generator_info` - claim generator info to use if not already specified in the builder when signing
    * `operating_system` - this is the only field that differs from the current claim generator info definition
      * `infer` - whether to infer the current operating system (default)
      * `name` - or override it manually with a specific string
  * `thumbnail`
    * `enabled` - whether to enable thumbnail generation
    * `ignore_errors` - whether to ignore errors on thumbnails generation (current behavior) 
    * `long_edge` - longest edge of the thumbnail preserving aspect ratio
    * `format` - by default we now generate thumbnails matching the input, otherwise force this format
    * `prefer_smallest_format` - prefer smaller file size formats, like png/tiff with no alpha->jpeg
    * `quality` - low, medium, or high; we choose sensible defaults for each format
  * `actions`
    * `all_actions_included` - whether or not all actions are included in the actions assertion (default true)
    * `actions` - list of additional actions to add to the actions assertion
    * `templates` - list of additional templates to add to the actions assertion
    * `auto_created_action` and `auto_opened_action` and `auto_placed_action`
      * `enabled` - whether to enable the auto_X_action
      * `source_tyoe` - source type of the action (required for created action)

### New API
TODO: doc

### TODO
* Auto generate example toml from doc comments and defaults for user-facing fields?
* Block scoped settings?
* Reader/Builder specific settings?

### Related Issues
* Closes #1215 
* Closes #1216
* Closes #1006
* Closes #755
* Closes #747
* Closes #586
* Blocked by #1197 

## Checklist


sdk/src/settings.rs
```diff
@@ -218,40 +233,327 @@ impl Default for Verify {
 
 impl SettingsValidate for Verify {}
 
+// TODO: thumbnails/previews for audio?
```
mauricefisher64: Seems like this belongs with thumbnail code and not the settings.

sdk/src/settings.rs
```diff
@@ -37,15 +37,25 @@ pub(crate) trait SettingsValidate {
...
+    private_key: Vec<u8>,
+    ta_url: Option<String>,
+}
+
 // Settings for trust list feature
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-#[cfg_attr(feature = "json_schema", derive(JsonSchema))]
 #[allow(unused)]
-pub(crate) struct Trust {
+pub struct Trust {
```
mauricefisher64: Why is this public now?

sdk/src/settings.rs
```diff
@@ -130,6 +141,7 @@ impl Default for Trust {
                 trust_anchors: None,
                 trust_config: None,
                 allowed_list: None,
+                signer_info: None,
```
mauricefisher64: This is not part of trust.  Is part of manifest generation so I would think Builder is the place for this.

sdk/src/settings.rs
```diff
@@ -165,6 +176,9 @@ pub(crate) struct Core {
     prefer_bmff_merkle_tree: bool,
     compress_manifests: bool,
     max_memory_usage: Option<u64>,
+
+    prefer_update_manifests: bool,
+    exclude_box_hash_metadata: bool,
```
mauricefisher64: This is probably not sufficient.  Which metadata?  I would think it would only be ones we currently look for.  A list of strings is probably appropriate but it would only have "XMP" and the currently support metadata type.

sdk/src/settings.rs
```diff
@@ -218,40 +233,327 @@ impl Default for Verify {
...
+    enabled: bool,
+    /// Whether to ignore thumbnail generation errors.
+    ///
+    /// This may occur, for instance, if the thumbnail media type or color layout isn't
+    /// supported.
+    ignore_errors: bool,
+    /// The size of the thumbnail.
+    ///
+    /// Note this function will preserve aspect ratio based on the longest edge.
+    size: (u32, u32),
```
mauricefisher64: If this is the case why do we need two numbers.

sdk/src/settings.rs
```diff
@@ -218,40 +233,327 @@ impl Default for Verify {
...
     builder: Builder,
+
+    #[serde(skip)]
+    config: Config,
 }
 
 impl Settings {
-    #[allow(unused)]
+    /// Returns the current [Settings] for the local thread.
+    pub fn thread_local() -> Option<Settings> {
```
mauricefisher64: How is this different than get_settings?
I would not call this thread_local because it means nothing.  The thread has no way to know of any other settings so IT IS the settings.  And technically it is a snapshot of the settings as such I am inclined to do away this this function since it could lead to some tricky bugs.  Just let user request the info they need from the Settings structure withouth holding a copy that could be stale.

sdk/src/settings.rs
```diff
@@ -218,40 +233,327 @@ impl Default for Verify {
...
+            &String::from_utf8(setting_buf)
+                .map_err(|_| Error::BadParam("invalid utf-8".to_owned()))?,
+        )
+    }
+
+    /// Merges the current [Settings] with thread local [Settings].
+    ///
+    /// Only fields that are present in the current [Settings] will override the fields
+    /// in the thread local [Settings].
+    pub fn set_thread_local(self) -> Result<()> {
```
mauricefisher64: Same comment as above.  You are setting the Settings.

sdk/src/settings.rs
```diff
@@ -218,40 +233,327 @@ impl Default for Verify {
...
+    }
+
+    /// Constructs a signer from the specified `trust.signer_info` in the settings.
+    ///
+    /// The returned signer can be passed to [Builder::sign][crate::Builder::sign]
+    /// and other related functions.
+    ///
+    /// This function will error with [Error::UnspecifiedSignerSettings][crate::Error::UnspecifiedSignerSettings]
+    /// if the `trust.signer_info` is unspecified.
+    pub fn signer(&self) -> Result<Box<dyn Signer>> {
```
mauricefisher64: Why is this part of settings?  Settings do work they hold values for other code that does the actual work.

sdk/src/settings.rs
ok-nick: I named it `thread_local` to make it clear that these settings only affect this thread, not others. If you create a new thread you might assume to expect the same settings, but that's not the case. I also agree with the stale settings point.

sdk/src/settings.rs
ok-nick: This should be `pub(crate)`.

sdk/src/settings.rs
ok-nick: Agreed.

sdk/src/settings.rs
ok-nick: I included it for convenience since it maps directly to the image-rs `thumbnail` function although this can be changed.

sdk/src/settings.rs
ok-nick: We need to find a way to get the `Signer` so it can be passed to `Builder::sign`, otherwise need to redesign some of the builder functions.

sdk/Cargo.toml
```diff
@@ -185,9 +186,18 @@ ureq = "2.4.0"
 url = "2.5.3"
 
 [target.'cfg(any(target_os = "wasi", not(target_arch = "wasm32")))'.dependencies]
-image = { version = "0.24.7", default-features = false, features = [
-    "jpeg",
+image = { version = "0.25.6", default-features = false, features = [
+    # TODO: eventually define feature flags for the asset handlers and match them
```
gpeacock, line 212: Nice idea, we will have to keep that in mind. Good to add the todo here!

sdk/Cargo.toml
```diff
@@ -185,9 +186,18 @@ ureq = "2.4.0"
...
+    #       up here.
     "png",
+    "jpeg",
+    "gif",
+    "webp",
+    "tiff",
+    "bmp",
```
gpeacock: I don't think there's any point in BMP and ico support, those aren't supported C2PA data formats and I don't think anyone is going to use those for thumbnails. I'm don't know about the size savings, but no sense in adding code we don't need.

sdk/src/error.rs
```diff
@@ -210,6 +210,12 @@ pub enum Error {
     #[error("type is unsupported")]
     UnsupportedType,
 
+    #[error("thumbnail format {0} is unsupported")]
+    UnsupportedThumbnailFormat(String),
```
gpeacock, line 229: We really need to work on subclassing errors. We have way too many already. I think these might fit under something like Settings errors.

sdk/src/ingredient.rs
```diff
@@ -894,13 +894,14 @@ impl Ingredient {
...
-                }
+
+            if let Some((output_format, image)) =
+                crate::utils::thumbnail::make_thumbnail_bytes_from_stream(
+                    std::io::BufReader::new(stream),
+                    format,
+                )?
+            {
+                self.set_thumbnail(output_format.to_string(), image)?;
             }
```
gpeacock, line 909: I'd like to at least log issues here. I've debated a lot about it being a hard error.

sdk/src/ingredient.rs
```diff
@@ -1378,13 +1380,14 @@ impl Ingredient {
...
             stream.rewind()?;
-            match crate::utils::thumbnail::make_thumbnail_from_stream(format, stream) {
-                Ok((format, image)) => {
-                    ingredient.set_thumbnail(format, image)?;
-                }
-                Err(err) => {
-                    log::warn!("Could not create thumbnail. {err}");
-                }
+
+            if let Some((output_format, image)) =
```
gpeacock, line 1376: Maybe this should be a hard error if we can't generate a requested thumbnail. Or maybe that's another setting?

sdk/src/utils/thumbnail.rs
```diff
@@ -11,79 +11,264 @@
...
+    /// An image in PNG format.
+    Png,
+    /// An image in JPEG format.
+    Jpeg,
+    /// An image in GIF format.
+    Gif,
+    /// An image in WEBP format.
+    WebP,
+    /// An image in TIFF format.
+    Tiff,
```
gpeacock: I'm not sure if anyone really wants to make a tiff, bmp or ico thumbnail.

sdk/src/utils/thumbnail.rs
```diff
@@ -11,79 +11,264 @@
...
+        }
     };
-    let thumbnail_bits = Vec::new();
-    let mut cursor = std::io::Cursor::new(thumbnail_bits);
-    img.write_to(&mut cursor, output_format)?;
 
-    let format = format.to_owned();
-    Ok((format, cursor.into_inner()))
+    let ignore_errors =
+        settings::get_profile_settings_value::<bool>("builder.thumbnail.ignore_errors")?;
```
gpeacock: Ah, ok you did add a setting for this!

sdk/src/utils/thumbnail.rs
```diff
@@ -11,79 +11,264 @@
...
+                "builder.thumbnail.format",
+            )?;
+            match global_format {
+                Some(global_format) => global_format,
+                None => {
+                    let prefer_smallest_format = settings::get_profile_settings_value::<bool>(
+                        "builder.thumbnail.prefer_smallest_format",
+                    )?;
+                    match prefer_smallest_format {
+                        true => match input_format {
```
gpeacock, line 195: there are probably similar tricks for large formats like tiff to add here.

cli/src/main.rs
```diff
@@ -138,6 +138,15 @@ struct CliArgs {
...
     #[clap(long, default_value("20000"))]
     reserve_size: usize,
+
+    // TODO: ideally this would be called config, not to be confused with the other config arg
+    /// Path to the config file.
+    #[clap(long)]
+    settings: Option<PathBuf>,
+
+    /// Which profile of the settings to use.
+    #[arg(long, requires = "settings")]
```
gpeacock: this should have a default value

cli/src/main.rs
```diff
@@ -138,6 +138,15 @@ struct CliArgs {
     /// will probably leave extra `0`s of unused space. Please specify a reserve-size if possible.
     #[clap(long, default_value("20000"))]
     reserve_size: usize,
+
+    // TODO: ideally this would be called config, not to be confused with the other config arg
+    /// Path to the config file.
+    #[clap(long)]
+    settings: Option<PathBuf>,
```
gpeacock: This needs an option to load from an environment variable and perhaps also from some known file locations depending on the os.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+trust_anchors = ""
+# String to trust configuration.
+trust_config = ""
+# Path to allowed certificate list (PEM format).
+allowed_list = ""
+
+# Core functionality.
+[settings.core]
+# Hash algorithm to use (sha256, sha384, sha512).
+hash_alg = "sha256"
```
gpeacock: we may need to add a soft hash value for soft bindings, I'm not sure.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+# The "default" profile is the base profile and is overridable by other profiles. Profiles are
+# set programatically via the API. Any fields not specified in the activated profile will redirect
+# to the "default" profile. Any fields not specified in the "default" profile will redirect to
+# internal defaults.
+[profile]
+
+# Claim generator info list.
+[[profile.default.claim_generator_info]]
+# A human readable name.
+name = "my_service"
```
gpeacock: No underscores needed or wanted in the name field here. We probably should have it capitalized as well

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+# A human readable string of the product's version.
+version = "1.0"
+# URL to the icon.
+icon = ""
```
gpeacock: Gotta do some thinking on this one. I don't currently support remote URL icons though the spec allows for it. We could always fetch and embed. We should also allow a file path here or perhaps just embedded content ( there's an embeded URI format)
If we do the download and embed feature, we may need to add an embed flag to indicate if they want the icon embedded or remote once we support that.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+# The operating system the claim generator is running on.
+operating_system = ""
+# Arbitrary fields can also be defined.
+some_other_field = ""
```
gpeacock, line 76: currently our SDK always embeds an extra identifier for the SDK here. We may want to have a way to disable that feature, but lets not worry about that yet. Maybe we add a comment here that says we do that though.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+
+# Any amount of claim generator infos can be defined.
+[[profile.default.claim_generator_info]]
+name = "my_other_service"
```
gpeacock: Yes for other profiles, but we need to make it clear somewhere that only one claim_generator_info can be added to a claim now. That's a new restriction from the v2 claim spec.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+
+# A signer can be loaded in the API with the function `c2pa::get_settings_signer`.
+[profile.default.signer]
+# Algorithm to use for signing.
+alg = "sha256"
+# Certificate used for signing (PEM format).
+sign_cert = ""
+# Private key used for signing (PEM format).
+private_key = ""
```
gpeacock: When we support other ways to configure signers, we should comment that the private_key mode here is only for testing.

sdk/src/assertions/actions.rs
```diff
@@ -29,6 +29,13 @@ use crate::{
 const ASSERTION_CREATION_VERSION: usize = 2;
 pub const CAI_INGREDIENT_IDS: &str = "org.cai.ingredientIds";
 
+// TODO: document, maybe make enum?
+pub mod digital_source_type {
+    pub const EMPTY: &str = "http://c2pa.org/digitalsourcetype/empty";
```
gpeacock, line 39: Yes this should become an Enum that is part of the public SDK.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,108 @@
...
+# Claim generator info list.
+[[profile.default.claim_generator_info]]
+# A human readable name.
+name = "my_service"
+# A human readable string of the product's version.
+version = "1.0"
+# URL to the icon.
+icon = ""
+# The operating system the claim generator is running on.
+operating_system = ""
```
gpeacock: I've never understood this field since there is no standard described for it. It also makes little sense to put in a profile since it should be determined by the system it is running on. Maybe we have some flag to indicate we want it added?

sdk/src/builder.rs
```diff
@@ -864,9 +877,66 @@ impl Builder {
...
+        let auto_created = settings::get_profile_settings_value::<bool>("auto_created_action")?;
+        let auto_opened = settings::get_profile_settings_value::<bool>("auto_opened_action")?;
+        if auto_created || auto_opened {
+            let has_parent = self
+                .definition
+                .ingredients
+                .iter()
+                .any(|ingredient| ingredient.is_parent());
+            let action = match (has_parent, auto_created, auto_opened) {
+                (true, _, true) => Some(Action::new(c2pa_action::OPENED)),
```
gpeacock: There's more to this. The opened action must also have an associated parentOf ingredient. The catch is that the the spec requires ingredient assertion urls that have already been added to the claim. So iterating through the ingredient definitions here is too early in the process. We need to add the actions after we add the ingredients. Another complication is that for opened, we may need to create the ingredient here.

 For created we will need to know what the digitalSourceType should be. I'm debating about how to get that info, but it could potentially be another default setting. In some configurations, the input content will always have the same source type (i.e. AI generated)

sdk/src/builder.rs
```diff
@@ -864,9 +877,66 @@ impl Builder {
...
+                }
+            }
+        }
+
+        // https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_relationship
+        let auto_placed = settings::get_profile_settings_value::<bool>("auto_placed_action")?;
+        if auto_placed {
+            for ingredient in &self.definition.ingredients {
+                if *ingredient.relationship() == Relationship::ComponentOf {
+                    actions.actions.push(Action::new(c2pa_action::PLACED));
```
gpeacock: Placed must be associated with a componentOf ingredient. I think the idea here is probably that if a componentOf ingredient doesn't have an associated placed action, we will add one for them.

sdk/src/builder.rs
```diff
@@ -1483,6 +1558,124 @@ mod tests {
...
 
+    #[test]
+    fn test_builder_auto_created() {
+        let _guard =
+            settings::set_scoped_profile_settings_value("auto_created_action", true).unwrap();
+
+        let mut output = Cursor::new(Vec::new());
+        Builder::new()
+            .sign(
+                &test_signer(SigningAlg::Ps256),
```
gpeacock: this is where we really should be using signer from settings

sdk/src/builder.rs
```diff
@@ -1483,6 +1558,124 @@ mod tests {
...
+        let reader = Reader::from_stream("image/jpeg", output).unwrap();
+
+        let actions: Actions = reader
+            .active_manifest()
+            .unwrap()
+            .find_assertion(Actions::LABEL)
+            .unwrap();
+
+        let action = actions.actions().first().unwrap();
+        assert_eq!(action.action(), c2pa_action::OPENED);
```
gpeacock, line 1791: good, we should confirm that the action refers to the appropriate parent ingredient here.

sdk/src/builder.rs
```diff
@@ -1483,6 +1558,124 @@ mod tests {
...
+        let reader = Reader::from_stream("image/jpeg", output).unwrap();
+
+        let actions: Actions = reader
+            .active_manifest()
+            .unwrap()
+            .find_assertion(Actions::LABEL)
+            .unwrap();
+
+        let action1 = actions.actions().get(1).unwrap();
+        assert_eq!(action1.action(), c2pa_action::PLACED);
```
gpeacock, line 1876: again lets check the associations of the ingredients

sdk/src/claim.rs
```diff
@@ -1975,25 +1975,29 @@ impl Claim {
 
             if let Some(assertion) = created_actions.first() {
                 let first_actions = Actions::from_assertion(assertion.assertion())?;
-                let first_actions_first_action = &first_actions.actions()[0];
+                let first_actions_first_action = &first_actions.actions().first();
```
gpeacock: much better!

sdk/src/claim.rs
```diff
@@ -1975,25 +1975,29 @@ impl Claim {
...
 
-                if first_actions_first_action.action() == c2pa_action::OPENED
-                    || first_actions_first_action.action() == c2pa_action::CREATED
-                {
-                    found_first_action = Some(assertion);
+                if let Some(first_actions_first_action) = first_actions_first_action {
```
gpeacock: might want to factor this out since we repeat the logic three times

sdk/src/builder.rs
```diff
@@ -632,6 +633,13 @@ impl Builder {
 
         claim_generator_info[0].insert("org.cai.c2pa_rs", env!("CARGO_PKG_VERSION"));
 
+        let profile_claim_generator_infos = settings::get_profile_settings_value::<
```
gpeacock: Doing all the updates in to_claim has an interesting issue. We are not updating the definition in the builder. So none of the changes we make here will be reflected there. If we add all this information and then archive and restore the builder. We will still only have the information that was in the definition. Maybe that's ok. It has the advantage that we will always use the current settings when you decide to sign.
I think this is probably how it should work, but we should be on the lookout for unexpected consequences.

sdk/src/settings.rs
```diff
@@ -224,32 +228,154 @@ impl Default for Verify {
 
 impl SettingsValidate for Verify {}
 
-// Settings for Builder API options
+/// Quality of the thumbnail.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum ThumbnailQuality {
```
gpeacock: i'm not sure how we determine this is a way that is useful to most people.

sdk/src/settings.rs
```diff
@@ -224,32 +228,154 @@ impl Default for Verify {
...
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct SignerInfo {
+    alg: SigningAlg,
+    sign_cert: Vec<u8>,
+    private_key: Vec<u8>,
+    tsa_url: Option<String>,
+}
+
+/// A configuration profile.
```
gpeacock: I was thinking we should add all of these values directly to settings. I think all of these new values would fit under builder in settings. 

I could make an argument that trust should be under verify, or maybe we flatten it all under settings. I just don't see a reason to create a new Profile setting.

sdk/src/settings.rs
```diff
@@ -369,8 +499,20 @@ pub(crate) fn load_settings<P: AsRef<Path>>(settings_path: P) -> Result<()> {
...
 #[allow(unused)]
+// TODO: when this is removed, remove the additional features (for all supported formats) from the Cargo.toml
+#[deprecated]
 pub fn load_settings_from_str(settings_str: &str, format: &str) -> Result<()> {
-    Settings::from_string(settings_str, format).map(|_| ())
+    Settings::from_string(settings_str, format, None).map(|_| ())
+}
+
+#[allow(unused)]
+pub fn load_settings_from_toml(toml: &str) -> Result<()> {
```
gpeacock: The current load_settings function (from file) is private to the crate and was never published, so we can reuse that name for this one and just call it load_settings. If we want to keep the other we should rename it to load_settings_from_file()
if the future is only to support toml, then we don't to specify it in the name.

sdk/src/settings.rs
```diff
@@ -437,6 +645,49 @@ pub(crate) fn get_settings_value<'de, T: serde::de::Deserialize<'de>>(
     })
 }
 
+#[allow(unused)]
+pub(crate) fn get_profile_settings_value<'de, T: serde::de::Deserialize<'de>>(
```
gpeacock: Just call this get_setting_value(). The profile is a global setting and will always apply.

sdk/src/settings.rs
```diff
@@ -437,6 +645,49 @@ pub(crate) fn get_settings_value<'de, T: serde::de::Deserialize<'de>>(
...
+                Err(err) => Err(err.into()),
+            }
+        })
+    })
+}
+
+/// Returns the signer specified in the "signer" field of the currently active settings profile.
+///
+/// If the signer settings aren't specified, this function will return [Error::UnspecifiedSignerSettings][crate::Error::UnspecifiedSignerSettings].
+pub fn get_profile_settings_signer() -> Result<Box<dyn Signer>> {
```
gpeacock: maybe just call this signer_from_settings(). In rust we don't usually use get_ for getters. we could use create_signer_from_settings() if we want an action for this.

gpeacock: I'm not crazy about having a bunch of isolated global functions, so maybe we should have a public Settings class that has static methods?

sdk/Cargo.toml
ok-nick: Removed.

sdk/src/error.rs
ok-nick: Yes I agree although I think there's a lot we can change here in regard to that. It may be better left off to a separate PR.

sdk/src/utils/thumbnail.rs
ok-nick: Yep I suspect there's a lot of little tricks we can do here. Will need to go over it and see what's ideal.

cli/src/main.rs
ok-nick: It does technically default to the "default" profile already, which is set with our internal defaults.

sdk/examples/c2pa.toml
ok-nick: There is a generalized soft hash field defined in the claim, I think that's a good idea to add here.

sdk/examples/c2pa.toml
ok-nick: I will remove the array of claim generator infos then.

sdk/src/settings.rs
ok-nick: Originally I had a setting for every file format, compression, quality, and encoding option. That's a lot of code, I think if you want anything more advanced then you should probably be generating thumbnails yourself.

sdk/src/settings.rs
ok-nick: I agree, I had something similar implemented earlier but removed it to narrow the scope of the PR. I think ideally the settings would be more self contained so we can add it as options to builders/readers in the future.
https://github.com/contentauth/c2pa-rs/blob/7893cac3b4a353f05c841f56ee5a0f9ff84659f4/sdk/src/settings.rs#L469-L602

sdk/examples/c2pa.toml
ok-nick: The "icon" field actually refers to `ClaimGeneratorInfo::icon`, which is a `UriOrResource`. So I should clarify here that it does accept a file path, URL, or technically a JUMBF URI although I don't think the latter is useful here.

sdk/examples/c2pa.toml
ok-nick: Good idea.

sdk/src/assertions/actions.rs
ok-nick: The spec also supports source types from https://cv.iptc.org/newscodes/digitalsourcetype/.

sdk/examples/c2pa.toml
ok-nick: I think we should keep the ability to manually specify it as it's defined in the spec, but add another option (set by default) to auto specify it.

sdk/src/builder.rs
ok-nick: Yes, I think ideally the builder wouldn't change while signing so that it can be reusable.

sdk/src/settings.rs
ok-nick: The problem is that our current `get_settings_value` takes an absolute path to the field. If we use that we'd need a way to fetch the profile every time, add it to the path string, then if that field doesn't exist, check the default profile.

sdk/src/builder.rs
ok-nick: Note that we should also add this to V1 and V2 action validation, see below:
> When using a c2pa.opened or c2pa.placed action, the ingredient field (for v1) or ingredients field (for v2) in the parameters object shall contain the hashed JUMBF URIs to one or more related ingredient assertions. In a c2pa.removed action, this field shall contain the hashed JUMBF URI to a componentOf ingredient assertion in a different manifest.

Ideally we'd somehow "force" them to specify digital source type in the API, but for the sake of simplicity adding a config option is the way to go.

sdk/src/claim.rs
ok-nick: See #1227

sdk/src/assertions/actions.rs
ok-nick: I'm going to leave this as constants for now otherwise we'll have to plumb through a lot of the code.

sdk/src/builder.rs
ok-nick: Another option is to set `auto_created_action` to `false` by default and force users to enable it and set the digital source type manually.

Edit: This is now the current behavior.

sdk/examples/c2pa.toml
ok-nick: Also FYI I don't think any of the settings in `settings.core` are implemented/used?

sdk/examples/c2pa.toml
gpeacock: If we allow a URL here, for now we would need to fetch it and then embed it. That might be a nice feature, but secondary to just making it work with a path.

sdk/src/builder.rs
```diff
@@ -632,6 +633,13 @@ impl Builder {
 
         claim_generator_info[0].insert("org.cai.c2pa_rs", env!("CARGO_PKG_VERSION"));
 
+        let profile_claim_generator_infos = settings::get_settings_value::<
+            Option<ClaimGeneratorInfo>,
+        >("builder.claim_generator_info");
+        if let Ok(Some(claim_generator_infos)) = profile_claim_generator_infos {
```
gpeacock: v2 claims only allow one claim generator and we only care about this feature for v2. So we can't push a second one here. If the manifest definition has a claim generator, we will use it, otherwise we add this one.

sdk/src/builder.rs
```diff
@@ -864,9 +877,105 @@ impl Builder {
...
+                (true, _, true) => {
+                    let source_type = settings::get_settings_value::<Option<String>>(
+                        "builder.auto_opened_action.source_type",
+                    )?;
+                    let action = {
+                        let action = Action::new(c2pa_action::OPENED);
+                        match source_type {
+                            Some(source_type) => action.set_source_type(source_type),
+                            None => action,
+                        }
```
gpeacock: a OPENED action is required to have an associated ingredient hashed uri. We may need to pass the parent hashed_uri into this method. The builder doesn't keep track of those since they are part of the generated claim. I have an ingredient_map for this, but it doesn't currently identify the parent.

sdk/src/builder.rs
```diff
@@ -864,9 +877,105 @@ impl Builder {
...
+                    };
+
+                    Some(action)
+                }
+                (false, true, _) => {
+                    let source_type = settings::get_settings_value::<Option<String>>(
```
gpeacock, line 1006: I think we need to add an optional created_source_type field to builder or the manifest defintion. If that exists, we will presume this is created and use that source type. That field should be an enum of the possible source types so we don't burden the user with knowing all these urls.

You don't need to include that in this PR, but I think it is part of the whole story.

gpeacock: I need to add extra logic that would automatically create a parent if not specified, and the key to knowing if I should do that would be the lack of the above created_source_type field.

sdk/src/builder.rs
```diff
@@ -864,9 +877,105 @@ impl Builder {
...
+    /// * `builder.auto_opened_action`
+    /// * `builder.auto_placed_action`
+    fn add_auto_actions_assertions(&self, actions: &mut Actions) -> Result<()> {
+        // https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_mandatory_presence_of_at_least_one_actions_assertion
+        let auto_created =
+            settings::get_settings_value::<bool>("builder.auto_created_action.enabled")?;
+        let auto_opened =
+            settings::get_settings_value::<bool>("builder.auto_opened_action.enabled")?;
+        if auto_created || auto_opened {
+            let has_parent = self
```
gpeacock: We need to base this not on the manifest defintion but on the generated ingredient hashed uri map. We need that in order to add the ingredient references to the actions. It probably makes sense to update all the actions here, not just the ones we are adding. But that can come later.

gpeacock: With all of these we need to check if the action is already defined. The user may have added it and we don't want to add it twice.

gpeacock (changes requested):

sdk/src/builder.rs
ok-nick: I am able to map the found parent ingredients id (label/instance) to its hashed URI via the ingredient_map.

sdk/examples/c2pa.toml
ok-nick: The spec allows specifying the icon as URL I believe without having to embed it? I think this feature is already supported by the SDK?

See here: https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_schema

sdk/src/builder.rs
```diff
@@ -627,7 +629,17 @@ impl Builder {
         let metadata = definition.metadata.clone();
         // add the default claim generator info for this library
         if claim_generator_info.is_empty() {
-            claim_generator_info.push(ClaimGeneratorInfo::default());
+            let claim_generator_info_settings = settings::get_settings_value::<
+                Option<ClaimGeneratorInfoSettings>,
+            >("builder.claim_generator_info");
+            match claim_generator_info_settings {
```
gpeacock, line 654: we can only have one claim_generator even though it is an array for backward compatibility. V2 only allows one. So we need to check if the user already defined one. If they have, we keep it. Otherwise if we have settings, we use that, otherwise we can use the default, or perhaps return an error, since they really should not use the default.

sdk/src/utils/test_signer.rs
```diff
@@ -78,41 +78,41 @@ pub(crate) fn async_test_signer(alg: SigningAlg) -> Box<dyn AsyncSigner> {
     ))
 }
 
-fn cert_chain_and_private_key_for_alg(alg: SigningAlg) -> (Vec<u8>, Vec<u8>) {
+fn cert_chain_and_private_key_for_alg(alg: SigningAlg) -> (&'static [u8], &'static [u8]) {
     match alg {
         SigningAlg::Ps256 => (
-            include_bytes!("../../tests/fixtures/certs/ps256.pub").to_vec(),
```
gpeacock, line 84: I like this!

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,210 @@
...
+[[settings.builder.actions.templates]]
+action = "c2pa.cropped"
+
+# Actions to be added to every "Actions" assertion.
+[[settings.builder.actions.actions]]
+# The label of the action.
+action = "c2pa.drawing"
+# TODO: probably don't want to relist all of the fields, see previous todo comment
+
+# Similarly, multiple actions can be defined.
```
gpeacock, line 111: Shouldn't this just be an array of actions?

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,210 @@
...
+# or if some are missing.
+all_actions_included = true
+
+# A template to use as the base values for a particular action.
+[[settings.builder.actions.templates]]
+# The label of the action.
+action = "c2pa.edited"
+# TODO: do we want to document these fields here or just include links to the docs
+#       rust docs or c2pa docs?
+software_agent = ""
```
gpeacock: A software agent is a structure, how do we define that here? in JSON or toml or ??

cli/src/main.rs
```diff
@@ -138,6 +139,23 @@ struct CliArgs {
...
+    /// Path to the config file.
+    #[clap(
+        long,
+        env = "C2PATOOL_SETTINGS",
+        default_value = default_settings_path().into_os_string()
+    )]
+    settings: PathBuf,
+}
+
+fn default_settings_path() -> PathBuf {
```
gpeacock, line 155: Can you add a comment on this about where this default file is?

cli/src/main.rs
```diff
@@ -375,52 +393,79 @@ fn configure_sdk(args: &CliArgs) -> Result<()> {
...
         }
 
         if let Some(allowed_list) = &allowed_list {
-            let data = load_trust_resource(allowed_list)?;
             debug!("Using allowed list from {allowed_list:?}");
-            let replacement_val = serde_json::Value::String(data).to_string(); // escape string
-            let setting = AL.replace("replacement_val", &replacement_val);
 
-            c2pa::settings::load_settings_from_str(&setting, "json")?;
+            let data = load_trust_resource(allowed_list)?;
```
gpeacock, line 415: We don't need to do this yet, but we may want to support reading from urls like this in the regular settings (or at least use settings to load these values instead of the command trust_list command  option.

sdk/examples/c2pa.toml
gpeacock: yes it does, but we don't support that in our implementation. There are some serious security issues to that feature.

sdk/src/assertions/actions.rs
```diff
@@ -29,6 +29,20 @@ use crate::{
 const ASSERTION_CREATION_VERSION: usize = 2;
 pub const CAI_INGREDIENT_IDS: &str = "org.cai.ingredientIds";
 
+// TODO: make enum
+/// Constants indicating from which source a digital image was created.
+///
+/// These constants are used in conjunction with [`Actions::source_type`].
+pub mod source_type {
```
gpeacock: make this private or pub(crate) if you need it. I don't want to publish it like this.

cli/src/main.rs
mauricefisher64: This is why I said loading the toml may require additional processing to hand special things we want to support like URL or clearing values.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,172 @@
...
+
+# Configuration for a signer.
+#
+# The signer can be retrieved via the `Settings::signer` function.
+[signer]
+
+# A signer can be loaded in the API with the function `Settings::signer`.
+[signer.local]
+# Algorithm to use for signing.
+alg = "ps256"
```
mauricefisher64, line 35: Why are we setting any alg here since we don't know what cert they are using?

ok-nick: This is just an example config file to understand how to use it, these aren't the defaults.

sdk/examples/c2pa.toml
```diff
@@ -0,0 +1,172 @@
...
+# Version information.
+version_major = 1
+version_minor = 0
+
+# # Trust settings for certificate validation.
+# [trust]
+# # String to user-provided trust anchors (PEM format).
+# user_anchors = ""
+# # String to system trust anchors (PEM format).
+# trust_anchors = ""
```
mauricefisher64, line 12: Like Gavin hinted at.  May this could also take a URL and the validation code could fetch if it found a URL here instead of an anchor list.  Not a requirement for this PR.

sdk/src/builder.rs
mauricefisher64: Yes, no CGI should be an error.  Not sure if we want a default.

sdk/src/ingredient.rs
mauricefisher64: 👍

cli/src/main.rs
```diff
@@ -138,6 +139,25 @@ struct CliArgs {
...
+    // TODO: ideally this would be called config, not to be confused with the other config arg
+    /// Path to the config file.
+    ///
+    /// By default config files are read from `$XDG_CONFIG_HOME/c2pa/c2pa.toml`.
+    #[clap(
+        long,
+        env = "C2PATOOL_SETTINGS",
+        default_value = default_settings_path().into_os_string()
+    )]
+    settings: PathBuf,
```
gpeacock, line 152: I'm not sure if this will end up like this, but I'm ok with it for now.

sdk/src/builder.rs
```diff
@@ -870,9 +891,201 @@ impl Builder {
...
+                        _ => None,
+                    }
+                }
+                _ => None,
+            };
+
+            // If the first action isn't "c2pa.created" or "c2pa.opened" then add ours,
+            // or if there are no actions then add our action.
+            if let Some(action) = action {
+                if let Some(first_action) = actions.actions.first() {
```
gpeacock, line 1026: We may need to do more here if someone didn't get the order right when setting these., maybe we have an initial pass that moves those actions, if they exist, to the top. IDK

sdk/src/builder.rs
```diff
@@ -1491,6 +1711,356 @@ mod tests {
         assert_eq!(test_assertion.answer, 42);
     }
 
+    #[test]
+    fn test_builder_settings_auto_created() {
+        #[cfg(target_os = "wasi")]
+        Settings::reset().unwrap();
+
+        Settings::from_toml(
+            &toml::toml! {
```
gpeacock, line 1720: Nice, though I think we want to make the auto settings default, so we only need to override this if we are testing the disabled case.

sdk/src/builder.rs
```diff
@@ -1491,6 +1711,356 @@ mod tests {
...
+        let action1 = actions.actions().get(1).unwrap();
+        assert_eq!(action1.action(), c2pa_action::PLACED);
+
+        let action2 = actions.actions().get(2).unwrap();
+        assert_eq!(action2.action(), c2pa_action::PLACED);
+
+        let reader_json = reader.json();
+
+        for action in [action1, action2] {
+            let ingredient_uris = action.get_parameter("ingredients").unwrap();
```
gpeacock, line 1884: I think there's a couple of things here:
- action should have a get_ingredients method that either returns the ingredient uri or the actual ingerdients.
- we need a reader.get_assertion_from_uri() method.

sdk/src/error.rs
```diff
@@ -207,6 +210,12 @@ pub enum Error {
     #[error("must fetch remote manifests from url {0}")]
     RemoteManifestUrl(String),
 
+    #[error("failed to fetch the remote settings")]
```
gpeacock, line 213: I don't want to release all these new errors before an error rework, but lets get them in and I'll work on that.

gpeacock (approved):
This looks good, we have some work to do before publishing, but this is ready to merge now. Thanks!

PR #1203, opened 2025-07-07
feat: Asset reference assertion

### LINKED ISSUE #1212: Asset Reference Assertion support
Add support for the [Asset Reference Assertion](https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_asset_reference) .

We need to verify that a user can create one of these assertions and that they can be read back successfully after being stored in the correct format. These are used to include a reference to the source image for a manifest. An example would be the url that Adobe Stock used to provide as part of the CreativeWork assertion. This should be displayed in L3 UI.

alextrnnn:
Added helper struct for asset reference assertion
Closes #1212
## Checklist


sdk/src/assertions/asset_reference.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn add_reference(mut self, reference: Reference) -> Self {
+        self.references.push(reference);
+        self
+    }
+}
+
+/// Defines a single location of where a copy of the asset may be obtained.
+#[derive(Deserialize, Serialize, Debug, Default, PartialEq, Eq)]
+pub struct Reference {
+    pub reference: InnerReference,
```
gpeacock: it is unclear to me why the spec doesn't just have a url field in the reference structure, but I'd guess it is so they can add other fields here later (maybe a hash?)
We have a HashedUri structure that contains a url field, so I'm going to suggest something like ReferenceUri for this.

gpeacock (changes requested):
This looks good. I added a suggestion for a renamed field, even though I can see it isn't public. I'd really like to see an integration level test for this.

sdk/src/assertions/asset_reference.rs
```diff
@@ -0,0 +1,145 @@
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    assertion::{Assertion, AssertionBase, AssertionCbor},
+    assertions::labels,
+    error::Result,
+};
+
+const ASSERTION_CREATION_VERSION: usize = 2;
```
gpeacock: There was no previous version of this assertion, so we don't need to specify one here. We can use the default.

sdk/src/assertions/asset_reference.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub struct AssetReference {
+    pub references: Vec<Reference>,
+}
+
+impl AssetReference {
+    pub const LABEL: &'static str = labels::ASSET_REFERENCE;
+
+    /// Creates an AssetReference to a location.
+    pub fn new(reference: Reference) -> Self {
```
gpeacock: Given that the Reference is just a struct with a URL, I think we can just pass the Url as a String here and not bother with exporting the Reference type.

sdk/src/assertions/asset_reference.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Deserialize, Serialize, Debug, Default, PartialEq, Eq)]
+pub struct ReferenceUri {
+    pub uri: String,
+}
+
+impl AssertionCbor for AssetReference {}
+
+impl AssertionBase for AssetReference {
+    const LABEL: &'static str = Self::LABEL;
+    const VERSION: Option<usize> = Some(ASSERTION_CREATION_VERSION);
```
gpeacock: You can leave out the version and let it use the default here. We only need this when there are newer versions to support.

gpeacock (changes requested):

sdk/src/assertions/asset_reference.rs
```diff
@@ -0,0 +1,128 @@
+use serde::{Deserialize, Serialize};
```
gpeacock, line 14: Every source code file must have the open source copyright at the top. You need to add that here. Copy it from another of the assertions.

gpeacock (changes requested):
Nice work! Add the copyright to your source file and we are good to go.

PR #1220, opened 2025-07-08
fix: Same file sign

### LINKED ISSUE #1117: Support input and output being the same file?
Currently it's unsupported to have the input and output file be the same file even with the `--force` flag set. This is due to early deleting of the output file if the file exists and `--force` is set.

This worked in 0.9.12 but doesn't work in the latest version, so could technically be classified as a regression.

## Current behavior

```
$ c2patool image.jpg -m manifest.json -o image.jpg -f


-----------

Note: Using default private key and signing certificate. This is only valid for development.
A permanent key and cert should be provided in the manifest definition or in the environment variables.

Error: embedding manifest

Caused by:
    No such file or directory (os error 2)
```

Additionally, `image.jpg` is deleted.

## Expected behavior

Either:

* Successful signing, a manifest is applied to `image.jpg`.
* Error message indicating that input and output cannot be the same file. Input file is not deleted.

alextrnnn:
Currently, signing a file and outputting to an existing file will:
1. Check to see if the file exists, and if it does, bail if there was no force flag. If there is a force flag, remove the file.
2. Makes a call to the `sign_file` method

My proposed solution will:
1. Check to see if the file exists and there was no force flag, if so it will bail. 
2. Write to a temporary file and copies it to the output path. 
3. Makes a call to the `sign` method

This allows signing where the input and output are the same file. The main difference is, no matter what, it will follow the pattern of writing to a temporary file and copying it to the output path (even if there is no need to overwrite an output). It will also now use the `sign` method rather than the `sign_file` method.

NOTE: NamedTempFile has a security warning mentioned in the rust [docs](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html). I talked with Andrew and he said it probably won't be applicable to us.
>The timescales for a file cleaner deleting a file are on the order of days, and we only need to file to survive for the runtime of a single invocation, which I would say tops out at a minute for a very big file."

NOTE: Persisting did not work for instances where `tmp/` is a different file system.

CALLOUT: Should there be concerns of performance with writing to a tempfile and copying it? The previous way was just to write to a file directly.

Resolves #1117

## Checklist


cli/src/main.rs
```diff
@@ -710,24 +712,35 @@ fn main() -> Result<()> {
...
+                    builder.sign(signer.as_ref(), &format, &mut source, &mut file)?;
+
+                if !output.exists() {
+                    // ensure the path to the file exists
+                    if let Some(output_dir) = &output.parent() {
+                        create_dir_all(output_dir)?;
+                    }
+                }
+
+                copy(file, &output)?;
```
gpeacock: our practice with temp files is to first try to rename the temp to the output. and if that fails, we do a copy. This avoids a copy when they are on the same volume.

cli/src/main.rs
```diff
@@ -710,24 +712,35 @@ fn main() -> Result<()> {
...
                     bail!("Missing extension output");
                 }
 
-                let manifest_data = builder
-                    .sign_file(signer.as_ref(), &args.path, &output)
-                    .context("embedding manifest")?;
+                let mut file = NamedTempFile::new()?;
+                let format = format_from_path(&args.path).unwrap();
+                let mut source = File::open(&args.path)?;
+                if builder.definition.title.is_none() {
```
gpeacock: We have a builder.sign_file which does most of this work for you. The catch is that you would need to set the title before calling it since it will otherwise use the title of the destination temp file.

cli/src/main.rs
```diff
@@ -710,24 +712,35 @@ fn main() -> Result<()> {
...
                     bail!("Missing filename on output");
                 }
                 if output.extension().is_none() {
                     bail!("Missing extension output");
                 }
 
-                let manifest_data = builder
-                    .sign_file(signer.as_ref(), &args.path, &output)
-                    .context("embedding manifest")?;
+                let mut file = NamedTempFile::new()?;
```
gpeacock: We should only do this if the output is the same file as the input. 
If the file is under a certain size we could use an in memory vector with a cursor so that no other file is needed. 
We used to support this in the SDK but I removed it since there are a number of issues to work through, but I think it is ok to support in c2patool.
The output file path could be on a network drive so performance can be a significant issue in some cases. We also don't know much about how temporary storage is handled (it could be limited). Hence is is good idea to make this feature optional.

cli/src/main.rs
alextrnnn: I tried using sign_file before, but in the call to `set_asset_from_dest`, it has the following logic that causes it to error out, since the temp file's path exists:
```
        if !path.exists() {
            // ensure the path to the file exists
            if let Some(output_dir) = path.parent() {
                std::fs::create_dir_all(output_dir)?;
            }
        } else {
            // if the file exists, we need to remove it to avoid appending to it
            return Err(crate::Error::BadParam(
                "Destination file already exists".to_string(),
            ));
        };
```

cli/src/main.rs
alextrnnn: I did previously use a Cursor, but Andrew suggested a tempfile to prevent the case of storing large videos in memory. Do you have a recommendation for the size before we switch to using tempfiles?

By making the feature optional, do you mean creating a feature flag for this? Or even as a CLI argument?

gpeacock (approved):
Nice work. There is always a danger that a NamedTempFile will not get deleted if we hae a crash, and someone will probably complain, but the workaround is to not use this feature.

PR #1222, opened 2025-07-09
fix: Force update of c2pa_c_ffi and c2patool crates if only c2pa-rs is updated

scouten-adobe:
Based on testing over in my [rp-sandbox](https://github.com/scouten-adobe/rp-sandbox/) repo, I believe this will force updates of the downstream crates `c2pa-c-ffi` and `c2patool` any time the core `c2pa` crate is updated.

It's a bit hacky in that it creates additional commit traffic, but that seems to the only way with the currently-available release-plz tool to achieve our desired result.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,50 @@ jobs:
           echo "RP output: $PR"
           echo "========================="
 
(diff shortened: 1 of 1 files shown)


gpeacock (approved):
Ugly but needed for now. I hope we get the feature you asked for!

PR #1225, opened 2025-07-09
fix: Trigger c2pa-c-ffi binary releases based on c2pa-c-ffi-v*** tags

scouten-adobe:


scouten-adobe:
OK, I'll withdraw for now since #1222 should cause the release to happen as desired.

PR #1227, opened 2025-07-09
fix: panic when adding empty actions assertion

### LINKED ISSUE #1197: c2patool 0.19.0 index out of bounds
Testing c2patool 0.19.0 against my suite of forgeries:

```
$ c2patool-0.19.0 -d shmoocon.jpg
[2025-07-01T05:03:45Z ERROR c2pa::status_tracker] Validation error: LogItem {
        kind: Failure,
        label: "self#jumbf=/c2pa/urn:uuid:88eb2bb3-d5f9-4516-9265-b65277b4b861/c2pa.assertions/c2pa.actions.v2",
        description: "Actions missing action array",
        crate_name: "c2pa",
        crate_version: "0.57.0",
        file: "sdk/src/claim.rs",
        function: "verify_actions",
        line: 1938,
        err_val: Some(
            "ValidationRule(\"No Action array in Actions\")",
        ),
        validation_status: Some(
            "assertion.action.malformed",
        ),
        ingredient_uri: None,
    }

thread 'main' panicked at sdk/src/claim.rs:1981:51:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

$ RUST_BACKTRACE=full c2patool-0.19.0 -d shmoocon.jpg
[2025-07-01T05:06:08Z ERROR c2pa::status_tracker] Validation error: LogItem {
        kind: Failure,
        label: "self#jumbf=/c2pa/urn:uuid:88eb2bb3-d5f9-4516-9265-b65277b4b861/c2pa.assertions/c2pa.actions.v2",
        description: "Actions missing action array",
        crate_name: "c2pa",
        crate_version: "0.57.0",
        file: "sdk/src/claim.rs",
        function: "verify_actions",
        line: 1938,
        err_val: Some(
            "ValidationRule(\"No Action array in Actions\")",
        ),
        validation_status: Some(
            "assertion.action.malformed",
        ),
        ingredient_uri: None,
    }

thread 'main' panicked at sdk/src/claim.rs:1981:51:
index out of bounds: the len is 0 but the index is 0
```

The sample picture: https://fotoforensics.com/analysis.php?id=dc1be89f99cd320c5f7dbee556277e82ddd30b03.103760&fmt=orig

ok-nick:
Fixes an index out of bounds panic when adding an empty actions assertion (although note that this is still invalid and will be caught gracefully afterwards).

Closes #1197

## Checklist


sdk/src/claim.rs
```diff
@@ -2003,12 +2007,13 @@ impl Claim {
             // for v1 claims check the single assertion store
             if let Some(assertion) = all_actions.first() {
                 let first_actions = Actions::from_assertion(assertion.assertion())?;
-                let first_actions_first_action = &first_actions.actions()[0];
-
-                if first_actions_first_action.action() == c2pa_action::OPENED
-                    || first_actions_first_action.action() == c2pa_action::CREATED
-                {
-                    found_first_action = Some(assertion);
+                let first_actions_first_action = &first_actions.actions().first();
```
gpeacock, line 1992: For v1 claims, this rule does not apply and we should not try to enforce it here.

gpeacock (changes requested):
We need to not crash, but we really should not be applying this rule to v1 claims. They were not required to follow this.

sdk/src/claim.rs
ok-nick: Yes, ~~additionally I think we are missing this check:
The full set of actions assertions in a C2PA Manifest shall contain no more than one action whose type is either c2pa.created or c2pa.opened. If one of these actions appears within created_assertions, then neither shall appear within gathered_assertions, and if one appears within gathered_assertions, then neither shall appear within created_assertions.~~

tmathern (commented):
Please consider creating a release (sdk & C ffi) once this is merged, as this change needs tot rickle down to the other SDKs.

sdk/src/claim.rs
```diff
@@ -1976,57 +1976,29 @@ impl Claim {
...
-                let first_actions_first_action = &first_actions.actions()[0];
-
-                if first_actions_first_action.action() == c2pa_action::OPENED
-                    || first_actions_first_action.action() == c2pa_action::CREATED
-                {
-                    found_first_action = Some(assertion);
-                }
+        // Skip further checks for v1 claims if not in strict validation mode
+        if claim.version() == 1 {
+            if let Ok(false) = get_settings_value::<bool>("verify.strict_v1_validation") {
```
gpeacock, line 1981: I guess your reasoning behind checking these two arrays vs the actions array is that v1 claims won't have any assertions here. But I'm curious about why we need to support v1 strict validation at all.

ok-nick: We do check all the assertions later on, but this code is just finding what the first action is in either array. I moved the v1 check before it since there was no validation being done at that point anyways.

PR #1234, opened 2025-07-14
chore: Improve manifest comparison

gpeacock:
The make_test_images manifest comparison ignores things like version changes so that 
it can report no unexpected changes
Make_test_images has been updated to generate v2 images by default now.
The v1 and v2 known good json folders have been updated after validating the updates.


make_test_images/src/compare_manifests.rs
```diff
@@ -149,129 +146,258 @@ fn gather_manifests(manifest_store: &Reader, manifest_label: &str, labels: &mut
...
+                    if let Some(arr) = ingredients.as_array() {
+                        // Use first ingredient url if present
+                        if let Some(first) = arr.first() {
+                            if let Some(url) = first.get("url").and_then(|v| v.as_str()) {
+                                return Some(format!("{action}:{url}"));
+                            }
+                        }
+                    }
+                }
+            }
```
ok-nick: We can also use `serde_json::from_value()` with the `Actions` struct here. Not blocking, but it gives us a good indication of how easy it is to use our assertion definitions.

Edit: Ah maybe not, we don't care about any additional validation/behavior, we just want the raw structure here.

make_test_images/src/compare_manifests.rs
```diff
@@ -149,129 +146,258 @@ fn gather_manifests(manifest_store: &Reader, manifest_label: &str, labels: &mut
...
-        r"^(xmp:iid:)([0-9a-fA-F\-]{36})$"
-        ).unwrap();
-        static ref UTC_TIMESTAMP_RE: Regex = Regex::new(
-            r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?([+-]\d{2}:\d{2}|Z)$"
-        ).unwrap();
-        static ref BASE64_RE: Regex = Regex::new(r"^[A-Za-z0-9+/=]{16,}$").unwrap();
+        static ref GUID_RE: Regex = Regex::new(r"([a-zA-Z0-9\._-]+:)?(urn:(c2pa|uuid):|xmp.iid:)[0-9a-fA-F\-]{36}").unwrap();
+        static ref JUMBF_GUID_RE: Regex = Regex::new(r"self#jumbf=/c2pa/[^:]+:urn:uuid:[0-9a-fA-F\-]{36}(/.*)?").unwrap();
+        static ref UTC_TIMESTAMP_RE: Regex = Regex::new(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?([+-]\d{2}:\d{2}|Z)$").unwrap();
+        static ref VERSION_RE: Regex = Regex::new(r"0\.\d+\.\d+").unwrap();
```
ok-nick: Always cool to see.

make_test_images/src/compare_manifests.rs
```diff
@@ -149,129 +146,258 @@ fn gather_manifests(manifest_store: &Reader, manifest_label: &str, labels: &mut
...
-        }
-        (serde_json::Value::Array(arr1), serde_json::Value::Array(arr2)) => {
-            for (i, (val1, val2)) in arr1.iter().zip(arr2.iter()).enumerate() {
-                compare_json_values(&format!("{path}[{i}]"), val1, val2, issues);
+
+            for key in keys2 {
+                if !map1.contains_key(key) {
+                    issues.push(format!("Added {path}.{key}: {}", map2.get(key).unwrap()));
+                }
             }
```
ok-nick: It might be better with something like:
```rust
for (key, value) in map1 {
    // todo
}
```
and same with keys2 so that it doesn't need to be collected, nor the value fetched from the map and unwrapped.

make_test_images/src/compare_manifests.rs
```diff
@@ -149,129 +146,258 @@ fn gather_manifests(manifest_store: &Reader, manifest_label: &str, labels: &mut
...
-                    PREFIXED_URN_UUID_CAPTURE.captures(s1),
-                    PREFIXED_URN_UUID_CAPTURE.captures(s2),
-                ) {
-                    if cap1.get(1).map_or("", |m| m.as_str())
-                        == cap2.get(1).map_or("", |m| m.as_str())
-                    {
-                        return;
-                    }
+            } else if use_custom_id {
+                use std::collections::HashMap;
```
ok-nick: Why import here?

make_test_images/src/compare_manifests.rs
```diff
@@ -149,129 +146,258 @@ fn gather_manifests(manifest_store: &Reader, manifest_label: &str, labels: &mut
...
+                        for v in arr2 {
+                            if let Some(action) = v.get("action").and_then(|a| a.as_str()) {
+                                map2.insert(action, v);
+                            }
+                        }
+                        for action in map1.keys() {
+                            if let (Some(v1), Some(v2)) = (map1.get(action), map2.get(action)) {
+                                compare_json_values(&format!("{path}[{action}]"), v1, v2, issues);
+                            }
+                        }
```
ok-nick: This is fine but it would be nice if it was separated out into its own function.

sdk/Cargo.toml
```diff
@@ -137,7 +137,7 @@ memchr = "2.7.4"
 mp4 = "0.14.0"
 nom = "7.1.3"
 non-empty-string = { version = "=0.2.4", features = ["serde"] }
-nonempty-collections = { version = "1.0", features = ["serde"] }
+nonempty-collections = { version = "0.3.1", features = ["serde"] }
```
ok-nick, line 140: Regressed?

gpeacock: no, this is intentional, the 1.0 release has breaking dependencies.

PR #1235, opened 2025-07-14
feat: Store icon references into c2pa.icon assertions for v2 instead of using data_boxes.

gpeacock:
This is required for 2.2 compliance


sdk/src/resource_store.rs
```diff
@@ -61,7 +63,17 @@ impl UriOrResource {
...
             UriOrResource::ResourceRef(r) => {
                 let data = resources.get(&r.identifier)?;
-                let hash_uri = claim.add_databox(&r.format, data.to_vec(), None)?;
+                let hash_uri = match claim.version() {
+                    1 => claim.add_databox(&r.format, data.to_vec(), None)?,
+                    _ => {
+                        let icon_assertion = EmbeddedData::new(
+                            labels::ICON,
+                            format_to_mime(&r.format),
+                            data.to_vec(),
```
ok-nick, line 72: Just something to watch out for, but this could be an expensive operation if we have large or a lot of resources.

sdk/src/ingredient.rs
```diff
@@ -1196,9 +1198,11 @@ impl Ingredient {
                     HashedUri::new(thumb_ref.identifier.clone(), thumb_ref.alg.clone(), &hash)
                 }
                 None => {
+                    // This is a resource reference so we need so we need to get the data from the resource
+                    // create thumbnail assertion and then return the hashed uri to that
                     let data = match self.thumbnail.as_ref() {
                         Some(thumbnail) => get_resource(&thumbnail.identifier),
-                        None => Err(Error::NotFound),
+                        None => Err(Error::ResourceNotFound(thumb_ref.identifier.to_string())),
```
ok-nick, line 1205: Nice.

sdk/src/resource_store.rs
gpeacock: Understood, it was always this way and hopefully icons are not too large. Resource.get actually returns a Cow here so if it is already a vec this won't cause any extra copy. It is a vec if we loaded the resource from file. I'm not sure what alternative we have the resource is stored in memory. It doesn't seem safe to just consume it. Ultimately we may want to make resources all stream based, and I've been moving in that direction.

PR #1236, opened 2025-07-16
feat: Certificate status assertion

alextrnnn:
Implements the Certificate Status assertion definition and support for checking for certificate revocation from ocsp values found in Certificate Status assertions. 

## Checklist


sdk/src/store.rs
```diff
@@ -4619,6 +4691,33 @@ pub mod tests {
         Ok(claim)
     }
 
+    #[test]
+    #[cfg(feature = "v1_api")]
```
gpeacock: Let's not write new tests that depend on v1_apis that are going away.

sdk/src/store.rs
```diff
@@ -4473,6 +4513,38 @@ impl Store {
...
+    #[allow(dead_code)]
+    /// Retrieves OCSP responses for certificate validation from the specified manifests.
+    ///
+    /// # Arguments
+    /// * `manifest_labels` - Vector of manifest labels to check for certificate assertions
+    /// * `validation_log` - Status tracker for logging validation events
+    ///
+    /// # Returns
+    /// A `Result` containing a vector of [`OcspResponse`] objects
+    pub fn get_certificate_assertion(
```
gpeacock: This seems oddly named, Isn't something more like get_oscp_responses?

sdk/src/store.rs
```diff
@@ -4473,6 +4513,38 @@ impl Store {
         )?;
         Ok(i_store)
     }
+
+    #[allow(dead_code)]
```
gpeacock, line 4875: Why is this dead_code?

sdk/src/assertions/mod.rs
```diff
@@ -29,6 +29,9 @@ pub use box_hash::{BoxHash, BoxMap, C2PA_BOXHASH};
 mod data_hash;
 pub use data_hash::DataHash;
 
+mod certificate_status;
+pub use certificate_status::CertificateStatus;
```
gpeacock: This should probably be pub(crate) since I don't know how anyone would use this directly.

sdk/src/store.rs
```diff
@@ -1777,6 +1784,39 @@ impl Store {
...
                 }
             }
+
+            // get the certificate status assertions
+            let certificate_status_assertions = found_claim.certificate_status_assertions();
+            for csa in certificate_status_assertions {
+                let certificate_status_assertion =
+                    CertificateStatus::from_assertion(csa.assertion()).map_err(|_e| {
+                        log_item!(
+                            csa.label(),
```
gpeacock: This should be the full uri to the assertion, you can construct it using the claim label and the csa_label().

sdk/src/cose_validator.rs
```diff
@@ -153,6 +154,14 @@ fn extract_serial_from_cert(cert: &X509Certificate) -> BigUint {
     cert.serial.clone()
 }
 
+/// Returns the unique serial number from the provided CoseSign1
+pub(crate) fn get_serial_num(sign1: &CoseSign1) -> Result<BigUint> {
```
mauricefisher64: Let's call this get_signing_cert_serial_num

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -68,7 +73,36 @@ pub fn check_ocsp_status(
             OcspFetchPolicy::FetchAllowed => {
                 fetch_and_check_ocsp_response(sign1, data, ctp, tst_info, validation_log)
             }
-            OcspFetchPolicy::DoNotFetch => Ok(OcspResponse::default()),
+            OcspFetchPolicy::DoNotFetch => {
+                if let Some(ocsp_response_ders) = ocsp_responses {
```
mauricefisher64, line 107: Add a comment here what we are doing.  

// If we are not fetching use the supplied OCSP responses.  If any not revoked found then that is a good status.  All other status are just informational.

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -68,7 +73,36 @@ pub fn check_ocsp_status(
...
+                    for ocsp_response_der in ocsp_response_ders {
```
mauricefisher64: Use a temporary StatusTrack in the loop.  Only merge in the results of the last loop that way you do not pollute the validation_log in a bunch of results we don't care about.  See how logging is done in verify_time_stamp.

sdk/src/store.rs
```diff
@@ -4502,6 +4542,38 @@ impl Store {
...
+    #[allow(dead_code)]
+    /// Retrieves OCSP responses for certificate validation from the specified manifests.
+    ///
+    /// # Arguments
+    /// * `manifest_labels` - Vector of manifest labels to check for certificate assertions
+    /// * `validation_log` - Status tracker for logging validation events
+    ///
+    /// # Returns
+    /// A `Result` containing a vector of [`OcspResponse`] objects
+    pub fn get_certificate_assertion(
```
mauricefisher64: Why is this function called get_certificate_assertion.  It looks like it returns OCSPResponses.  Should it return the actual created assertions?

sdk/src/validation_results.rs
```diff
@@ -853,6 +853,12 @@ pub mod validation_codes {
     /// Any corresponding URL should point to a C2PA assertion box.
     pub const ASSERTION_TIMESTAMP_MALFORMED: &str = "assertion.timestamp.malformed";
 
+    /// The certificate status assertion is malformed.
+    ///
```
mauricefisher64: The values in this file are from the C2PA spec.  This is not a spec value.

sdk/src/store.rs
alextrnnn: This method isn't called by any of the other code, iirc it was meant as a method Gavin would later call/make use of. I'm not sure exactly what he would find more convenient (retrieving a vec of `ocsp_ders` vs `ocsp_responses` vs a singular certificate status assertion. I've made it return a singular certificate status assertion for now.

mauricefisher64 (approved):
Looks good

sdk/src/ingredient.rs
```diff
@@ -328,6 +333,11 @@ impl Ingredient {
         self.data.as_ref()
     }
 
+    /// Returns a reference to the ocsp responses if it exists.
+    pub fn oscp_responses_ref(&self) -> Option<&Vec<ResourceRef>> {
```
gpeacock: Probably make this pub(crate) Ingredient is a public API, but I guess it doesn't do much harm

sdk/tests/integration.rs
```diff
@@ -364,6 +364,46 @@ mod integration_1 {
...
+        let signer = Settings::signer()?;
+        let mut source = Cursor::new(include_bytes!("fixtures/cai.jpg"));
+        builder.add_ingredient_from_stream(parent_json, "image/jpeg", &mut source)?;
+        builder.sign_file(signer.as_ref(), &parent_path, &output_path)?;
+
+        // read our new file with embedded manifest
+        let reader = Reader::from_file(&output_path)?;
+
+        let mut file = File::create("foo.txt")?;
+        file.write_all(reader.to_string().as_bytes())?;
```
gpeacock: This test needs to do something to validate that the assertion was added.

gpeacock (changes requested):
We need a test that validates that the expected certificate status assertion gets added on signing and that it validates correctly when reading.
Reading validation could be done in a separate PR, but we need to ensure that the assertion is created under the right conditions here.

Issue #1231, opened 2025-07-11
Add builder.set_resource_path


gpeacock:
We need a way to support users who leveraged the base_path method of reading Ingredients and manifests. The new API allows adding resources into a memory map so that they can be read, but the existing ingredient..json and manifest.json files have resource references that are difficult to process without this feature. There is already a supported base_path field in the Builder, but we have not made it accessible from other languages. This would add a setter to the builder called set_resource_dir(). If this is set, resources that are not found in memory will be searched for in the given directory. 

Builder should not write files to that folder if add_resource is used, added resources should go into memory unless archived. In memory resources are preferred when reading and always used when writing.
*This is a change in current behavior.

This api should also be exposed in the c2pa_c_ffi crate. Then we will need to update other crates that that use it, starting with c2pa-c for c++.
 
I see this as a temporary feature because in the long run I'd like to have a add_resource_resolver() method and you could add a resolver for any number of different systems, not just files.

PR #1237, opened 2025-07-16
feat: Add C binding for setting the base path

alextrnnn:
Add a C binding for setting the base path, which will be used to show an example of loading ingredients from a folder.

Related to #1231 

## Checklist


sdk/src/builder.rs
```diff
@@ -2230,4 +2249,32 @@ mod tests {
...
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
+
+    #[cfg(feature = "file_io")]
+    #[test]
+    fn test_base_path() -> Result<()> {
+        let mut builder = Builder::new();
+        let ingredient_folder = fixture_path("ingredient");
+        builder.set_base_path(&ingredient_folder);
+        assert_eq!(builder.base_path, Some(&ingredient_folder).cloned());
```
ok-nick:
```suggestion
        assert_eq!(builder.base_path.as_ref(), Some(&ingredient_folder));
```

sdk/src/builder.rs
```diff
@@ -2230,4 +2249,32 @@ mod tests {
         assert!(reader_json.contains("Test Ingredient"));
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
+
+    #[cfg(feature = "file_io")]
+    #[test]
+    fn test_base_path() -> Result<()> {
```
ok-nick: I think this should return nothing to stay consistent with the other tests. With that, everything would need to be unwrapped. This is helpful because if there's a problem in the test it tells us where in the test.

sdk/src/builder.rs
```diff
@@ -2230,4 +2249,32 @@ mod tests {
...
+
+        builder.sign(&signer, "image/jpeg", &mut source, &mut dest)?;
+
+        let reader = Reader::from_stream("jpeg", &mut dest)?;
+        let reader_json = reader.json();
+        println!("{}", &reader);
+        assert!(reader_json.contains("C.jpg"));
+        // manifest label of ingredient
+        assert!(reader_json.contains("contentauth:urn:uuid:b2b1f7fa-b119-4de1-9c0d-c97fbea3f2c3"));
+        Ok(())
```
ok-nick: I think it would be better to use the public API to find if these things exist. For example, something like `reader.active_manifest().unwrap().ingredients().first().unwrap().title()`.

gpeacock (approved):

PR #1239, opened 2025-07-16
feat: Add option for configuring trust when validating identity assertions (CAI-7980)

scouten-adobe:


scouten-adobe:
This builds on #1238.

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -325,7 +325,8 @@ impl IdentityAssertion {
         let sig_type = self.signer_payload.sig_type.as_str();
 
         if sig_type == "cawg.x509.cose" {
-            let verifier = X509SignatureVerifier {};
+            // TO DO BEFORE MERGING: Should be able to configure this.
```
scouten-adobe: @gpeacock when you're back, I'd like to discuss moving this into settings. I started on this but hit some dead ends.

cli/tests/integration.rs
```diff
@@ -536,6 +536,13 @@ fn tool_tree() -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
+// DO NOT MERGE with this test disabled. The problem is that the
```
tmathern: Why not let if fail then to block merging?

sdk/src/crypto/cose/certificate_trust_policy.rs
```diff
@@ -294,6 +294,11 @@ impl CertificateTrustPolicy {
         Ok(())
     }
 
+    /// Add default extended key usage (EKU) values.
+    pub fn add_default_valid_ekus(&mut self) {
```
tmathern, line 298: Not sure about how I feel on the acronym in the function name (ekus), but the whole thing is a moutful too.

cli/tests/integration.rs
```diff
@@ -549,6 +556,13 @@ fn tool_read_image_with_cawg_data() -> Result<(), Box<dyn Error>> {
...
+// underlying CAWG code now enforces trust list checks on any X.509
+// certificate used in an identity assertion. I _think_ this test
+// asset uses a test-quality cert and that isn't enabled by default.
+// Will need to consult with Gavin on a new fix approach before
```
gpeacock: I updated this recently I created it with the cawg_identity example, 
cargo r --example cawg_identity tests/fixtures/C.jpg target/C_with_CAWG_data.jpg
It uses test certs like this:
For testing we allow test certs to be in the trust list. Ask Maurice about that.

    const CAWG_CERTS: &[u8] = include_bytes!("../../sdk/tests/fixtures/certs/ed25519.pub");
    const CAWG_PRIVATE_KEY: &[u8] = include_bytes!("../../sdk/tests/fixtures/certs/ed25519.pem");

sdk/src/cose_sign.rs
```diff
@@ -151,6 +151,7 @@ fn signing_cert_valid(signing_cert: &[u8]) -> Result<()> {
     let mut passthrough_cap = CertificateTrustPolicy::default();
 
     // allow user EKUs through this check if configured
+    // TODO: Need to determine if we're using C2PA or CAWG trust config here.
```
mauricefisher64: Do you have a different set of EKUs?

scouten-adobe:
Per conversation with @gpeacock and @mauricefisher64, this is ready to go once the c2patool issues are resolved. Use `--settings (path)` in c2patool invocations to pass in the appropriate new settings file.

PR #1246, opened 2025-07-21
feat(sdk): c2pa_reader_remote_url returns *mut c_char

dyro:
This PR allows C clients to free const strings from Rust without having to `const_cast<char *>`the value passed into `c2pa_release_string`. This functionality will be used by clients who've called `Reader::remote_url(&self) -> *const char`. 

## Checklist


c2pa_c_ffi/src/c_api.rs
```diff
@@ -431,6 +431,19 @@ pub unsafe extern "C" fn c2pa_string_free(s: *mut c_char) {
...
+/// * Reads from NULL-terminated C strings.
+/// * The string can only be freed once and is invalid after this call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_const_string_free(s: *const c_char) {
+    if !s.is_null() {
+        // Cast is okay as strings allocated by Rust and then passed to C are mutable.
+        drop(CString::from_raw(s as *mut c_char));
+    }
+}
+
```
ok-nick, line 446: Hm, it doesn't seem right to make another string free function. Maybe we should instead be returning a `*mut ptr` from `Reader::remote_url` and save `*const ptr` for pointers that don't need to be freed.

ok-nick (changes requested):

c2pa_c_ffi/src/c_api.rs
dyro: Went back and forth on this. I chose to introduce the const version because returning a `*mut char` indicates to the user that they can modify the value being returned. I think returning the `*const`is the right call for the remote_url API

ok-nick: The way I was thinking about it is that we are giving ownership of the string to C, so they are free to do with it what they want and that would align better with free being a mutable operation. If we had some type of static Rust string that definitely shouldn't be modified, that would make sense to return a const pointer (which shouldn't be freed). 

But if we did go with this change, would we need a `*mut ptr` free API then? Wouldn't C implicitly cast it to a `*const ptr` anyways?

dyro: But we do have a precedent for using `*mut c_chars` as return types for values we'd probably think of as `const`. For example, the `Builder::json` API. I'll modify the `remote_url` API to return a `*const c_char`.

dyro:
Addressed your feedback, @ok-nick.

ok-nick (approved):

c2pa_c_ffi/src/c_api.rs
```diff
@@ -603,15 +603,16 @@ pub unsafe extern "C" fn c2pa_reader_json(reader_ptr: *mut C2paReader) -> *mut c
 ///
 /// # Safety
 /// reader_ptr must be a valid pointer to a C2paReader.
+/// The returned value MUST be released by calling c2pa_string_free.
 #[no_mangle]
-pub unsafe extern "C" fn c2pa_reader_remote_url(reader_ptr: *mut C2paReader) -> *const c_char {
+pub unsafe extern "C" fn c2pa_reader_remote_url(reader_ptr: *mut C2paReader) -> *mut c_char {
```
gpeacock: This should be treated as a constant by the caller, making any modification would be a very bad thing. Why is the caller treating it as mutable?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -586,7 +586,7 @@ pub unsafe extern "C" fn c2pa_reader_free(reader_ptr: *mut C2paReader) {
 /// Returns a JSON string generated from a C2paReader.
 ///
 /// # Safety
-/// The returned value MUST be released by calling c2pa_string_free
+/// The returned value MUST be released by calling c2pa_string_free,
 /// and it is no longer valid after that call.
 #[no_mangle]
 pub unsafe extern "C" fn c2pa_reader_json(reader_ptr: *mut C2paReader) -> *mut c_char {
```
gpeacock, line 605: The real bug here is that the Json string is not being treated as a constant. We should change this!

gpeacock (changes requested):
I think the right thing is to alway treat these returned strings as constants, so I don't agree with this change. Why is the client referring to it a mutable?

c2pa_c_ffi/src/c_api.rs
dyro: @gpeacock : I re-pushed my original commit. What do you think of these changes?

ok-nick (commented):
> But if we do go with this change, would we need a *mut ptr free API then? Wouldn't C implicitly cast it to a *const ptr anyways?

If we are going back to a const ptr, we should consider having only one free string function then?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -431,6 +431,19 @@ pub unsafe extern "C" fn c2pa_string_free(s: *mut c_char) {
     }
 }
 
+/// Frees a const string allocated by Rust.
+///
+/// # Safety
+/// * Reads from NULL-terminated C strings.
+/// * The string can only be freed once and is invalid after this call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_const_string_free(s: *const c_char) {
```
gpeacock, line 440: We should just change c2pa_string_free to take a *const c_char. There is no valid case where it should be treated as a non constant. The bug here is that I didn't change c2pa_string_free to take a *const c_char when I updated the other functions to return one. We don't need a new function for this, we just need to fix the current one.

gpeacock (changes requested):

c2pa_c_ffi/src/c_api.rs
gpeacock: None of the string values returned to C from Rust should be considered mut, they are all const from the C point of view.

gpeacock (changes requested):

PR #1253, opened 2025-07-23
fix: Support for BMFF V3 hashing options

mauricefisher64:
Support for validation of V3 features (fixed and variable sized mdat Merkle trees)
Support for generation of fixed size mdat Merkle trees
Fix for bug when top level exclusion is not follow immediately have by the top level box (subset exclusions)
Change HashRanges to be u64 by default to handle large mdata in 32 bit environments
Change streaming path to use save_vec when generating intermediate stream

## Checklist


sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -551,20 +565,22 @@ where
...
+                            if subset.length == 0 {
+                                exclusion_length - subset.offset
                             } else {
-                                min(subset.length as u64, exclusion_length)
-                            }) as usize,
+                                min(
+                                    subset.length,
+                                    exclusion_length.saturating_sub(subset.offset),
+                                )
+                            },
```
ok-nick: How come we don't use `saturating_sub` on line 570 as well?

sdk/src/store.rs
```diff
@@ -1911,7 +1941,10 @@ impl Store {
 
             if calc_hashes {
                 if block_end > block_start && (block_end as u64) <= stream_len {
-                    dh.add_exclusion(HashRange::new(block_start, block_end - block_start));
+                    dh.add_exclusion(HashRange::new(
+                        block_start as u64,
+                        (block_end - block_start) as u64,
+                    ));
```
ok-nick, line 1947: Maybe unrelated to this PR,  but should `HashObjectPositions` also have `u64` fields for block_start/end?

sdk/src/store.rs
```diff
@@ -2027,8 +2061,13 @@ impl Store {
         exclusions.push(trun);
         */
 
-        // enable flat flat files with Merkle trees
-        if flat_fragmented_w_merkle {
+        // enable flat flat files with Merkle trees if desired
+        // we do this here because the UUID boxes must be in place
+        // for the later hash generation
+        if let Ok(Some(merkle_chunk_size)) =
+            get_settings_value::<Option<usize>>("core.merkle_tree_chunk_size_in_kb")
```
ok-nick, line 2068: Should this be defined as a `u64` since we cast it to that anyways?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -985,10 +1007,10 @@ impl BmffHash {
         asset_path: &std::path::Path,
         fragment_paths: &Vec<std::path::PathBuf>,
         output_dir: &std::path::Path,
-        local_id: u32,
-        unique_id: Option<u32>,
+        local_id: usize,
+        unique_id: Option<usize>,
     ) -> crate::Result<()> {
-        let max_proofs: usize = 4; // todo: calculate (number of hashes to perform vs size of manifest) or allow to be set
+        let max_proofs = get_settings_value::<usize>("core.merkle_tree_max_proofs").unwrap_or(5);
```
ok-nick: Can we default this in settings instead?

sdk/src/settings.rs
```diff
@@ -162,7 +162,8 @@ pub(crate) struct Core {
     hash_alg: String,
     salt_jumbf_boxes: bool,
     prefer_box_hash: bool,
-    prefer_bmff_merkle_tree: bool,
+    merkle_tree_chunk_size_in_kb: Option<usize>,
+    merkle_tree_max_proofs: usize,
```
ok-nick: I think logically it would make sense if these are `u64`, if you are specifying a setting or embedding a manifest, it shouldn't be influenced by the platform. At least for the c2pa manifest...

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1208,11 +1230,310 @@ impl BmffHash {
...
+            .iter()
+            .map(|cbor_data| {
+                let mut uuid_box_data: Vec<u8> = Vec::with_capacity(cbor_data.len() + 32);
+                crate::asset_handlers::bmff_io::write_c2pa_box(
+                    &mut uuid_box_data,
+                    &[],
+                    false,
+                    cbor_data,
+                )
+                .unwrap_or_default(); // this cannot fail unless the vec allocation above fails
```
ok-nick, line 1419: `unwrap_or_default()` is returning `()` either way, should be equivalent when removing this line.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1208,11 +1230,310 @@ impl BmffHash {
...
+            uuid_boxes.push(mm_cbor);
+        }
+
+        // pad the uuid boxes  with zeros to the same length as the longest box per the spec
+        let max_length = uuid_boxes.iter().map(|b| b.len()).max().unwrap_or(0);
+        for uuid_box in &mut uuid_boxes {
+            if uuid_box.len() < max_length {
+                uuid_box.resize(max_length, 0);
+            }
+        }
```
ok-nick, line 1406: Not blocking but the iteration can be avoided if we used `if let Some(..` rather than `unwrap_or(0)`.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1208,11 +1230,310 @@ impl BmffHash {
...
+
+    // save desired Merkle tree row
+    // create a MerkleMap for a specific range of mdat box
+    pub(crate) fn create_merkle_map_for_mdat_box(
+        &self,
+        reader: &mut dyn CAIRead,
+        box_info: &BoxInfoLite,
+        merkle_map: &mut MerkleMap,
+    ) -> crate::Result<Vec<Vec<u8>>> {
+        let max_proofs = get_settings_value::<usize>("core.merkle_tree_max_proofs").unwrap_or(5);
```
ok-nick: Same here, can we default this in settings?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1208,11 +1230,310 @@ impl BmffHash {
...
+    ) -> crate::Result<C2PAMerkleTree> {
+        let alg = merkle_map
+            .alg
+            .as_ref()
+            .or(self.alg.as_ref())
+            .ok_or(Error::BadParam("alg is required".to_string()))?;
+
+        // generate leaves in the Merkle tree based on the box length and fixed or variable block sizes
+        let mut leaves = Vec::new();
+        if let Some(fixed_block_size) = merkle_map.fixed_block_size {
```
ok-nick, line 1290: Maybe we should do additional validation somewhere so someone can't accidentally specify both fixed block and variable block size.

sdk/src/assertions/bmff_hash.rs
mauricefisher64: We do default it in settings.

sdk/src/store.rs
mauricefisher64: Not sure it matters but since this in kb what would it mean to have 64bit worth of kbs.

sdk/src/settings.rs
mauricefisher64: Neither would ever be 64 bits in practice.

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: No harm in checking there too.

sdk/src/store.rs
mauricefisher64: Maybe when I get back to BoxHash stuff

sdk/src/assertions/bmff_hash.rs
mauricefisher64: Making it an errror if not set.

sdk/src/assertions/bmff_hash.rs
mauricefisher64: see this check
```
}  else if mm_vec
                .iter()
                .all(|mm| mm.fixed_block_size.is_some() || mm.variable_block_sizes.is_some())
```

sdk/src/assertions/bmff_hash.rs
```diff
@@ -487,24 +506,63 @@ impl BmffHash {
         self.verify_stream_hash(&mut data, alg)
     }
 
+    fn verify_self(&self) -> crate::error::Result<()> {
+        if self.is_remote_hash() {
+            return Err(Error::BadParam(
```
gpeacock, line 511: maybe make this a C2PAValidation error for consistency

sdk/src/assertions/bmff_hash.rs
```diff
@@ -550,10 +608,8 @@ impl BmffHash {
                 if let Some(init_hash) = &mm.init_hash {
                     if let Some(moof_box) = first_moof {
                         // add the moof to end exclusion
-                        let moof_exclusion = HashRange::new(
-                            moof_box.offset as usize,
-                            (size - moof_box.offset) as usize,
-                        );
+                        let moof_exclusion =
```
gpeacock, line 611: I can't see moof without thinking about Susan Kare's [DogCow](https://en.wikipedia.org/wiki/Dogcow).

sdk/src/assertions/bmff_hash.rs
mauricefisher64: Hmm, maybe but it used to be supported.  What is really saying is that we don't support it.

mauricefisher64: C2PAValidation means it actually has a C2PA validation code for the error.  I don't think there is one for this case.

sdk/src/utils/io_utils.rs
```diff
@@ -16,12 +16,33 @@ use std::path::PathBuf;
...
 #[allow(unused)] // different code path for WASI
 use tempfile::{tempdir, Builder, NamedTempFile, TempDir};
 
-use crate::{Error, Result};
+use crate::{asset_io::rename_or_move, Error, Result};
+// Replace data at arbitrary location and len in a file.
+// start_location is where the replacement data will start
+// replace_len is how many bytes from source to replaced starting a start_location
+// data is the data that will be inserted at start_location
+#[allow(dead_code)]
```
gpeacock, line 30: does this need a file_io feature setting?

sdk/src/validation_results.rs
```diff
@@ -771,7 +771,12 @@ pub mod validation_codes {
     /// A BMFF hash assertion is malformed.
     ///
     /// Any corresponding URL should point to a C2PA assertion box.
-    pub const ASSERTION_BOXHASH_MALFORMED: &str = "assertion.bmffHash.malformed";
+    pub const ASSERTION_BMFFHASH_MALFORMED: &str = "assertion.bmffHash.malformed";
+
+    /// A Box hash assertion is malformed.
+    ///
+    /// Any corresponding URL should point to a C2PA assertion box.
+    pub const ASSERTION_BOXESHASH_MALFORMED: &str = "assertion.boxesHash.malformed";
```
gpeacock, line 779: And I just now realized the the spec calls them BOXES Hashes!

sdk/src/store.rs
```diff
@@ -88,6 +94,7 @@ use crate::{
 use crate::{external_manifest::ManifestPatchCallback, RemoteSigner};
 
 const MANIFEST_STORE_EXT: &str = "c2pa"; // file extension for external manifests
+const MANIFEST_RESERVE_SIZE: usize = 10 * 1024 * 1024; // 10MB reserve size for manifest
```
gpeacock, line 97: we really should make this a settings option. Some may want it bigger.

sdk/src/store.rs
```diff
@@ -172,6 +179,29 @@ impl Store {
         store
     }
 
+    /* for coming PR
+    // Append new Store to the current Store preserving the order from the new Store
+    pub(crate) fn append_store(&mut self, store: &Store) {
```
gpeacock, line 184: Don't we always preserve the order?

sdk/src/store.rs
```diff
@@ -2621,7 +2708,10 @@ impl Store {
                 .await?
         };
 
-        let intermediate_output: Vec<u8> = Vec::new();
+        let intermediate_output: Vec<u8> = safe_vec(
+            stream_len(input_stream)? + MANIFEST_RESERVE_SIZE as u64,
+            None,
+        )?;
```
gpeacock, line 2716: What is the error here? We probably want something more specific to know if we hit this.

sdk/src/store.rs
```diff
@@ -3129,7 +3219,11 @@ impl Store {
         output_stream: &mut dyn CAIReadWrite,
         reserve_size: usize,
     ) -> Result<Vec<u8>> {
-        let intermediate_output: Vec<u8> = Vec::new();
+        // make sure we can hold the intermediate stream before attempting
+        let intermediate_output: Vec<u8> = safe_vec(
```
gpeacock, line 3225: maybe factor this out into a function that does a safe large vec allocate and returns an API specific error?

gpeacock: Actually I'd go a step further and have it return a stream, where that stream could be a temp file over some size if the file_io feature is available. Plus it could check for settings on size limits and file_backup options.

gpeacock (approved):

PR #1258, opened 2025-07-24
feat: Add metadata assertion and support for validation

alextrnnn:
Supports c2pa.metadata assertion with validation of allowed fields

## Checklist


sdk/src/assertions/meta.rs
```diff
@@ -0,0 +1,497 @@
...
+use serde_json::Value;
+
+use crate::{
+    assertion::{Assertion, AssertionBase, AssertionJson},
+    assertions::labels,
+    Error, Result,
+};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Meta {
```
mauricefisher64: Let's get a better name for this and the file.  Perhaps MetadataAssertion

sdk/src/claim.rs
```diff
@@ -3170,9 +3170,30 @@ impl Claim {
         // check action rules
         Claim::verify_actions(claim, svi, validation_log)?;
```
mauricefisher64, line 3259: Put this chunk in its own function lime verify_actions

sdk/src/claim.rs
```diff
@@ -3170,9 +3170,30 @@ impl Claim {
...
+                    Error::ValidationRule("fields must be in allowed list".into()),
+                )?;
+            }
+        }
         Ok(())
     }
 
+    pub fn metadata_assertions(&self) -> Vec<&ClaimAssertion> {
+        let dummy_data = AssertionData::Cbor(Vec::new());
+        let dummy_ingredient = Assertion::new(assertions::labels::METADATA, None, dummy_data);
```
mauricefisher64: There is actually. more to it than this.  You have not accounted for user metadata assertions.  If you look at the spec you should allow include labels that end in .metadata.  For validation you need to make sure they are valid c2pa.metadata sturctures (i.e. it has a @context and a list of key value pairs that match name spaces listed in the @context).  You do not have to check those  user keys against the allowed lists.

sdk/src/assertions/meta.rs
alextrnnn: There's a pending review on https://github.com/contentauth/c2pa-rs/pull/1261, once it's merged I can name this `Metadata`

sdk/src/assertions/meta.rs
```diff
@@ -0,0 +1,541 @@
...
+        if self.label == labels::METADATA {
+            for (namespace, url) in &self.context {
+                if let Some(expected_url) = ALLOWED_SCHEMAS.get(namespace.as_str()) {
+                    if url != expected_url {
+                        return false;
+                    }
+                }
+            }
+        }
+
```
mauricefisher64: We should add a check for an empty @context.  That would be malformed.

sdk/src/assertions/meta.rs
```diff
@@ -0,0 +1,541 @@
...
+    #[serde(rename = "@context")]
+    pub context: HashMap<String, String>,
+    #[serde(flatten)]
+    pub value: HashMap<String, Value>,
+    #[serde(skip)]
+    pub label: String,
+}
+
+impl Meta {
+    pub fn new(json: &str, label: &str) -> Result<Self, Error> {
```
mauricefisher64: let's call the param jsonld instead of json

sdk/src/claim.rs
```diff
@@ -3170,9 +3171,46 @@ impl Claim {
...
         Ok(())
     }
 
+    // Perform metadata validation check
+    fn verify_metadata(claim: &Claim, validation_log: &mut StatusTracker) -> Result<()> {
+        for metadata_assertion in claim.metadata_assertions() {
+            let metadata_assertion = Meta::from_assertion(metadata_assertion.assertion())?;
+            if !metadata_assertion.is_valid() {
+                log_item!(
+                    claim.uri(),
```
mauricefisher64: Use the full metadata URI here

sdk/src/assertions/metadata.rs
```diff
@@ -0,0 +1,556 @@
...
+    #[serde(rename = "@context")]
+    pub context: HashMap<String, String>,
+    #[serde(flatten)]
+    pub value: HashMap<String, Value>,
+    #[serde(skip)]
+    pub label: String,
+}
+
+impl Metadata {
+    pub fn new(jsonld: &str, label: &str) -> Result<Self, Error> {
```
gpeacock: functions should have doc comments, and we generally will pass labels before values.

sdk/src/assertions/metadata.rs
```diff
@@ -0,0 +1,556 @@
...
+use serde_json::Value;
+
+use crate::{
+    assertion::{Assertion, AssertionBase, AssertionJson},
+    assertions::labels,
+    Error,
+};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct Metadata {
```
gpeacock, line 34: This needs doc comments, both on the struct and on each public  field

PR #1260, opened 2025-07-24
feat: define `DigitalSourceType` as enum

ok-nick:
Define `Actions::source_type` as an enum so user's no longer need to manually specify the entire URL.

### Questions
* Do we want to provide the `Other(String)` variant or ensure they are only what we support?
  * This has the flexibility that it supports future c2pa versions but it can allow for incorrect URLs. 
* Do we want the `DigitalSourceType` variants in PascalCase, camelCase, snake_case, something else?
* A side-effect of this implementation is that we allow digital source types to be read as, e.g. "Empty" in addition to the URL, which is technically incorrect according to the spec but offers a much more convenient JSON manifest/settings definition.

## Checklist

### DIFF
--- a/sdk/examples/c2pa.toml
+++ b/sdk/examples/c2pa.toml
@@ -90,7 +90,7 @@ action = "c2pa.edited"
 # The source type field is required for the c2pa.created action.
 #
 # For more information, see `c2pa::assertions::actions::source_type`.
-source_type = "http://c2pa.org/digitalsourcetype/empty"
+source_type = "Empty"
 # Description for the action
 description = "Some edit action."
 # Arbitrary key/value pairs to store in the action.
@@ -122,7 +122,7 @@ enabled = true
 # The source type field is required for the c2pa.created action.
 #
 # For more information, see `c2pa::assertions::actions::source_type`.
-source_type = "http://c2pa.org/digitalsourcetype/empty"
+source_type = "Empty"
 
 # Settings for configuring how c2pa.opened actions are auto created.
 #
@@ -134,7 +134,7 @@ enabled = true
 # For more information, see `c2pa::assertions::actions::source_type`.
 #
 # Note this field is optional for the c2pa.opened action.
-source_type = "http://c2pa.org/digitalsourcetype/empty"
+source_type = "Empty"
 
 # Settings for configuring how c2pa.placed actions are auto created.
 #
@@ -146,7 +146,7 @@ enabled = true
 # For more information, see `c2pa::assertions::actions::source_type`.
 #
 # Note this field is optional for the c2pa.placed action.
-source_type = "http://c2pa.org/digitalsourcetype/empty"
+source_type = "Empty"
 
 # Settings for automatic thumbnail generation.
 [builder.thumbnail]
--- a/sdk/src/assertions/actions.rs
+++ b/sdk/src/assertions/actions.rs
@@ -11,7 +11,7 @@
 // specific language governing permissions and limitations under
 // each license.
 
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt};
 
 use log::error;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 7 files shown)


sdk/src/assertions/actions.rs
```diff
@@ -29,19 +29,186 @@ use crate::{
...
-    pub const EMPTY: &str = "http://c2pa.org/digitalsourcetype/empty";
+    #[serde(alias = "Empty", rename = "http://c2pa.org/digitalsourcetype/empty")]
+    Empty,
     /// Data that is the result of algorithmically using a model derived from sampled content and data.
     /// Differs from <http://cv.iptc.org/newscodes/digitalsourcetype/>trainedAlgorithmicMedia in that
     /// the result isn’t a media type (e.g., image or video) but is a data format (e.g., CSV, pickle).
-    pub const TRAINED_ALGORITHMIC_DATA: &str =
-        "http://c2pa.org/digitalsourcetype/trainedAlgorithmicData";
+    #[serde(
+        alias = "TrainedAlgorithmicMedia",
```
gpeacock: I'm glad you caught this, we could read this, but we shouldn't write it.

sdk/src/assertions/actions.rs
```diff
@@ -29,19 +29,186 @@ use crate::{
...
+    }
+}
+
+/// C2PA description of the source of an asset.
+///
+/// The full list of possible C2PA digital source types are found below:
+/// <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_digital_source_type>
+#[non_exhaustive]
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+pub enum C2paDigitalSourceType {
```
gpeacock: This reminds me that we really should have a type like this for format/content-type/mime-type

sdk/src/assertions/actions.rs
```diff
@@ -29,19 +29,186 @@ use crate::{
...
-pub(crate) mod source_type {
+/// The full list of possible digital source types are found below:
+/// <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_digital_source_type>
+/// <https://cv.iptc.org/newscodes/digitalsourcetype>
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+#[serde(untagged)]
+pub enum DigitalSourceType {
+    C2pa(C2paDigitalSourceType),
+    Iptc(IptcDigitalSourceType),
+    Other(String),
```
gpeacock: I'm not sure what benefit we get from breaking this into C2pa vs Iptc. I realize the spec says only Iptc source types are supported for v1 actions, but I'm not planning to write any of those soon. Also, there used to be other now-deprecated iptc source types that used to be supported for v1 actions. I don't think end users will really care if this value is defined by Iptc or c2pa, they just want to set a value that makes sense.

Should we allow for other or call it an error?

ok-nick: I agree, I can flatten it out then, I didn't know if we wanted it restricted to v1/v2.

> Should we allow for other or call it an error?

If the IPTC source type list can update independent of the C2PA spec, then using an older version of the SDK shouldn't error if a new type is added. The most ideal scenario is that we auto-generate this enum using the machine-readable list IPTC provides, although that is probably more effort than it's worth.

sdk/src/assertions/actions.rs
ok-nick: That would be nice, I'll make an issue to track it.

ok-nick: https://github.com/contentauth/c2pa-rs/issues/1268

sdk/src/assertions/actions.rs
ok-nick: Hm, there is a bit of a conflict flattening out the enum. The spec references a "trainedAlgorithmicData" source type, but the actual link: http://c2pa.org/digitalsourcetype/trainedAlgorithmicData shows a "trainedAlgorithmicMedia" source type which conflicts with the IPTC one.

sdk/examples/c2pa.toml
```diff
@@ -90,7 +90,7 @@ action = "c2pa.edited"
 # The source type field is required for the c2pa.created action.
 #
 # For more information, see `c2pa::assertions::actions::source_type`.
-source_type = "http://c2pa.org/digitalsourcetype/empty"
+source_type = "digsrctype:empty"
 # Description for the action
```
gpeacock, line 94: where is this defined? shouldn't it be DigitalSourceType::Empty?

ok-nick: Every digital source type has an "ID (QCode)" so I figured it would make more sense to use that.

Screenshot 2025-07-30 at 9 48 33 AM

gpeacock: Leveraging existing naming is often a good idea, and could help with looking up the meaning, but the **empty** type specifically is not defined as part of that namespace, it is only defined in C2PA. So I'm not sure it makes sense to extend that this way. We may want to sync up with Leonard on this. Maybe IPTC has any plans to adopt the C2PA values, but I suspect they won't.

tmathern: (Also some checks in some places expect the URLs too - would these checks still work through the aliasing or not?)

ok-nick: Yes.

sdk/src/assertions/actions.rs
```diff
@@ -29,19 +29,178 @@ use crate::{
...
+    /// The full list of possible digital source types are found below:
+    /// <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_digital_source_type>
+    /// <https://cv.iptc.org/newscodes/digitalsourcetype>
+    #[non_exhaustive]
+    #[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+    #[cfg_attr(feature = "json_schema", derive(JsonSchema))]
+    pub enum DigitalSourceType {
+        /// Media whose digital content is effectively empty, such as a blank canvas or zero-length video.
+        #[serde(
+            alias = "digsrctype:empty",
```
gpeacock: where would these aliases be used and do we need to have the digscrtype prefix on them? If the purpose is to simplify the JSON definitions for these values, then why not make them as simple as possible?

ok-nick: These IDs are defined in the spec, although I don't feel strongly about keeping them.

PR #1264, opened 2025-07-26
feat: No Send for CAIRead on wasm

cdmurph32:
`unsafe impl Send` will no longer be required on Wasm. Wasm is single threaded.

## Checklist

### DIFF
--- a/sdk/src/asset_io.rs
+++ b/sdk/src/asset_io.rs
@@ -40,9 +40,15 @@ pub struct HashObjectPositions {
     pub htype: HashBlockObjectType, // type of hash block object
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 pub trait CAIRead: Read + Seek + Send {}
+#[cfg(target_arch = "wasm32")]
+pub trait CAIRead: Read + Seek {}
 
+#[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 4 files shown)


sdk/src/asset_io.rs
```diff
@@ -40,9 +40,15 @@ pub struct HashObjectPositions {
     pub htype: HashBlockObjectType, // type of hash block object
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 pub trait CAIRead: Read + Seek + Send {}
```
scouten-adobe, line 44: A topic for another day and another PR: I'd like to see us discontinue the use of `CAI` prefixes for public APIs.

sdk/src/jumbf_io.rs
```diff
@@ -324,6 +324,7 @@ where
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
```
scouten-adobe, line 327: This makes me wonder if we should have a `send-generic` crate similar to the `async-generic` trait that I built a year or two ago. So much code duplication 🙃.

Again, not a blocker for this PR.

tmathern: Where would the crate live? Own repo aka not here, right?

tmathern (approved):
With the caveat that all tests & builds should still work after that!

PR #1269, opened 2025-07-29
feat: Bump MSRV to 1.86.0

scouten-adobe:


mauricefisher64 (approved):

gpeacock (changes requested):
We need to update this pinned test in ci.yml. It ties us to an older rust version.

https://github.com/contentauth/c2pa-rs/actions/runs/16588277907/job/46917859474?pr=1266

- name: Install Rust toolchain
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: nightly-2025-01-24
          # Pinning to specific nightly build for now. More recent versions
          # (not sure of start date) cause linker problems with OpenSSL native
          # code on Mac and Windows, but only when built with minimal dependencies.
          # Problem discovered on 2025-01-27.
          # TO DO: Investigate and file bugs if necessary against rust-lang.

PR #1272, opened 2025-07-30
chore: Update expired CAWG test file and related unit tests

gpeacock:
the C_with_CAWG_data.jpg file was created from the cawg_identity demo app but that app did not add a timestamp, so when the cert expired, the tests started failing. This rebuilds the test file with a new cert and a timestamp. It also updates the related unit tests since the cawg content in that example has changed since then.


scouten-adobe (approved):

gpeacock:
this ended up a duplicate, so deleting it.

PR #1273, opened 2025-07-30
chore: Fix CAWG test (was lacking a timestamp) and update tests

scouten-adobe:


c2pa_c_ffi/src/json_api.rs
```diff
@@ -117,10 +117,11 @@ mod tests {
     fn test_verify_from_file_cawg_identity() {
         let path = test_path("tests/fixtures/C_with_CAWG_data.jpg");
         let result = read_file(&path, None);
+        dbg!(&result);
```
gpeacock, line 120: forgot to remove this dbg!(), but little harm

gpeacock (approved):

PR #1274, opened 2025-07-30
feat: add `ParametersMap` to `Actions`

ok-nick:
Changes `Actions::parameters` to use a `ParametersMap` struct as defined in the spec so we don't need to manually decode a `serde_cbor::Value` each time we use a field.

~~**OPEN QUESTION:** How do we want to handle the `Actions::get_parameter`/`Actions::set_parameter` methods. With this change `Actions::get_parameter("ingredients")`, for example, will return `None`. Whereas the previous behavior it would return a `serde_cbor::Value`.~~

## Checklist


gpeacock (approved):

sdk/src/assertions/actions.rs
```diff
@@ -325,10 +363,9 @@ impl Action {
     /// Sets the value of the `xmpMM:InstanceID` property for the
     /// modified (output) resource.
     #[deprecated(since = "0.37.0", note = "Use `add_ingredient_id()` instead")]
```
gpeacock, line 536: lets just rip this out, it has been deprecated long enough

sdk/src/assertions/actions.rs
```diff
@@ -255,21 +289,25 @@ impl Action {
...
     ///
     /// These vary by the type of action.
-    pub fn parameters(&self) -> Option<&HashMap<String, Value>> {
+    pub fn parameters(&self) -> Option<&ParametersMap> {
         self.parameters.as_ref()
     }
 
     /// Returns an individual action parameter if it exists.
-    pub fn get_parameter(&self, key: &str) -> Option<&Value> {
+    pub fn get_parameter(&self, key: &str) -> Option<Value> {
```
gpeacock: Since this PR breaks the API with the parameters() method anyway, we should also break it here and change the value to serde_json as it should have been. Another option would be to deprecate it and have them just reference the common map with a get? Ideally I'd like to remove any serde_cbor Value references from the API.

gpeacock: We could change it to something more like this from manifest. It can deserialize into any data type without requiring us to expose serde_cbor value.

pub fn get_parameter<T: DeserializeOwned>(&self, label: &str) -> Result<T> {

sdk/src/assertions/actions.rs
```diff
@@ -368,26 +398,51 @@ impl Action {
...
-            }
+        let parameters = match &mut self.parameters {
+            Some(parameters) => parameters,
             None => {
-                let mut p = HashMap::new();
-                p.insert(key.into(), value);
-                p
+                self.parameters = Some(ParametersMap::default());
+                #[allow(clippy::unwrap_used)]
+                self.parameters.as_mut().unwrap()
```
gpeacock: One method I've used is this:
let parameters = self.parameters.take().or_else(ParametersMap::default());

parameters.foo = ....

self.parameters = Some(parameters)

sdk/src/assertions/actions.rs
```diff
@@ -397,17 +452,18 @@ impl Action {
...
+
+        let parameters = match &mut self.parameters {
+            Some(parameters) => parameters,
             None => {
-                let mut p = HashMap::new();
-                p.insert(key.into(), value);
-                p
+                self.parameters = Some(ParametersMap::default());
+                #[allow(clippy::unwrap_used)]
+                self.parameters.as_mut().unwrap()
```
gpeacock: best to avoid the unwraps, though I can see this is safe.

sdk/src/assertions/actions.rs
gpeacock: This would mirror set_parameter

gpeacock (changes requested):

sdk/src/assertions/actions.rs
ok-nick: Actually, it turns out there's a method called `Option::get_or_insert_default`. Pretty cool trick.

sdk/src/assertions/actions.rs
```diff
@@ -135,6 +135,40 @@ impl From<ClaimGeneratorInfo> for SoftwareAgent {
     }
 }
 
+/// Additional parameters of the action.
+#[derive(Deserialize, Serialize, Clone, Debug, Default, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ParametersMap {
```
gpeacock: Can we just name this Parameters?

gpeacock: ActionParameters would be ok as well.

sdk/src/utils/cbor_types.rs
```diff
@@ -162,6 +162,7 @@ pub(crate) fn map_cbor_to_type<T: serde::de::DeserializeOwned>(
 }
 
 // Convert Value to concrete value.  v must be a serde_cbor::Value
+#[allow(unused)]
 pub(crate) fn value_cbor_to_type<T: serde::de::DeserializeOwned>(
```
gpeacock: do we still need this? Delete if no longer referenced

sdk/src/assertions/actions.rs
gpeacock: Use this instead:

let parameters = self.parameters.get_or_insert_with(ParametersMap::default);
parameters.common.insert(key.into(), value);

gpeacock (approved):

PR #1281, opened 2025-08-01
chore: Divide testing into PR validation and release readiness workflows

scouten-adobe:
PR validation (aka "CI") should be MUCH faster now since we're only running on ubuntu-latest.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,6 @@ on:
   push:
     branches:
       - main
-  schedule:
-    - cron: "0 18 * * 1,4,6" # 1800 UTC every Monday, Thursday, Saturday
 
 jobs:
   get-features:
@@ -39,8 +37,8 @@ jobs:
           echo "rust-native-features=$RUST_NATIVE_FEATURES" >> "$GITHUB_OUTPUT"
           echo "openssl-features=$OPENSSL_FEATURES" >> "$GITHUB_OUTPUT"
 
-  tests:
-    name: Unit tests
+  tests-openssl:
+    name: Unit tests (OpenSSL installed)
     needs: get-features
     if: |
       github.event_name != 'pull_request' ||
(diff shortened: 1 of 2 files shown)


dyro:
The linux jobs seem to complete pretty quickly, sans the tests using cross. Should we try switching to aarch64 linux runners before removing tests from our PR checks?

.github/workflows/ci.yml
```diff
@@ -488,45 +333,6 @@ jobs:
...
-        # environment. The goal is to fail PR validation
-        # if the subsequent release would result in a failed
-        # documentation build on docs.rs.
-        run: cargo +nightly-2025-06-05 doc --workspace --features "$FEATURES" --no-deps
-        env:
-          FEATURES: ${{needs.get-features.outputs.openssl-features}}
-          RUSTDOCFLAGS: --cfg docsrs
-          DOCS_RS: 1
-
   cargo-deny:
```
gpeacock, line 368: I don't know how much time this takes, but this is very much a release ready kind of issue

.github/workflows/ci.yml
```diff
@@ -39,8 +37,8 @@ jobs:
           echo "rust-native-features=$RUST_NATIVE_FEATURES" >> "$GITHUB_OUTPUT"
           echo "openssl-features=$OPENSSL_FEATURES" >> "$GITHUB_OUTPUT"
 
-  tests:
-    name: Unit tests
+  tests-openssl:
```
gpeacock, line 46: Do we need to do openssl here? I'll defer to maurice on that, but I don't think it will matter for most things.

.github/workflows/ci.yml
```diff
@@ -253,187 +279,6 @@ jobs:
...
-      # it isn't readily accessible from cross-compilation
-      # environment. (A PR to fix this would be welcomed!)
-
-      - name: Run unit tests (cross build)
-        env:
-          FEATURES: ${{needs.get-features.outputs.openssl-features}}
-        run: |
-          cross test --all-targets --features "$FEATURES" --target ${{ matrix.target }}
-
-  tests-wasm:
```
gpeacock, line 300: I kind of think we may still want the wasm tests here, since it is so often an indicator. But I could be argued out of it.

.github/workflows/ci.yml
scouten-adobe: Are you referring to the license audit task? I specifically left that one in PR validation. It's also a very fast test (~30 seconds on average).

scouten-adobe: Also note that the CI tests will also run on the release PRs, so we get both sets of tests.

.github/workflows/ci.yml
scouten-adobe: Speaking as the c2pa-crypto guy, I think we should retain it because we have fairly different code paths for the two. The run time on Linux is not that much longer (maybe an additional 30 seconds).

.github/workflows/ci.yml
scouten-adobe: You're absolutely right! I've moved Wasm back into the CI run.

scouten-adobe:
@gpeacock as an alternative to your Makefile suggestion, I've added a new label "check-release" which you can apply to any PR to invoke the full release-readiness test suite. (IMHO the Makefile approach is necessarily incomplete because we need to be able to run on multiple platforms and you can't do that on any laptop.)

.github/workflows/ci.yml
gpeacock: yes I was referring to the license audit. It is just if something goes wrong here it will force us to fix it immediately and it isn't really and issue until we need to publish. I don't mind leaving it in since it is fast, but it's often the case this is triggered due to some third party crate dependency and could really be resolved later.

ok-nick: I previously ran into an issue updating the zip crate causing an incompatible license. This is helpful to run on PR validation so we don't merge something that requires us to wait on approval. If we can make this job run only when the Cargo.lock changes, that would be the most ideal.

.github/workflows/release_readiness.yml
```diff
@@ -0,0 +1,325 @@
...
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@v2
+
+      - name: Run unit tests (cross build)
+        env:
+          FEATURES: ${{needs.get-features.outputs.openssl-features}}
+        run: |
+          cross test --all-targets --features "$FEATURES" --target ${{ matrix.target }}
+
+  tests-wasi:
```
ok-nick, line 139: Just as a heads up, I ran into some WASI-specific issues with settings because it runs single threaded tests, also some issues around networking. It might be useful to also run this on validation so we don't have to go in and debug on release.

.github/workflows/ci.yml
mauricefisher64: Ageed

mauricefisher64 (approved):

.github/workflows/release_readiness.yml
scouten-adobe: @ok-nick happy to tune this over time. I wanted to push this out for now because the WASI test suite has been more flaky than others for reasons that aren't tied to the underlying code.

tmathern (approved):

PR #1282, opened 2025-08-02
fix: v2 redactions were reporting validation error if the assertion was redacted

gpeacock:
Fixed the validation logic to validate that the redaction URI refers to a valid claim and the claim has a reference to the assertion, and then validate that the assertion has been removed from the relevant assertion store.

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -2500,7 +2500,7 @@ impl Claim {
                 // 2.d if redacted actions contains a redacted parameter if must be a resolvable reference
                 if action.action() == c2pa_action::REDACTED {
                     if let Some(params) = action.parameters() {
-                        let mut parent_tested = None; // on exists if action actually pointed to an ingredient
+                        let mut parent_tested = None; // only exists if action actually pointed to an ingredient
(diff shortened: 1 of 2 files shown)


mauricefisher64 (approved):

sdk/src/claim.rs
```diff
@@ -2522,41 +2522,63 @@ impl Claim {
...
+                                        if ingredient_claim
+                                            .assertion_hashed_uri_from_label(&redaction_label)
+                                            .is_some()
+                                        {
+                                            // The url reference is valid, now check if it was actually redacted
+                                            parent_tested = Some(false);
+                                            // Now if the assertion is not in the assertion store we are ok.
+                                            // Todo: would a zeroed out assertion show up here? if so we need to do a zero check
+                                            if ingredient_claim
+                                                .get_claim_assertion(&redaction_label, 0)
```
mauricefisher64, line 2542: I left this comment in the other PR. But the test is wether the Claim has an assertion in its assertion list (gathered, created, or 1.x lists).  It should/must be there.  Now what should fail is the second part that you have done here that is no matching assertion in the assertion store.  0s are already handled for you at that level.

PR #1284, opened 2025-08-04
fix: bmff update manifest support

mauricefisher64:
This change add support for C2PA 2.2 changes to how we handle update manifests in for BMFF bases assest.  

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/bmff_io.rs
+++ b/sdk/src/asset_handlers/bmff_io.rs
@@ -1482,7 +1482,7 @@ impl CAIWriter for BmffIO {
             ));
         }
 
-        // if is an ordinary manifest then it should not have an update manifest
+        // if is an ordinary manifest store then it should not have an update manifest store
         if has_manifest && has_update {
             return Err(Error::BadParam(
                 "BMFF save failure, found manifest store with update manifest store".to_string(),
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -1466,10 +1466,10 @@ impl Store {
                     let ingredient_version = ingredient.version();
                     let has_redactions = svi.redactions.iter().any(|r| r.contains(&label));
 
-                    // only allow the extra ingredient trust checks for 1.x ingredients
+                    // allow the extra ingredient trust checks
                     // these checks are to prevent the trust spoofing
-                    let check_ingredient_trust: bool = ingredient_version < 2
-                        && crate::settings::get_settings_value("verify.check_ingredient_trust")?;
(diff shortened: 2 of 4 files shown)


sdk/src/store.rs
```diff
@@ -6209,6 +6213,160 @@ pub mod tests {
         assert!(!um_report.has_any_error());
     }
 
+    #[test]
+    #[cfg(feature = "file_io")]
+    #[cfg(feature = "v1_api")]
+    fn test_update_manifest_v2_bmff() {
```
ok-nick, line 6287: Is this testing v2 bmff or v3 bmff?

sdk/src/store.rs
```diff
@@ -6209,6 +6213,160 @@ pub mod tests {
...
+
+        let mut collapsed_report =
+            StatusTracker::with_error_behavior(ErrorBehavior::StopOnFirstError);
+
+        // read back in store with update manifest
+        let collapsed_store =
+            Store::load_from_asset(op2.as_path(), true, &mut collapsed_report).unwrap();
+
+        let cm = collapsed_store.provenance_claim().unwrap();
+        assert!(!cm.update_manifest());
```
ok-nick, line 6432: With this case, the update manifest still exists in the store just not as the provenance claim?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1390,6 +1455,119 @@ impl CAIWriter for BmffIO {
...
+        // if is an ordinary manifest then it should not have an update manifest
+        if has_manifest && has_update {
+            return Err(Error::BadParam(
+                "BMFF save failure, found manifest store with update manifest store".to_string(),
+            ));
+        }
+
+        // if we already have an "original" manifest store and an "update" manifest store
+        // then we can just apppend to the update store
+        if has_original && has_update && is_update {
```
ok-nick, line 1494: What would happen if you had a file with a split original and update manifest and wanted to append a new standard manifest on top, but there were new data boxes appended after the update manifest by the user (e.g. two videos merged)? The update manifest would no longer be the last box, does this affect the code? Is that update manifest box just deleted and appended with the new standard manifest at the end of the file?

sdk/src/store.rs
mauricefisher64: This actually update manifests for V2 claims for bmff.

tmathern (approved):

tmathern:
Clippy seems unhappy - not sure if this is due to the Rust update or not (I think not, since it ran after the new Rust version happened?).

PR #1296, opened 2025-08-08
feat: multi asset hash

alextrnnn:

## Checklist


sdk/src/claim.rs
```diff
@@ -3065,16 +3065,73 @@ impl Claim {
...
-                                    claim.assertion_uri(&hash_binding_assertion.label()),
-                                    format!("asset hash error, name: {name}, error: {e}"),
-                                    "verify_internal"
-                                )
-                                .validation_status(validation_status::ASSERTION_DATAHASH_MISMATCH)
-                                .failure(
-                                    validation_log,
-                                    Error::HashMismatch(format!("Asset hash failure: {e}")),
-                                )?;
+                                // If standard asset hard binding fails, try multi-asset hash validation.
```
mauricefisher64: Let's move the mult-hash test and error handling into its own function.

sdk/src/claim.rs
```diff
@@ -3065,16 +3065,73 @@ impl Claim {
...
+                                // Only one multi-asset hash assertion is allowed per manifest.
+                                if let Some(assertion) =
```
mauricefisher64: Where is the test for multiple MultiAssetHash assertions.  The spec say there should be only one.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        for part in &self.parts {
+            if let Some(optional) = part.optional {
+                if optional {
+                    continue;
+                };
+            }
+
+            // Retrieve the assertion linked in the multi-asset assertions.
+            if let Some(assertion) = claim.get_assertion_from_link(&part.hash_assertion.url()) {
```
mauricefisher64: Where is the error case for when the if fails and you cannot fetch the assertion?  This is supposed to be a assertion.multiAssetHash.missingPart error.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,300 @@
...
+                                    Some(alg) => alg,
+                                    None => claim.alg(),
+                                };
+                                dh.verify_stream_hash(&mut part_reader, Some(alg))?;
+                            }
+                            l if l.starts_with(BmffHash::LABEL) => {
+                                return Err(Error::NotImplemented(
+                                    "BmffHash not yet implemented for Multi-Asset hashes"
+                                        .to_string(),
+                                ));
```
mauricefisher64: Where is support for the BoxHash case?

sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -69,10 +84,80 @@ fn extract_xmp(seg: &JpegSegment) -> Option<&str> {
...
+
+fn extract_xmp_extensions(seg: &JpegSegment) -> Option<XmpExtension> {
+    let (sig, rest) = seg
+        .contents()
+        .split_at_checked(XMP_EXTENSION_SIGNATURE_BUFFER_SIZE)?;
+
+    if sig.starts_with(XMP_EXTENSION_SIGNATURE.as_bytes()) && rest.len() > 40 {
+        // 32 byte GUID, 4 byte length, 4 byte offset
+        let guid = std::str::from_utf8(&rest[..32]).ok()?.to_string();
+        let length = read_be_u32(&mut &rest[32..36]).ok()?;
```
mauricefisher64: Why did you need a new function to read U32 BE
let length = BigEndian::read_u32(&rest[32..36]);

sdk/src/claim.rs
alextrnnn: Moved some more things into the method such as the logging of the errors that otherwise would've been logged if the multi-asset-hash didn't exist/failed. Let me know if there's a cleaner way, I wasn't too sure how to make it cleaner

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,328 @@
...
+                        }
+                        l if l.starts_with(BoxHash::LABEL) => {
+                            let bh = BoxHash::from_assertion(assertion)?;
+                            let box_hash_processor = asset_handler
+                                .ok_or(Error::UnsupportedType)?
+                                .asset_box_hash_ref()
+                                .ok_or(Error::HashMismatch("Box hash not supported".to_string()))?;
+                            bh.verify_stream_hash(
+                                &mut part_reader,
+                                Some(claim.alg()),
```
mauricefisher64, line 173: Get the alg the same way as for DataHash

sdk/src/claim.rs
mauricefisher64: You also need the handle mulit-part for the BmmHash case

sdk/src/claim.rs
```diff
@@ -3283,7 +3282,101 @@ impl Claim {
...
+                    .failure(
+                        validation_log,
+                        Error::HashMismatch(format!("Asset hash failure: {err_str}")),
+                    )?;
+                }
+            }
+            return multi_hash_result;
+        }
+
+        let description = if let Some(name) = hash_binding_name {
```
mauricefisher64, line 4070: Let's add a comment here that that is is passthrough error handling/reporting from the caller.

sdk/src/assertions/multi_asset_hash.rs
alextrnnn: The `BoxHash` itself doesn't have an alg, but it contains multiple `BoxMaps` which then contain an alg. From the look of it, it looks like the `alg` passed in is used as a fallback if the `BoxMap` (similar to DataHash)

https://github.com/contentauth/c2pa-rs/blob/4b1d80c858dac17f72b88107a5a879f4e09b8ff8/sdk/src/assertions/box_hash.rs#L157-L163

I am not too sure how to perform the same method as the DataHash, or if it was necessary to do that same way for both?

mauricefisher64: Makes sense

sdk/src/assertions/mod.rs
```diff
@@ -59,6 +59,9 @@ pub use assertion_metadata::{
     c2pa_source, Actor, AssertionMetadata, AssetType, DataBox, DataSource, ReviewCode, ReviewRating,
 };
 
+mod multi_asset_hash;
+pub use multi_asset_hash::MultiAssetHash;
```
ok-nick: We also need to export the other public structs.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,360 @@
...
+
+/// A `MultiAssetHash` assertion provides information on hash values for multiple parts of an asset.
+///
+/// This assertion contains a list of parts, each one declaring a location within the asset and
+/// the corresponding hash assertion for that part.
+///
+/// See <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_multi_asset_hash>
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub struct MultiAssetHash {
+    pub parts: Vec<PartHashMap>,
```
ok-nick, line 43: We should add docs for these fields from the spec.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,360 @@
...
+                    if locator.byte_offset != expected_offset {
+                        return Err(Error::C2PAValidation(
+                            ASSERTION_MULTI_ASSET_HASH_MALFORMED.to_string(),
+                        ));
+                    }
+                    // Keep track of the size of optional parts.
+                    if part.optional.unwrap_or(false) {
+                        optional_sizes += locator.length;
+                    }
+                    expected_offset += locator.length;
```
ok-nick, line 75: We should also verify the byte offsets so that each part's length + byte offset doesn't overlap and also doesn't extend past the end of the file.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    /// Verifies each part of the multi-asset hash through comparing computed hashes.
+    /// Validates part locations, reads the specified byte ranges, and verifies against referenced hash assertions.
+    fn verify_stream_hash(
+        &self,
+        mut reader: &mut dyn CAIRead,
+        claim: &Claim,
+        asset_handler: Option<&dyn AssetIO>,
+    ) -> Result<()> {
+        let length = stream_len(reader)?;
```
ok-nick, line 129: We may want to look into this for other parts of the SDK, there is mixed usage of this and `.stream_position()`.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,360 @@
...
+        self.verify_self(length)?;
+
+        for part in &self.parts {
+            if let Some(optional) = part.optional {
+                if optional {
+                    continue;
+                };
+            }
```
ok-nick:
```suggestion
            if part.optional.unwrap_or(false) {
                continue;
            }
```

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,360 @@
...
+                LocatorMap::ByteRangeLocator(locator) => {
+                    let offset = locator.byte_offset;
+                    let length = locator.length;
+
+                    // Read only the specified parts within the larger stream.
+                    reader.seek(std::io::SeekFrom::Start(offset))?;
+                    let buf = reader.read_to_vec(length).map_err(|_| {
+                        Error::C2PAValidation(ASSERTION_MULTI_ASSET_HASH_MISSING_PART.to_string())
+                    })?;
+                    let mut part_reader = Cursor::new(buf);
```
ok-nick: Might be able to use `Read::take` here instead of allocating a vec so we keep it purely streams.

sdk/src/assertions/multi_asset_hash.rs
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq)]
+pub struct PartHashMap {
+    pub location: LocatorMap,
+    #[serde(rename = "hashAssertion")]
```
ok-nick: Not a blocker, but I think it makes more sense to have `#[serde(rename_all = "camelCase")]` on the whole struct. Same goes for the other places this pattern is used.

sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -68,11 +82,72 @@ fn extract_xmp(seg: &JpegSegment) -> Option<&str> {
     }
 }
 
-// Extract XMP from bytes.
+fn extract_xmp_extensions(seg: &JpegSegment) -> Option<XmpExtension> {
```
ok-nick, line 87: Nice work here.

ok-nick (commented):
Great work! Just a few thoughts that we will have to figure out how to address.

alextrnnn:
Sorry for the delay and thanks for the feedback @ok-nick ! I'll try to address these over the weekend

sdk/src/assertions/multi_asset_hash.rs
alextrnnn: Thanks for the catch! I added some tests to this and changed the logic slightly

scouten-adobe:
Closing for lack of activity. @alextrnn feel free to re-open if you think we'll revisit this.

PR #1298, opened 2025-08-11
feat: Updates API and adds new examples

gpeacock:
Builder::open() renamed to Builder::Edit()
Breaking change: Builder.add_ingredient() now returns a Result that must be checked.
It also allows passing anything that can be serialized into an Ingredient.
Updated Builder.add_ingredient_with_stream to accept a serializable Ingredient parameter.
Updates example and test code to use the new methods.


sdk/src/builder.rs
```diff
@@ -190,71 +188,64 @@ impl DigitalSourceType {
...
 #[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
 #[cfg_attr(feature = "json_schema", derive(JsonSchema))]
-enum BuilderFlow {
-    // Represents a builder for creating new assets.
+enum BuilderIntent {
+    /// This is a new digital creation, a DigitalSourceType is required.
+    ///
+    /// The Manifest must not have have a parent ingredient.
+    /// A `c2pa.created` action will be added if not provided.
+    #[serde(rename = "create")]
```
ok-nick, line 179: You can also do `#[serde(rename_all = "lowercase")]` on the struct definition.

sdk/Cargo.toml
```diff
@@ -190,6 +190,7 @@ x509-certificate = "0.24.0"
 x509-parser = "0.16.0"
 zeroize = { version = "1.8", features = ["zeroize_derive"] }
 zip = { version = "3.0.0", default-features = false }
+clap = { version = "4.0", features = ["derive"] }
```
ok-nick, line 193: Did you mean to add this to the SDK dependencies?

sdk/examples/client/client.rs
```diff
@@ -139,27 +128,19 @@ pub fn main() -> Result<()> {
...
-    builder.definition.claim_version = Some(2);
-    let mut generator = ClaimGeneratorInfo::new(GENERATOR);
+    // create a new Manifest - the source file will be the parent
+    let mut builder = Builder::edit();
+    let mut generator = ClaimGeneratorInfo::new("c2pa-rs client example");
     generator.set_version("0.1");
     #[allow(deprecated)]
     builder
         .set_claim_generator_info(generator)
-        .add_ingredient(parent)
```
ok-nick, line 150: Do we add the parent ingredient automatically now?

sdk/examples/edit.rs
```diff
@@ -0,0 +1,64 @@
...
+    arg_required_else_help = true
+)]
+struct Args {
+    /// Source file path
+    source: String,
+    /// Destination file path (optional)
+    dest: Option<String>,
+    /// Settings file path
+    #[arg(long)]
+    settings: Option<String>,
```
ok-nick, line 29: You may be able to use a `PathBuf` here.

sdk/examples/edit.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        builder.add_action(Action::new("c2pa.published"))?;
+
+        builder.add_action(json!({
+            "action": "c2pa.edited",
+            "description": "my special edits",
+            "parameters": {
+                "name": "any value"
+            },
+        }))?;
```
ok-nick, line 54: Do we want the examples to show passing in json or building the `Actions` struct directly? I'm leaning a bit more towards the latter to keep everything more Rusty.

sdk/src/ingredient.rs
```diff
@@ -1459,6 +1459,38 @@ impl std::fmt::Display for Ingredient {
...
+}
+
+impl TryFrom<&String> for Ingredient {
+    type Error = crate::Error;
+
+    fn try_from(json: &String) -> std::result::Result<Self, Self::Error> {
+        Self::from_json(json)
+    }
+}
+
```
ok-nick, line 1493: You may be able to do `impl TryFrom<AsRef<str>>` rather than 3 implementations. I'm not too sure how I feel about this either, I think it makes more sense to have an explicit function to convert from json to an `Ingredient`.

sdk/src/builder.rs
```diff
@@ -593,13 +611,48 @@ impl Builder {
...
+    ///
+    /// // Add from JSON string
+    /// builder.add_ingredient(r#"{"title": "Another", "relationship": "componentOf"}"#)?;
+    /// # Ok::<(), c2pa::Error>(())
+    /// ```
+    pub fn add_ingredient<I>(&mut self, ingredient: I) -> Result<&mut Self>
     where
-        I: Into<Ingredient>,
+        I: TryInto<Ingredient>,
+        I::Error: std::fmt::Display,
```
ok-nick, line 649: I'm leaning a bit more towards having `Builder.add_ingredient(Ingredient::from_json(json))` or something of the sort. I think this makes the most sense for Rust, but in our C bindings it would be fine if we had something like `c2pa_builder_add_ingredient(json)`

sdk/examples/edit.rs
gpeacock: This is a little of both, to show that it works both ways.

PR #1304, opened 2025-08-13
fix: Mark the CreativeWork assertion as deprecated

scouten-adobe:
CreativeWork assertion is no longer part of the C2PA Technical Specification and its use should be avoided when creating new manifests.

### DIFF
--- a/sdk/src/assertions/creative_work.rs
+++ b/sdk/src/assertions/creative_work.rs
@@ -36,7 +36,7 @@ const CW_AUTHOR: &str = "author";
 /// new manifests.
 #[deprecated(
     since = "0.59.0",
-    note = "The CreativeWork assertion is no longer part of the C2PA Technical Specification. Please use the CAWG identity assertion instead (https://opensource.contentauthenticity.org/docs/manifest/cawg-id)."
+    note = "The CreativeWork assertion is no longer part of the C2PA Technical Specification. Please use the CAWG identity and/or metadata assertion instead (https://opensource.contentauthenticity.org/docs/manifest/cawg-id)."
 )]
 #[derive(Serialize, Deserialize, Debug)]
 pub struct CreativeWork(SchemaDotOrg);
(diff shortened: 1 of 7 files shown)


sdk/src/assertions/creative_work.rs
```diff
@@ -25,6 +27,17 @@ use crate::{
...
+/// Assertion that implements various schema.org-based assertions,
+/// including the now-deprecated CreativeWork assertion.
+///
+/// This structure is here to allow parsing of existing C2PA 1.x manifests
+/// this assertion. The CreativeWork assertion is no longer part of the
+/// C2PA Technical Specification and it should not be used when creating
+/// new manifests.
+#[deprecated(
+    since = "0.59.0",
+    note = "The CreativeWork assertion is no longer part of the C2PA Technical Specification. Please use the CAWG identity assertion instead (https://opensource.contentauthenticity.org/docs/manifest/cawg-id)."
```
gpeacock: This should refer to the cawg.metadata assertion as well.

sdk/src/assertions/creative_work.rs
```diff
@@ -25,6 +27,17 @@ use crate::{
...
+/// Assertion that implements various schema.org-based assertions,
+/// including the now-deprecated CreativeWork assertion.
+///
+/// This structure is here to allow parsing of existing C2PA 1.x manifests
+/// this assertion. The CreativeWork assertion is no longer part of the
+/// C2PA Technical Specification and it should not be used when creating
+/// new manifests.
+#[deprecated(
+    since = "0.59.0",
+    note = "The CreativeWork assertion is no longer part of the C2PA Technical Specification. Please use the CAWG identity and/or metadata assertion instead (https://opensource.contentauthenticity.org/docs/manifest/cawg-id)."
```
gpeacock, line 39: should clarify with CAWG metata here, since people could confuse the two.

gpeacock: Please use the CAWG identity and/or CAWG metadata assertion instead...

PR #1306, opened 2025-08-13
feat: Add new soft binding fields to the ingredient assertion structure

gpeacock:
Todo, I haven't ported these over to the higher level ingredient yet, but we need them at this level regardless.


sdk/src/assertions/ingredient.rs
```diff
@@ -66,6 +66,8 @@ pub struct Ingredient {
     pub active_manifest: Option<HashedUri>,
     pub claim_signature: Option<HashedUri>,
 
+    pub soft_bindings_matched: Option<bool>,
+    pub soft_bindings_algorithms_matched: Option<Vec<String>>,
```
ok-nick: I'm not sure how we serialize/deserialize this assertion but as an FYI, the spec names this field `softBindingAlgorithmsMatched` instead of `softBindingsAlgorithmsMatched`.

gpeacock: Good catch, thanks, I fixed that and added serialize/deserialize logic.

PR #1309, opened 2025-08-14
fix: Fix broken ingredient recursion

mauricefisher64:
Updates to tweaks to hash binding logging
Fix accidentally removed unknown ingredient status
Better update manifest handling

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -69,8 +69,8 @@ use crate::{
         self,
         boxes::*,
         labels::{
-            manifest_label_from_uri, manifest_label_to_parts, to_absolute_uri, to_assertion_uri,
-            to_manifest_uri, ASSERTIONS, CREDENTIALS, DATABOXES, SIGNATURE,
+            manifest_label_from_uri, manifest_label_to_parts, to_assertion_uri, to_manifest_uri,
+            ASSERTIONS, CREDENTIALS, DATABOXES, SIGNATURE,
         },
     },
     jumbf_io::{
@@ -1851,8 +1851,8 @@ impl Store {
             }
         }
 
-        // get the timestamp assertions
         for found_claim in svi.manifest_map.values() {
+            // get the timestamp assertions
             let timestamp_assertions = found_claim.timestamp_assertions();
             for ta in timestamp_assertions {
                 let timestamp_assertion =


sdk/src/store.rs
```diff
@@ -1480,6 +1481,11 @@ impl Store {
     ) -> Result<()> {
         // walk the ingredients
         for i in claim.ingredient_assertions() {
+            // allow for zero out ingredient assertions
+            if is_zero(i.assertion().data()) {
```
gpeacock, line 1485: I thought you said we would not see these as assertions. Is there a case where we do?

sdk/src/store.rs
```diff
@@ -1500,6 +1506,25 @@ impl Store {
 
             // is this an ingredient
             if let Some(c2pa_manifest) = ingredient_assertion.c2pa_manifest() {
+                // if this is a v3 ingredient then it must have validation report indicating it was validated
+                if let Some(ingredient_version) = ingredient_assertion.version() {
+                    if ingredient_version >= 3 && ingredient_assertion.validation_results.is_none()
+                    {
+                        log_item!(
+                            i.label().clone(),
```
gpeacock: can we use the full jumbf uri here?

sdk/src/store.rs
```diff
@@ -1755,76 +1794,60 @@ impl Store {
...
-                is_bmff = self
-                    .get_claim(&label)
-                    .ok_or(Error::ClaimMissingHardBinding)?
-                    .hash_assertions()
-                    .iter()
-                    .any(|a| a.label_raw().starts_with(labels::BMFF_HASH));
-                label
+        svi.binding_claim = self.get_hash_binding_manifest(claim).ok_or_else(|| {
+            log_item!(
+                claim.label().to_owned(),
```
gpeacock: this should be a jumbf uri, maybe we need a helper function for this.

gpeacock (approved):
looks good, there's a few places where we should be logging uris instead of labels.

sdk/src/store.rs
mauricefisher64: We don't but it is a new required test, so I added it.

sdk/src/store.rs
mauricefisher64: Sure, will add now

PR #1311, opened 2025-08-14
feat: Add benchmarking

alextrnnn:
Adds benchmarking into CI using Codspeed. Specifically benchmarks instances of signing a 100kb file and reading a 100kb signed file with a simple manifest (as defined by tests/fixtures/simple_manifest.json). 

NOTE: This will only run when the 'benchmark' label is added.

## Checklist


.github/workflows/ci.yml
```diff
@@ -311,6 +311,142 @@ jobs:
         env:
           WASM_BINDGEN_TEST_TIMEOUT: 60
 
+  tests-wasi:
```
scouten-adobe, line 350: I'd prefer that we have the WASI and minimal deps tests in a separate PR. Mind removing those here?

.github/workflows/ci.yml
```diff
@@ -311,6 +311,142 @@ jobs:
...
+          cargo +nightly-2025-01-24 test -Z direct-minimal-versions --all-targets --features "$FEATURES"
+
+  benchmarks:
+    name: Run benchmarks on signing and reading
+    if: |
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.author_association == 'COLLABORATOR' ||
+      github.event.pull_request.author_association == 'MEMBER' ||
+      github.event.pull_request.user.login == 'dependabot[bot]' ||
+      contains(github.event.pull_request.labels.*.name, 'benchmark')
```
scouten-adobe: To discuss with the group: If these run sufficiently quickly, should we run them on _all_ PRs? I'm inclined to say yes.

Based on my experience with Codspeed (for example here: https://github.com/scouten/asciidoc-parser/pull/288#issuecomment-3123719670), I think they are reliable and reasonably fast.

scouten-adobe (changes requested):
Other than the extra tests, I think this is great!

.github/workflows/ci.yml
alextrnnn: Ah whoops, I must've included it incorrectly when trying to resolve a merge conflict. Thanks for the catch!

.github/workflows/ci.yml
gpeacock: I'm for that, if they don't delay things. But we need to think about how we update any triggers when a delta is decided to be ok.

sdk/benches/fixtures/100kb-signed.gif
ok-nick, line 1: We should think about how to store all these assets outside of the repo.

ok-nick (approved):
Before we merge we need to make sure to add `CODSPEED_TOKEN` to the env.

sdk/benches/fixtures/100kb-signed.gif
scouten-adobe: Agreed.

.github/workflows/ci.yml
ok-nick: I agree that it should run on all PRs. I also agree with Gavin and to add to that, we should introduce triggers for changes to `*.rs` or `Cargo.*` in the `sdk/` folder.

PR #1325, opened 2025-08-18
feat: Fetch ocsp wasm

cdmurph32:
* Make `ocsp_fetch` available as async.
* Add ocsp_fetch for Wasm. as with fetching the manifest, the browser through wasm-bindgen only supports async HTTP GET. WASI supports synchronous and asynchronous fetching.

* Remove Actix as a dependency. Actix for our purposes is redundant with tokio, which is used for most async tests. The `c2pa_test_async` macro can be used for these tests. I can break this part up into another PR if that's easier.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.63.0](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.62.0...c2pa-v0.63.0)
+_10 September 2025_
+
+### Added
+
+* Add `cawg_trust.verify_trust_list` setting ([#1356](https://github.com/contentauth/c2pa-rs/pull/1356))
+* Remove the v1_api feature and all associated code ([#1387](https://github.com/contentauth/c2pa-rs/pull/1387))
+
+### Fixed
+
+* Fix/trust list check ([#1398](https://github.com/contentauth/c2pa-rs/pull/1398))
+* Remove unmaintained dependencies conv and custom_derive ([#1393](https://github.com/contentauth/c2pa-rs/pull/1393))
+
 ## [0.62.0](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.61.0...c2pa-v0.62.0)
 _06 September 2025_
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,12 +46,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -300,7 +294,7 @@ dependencies = [
  "futures-lite",
  "parking",
  "polling",
- "rustix 1.0.8",
+ "rustix 1.1.2",
  "slab",
  "windows-sys 0.60.2",
 ]
@@ -340,7 +334,7 @@ dependencies = [
  "cfg-if",
  "event-listener 5.4.1",
  "futures-lite",
- "rustix 1.0.8",
+ "rustix 1.1.2",
 ]
 
 [[package]]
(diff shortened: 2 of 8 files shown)


sdk/src/crypto/cose/ocsp.rs
```diff
@@ -276,54 +277,57 @@ fn check_stapled_ocsp_response(
...
-        };
+    let ocsp_der: Vec<u8> = if _sync {
+        // No sync version of fetch_ocsp_response for wasm-bindgen
+        #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
+        return Ok(OcspResponse::default());
+        #[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
+        match crate::crypto::ocsp::fetch_ocsp_response(&certs) {
+            Some(der) => der,
+            None => return Ok(OcspResponse::default()),
+        }
```
ok-nick: I think it would be more clear if we had a blocking function that returns `OcspResponse::default()` instead of the platform specific behavior here.

sdk/src/crypto/ocsp/fetch.rs
```diff
@@ -29,7 +33,99 @@ use crate::crypto::base64;
...
+
+    Some(OcspRequestData { request_str, url })
+}
+
+// Common function to process OCSP responders
+#[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
+fn process_ocsp_responders<F>(certs: &[Vec<u8>], mut fetch_fn: F) -> Option<Vec<u8>>
+where
+    F: FnMut(&OcspRequestData) -> Option<Vec<u8>>,
+{
```
ok-nick: Maybe this function should just return a list of ocsp requests? It doesn't seem right to have it only for sync code yet inline the same code for async. It also doesn't seem right to have this function support both async/sync.

sdk/src/store.rs
```diff
@@ -3988,51 +4031,26 @@ impl Store {
 
     /// Handles remote manifests asynchronously when fetch_remote_manifests feature is enabled
     /// Required because wasm-bindgen cannot use async functions in a sync context.
-    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
+    #[cfg(target_arch = "wasm32")]
     async fn handle_remote_manifest_async(ext_ref: &str) -> Result<Vec<u8>> {
```
ok-nick: Isn't this already supported for non-wasm32 as well?

sdk/src/store.rs
```diff
@@ -4055,10 +4072,19 @@ impl Store {
                     crate::utils::xmp_inmemory_utils::XmpInfo::from_source(stream, asset_type)
                         .provenance
                 {
-                    Ok((
-                        Store::handle_remote_manifest_async(&ext_ref).await?,
-                        Some(ext_ref),
-                    ))
+                    #[cfg(not(target_arch = "wasm32"))]
+                    return Ok((Store::handle_remote_manifest(&ext_ref)?, Some(ext_ref)));
+                    #[cfg(target_arch = "wasm32")]
```
ok-nick: Here too, I think we may be able to naturally support async remote manifest calls for non-wasm32 targets.

sdk/src/store.rs
cdmurph32: I'm combining both wasm-bindgen and WASI into the same function.

sdk/src/crypto/ocsp/fetch.rs
cdmurph32: Good idea. Done.

sdk/src/crypto/cose/ocsp.rs
cdmurph32: Good idea. Done.

sdk/src/store.rs
cdmurph32: Yes. We can discuss and raise a second PR if we want a non-wasm async version.

sdk/src/store.rs
cdmurph32: Yes. We can discuss and raise a second PR if we want a non-wasm async version.

sdk/src/builder.rs
```diff
@@ -2939,6 +2942,53 @@ mod tests {
         assert_eq!(parent.assertions().len(), 1);
     }
 
+    #[c2pa_test_async]
+    async fn test_redaction_async() {
```
ok-nick: Did you mean to include this test case in this PR?

ok-nick (approved):
Nice work.

sdk/src/builder.rs
cdmurph32: Yes. There were no tests for the async ocsp fetch codepath.

PR #1329, opened 2025-08-24
chore: replace actix with tokio for tests

cdmurph32:
Remove Actix as a dependency.
Actix is redundant with tokio for testing purposes. 

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -1608,10 +1608,7 @@ mod tests {
         crypto::raw_signature::SigningAlg,
         hash_stream_by_alg,
         settings::Settings,
-        utils::{
-            test::write_jpeg_placeholder_stream,
-            test_signer::{async_test_signer, test_signer},
-        },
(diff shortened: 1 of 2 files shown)


sdk/src/crypto/raw_signature/tests/async_signers.rs
```diff
@@ -13,19 +13,15 @@
 
 #![allow(unused)] // test fns appear unused on WASM
 
+use c2pa_macros::c2pa_test_async;
 #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
 use wasm_bindgen_test::wasm_bindgen_test;
```
ok-nick, line 18: In the future, I think in our macro we can use the absolute `wasm_bindgen_test` path rather than needing to import this.

ok-nick (approved):
Looks good, great change.

sdk/src/crypto/raw_signature/tests/async_signers.rs
cdmurph32: I'll give that a try in my next PR

PR #1332, opened 2025-08-25
fix: ingredient thumbnail generation respect settings

### LINKED ISSUE #1330: Is there a way to skip adding thumbnails for ingredients?
It currently looks like when using the setting `load_settings('{"builder": { "thumbnail": {"enabled": false}}}')`, the Builder will not add thumbnails to the generated/added manifest. But when adding ingredients to a builder with that same setting, the ingredients still get thumbnails (despite that setting asking it not to). I would expect that setting to deactivate all thumbnails? 

So, the question: Is there a way to also skip adding ingredients on thumbnails?

I am observing that behavior in this Python test [here](https://github.com/contentauth/c2pa-python/blob/d6e3c524a41df71360fe138f5e2536d56928f9f6/tests/test_unit_tests.py#L858-L898), when using an ingredient that does not contain any content credentials (in case this would change something, but I wouldn't think so!). The resulting active manifest overall doesn't have a thumbnail as expected, but the ingredient still gets one (print the manifest to observe the ingredient in the ingredients array having a thumbnail).

Additional notes:
- Looking at the code, I think maybe around [these lines](https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/ingredient.rs#L795-L800) we should also check the setting like [here](https://github.com/contentauth/c2pa-rs/blob/903094b1b6ea2e884ed3ac37525c8d72cb0ce262/sdk/src/builder.rs#L1223-L1225). Or maybe it is just not propagated down enough?

ok-nick:
If `builder.thumbnail.enabled` is false then a thumbnail will not be automatically generated for ingredients. Should we introduce a separate "ingredient thumbnail" setting or use the builder?

* Closes #1330

## Checklist

### DIFF
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -796,6 +796,14 @@ impl Ingredient {
         if ingredient.thumbnail.is_none() {
             if let Some((format, image)) = options.thumbnail(path) {
                 ingredient.set_thumbnail(format, image)?;
+            } else {
+                #[cfg(feature = "add_thumbnails")]
+                if let Some(format) = crate::format_from_path(path) {
+                    ingredient.maybe_add_thumbnail(
+                        &format,
+                        &mut std::io::BufReader::new(std::fs::File::open(path)?),
+                    )?;
+                }
             }
         }
         Ok(ingredient)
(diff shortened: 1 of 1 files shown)


tmathern (approved):
Approving the code because it fixes the issues/questions, but the question if it should be another setting is totally legitimate. 

I think it should, because I see it coming that people will want thumbnails for the manifest epr se, but not the ingredients on that manifest.

PR #1339, opened 2025-08-26
fix: Reset settings after test_stream_thumbnail

cdmurph32:

Wasm is single threaded and runs tests in sequence.

* Add WASI back into CI.

## Checklist


.github/workflows/release_readiness.yml
```diff
@@ -151,9 +151,7 @@ jobs:
       - name: Install Rust toolchain
```
ok-nick: Do we need the WASI test in release_readiness if we have it in regular CI?

cdmurph32: I guess not. I must have misread the files.

tmathern (approved):
Approving so all test jobs pass again (in combo with https://github.com/contentauth/c2pa-rs/pull/1343)

PR #1340, opened 2025-08-26
style: fix docs and dead code lint

ok-nick:
Fixes lints and dead code references on certain targets.

## Checklist


sdk/src/utils/thumbnail.rs
```diff
@@ -99,6 +99,7 @@ impl fmt::Display for ThumbnailFormat {
 /// * `builder.thumbnail.ignore_errors`
 ///
 /// Read [make_thumbnail_from_stream] for more information.
+#[allow(dead_code)]
```
gpeacock: The better fix here it to tag this a v1_api so that it will go away when we strip that out. I have fixes for both of these in an upcoming pr, so you can either update this or I'll update it when I merge my pr.

PR #1341, opened 2025-08-26
feat(sdk): falls back to fs for large intermediate streams

dyro:
Adds  the `stream_with_fs_fallback` function to io_utils which allows clients to create a `Read`, `Write`, and `Seek`able in-memory stream that will fall back to the file system if the stream becomes too large. This function is used during the signing process to write intermediate streams. The motivation for this change is to support signing large files. 

if the output stream, provided by the client, is a `Cursor<Vec<u8>>` they will still be allocating the entire asset in-memory. _When signing very large files, we should direct clients to use a stream that's writing to the filesystem._

## Checklist


sdk/src/utils/io_utils.rs
```diff
@@ -118,6 +119,26 @@ pub(crate) fn stream_len<R: Read + Seek + ?Sized>(reader: &mut R) -> Result<u64>
...
+/// # Errors
+/// - Returns an error if the threshold value from settings is not valid.
+pub(crate) fn stream_with_fs_fallback(
+    threshold_override: Option<usize>,
+) -> Result<SpooledTempFile> {
+    let threshold = threshold_override.unwrap_or(get_settings_value::<usize>(
+        "core.backing_store_memory_threshold_in_mb",
+    )?);
+
+    Ok(SpooledTempFile::new(threshold))
```
mauricefisher64: Nice that a crate already has this feature!

sdk/src/settings/mod.rs
```diff
@@ -189,6 +190,7 @@ impl Default for Core {
             merkle_tree_max_proofs: 5,
             compress_manifests: true,
             max_memory_usage: None,
+            backing_store_memory_threshold_in_mb: 128,
```
mauricefisher64: I wonder if this should be upped to say 512 so that most images and videos stay in memory.

dyro: Makes sense to me. I have to modify the PR to support wasm / wasi. I'll make this change as well.

PR #1348, opened 2025-08-27
fix: gif global color table size

### LINKED ISSUE #1342: GIF - LSD reading incorrect packed bytes to determine size of the global color table
It appears that the [GIF LSD Packed Byte Handling Code](https://github.com/contentauth/c2pa-rs/blob/e6bc2be3c1ede0fa43c3aca4f68ae7c57d6a5fc9/sdk/src/asset_handlers/gif_io.rs#L841) is using the `Color Resolution` bits rather than the `Size of Global Color Table` bits to determine the size of the Global Color Table.

The fix is, on the referred line above, to just not byte shift as the size is the 3 least significant bits. ( `(packed >> 4) & 0b111` -> `packed & 0b111`).

Spec: https://www.w3.org/Graphics/GIF/spec-gif89a.txt

Section 18.c:
```
<Packed Fields>  = Global Color Table Flag       1 Bit
                   Color Resolution              3 Bits
                   Sort Flag                     1 Bit
                   Size of Global Color Table    3 Bits
```

Section 18.c.iv:
> iv) Color Resolution - Number of bits per primary color available to the original image, minus 1. This value represents the size of the entire palette from which the colors in the graphic were selected, not the number of colors actually used in the graphic. For example, if the value in this field is 3, then the palette of the original image had 4 bits per primary color available to create the image.  This value should be set to indicate the richness of the original palette, even if not every color from the whole palette is available on the source machine.

Section 18.c.vi
> vi) Size of Global Color Table - If the Global Color Table Flag is set to 1, the value in this field is used to calculate the number of bytes contained in the Global Color Table. To determine that actual size of the color table, raise 2 to [the value of the field + 1].  Even if there is no Global Color Table specified, set this field according to the above formula so that decoders can choose the best graphics mode to display the stream in.  (This field is made up of the 3 least significant bits of the byte.)

ok-nick:
Fix global color table size to use the proper variable.

* Closes #1342

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/gif_io.rs
+++ b/sdk/src/asset_handlers/gif_io.rs
@@ -387,8 +387,11 @@ impl GifIO {
 
         Header::from_stream(stream)?;
         let logical_screen_descriptor = LogicalScreenDescriptor::from_stream(stream)?;
-        if logical_screen_descriptor.color_table_flag {
-            GlobalColorTable::from_stream(stream, logical_screen_descriptor.color_resolution)?;
+        if logical_screen_descriptor.global_color_table_flag {
+            GlobalColorTable::from_stream(
+                stream,
+                logical_screen_descriptor.global_color_table_size,
+            )?;
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


mauricefisher64 (approved):
I don't know GIF so I have to assume this correct

PR #1354, opened 2025-08-28
fix: Fix label handling for custom metadata

mauricefisher64:
Labels serialization is handled by the base structures, fix the code to return the correct label so that the correct value is presented.

## Checklist

### DIFF
--- a/sdk/src/assertions/metadata.rs
+++ b/sdk/src/assertions/metadata.rs
@@ -81,7 +81,14 @@ impl Metadata {
             for (namespace, uri) in &self.context {
                 if let Some(expected_uri) = ALLOWED_SCHEMAS.get(namespace.as_str()) {
                     if uri != expected_uri {
-                        return false;
+                        // check the backcompat list
+                        if let Some(bcl) = BACKCOMPAT_LIST.get(namespace.as_str()) {
+                            if !bcl.iter().any(|v| v == uri) {
+                                return false;
+                            }
+                        } else {
+                            return false;
+                        }
                     }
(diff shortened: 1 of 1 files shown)


sdk/src/assertions/metadata.rs
```diff
@@ -83,29 +93,53 @@ impl Metadata {
...
 
     fn to_assertion(&self) -> Result<Assertion, Error> {
         Self::to_json_assertion(self)
     }
 
     fn from_assertion(assertion: &Assertion) -> Result<Self, Error> {
-        let mut metadata = Self::from_json_assertion(assertion)?;
-        metadata.label = assertion.label().to_owned();
-        Ok(metadata)
+        match assertion.decode_data() {
```
gpeacock: why don't you use Self::from_json_asserrtion() here and then assign metadata based on this: 
metadata.abel = (assertion.label() != labels::METADATA)
    .then(|| assertion.label().to_owned());

gpeacock (changes requested):
I had a feeling something was missing here, I just wasn't sure what.
I've got a suggestion for how to simplify the change.

sdk/src/assertions/metadata.rs
mauricefisher64: I can do that.

mauricefisher64: fixed

ok-nick (approved):

ok-nick:
Approved although there seems to be an issue with this change on WASI.

PR #1355, opened 2025-08-28
feat: sync/async HTTP resolvers API

ok-nick:
Introduces new sync and async http resolver traits to abstract networking from the API and naturally support multiple networking libraries (e.g. `ureq`, `reqwest`, `reqwest_blocking`, `wasi`, `wstd`) and WASM/WASI everywhere. This change also adds full async/await support for fetching remote manifests, ocsp, and timestamps for **every supported platform**.

This implementation can reduce a significant number of dependencies by narrowing down to a single networking library or optionally no networking library at all. It also bubbles up the http resolvers to the builder/reader (more work pending...) so they can be custom implemented or  optionally reused by the caller which can result in better performance.  A final result of this work is that it reduces a lot of code duplication between desktop and WASM/WASI.

### Dependency Count
- with `ureq` + `reqwest`: 404
- with `reqwest`: 399
- with `ureq`: 365
- with neither: 345

## New API
Introduces a `SyncHttpResolver` trait and `SyncGenericResolver` struct that automatically constructs an ideal sync http resolver based on features enabled and platform. The features are as listed:
* `http_reqwest_blocking`
* `http_ureq`
* `http_wasi`

Additionally introduces a `AsyncHttpResolver` trait and `AsyncGenericResolver` struct that automatically constructs an ideal async resolver based on features enabled. The features are as listed:
* `http_reqwest`
* `http_wstd`

### Traits
```rust
/// A resolver for sync (blocking) HTTP requests.
pub trait SyncHttpResolver {
    /// Resolve a [`http::Request`] into a [`http::Response`] with a streaming body.
    fn http_resolve(
        &self,
        request: Request<Vec<u8>>,
}

/// A resolver for non-blocking (async) HTTP requests.
pub trait AsyncHttpResolver {
    /// Resolve a [`http::Request`] into a [`http::Response`] with a streaming body.
    async fn http_resolve_async(
        &self,
        request: Request<Vec<u8>>,
}
```

## Future Work
#1364: Unfortunately our test cases create their own tokio runtime, but call into the sync c2pa APIs. Tokio does not allow multiple async runtimes to be created within each other. As a result, those tests cannot use `reqwest_blocking` until they are fixed and we must enable both `ureq` (sync) and `reqwest` (async) by default.

#1366: We should also expose a Builder/Reader API for the caller to specify their own sync/async http resolvers.

#1367: The final thing we should look into is matching `openssl` and `rustls` with our networking library features. I believe by default we always use `rustls` for them even if we have the `openssl` feature enabled.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -181,7 +181,7 @@ sha1 = "0.10.6"
 spki = { version = "0.7.3", optional = true }
 static-iref = "3.0"
 tempfile = "=3.15.0"
-thiserror = "2.0.8"
+thiserror = "2.0.17"
 url = "2.5.3"
 uuid = { version = "1.18.0", features = ["serde", "v4"] }
 web-time = "1.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -532,7 +532,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
- "thiserror 1.0.69",
+ "thiserror 2.0.17",
  "tokio",
 ]
 
--- a/c2pa_c_ffi/Cargo.toml
+++ b/c2pa_c_ffi/Cargo.toml
@@ -33,7 +33,7 @@ c2pa = { path = "../sdk", version = "0.69.0", default-features = false, features
 scopeguard = "1.2.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-thiserror = "1.0.64"
+thiserror = "2.0.17"
 tokio = { version = "1.36", features = ["rt-multi-thread", "rt"] }
 
 [dev-dependencies]


Makefile
```diff
@@ -26,17 +26,17 @@ test-local:
 	cargo test --features="file_io, fetch_remote_manifests, add_thumbnails, v1_api" --all-targets
 
 test-wasm:
-	cd sdk && wasm-pack test --node -- --no-default-features --features="rust_native_crypto, fetch_remote_manifests"
+	cd sdk && wasm-pack test --node -- --no-default-features --features="rust_native_crypto, fetch_remote_manifests, http_reqwest"
```
tmathern, line 29: It would be good to document what to enable when for now. This doc will be useful for others using the SDK (as I wager this change may break builds or at least some integration tests if the proper features are missing).

sdk/Cargo.toml
```diff
@@ -36,18 +36,27 @@ features = [
     "pdf",
     "v1_api",
     "diagnostics",
+    "http_ureq",
+    "http_reqwest",
```
tmathern: (Need for doc as emntioned above: Especially since as a developer discovering the SDJ I would have the immediate naive question if those are mutually exclusive or not)

sdk/Cargo.toml
```diff
@@ -36,18 +36,27 @@ features = [
...
 ]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["openssl"]
+# The reason we use `ureq` instead of `reqwest_blocking` is because the blocking version
+# of reqwest creates a new tokio runtime. A lot of our tests create their own tokio runtime
+# but do not call the async versions of our API. Thus, we use ureq, which has a complete
+# blocking implementation, no runtimes included.
+# TODO: make our async tests always use the async API so we don't need ureq + reqwest
```
tmathern: Link the newly created ticket here

sdk/Cargo.toml
tmathern: (Doc as in a readme, or developer setup part of a readme)

sdk/examples/v2show.rs
```diff
@@ -14,12 +14,14 @@
...
 
 use anyhow::Result;
+
 #[cfg(target_arch = "wasm32")]
 fn main() -> Result<()> {
     Ok(())
 }
 
-#[cfg(not(target_arch = "wasm32"))]
+// TODO: use SyncGenericResolver when we export public API for it to support wasi
```
tmathern, line 23: Link ticket?

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -12,19 +12,15 @@
 // each license.
 
 use async_generic::async_generic;
+use chrono::{DateTime, Utc};
```
tmathern, line 15: Is it OK now to remove that config `#[cfg(not(target_arch = "wasm32"))]`? Why?

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -276,54 +277,51 @@ fn check_stapled_ocsp_response(
...
-            // TO REVIEW: This is how the old code worked, but is it correct to ignore a
-            // malformed OCSP response?
-            return Ok(OcspResponse::default());
-        };
+    // Check the OCSP response, but only if it is well-formed.
+    // Revocation errors are reported in the validation log.
+    let Ok(ocsp_data) =
+        OcspResponse::from_der_checked(&ocsp_response_der, signing_time, validation_log)
+    else {
+        // TO REVIEW: This is how the old code worked, but is it correct to ignore a
```
tmathern: @mauricefisher64 Is the question in the comment something we know the answer of by now?

sdk/src/reader.rs
```diff
@@ -125,30 +126,56 @@ impl Reader {
     /// let reader = Reader::from_stream("image/jpeg", stream).unwrap();
     /// println!("{}", reader.json());
     /// ```
-    #[async_generic()]
+    #[async_generic]
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn from_stream(format: &str, mut stream: impl Read + Seek + Send) -> Result<Reader> {
+    pub fn from_stream(format: &str, stream: impl Read + Seek + Send) -> Result<Reader> {
```
gpeacock: This is technically a breaking change, since Rust really cares about mutability, but it is a good change. We don't really want to change the input stream.

sdk/src/ingredient.rs
```diff
@@ -938,51 +966,56 @@ impl Ingredient {
...
-        Self::from_stream_async(format, &mut stream).await
+        Self::from_stream_async(format, &mut stream, &AsyncGenericResolver::new()).await
     }
 
     /// Creates an `Ingredient` from a stream (async version).
     ///
     /// This does not set title or hash.
     /// Thumbnail will be set only if one can be retrieved from a previous valid manifest.
-    pub async fn from_stream_async(format: &str, stream: &mut dyn CAIRead) -> Result<Self> {
+    pub async fn from_stream_async(
```
gpeacock: This is a public API, so a breaking change and we would need to publish and document AsyncHttpResolver. Was that the intention?

gpeacock (commented):
I've got some questions.

sdk/src/crypto/time_stamp/error.rs
```diff
@@ -68,4 +70,8 @@ pub enum TimeStampError {
     /// response.
     #[error("internal error ({0})")]
     InternalError(String),
+
+    /// An error occurred while resolving the http request.
+    #[error(transparent)]
```
tmathern, line 75: Isn't adding an error going to mean possible breaking changes (for whoever checks that error enum)?

sdk/src/error.rs
```diff
@@ -300,6 +303,12 @@ pub enum Error {
     #[error(transparent)]
     Utf8Error(#[from] std::str::Utf8Error),
 
+    #[error(transparent)]
```
tmathern, line 306: Same note as previous - going to be a breaking change?

sdk/src/ingredient.rs
tmathern: +1

ok-nick: No, I didn't mean to expose this to the public API yet, I will change it.

sdk/src/reader.rs
ok-nick: I don't think this is actually a breaking change. It only affects the local variable in the function, not the underlying value of the variable. To change the signature to be mutable we'd have to do `stream: &mut impl Read + Seek + Send`.

Makefile
ok-nick: I don't think we document features anywhere, although I agree that we should. Most Rust libraries describe their features in the main file to be generated by the docs so I think that would be the way to go.

sdk/src/crypto/cose/ocsp.rs
ok-nick: Yes, OCSP fetching wasn't implemented for WASM and now that it is we use these parameters.

sdk/src/crypto/time_stamp/error.rs
ok-nick: No, because we mark the `Error` type as `#[non_exhaustive]`.

sdk/src/error.rs
ok-nick: No, see above.

sdk/Cargo.toml
ok-nick: Yes, good idea, see above.

Makefile
gpeacock: We do document features on the main readme and in rust docs.
This introduces some breaking changes at the c2pa-rs level, but it doesn't directly impact the c-ffi-api and associated bindings since we don't directly expose Ingredient there. It may introduce an new binding api for adding resolvers in other languages which could take some effort to figure out.

ok-nick: I'm not a huge fan of the current feature situation. Rust docs state that features [should be additive](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification) and mutually exclusive features [should be avoided if at all possible](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#mutually-exclusive-features) because if you depend on two crates with both features, the features will use the union of them (thus breaking the crate).

I'm going to rework the generic resolvers to default to a particular http library if multiple features are enabled. I think we should also fix the situation with openssl and rust_native_crypto, I'll create an issue to track it.

Edit: #1372

sdk/Cargo.toml
scouten-adobe: I'm concerned about this comment:

> TODO: make our async tests always use the async API so we don't need ureq + reqwest

If I understand correctly, that would mean we would have no test coverage for the synchronous APIs. If so, that seems like a bug waiting to happen.

sdk/Cargo.toml
```diff
@@ -189,6 +209,30 @@ x509-certificate = "0.24.0"
 x509-parser = "0.16.0"
 zeroize = { version = "1.8", features = ["zeroize_derive"] }
 zip = { version = "3.0.0", default-features = false }
+# TODO: this can eventually be optional
+http = "1.3.1"
```
scouten-adobe: We mostly alphabetize the dependencies so as to avoid having duplicates. Would you mind moving this to its place in the sorted order?

sdk/src/crypto/asn1/mod.rs
```diff
@@ -10,10 +10,6 @@
 // cryptographic-message-syntax crate located at:
 // https://github.com/indygreg/cryptography-rs/tree/main/cryptographic-message-syntax/src/asn1
 
-// We can not incorporate the entire crate directly because other parts of the
```
scouten-adobe, line 13: Why delete this comment?

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -276,54 +277,51 @@ fn check_stapled_ocsp_response(
...
     ctp: &CertificateTrustPolicy,
     _tst_info: Option<&TstInfo>,
     validation_log: &mut StatusTracker,
 ) -> Result<OcspResponse, CoseError> {
-    #[cfg(target_arch = "wasm32")]
-    {
-        let _ = (sign1, data, ctp, validation_log);
-        Ok(OcspResponse::default())
-    }
+    let certs = cert_chain_from_sign1(sign1)?;
```
scouten-adobe, line 348: Really nice to see this implemented!

sdk/src/crypto/cose/ocsp.rs
scouten-adobe: IIRC I wrote this comment in December when I did the c2pa_crypto refactor. Question is still open.

sdk/src/crypto/time_stamp/http_request.rs
```diff
@@ -65,10 +73,11 @@ pub fn default_rfc3161_request(
     Ok(ts)
 }
 
+#[async_generic]
 fn time_stamp_request_http(
     url: &str,
     headers: Option<Vec<(String, String)>>,
-    request: &TimeStampReq,
+    timestamp_request: &TimeStampReq,
```
scouten-adobe, line 95: :+1: to more explicit variable names.

sdk/src/http.rs
```diff
@@ -0,0 +1,545 @@
...
+use std::io::{self, Read};
+
+use async_trait::async_trait;
+use http::{Request, Response};
+
+use crate::Result;
+
+/// A resolver for sync (blocking) HTTP requests.
+pub trait SyncHttpResolver {
+    /// Resolve a [`http::Request`] into a [`http::Response`] with a streaming body.
```
scouten-adobe:
```suggestion
    /// Resolve a [`Request`] into a [`Response`] with a streaming body.
    ///
    /// [`Request`]: http::Request
    /// [`Response`]: http::Response
```

scouten-adobe: … and similar throughout this file, if you're so inclined. 🙂

sdk/src/http.rs
```diff
@@ -0,0 +1,546 @@
...
+/// Note that WASM (non-WASI) does not have a built-in [`SyncHttpResolver`].
+pub struct SyncGenericResolver {
+    http_resolver: Box<dyn SyncHttpResolver>,
+}
+
+impl SyncGenericResolver {
+    /// Create a new [`SyncGenericResolver`] with an auto-specified [`SyncHttpResolver`].
+    ///
+    /// Note that if `http_ureq` and `http_reqwest_blocking` are enabled at the same time, this
+    /// function will panic.
```
scouten-adobe: Rather than panic, could we make this a compile-time error?

scouten-adobe:
More specific review comments pending as I read the detailed diffs, but I'll post this separately since it doesn't apply to any specific portion of the PR:

I'd like to explore the idea that this is a valuable interim step along the way to where we eventually want to land, but this may not be the right API for our 1.0 release. I'm in general concerned with our use of global state (including thread-local storage) and this PR adds to that state of affairs.

The way that `SyncGenericResolver::new` is baked in at the core of several call sites doesn't seem to offer a way to substitute in a custom implementation of the `SyncHttpResolver` trait should that be necessary. (I _think_ Eli is going to want to do that in the JS SDK, as a specific example.)

To be clear, this comment is _not_ intended to block merging of this PR, but rather to flag this as a new part of a larger discussion about globals and thread-locals that I've already requested happen before we declare 1.0.

sdk/src/http.rs
```diff
@@ -0,0 +1,546 @@
...
+    Other(Box<dyn std::error::Error + Send + Sync>),
+}
+
+#[cfg(all(feature = "http_reqwest_blocking", not(target_os = "wasi")))]
+mod sync_reqwest_resolver {
+    use std::io::Cursor;
+
+    use super::*;
+
+    impl SyncHttpResolver for reqwest::blocking::Client {
```
scouten-adobe: I _love_ this approach!

sdk/src/http.rs
```diff
@@ -0,0 +1,546 @@
...
+    use super::*;
+
+    impl From<reqwest::Error> for HttpResolverError {
+        fn from(value: reqwest::Error) -> Self {
+            Self::Other(Box::new(value))
+        }
+    }
+}
+
+// TODO: Switch to reqwest_blocking once it supports WASI https://github.com/seanmonstar/reqwest/issues/2294
```
scouten-adobe: Can we add a tracking issue for this in our project?

sdk/src/http.rs
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    impl From<outgoing_handler::ErrorCode> for HttpResolverError {
+        fn from(value: outgoing_handler::ErrorCode) -> Self {
+            Self::Other(Box::new(value))
+        }
+    }
+}
+
+// TODO: Switch to reqwest once it supports WASI https://github.com/seanmonstar/reqwest/issues/2294
```
scouten-adobe: Please add a tracking issue in our project for this, too.

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -79,71 +77,28 @@ pub(crate) async fn resolve(did: &Did<'_>) -> Result<DidDocument, DidWebError> {
 }
 
 async fn get_did_doc(url: &str) -> Result<Vec<u8>, DidWebError> {
-    #[cfg(not(target_os = "wasi"))]
```
scouten-adobe, line 82: Thank you for filling in the missing WASI implementation 🙏 .

sdk/src/lib.rs
```diff
@@ -86,6 +86,30 @@
...
+//!
+//! - **openssl** *(enabled by default)*: Enables the system `openssl` implementation for cryptography.
+//! - **rust_native_crypto** *(enabled by default)*: Enables the Rust native implementation for cryptography.
+//! - **http_ureq** *(enabled by default)*: Enables `ureq` for sync HTTP requests.
+//! - **http_reqwest** *(enabled by default)*: Enables `reqwest` for async HTTP requests.
+//! - **http_reqwest_blocking**: Enables the `blocking` feature of `reqwest` for sync HTTP requests.
+//! - **http_wasi** *(enabled by default)*: Enables `wasi` for sync HTTP requests on WASI.
+//! - **http_wstd** *(enabled by default)*: Enables `wstd` for async HTTP requests on WASI.
+//!
+// TODO: document other features...
```
scouten-adobe: I believe these are documented in the README.md file. Let's try to avoid duplication between the two lists.

sdk/src/reader.rs
scouten-adobe: @gpeacock I think the reason this isn't breaking is that `from_stream` takes ownership of `stream` rather than receiving it by reference. Since the caller transfers ownership, it doesn't really matter whether _this_ function uses it as a mutable variable or not.

If this were a `&dyn Read + Seek + Send`, then the change in mutability would be a breaking change.

sdk/src/store.rs
```diff
@@ -3826,131 +3839,30 @@ impl Store {
         }
     }
 
-    // fetch remote manifest if possible
```
scouten-adobe: Yay for deletion of platform-specific code! 🎉

sdk/src/http.rs
ok-nick: I changed this in a recent commit to default to `http_reqwest_blocking`. It's technically valid to enable multiple http features if different dependencies are using them. The traits are implemented for those dependency-specific structs as well so that may be useful when we allow resolvers to be passed in through the public API.

sdk/Cargo.toml
ok-nick: No problem.

sdk/src/crypto/cose/ocsp.rs
ok-nick: Hm, maybe we should report `signingCredential.ocsp.skipped` then?

sdk/src/http.rs
ok-nick: #1377

sdk/src/http.rs
ok-nick: #1377

sdk/src/lib.rs
ok-nick: I couldn't find it documented in the README, although regardless I do think we should document these features in the docs.rs (if not exclusively). That's the first place I look and it's a common pattern among popular Rust crates.

sdk/src/http.rs
ok-nick: No problem.

sdk/src/crypto/asn1/mod.rs
ok-nick: All networking in this codebase supports sync/async and additionally there is no networking within this set of modules.

sdk/src/lib.rs
ok-nick: Ok I sent a commit adding the new features to `usage.md`. However, I think we should still consider adding this to the main docs.rs.

sdk/Cargo.toml
```diff
@@ -36,18 +36,28 @@ features = [
...
     "v1_api",
     "diagnostics",
+    "http_ureq",
+    "http_reqwest",
 ]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["openssl"]
+# The reason we use `ureq` instead of `reqwest_blocking` is because the blocking version
```
cdmurph32: What is the basis of having so many options? Is the resolver used in all feature combinations? It would be nice if the features mapped across all targets, rather than target specific features.

ok-nick: > What is the basis of having so many options?

You only need to enable the features for the networking libraries you use.

> Is the resolver used in all feature combinations?

It can be, let's say for instance there are multiple sub-dependencies that use c2pa-rs with different http features, cargo will combine them all. You also may not use async but do use sync, so no reason to enable the async library. In our case, we must use both `ureq` (sync) and `reqwest` (async) because of how our tests are setup.

> It would be nice if the features mapped across all targets, rather than target specific features.

Well they are library-specific, e.g. `http_wasi` enables the `wasi` library but it just so happens it can only be used on WASI. Same with `http_wstd` and the `wstd` library.

tmathern: Mild concern about the test matrix here. What is the minimum set of combinations we need to think of to cover this with enough testing so we feel good here?

ok-nick: You can run `--all-features` (or all http features) on any given platform and it will test all of the networking libraries that are supported on that given platform with the tests included in the module. Unfortunately WASM doesn't support our testing framework, `httpmock`, although we don't currently have any alternatives in place besides real world calls.

Edit: although currently it defaults to `http_reqwest_blocking` when all features are enabled which causes issues with #1364

sdk/Cargo.toml
```diff
@@ -120,6 +138,7 @@ ed25519-dalek = { version = "2.1.1", features = [
 env_logger = "0.10.0"
 hex = "0.4.3"
 hex-literal = "0.4.1" # Might be only a dev dependency
+http = "1.3.1" # TODO: this can eventually be optional
```
tmathern: Eventually... when? Is it ticket'ed? (Otherwise it will fly under the radar).

ok-nick: I'll remove this comment for now, I don't think it's realistic or worthwhile to make this dependency optional. Thanks for pointing it out.

sdk/Cargo.toml
tmathern: I am not necessarily a fan of that, because users could still pick and choose. We should aime when possible for less features, not more. But here it might not be possible. Therefore: Ok.

sdk/src/builder.rs
```diff
@@ -2853,7 +2856,9 @@ mod tests {
 
         output.set_position(0);
 
-        let reader = Reader::from_stream("jpeg", &mut output).expect("from_bytes");
+        let reader = Reader::from_stream_async("jpeg", &mut output)
```
tmathern, line 3224: Why not keep the test for both sync and async paths?

sdk/Cargo.toml
ok-nick: It's fine to pick and choose features, that's how it's done by default, but all of them can be enabled for the sake of testing, due to dependency conflicts (which are bound to happen to someone), or if you do intend to use multiple networking libraries.

ok-nick: Also to answer your question more directly "the minimum set of combinations" is 1 to have testing for all of them.

sdk/src/lib.rs
ok-nick: I made an issue to track this here #1414 for further discussion.

sdk/src/http/wasi.rs
```diff
@@ -0,0 +1,171 @@
...
+// at your option.
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+// TODO: Switch to reqwest_blocking once it supports WASI https://github.com/contentauth/c2pa-rs/issues/1377
```
tmathern, line 14: Would it mean this whole file will disappear too? Will we be able tor efactor it and fold it into something else?

ok-nick: From my understanding `wasi` is the low-level library, `wstd` is an abstraction on top of it, and when WASI is fully supported in `std` we can use `reqwest` directly. To answer your question, I suspect we will no longer need this code but on the other hand, maybe we don't want to force our WASI users to pull in a big library like `reqwest`.

sdk/src/builder.rs
ok-nick: This was part of my effort to fix #1364 until I realized there was a lot more work involved. Our sync and async code paths should be equivalent besides the extra unpredictability of concurrency, so testing async is testing sync with more rigor.

sdk/Cargo.toml
scouten-adobe: I share @tmathern's concern and will amplify it a bit. Each feature should be tested in CI (at least in the release-readiness suite); otherwise we don't really know that it's still viable.

sdk/src/crypto/ocsp/fetch.rs
```diff
@@ -143,31 +141,45 @@ fn process_ocsp_responders(certs: &[Vec<u8>]) -> Option<Vec<OcspRequestData>> {
     }
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-#[async_generic()]
+/// Retrieve an OCSP response if available.
+///
+/// Checks for an OCSP responder in the end-entity certifricate. If found, it
```
scouten-adobe: Typo: certifricate -> certificate

sdk/src/http/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+use http::{Request, Response};
+
+use crate::Result;
+
+mod reqwest;
+mod ureq;
+mod wasi;
+
+// Since we use `http::Request` and `http::Response` we also expose the `http` crate.
+pub use http;
```
scouten-adobe, line 26: I'd prefer that we not rebroadcast an external crate's API. What would be the consequence of removing this?

sdk/src/http/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+pub enum HttpResolverError {
+    /// An error occured in the [`http`] crate.
+    #[error(transparent)]
+    Http(#[from] http::Error),
+
+    /// An error occured in during I/O.
+    #[error(transparent)]
+    Io(#[from] io::Error),
+
+    /// The sync http resolver is not implemented.
```
scouten-adobe: General style critique (and we're already very inconsistent on this): I'd prefer that we capitalize acronyms in our public documentation (i.e. HTTP instead of http).

sdk/src/http/reqwest.rs
```diff
@@ -0,0 +1,111 @@
...
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+#[cfg(all(feature = "http_reqwest_blocking", not(target_os = "wasi")))]
+pub mod sync_impl {
```
scouten-adobe, line 15: I'm curious why this is marked `pub` when the `request` mod is not otherwise exposed.

sdk/src/http/reqwest.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        #[test]
+        fn test_http_reqwest() {
+            assert_http_resolver(reqwest::blocking::Client::new());
+        }
+    }
+}
+
+#[cfg(all(feature = "http_reqwest", not(target_os = "wasi")))]
+pub mod async_impl {
```
scouten-adobe, line 55: Same comment about `pub` access here.

sdk/src/http/ureq.rs
```diff
@@ -0,0 +1,57 @@
...
+
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+#[cfg(all(feature = "http_ureq", not(target_arch = "wasm32")))]
+pub mod sync_impl {
```
scouten-adobe, line 15: And again, same comment. (I'll stop repeating myself here.)

sdk/src/store.rs
```diff
@@ -3086,201 +3098,39 @@ impl Store {
         }
     }
 
-    #[cfg(all(feature = "fetch_remote_manifests", target_os = "wasi"))]
```
scouten-adobe, line 3378: This is really nice to see. (Why we're doing this PR, right?) 😎

sdk/src/http/mod.rs
ok-nick: If we removed this, users of this part of the library must ensure they have the `http` dependency with the version we use as a dependency in `c2pa`. I think if we're exposing these definition structs in our public API then they should be accessible from there, otherwise you will have to pull it in as a dependency anyways.

sdk/src/http/mod.rs
ok-nick: I agree, thanks for pointing this out.

sdk/src/http/reqwest.rs
ok-nick: It's marked as `pub` so that it's accessible from the `sdk/src/http/mod.rs` module. I can do `pub(crate)` if that's preferred, but it's not being reexported to the public API.

sdk/src/store.rs
ok-nick: Of course!

sdk/src/http/mod.rs
scouten-adobe: OK, that makes sense. I ran `cargo doc --open` and did not find our `http` mod in the documented public API surface. But I also found a TODO that says we're intending to add that to the public API surface at some point. This is good prep for that time.

sdk/Cargo.toml
scouten-adobe: With the changes to release-readiness, I'm :+1:.

scouten-adobe (changes requested):
Thank you for addressing all of the concerns so far. I would like to understand what we're seeing with the performance benchmarks. Can you dig in on that a little bit before we merge?

ok-nick:
The benchmark differences are due to enabling all features during benchmarking CI, see fix in #1423.

sdk/src/crypto/cose/ocsp.rs
ok-nick: This was resolved on main already.

PR #1356, opened 2025-08-28
feat: Add `cawg_trust.verify_trust_list` setting

scouten-adobe:
Draft PR for now … needs refinement and a possible corresponding option for C2PA trust infrastructure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -669,7 +669,6 @@ dependencies = [
  "console_log",
  "const-hex",
  "const-oid",
- "conv",
  "coset",
  "der",
  "ecdsa",
@@ -1097,15 +1096,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "conv"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78ff10625fd0ac447827aa30ea8b861fead473bb60aeb73af6c1c58caf0d1299"
-dependencies = [
- "custom_derive",
-]
-
-[[package]]
 name = "core-foundation"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1238,12 +1228,6 @@ dependencies = [
  "quote",
  "syn 2.0.106",
 ]
-
-[[package]]
-name = "custom_derive"
-version = "0.1.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef8ae57c4978a2acd8b869ce6b9ca1dfe817bff704c220209fdef2c0b75a01b9"
 
 [[package]]
 name = "darling"
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -106,7 +106,6 @@ config = { version = "0.14.0", default-features = false, features = [
 ] }
 const-hex = "1.14"
 const-oid = { version = "0.9.6", optional = true }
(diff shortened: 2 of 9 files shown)


cli/src/main.rs
```diff
@@ -105,6 +105,10 @@ struct CliArgs {
     #[clap(long = "certs")]
     cert_chain: bool,
 
+    /// Do not enforce trust lists. (CAWG-only at the moment.)
+    #[clap(long)]
+    disable_cawg_trust: bool,
```
gpeacock: This does not need to be a command line option, it belongs only in settings.

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -333,25 +333,38 @@ impl IdentityAssertion {
             // Load the trust handler settings. Don't worry about status as these
             // are checked during setting generation.
 
-            if let Ok(Some(ta)) = get_settings_value::<Option<String>>("cawg_trust.trust_anchors") {
-                let _ = ctp.add_trust_anchors(ta.as_bytes());
-            }
+            let cose_verifier =
+                if let Ok(true) = get_settings_value::<bool>("cawg_trust.bypass_trust") {
```
gpeacock: why not just not configure cawg trust if you don't want it?
i.e. you can update the settings to not enable cawg trust lists.
If we need a special setting for this, I'd rather have cawg_trust enabled  by default and then disable it. it just seems clearer to me.

gpeacock (changes requested):
What is the use case for this, we should document the settings value with some reason to use it.

tmathern:
Does this mean we can enable/disable the trust verification using settings then, if I understand correctly?

cli/src/main.rs
tmathern: +1, I also think this should be "only" a setting. You can also use settings with the cli anyway.

andrewhalle:
@gpeacock This came out of some discussion I had with Eric so let me see if I can explain why something like this would be useful for me. Since I frequently look at images signed with self-signed CAWG X.509 certs (just test images), upgrading from `c2patool@0.18.1` to `c2patool@0.20.1` broke my workflow because an image that was previously "Valid" but untrusted now just shows an error.

I treat `c2patool` as a developer tool, so it's useful for me to be able to extract all available information regardless of trust. In suggested something like an `--allow-untrusted` flag, I drew analogy with `curl --insecure` for connecting to an HTTPS endpoint without validating the certificate.

I haven't yet taken a look to see if I can achieve the same workflow as I had before by just changing settings, but I do in general like when this sort of functionality is exposed via CLI flags.

gpeacock:
there was a bug in the CAWG validation logic that was causing it to error out in this case instead of generating a report that includes a error on the fact that the CAWG was not trusted (because of no trust list) but is valid. So now you will get a full report regardless and there is no need to pass a special option for this.

sdk/src/identity/identity_assertion/assertion.rs
gpeacock: you don't need to do a setting or anything, just update c2patool with the new sdk and it will go back to working as expected.

scouten-adobe: Removed the c2patool changes. This is only in settings now.

cli/src/main.rs
scouten-adobe: CL option removed.

PR #1370, opened 2025-09-02
Add CAWG validation to reader

scouten-adobe:
This is a sketch of the changes I'd like to see made to the `Reader` interface to make CAWG more directly part of the reading process from clients' point of view.

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -463,6 +463,10 @@ pub unsafe extern "C" fn c2pa_free_string_array(ptr: *const *const c_char, count
 
 // Run CAWG post-validation - this is async and requires a runtime.
 fn post_validate(result: Result<C2paReader, c2pa::Error>) -> Result<C2paReader, c2pa::Error> {
+    if true {
+        todo!("Remove me?");
+    }
+
     match result {
         Ok(mut reader) => {
             let runtime = match Runtime::new() {
--- a/sdk/examples/cawg_identity.rs
+++ b/sdk/examples/cawg_identity.rs
@@ -125,9 +125,7 @@ mod cawg {
 
         builder.sign_file_async(&signer, source, &dest).await?;
 
-        let mut reader = Reader::from_file(dest)?;
-
-        reader.post_validate_async(&CawgValidator {}).await?;
+        let reader = Reader::from_file_async(dest).await?;
 
         println!("{reader}");
         Ok(())
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -136,7 +136,13 @@ impl Reader {
             Store::from_stream_async(format, &mut stream, verify, &mut validation_log).await
         }?;
 
-        Self::from_store(store, &validation_log)
+        let /* mut */ result = Self::from_store(store, &validation_log)?;
+        if false {
+            // QUESTION: What to do if we're in the _sync version and there
+            // are identity assertions? Just report an error (needs async)?
+            todo!("Add identity assertion validation here");
+        }
+        Ok(result)
     }
 
     #[async_generic()]
@@ -151,7 +157,11 @@ impl Reader {
             Store::from_stream_async(format, &mut stream, verify, &mut validation_log).await
         }?;
 
-        Self::from_store(store, &validation_log)
+        let mut result = Self::from_store(store, &validation_log)?;
+        if false {
+            todo!("Add identity assertion validation here");
+        }
+        Ok(result)
     }
 
     #[cfg(feature = "file_io")]
@@ -735,6 +745,9 @@ impl Reader {
         validator: &impl AsyncPostValidator
     ))]
     pub fn post_validate(&mut self, validator: &impl PostValidator) -> Result<()> {
+        if true {
+            todo!("Remove me");
+        }
         let mut validation_log = StatusTracker::default();
         let mut validation_results = self.validation_results.take().unwrap_or_default();
         let mut assertion_values = HashMap::new();
@@ -997,6 +1010,9 @@ pub mod tests {
 
     #[test]
     fn test_reader_post_validate() -> Result<()> {
+        if true {
+            todo!("Remove me");
+        }
         use crate::{log_item, status_tracker::StatusTracker};
 
         let mut reader =
--- a/c2pa_c_ffi/src/json_api.rs
+++ b/c2pa_c_ffi/src/json_api.rs
@@ -24,6 +24,9 @@ use crate::{Error, Result, SignerInfo};
 pub fn read_file(path: &str, data_dir: Option<String>) -> Result<String> {
     let mut reader = Reader::from_file(path).map_err(Error::from_c2pa_error)?;
     let runtime = Runtime::new().map_err(|e| Error::Other(e.to_string()))?;
+    if true {
+        todo!("Remove post_validate_async here?");
+    }
     runtime
         .block_on(reader.post_validate_async(&CawgValidator {}))
         .map_err(Error::from_c2pa_error)?;
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -574,6 +574,9 @@ fn verify_fragmented(init_pattern: &Path, frag_pattern: &Path) -> Result<Vec<Rea
 
 // run cawg validation if supported
 fn validate_cawg(reader: &mut Reader) -> Result<()> {
+    if true {
+        todo!("Remove me?");
+    }
     #[cfg(not(target_os = "wasi"))]
     {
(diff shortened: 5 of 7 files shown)


sdk/src/reader.rs
```diff
@@ -136,7 +136,13 @@ impl Reader {
             Store::from_stream_async(format, &mut stream, verify, &mut validation_log).await
         }?;
 
-        Self::from_store(store, &validation_log)
+        let /* mut */ result = Self::from_store(store, &validation_log)?;
+        if false {
+            // QUESTION: What to do if we're in the _sync version and there
```
gpeacock: I think we need to support both sync and async identity validation if we are going to include it at this level in the API. If it is part of validation, it can't just be async validation.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -463,6 +463,10 @@ pub unsafe extern "C" fn c2pa_free_string_array(ptr: *const *const c_char, count
 
 // Run CAWG post-validation - this is async and requires a runtime.
 fn post_validate(result: Result<C2paReader, c2pa::Error>) -> Result<C2paReader, c2pa::Error> {
```
gpeacock, line 520: PostValidate could be useful for validating and formatting any 3rd party assertions we don't have in the SDK. 
Ideally there would be some way to integrate this into the definition of an Assertion such that you would just need to add an assertion handler to the SDK, but that's a problem for tomorrow.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -463,6 +463,10 @@ pub unsafe extern "C" fn c2pa_free_string_array(ptr: *const *const c_char, count
...
+    if true {
```
tmathern: Will you need sync APIs here?

sdk/examples/cawg_identity.rs
```diff
@@ -125,9 +125,7 @@ mod cawg {
 
         builder.sign_file_async(&signer, source, &dest).await?;
 
-        let mut reader = Reader::from_file(dest)?;
-
-        reader.post_validate_async(&CawgValidator {}).await?;
+        let reader = Reader::from_file_async(dest).await?;
```
tmathern, line 127: Why not keep the sync interface too? I am not sure we can make everything async

sdk/src/reader.rs
tmathern: Good question. I wouldn't want everything to become async. I wonder if we need the sync interface for some native bindings? It's not great but maybe we need a blocking API? Or a cawg validation opt-out for the sync path?

sdk/src/reader.rs
```diff
@@ -997,6 +1034,10 @@ pub mod tests {
 
     #[test]
     fn test_reader_post_validate() -> Result<()> {
+        if false {
```
scouten-adobe: Just noticed this was still here. Will remove shortly.

sdk/src/reader.rs
```diff
@@ -154,23 +183,109 @@ impl Reader {
...
+
+        Ok(reader)
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    pub async fn from_stream_with_cawg_async(
+        format: &str,
+        stream: impl Read + Seek,
+    ) -> Result<Reader> {
+        let mut reader = Self::from_stream_async(format, stream).await?;
```
tmathern: The implementations in 
pub async fn from_stream_with_cawg_async

and 

pub async fn from_stream_with_cawg_async

are the same. Is there no way to reuse instead of duplicating? No trick to bypass the differences in requesteds traits on the stream ?

scouten-adobe: I wish … I wish …

Sadly, there is not such a trick. (I've looked.)

sdk/src/reader.rs
```diff
@@ -154,23 +183,109 @@ impl Reader {
...
+    #[cfg(not(target_arch = "wasm32"))]
+    pub async fn from_stream_with_cawg_async(
+        format: &str,
+        stream: impl Read + Seek + Send,
+    ) -> Result<Reader> {
+        let mut reader = Self::from_stream_async(format, stream).await?;
+
+        if get_settings_value::<bool>("verify.verify_after_reading")? {
+            reader.post_validate_async(&CawgValidator {}).await?;
+        }
```
emensch: So without `verify.verify_after_reading`, this will behave the same as `Reader::from_stream_async`?

Unless I'm mistaken, that feels a bit unintuitive to me. I'd think that calling `from_stream_with_cawg_async` would be "enough" from an end-user perspective to opt in to CAWG validation.

sdk/src/reader.rs
```diff
@@ -154,23 +183,109 @@ impl Reader {
...
+    ///     .await
+    ///     .unwrap();
+    /// println!("{}", reader.json());
+    /// # }
+    /// ```
+    ///
+    /// [CAWG identity]: https://cawg.io/identity
+    /// [from_stream_with_cawg_async()]: Self::from_stream_with_cawg_async
+    #[cfg(not(target_arch = "wasm32"))]
+    pub async fn from_stream_with_cawg_async(
```
scouten-adobe: TO DO: Delete this and fold back into the from_stream_async call. Add a note (depending on outcome) that says not supported on sync.

scouten-adobe: TO DO: If CAWG identity assertion is encountered, do the asynchronous blocking call in the sync implementation (borrow from the C bindings). Gavin may add a setting to disable if the blocking behavior is undesirable.

scouten-adobe:
TO DO: Check in to whether verify-on-sign does CAWG identity assertion verification and if it becomes async as a result.

sdk/examples/cawg_identity.rs
scouten-adobe: IMHO we shouldn't be hiding the fact that some APIs are inherently asynchronous. We can and should push the cost of bridging that gap as far up the API chain as possible. Leaving this as is.

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -71,14 +71,15 @@ pub struct IdentityAssertion {
...
+#[allow(unused)] // TEMPORARY while considering API simplification
 impl IdentityAssertion {
     /// Find the `IdentityAssertion`s that may be present in a given
     /// [`Manifest`].
     ///
     /// Iterator returns a [`Result`] because each assertion may fail to parse.
     ///
     /// Aside from CBOR parsing, no further validation is performed.
-    pub fn from_manifest<'a>(
+    pub(crate) fn from_manifest<'a>(
```
scouten-adobe, line 82: @emensch @tmathern please review these access changes and LMK if any of these need to remain public.

tmathern: From my point of view the access changes are OK.

sdk/src/manifest.rs
```diff
@@ -535,6 +542,47 @@ impl Manifest {
...
                 }
+                label
+                    if decode_identity_assertions
+                        && (label == "cawg.identity" || label.starts_with("cawg.identity__")) =>
+                {
+                    let value = assertion.as_json_object()?;
+                    let mut ma = ManifestAssertion::new(label.to_string(), value)
+                        .set_instance(claim_assertion.instance());
+
+                    //dbg!(&identity_assertion);
```
tmathern: I think this can be removed for good then!

tmathern (approved):
Can you give the image a better name than cawgi? What is special about it, except that it has cawg data?

sdk/src/identity/identity_assertion/assertion.rs
cdmurph32: Can the fields of IdentityAssertion above be public, or can I have a more general constructor? I redefine it in the Node SDK, which is not a good practice.
https://github.com/contentauth/c2pa-node-v2/blob/main/src/neon_identity_assertion_builder.rs#L34

sdk/src/manifest.rs
```diff
@@ -535,6 +542,45 @@ impl Manifest {
...
+                    for a in claim.assertions() {
+                        partial_claim.add_assertion(a);
+                    }
+
+                    let uri = to_assertion_uri(manifest_label, label);
+                    validation_log.push_current_uri(&uri);
+                    let value: Option<serde_json::Value> = if _sync {
+                        crate::log_item!(
+                            uri,
+                            "decoding identity assertions not supported in sync",
```
gpeacock, line 563: We need to find a better solution for this as soon as we can.

scouten-adobe: Agreed. Tracking here: https://jira.corp.adobe.com/browse/CAI-9767.

PR #1371, opened 2025-09-03
fix: Expose underlying timestamp assertion support

mauricefisher64:
This exposes the generation of the timestamp assertion to the upper levels of the API.

## Checklist


sdk/src/store.rs
```diff
@@ -1870,28 +1925,21 @@ impl Store {
...
-                            continue;
+                        if let Some(signature) = self.get_cose_sign1_signature(referenced_claim) {
+                            if let Ok(tst_info) = verify_time_stamp(
+                                time_stamp_token,
+                                &signature,
+                                &self.ctp,
+                                validation_log,
+                            ) {
+                                svi.timestamps.insert(rc.label().to_owned(), tst_info);
+                                continue;
```
ok-nick: Did you mean to put continue here?

sdk/src/store.rs
```diff
@@ -498,11 +497,67 @@ impl Store {
...
+    fn send_timestamp_token_request(&self, tsa_url: &str, message: &[u8]) -> Result<Vec<u8>> {
+        let body = crate::crypto::time_stamp::default_rfc3161_message(message)?;
+        let headers = None;
+
+        let bytes =
+            crate::crypto::time_stamp::default_rfc3161_request(tsa_url, headers, &body, message)
+                .map_err(|_e| Error::OtherError("timestamp token not found".into()))?;
+
+        // make sure it is a good response
+        let ctp = CertificateTrustPolicy::passthrough();
```
ok-nick: Is there a trust list for TSAs?

sdk/src/store.rs
```diff
@@ -6455,12 +6503,10 @@ pub mod tests {
...
 
-            /* sample of adding timestamp assertion
-
+            /*
+            // sample of adding timestamp assertion
             // lets add a timestamp for old manifest
-            let timestamp = send_timestamp_request(pc.signature_val()).unwrap();
-            crate::crypto::time_stamp::verify_time_stamp(&timestamp, pc.signature_val()).unwrap();
-            let timestamp_assertion = crate::assertions::TimeStamp::new(pc.label(), &timestamp);
+            let timestamp_assertion = store.get_timestamp_assertion(&[pc.label()], "http://timestamp.digicert.com").unwrap();
```
ok-nick: Do we not test this to avoid a network call or it just isn't compatible with this testing?

sdk/src/store.rs
```diff
@@ -498,11 +497,67 @@ impl Store {
         }
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
+    fn send_timestamp_token_request(&self, tsa_url: &str, message: &[u8]) -> Result<Vec<u8>> {
```
gpeacock: I'm not clear on why these are in Store, they don't seem to use self. Maybe we should move these into the Timestamp.rs assertion. I found that worked well for building the Ingredient assertion.

sdk/src/store.rs
mauricefisher64: Yes, but we don't enforce trust lists on signers so that you can use the SDK with any certs.

sdk/src/store.rs
mauricefisher64: Yes, but the code it was skipping was removed so this line is no longer needed.  Good catch.

sdk/src/store.rs
mauricefisher64: We are avoiding the network call

sdk/src/store.rs
```diff
@@ -471,11 +470,77 @@ impl Store {
...
+        let data = manifest.data().ok()?;
+        let mut validation_log =
+            StatusTracker::with_error_behavior(ErrorBehavior::StopOnFirstError);
+
+        let sign1 = parse_cose_sign1(sig, &data, &mut validation_log).ok()?;
+
+        Some(sign1.signature)
+    }
+
+    /// Creates a TimeStamp (c2p.time-stamp) assertion containing the TimeStampTokens for each
```
tmathern:
```suggestion
    /// Creates a TimeStamp (c2pa.time-stamp) assertion containing the TimeStampTokens for each
```
A typo, maybe?

tmathern (approved):

sdk/src/store.rs
mauricefisher64: Done

sdk/src/store.rs
mauricefisher64: Yup

PR #1375, opened 2025-09-03
ci: fix rust native crypto test to use proper features

ok-nick:
Fixes our rust native crypto CI test to use the correct features during testing.


scouten-adobe (approved):
Good catch, thanks!

Issue #1374, opened 2025-09-03
Test all compilation features for release readiness


ok-nick:
Investigate [`cargo-hack`](https://github.com/taiki-e/cargo-hack) as a way to implement this.

PR #1379, opened 2025-09-04
fix: make `openssl` and `rust_native_crypto` features additive

### LINKED ISSUE #1372: `rust_native_crypto` and `openssl` should not be mutually exclusive features
Rust docs state that features [should be additive](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification) and mutually exclusive features [should be avoided if at all possible](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#mutually-exclusive-features) because if you depend on two crates with both features, the features will use the union of them (thus breaking the crate). As a solution, we can have openssl implemented by default UNLESS the `rust_native_crypto` feature flag is specified.

Ran into a similar situation in #1355

ok-nick:
Rust docs state that features [should be additive](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification) and mutually exclusive features [should be avoided if at all possible](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#mutually-exclusive-features) because if you depend on two crates with both features, the features will use the union of them (thus breaking the crate).

The solution is enable `openssl` by default UNLESS the `rust_native_crypto` feature flag is specified. If both are specified, use `rust_native_crypto`.  If neither are specified, there is a compile error (although I think in the future we may not want this).

The added benefit of this is that it allows `--all-features` to work (common especially with `rust-analyzer`), unblocks #1374, and fixes a related issue in #1375.

* Closes #1372

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -24,19 +24,7 @@ rust-version = "1.86.0"
 exclude = ["tests/fixtures"]
 
 [package.metadata.docs.rs]
-# Build with all features, but not with openssl and rust_native_crypto simulatenously.
-features = [
-    "openssl",
-    "add_thumbnails",
-    "file_io",
-    "serialize_thumbnails",
-    "no_interleaved_io",
-    "fetch_remote_manifests",
-    "json_schema",
-    "pdf",
-    "v1_api",
-    "diagnostics",
-]
+all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
@@ -47,8 +35,8 @@ serialize_thumbnails = []
 no_interleaved_io = ["file_io"]
 fetch_remote_manifests = ["dep:wasi"]
 json_schema = ["dep:schemars"]
-openssl = ["dep:openssl"]
 pdf = ["dep:lopdf"]
+openssl = ["dep:openssl"]
 rust_native_crypto = [
     "dep:const-oid",
     "dep:ecdsa",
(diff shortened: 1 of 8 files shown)


cdmurph32:
I'd prefer `rust_native_crypto` be removed as a feature and be the default.

cdmurph32:
This was the previous behavior, but we removed it because of build artifact size. It includes openssl in the build, even if `rust_native_crypto` is selected.

tmathern:
+1 if we could not have openssl full-blown back.

sdk/Cargo.toml
```diff
@@ -47,8 +35,8 @@ serialize_thumbnails = []
 no_interleaved_io = ["file_io"]
 fetch_remote_manifests = ["dep:wasi"]
 json_schema = ["dep:schemars"]
-openssl = ["dep:openssl"]
 pdf = ["dep:lopdf"]
```
scouten-adobe: Why move this? These were in (roughly) alphabetical order before

sdk/src/lib.rs
```diff
@@ -203,11 +203,5 @@ pub(crate) mod store;
 pub(crate) mod utils;
 pub(crate) use utils::{cbor_types, hash_utils};
 
-#[cfg(all(feature = "openssl", feature = "rust_native_crypto"))]
-compile_error!("Features 'openssl' and 'rust_native_crypto' cannot be enabled at the same time.");
-
 #[cfg(not(any(feature = "openssl", feature = "rust_native_crypto")))]
 compile_error!("Either 'openssl' or 'rust_native_crypto' feature must be enabled.");
-
-#[cfg(all(feature = "openssl", target_arch = "wasm32"))]
```
scouten-adobe, line 212: I wish there was a `compile_warning` macro similar to `compile_error`, but alas … there is not.

sdk/Cargo.toml
ok-nick: Artifact from changing things around. I fixed this in latest commit, but I think we should add a toml formatter in here somewhere.

scouten-adobe: Once upon a time, I was contributing to such a tool: https://github.com/FuelLabs/cargo-toml-lint

Looks like it's become abandon-ware. :-(

scouten-adobe (approved):
Thanks for making the changes. :+1: now.

sdk/Cargo.toml
ok-nick: Here's what I use: https://github.com/tamasfe/taplo

PR #1388, opened 2025-09-04
feat: Implement CAWG X.509 signing via settings

scouten-adobe:


sdk/src/settings/signer.rs
```diff
@@ -58,7 +58,25 @@ impl SignerSettings {
     ///
     /// If the signer settings aren't specified, this function will return [Error::MissingSignerSettings][crate::Error::MissingSignerSettings].
     pub fn signer() -> Result<Box<dyn Signer>> {
+        let c2pa_signer = Self::c2pa_signer()?;
+
+        // TO DISCUSS: What if get_value returns an Err(...)?
+        if let Ok(Some(cawg_x509_signer_info)) = Settings::get_value::<Option<SignerSettings>>("cawg_x509_signer") {
```
ok-nick: There's this weird scenario where calling `Settings::get_value` on an optional setting will return `config::Error::NotFound` if it doesn't exist (other scenarios as well). But if you were to call `Settngs::set_value("cawg_x509_signer", None);` beforehand, then `Settings::get_value` would return `None`.

sdk/src/settings/signer.rs
```diff
@@ -58,7 +58,25 @@ impl SignerSettings {
     ///
     /// If the signer settings aren't specified, this function will return [Error::MissingSignerSettings][crate::Error::MissingSignerSettings].
     pub fn signer() -> Result<Box<dyn Signer>> {
```
ok-nick, line 66: Hm, we should probably change this function name as well to represent the CAWG signer (if you weren't planning on doing that already).

scouten-adobe: Actually, I think the name is correct as it stands. (See implementation that added just now.)

The end result is a single `Signer` object. When CAWG settings are detected, it creates a new (internal) `CawgX509IdentitySigner` which wraps the existing `Signer` object and replaces its `dynamic_assertions` callback with one that creates an `X509CredentialHolder` which will then add the CAWG identity assertion in at the right place.

scouten-adobe:
Still TO DO: Add test coverage and documentation for newly-introduced settings.

sdk/src/settings/signer.rs
```diff
@@ -94,6 +137,35 @@ impl SignerSettings {
...
+            }
+
+            SignerSettings::Remote {
+                url: _url,
+                alg: _alg,
+                sign_cert: _sign_cert,
+                tsa_url: _tsa_url,
+            } => todo!("Remote signing not yet supported"),
+        }
+    }
```
ok-nick: Do we still need this function?

sdk/src/settings/signer.rs
```diff
@@ -107,6 +179,85 @@ impl SettingsValidate for SignerSettings {
...
+            self.cawg_sign_cert.as_bytes(),
+            self.cawg_private_key.as_bytes(),
+            self.cawg_alg,
+            self.cawg_tsa_url.clone(),
+        ) else {
+            // dynamic_assertions() API doesn't let us fail.
+            // signer_from_cert_chain_and_private_key rarely fails,
+            // so when it does, we do so silently.
+            return vec![];
+        };
```
ok-nick, line 218: Hm I see why this is done but it seems a bit awkward. We should consider storing a generic in `X509CredentialHolder` in the future, if possible.

Edit: Although I suspect we'd need many more changes in the `signer_from...` function.

scouten-adobe: There's a really nasty borrow-checker story behind this approach. We'll slay that dragon another time.

sdk/src/settings/signer.rs
scouten-adobe: You're right; we don't. Deleted.

cli/src/main.rs
```diff
@@ -740,7 +740,11 @@ fn main() -> Result<()> {
 
             Box::new(signer)
         } else {
-            sign_config.signer()?
+            match Settings::signer() {
```
scouten-adobe, line 743: Thanks to @ok-nick for this patch!

PR #1394, opened 2025-09-08
fix: Update library-release.yml

tmathern:
- Change merge-multiple from true to false, as there are suspicions this corrupts files on release

## Checklist


tmathern:
It seems the only way to test this i through a fork. So first testing, then possibly reopening if it fixes the issue.

PR #1406, opened 2025-09-11
feat(sdk): Introduce new API to retrieve detailed manifest JSON

magaylor:
Introduced a new API that allows for getting detailed manifest json from a C2paReader.

Additionally, I fixed some errors that make test exposed and fixed some typos.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,9 @@ name: CI
 on:
   workflow_dispatch:
   pull_request:
+    branches:
+      - main
+      - v1_api
     types:
       - opened
       - reopened
@@ -17,6 +20,7 @@ on:
   push:
     branches:
       - main
+      - v1_api
 
 jobs:
   get-features:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.64.0](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.63.0...c2pa-v0.64.0)
+_15 September 2025_
+
+### Added
+
+* Fetch OCSP wasm ([#1402](https://github.com/contentauth/c2pa-rs/pull/1402))
+
+### Fixed
+
+* Allow v1 claim  c2pa.manifest cbor assertions without full valid… ([#1412](https://github.com/contentauth/c2pa-rs/pull/1412))
+
 ## [0.63.0](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.62.0...c2pa-v0.63.0)
 _10 September 2025_
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,11 +283,11 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "2.5.0"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "19634d6336019ef220f09fd31168ce5c184b295cbf80345437cc36094ef223ca"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
 dependencies = [
- "async-lock",
+ "autocfg",
  "cfg-if",
  "concurrent-queue",
  "futures-io",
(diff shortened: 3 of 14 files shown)


sdk/src/reader.rs
```diff
@@ -449,6 +449,15 @@ impl Reader {
         }
     }
 
+    /// Get the Reader as a detailed JSON string
+    /// This just calls to_
```
scouten-adobe: Incomplete sentence?

magaylor: This is the same formatting that `pub fn json(&self) -> String` uses (immediately above this).

I can edit both comments if you'd like

sdk/src/reader.rs
```diff
@@ -995,6 +1004,19 @@ pub mod tests {
         Ok(())
     }
 
+    #[test]
+    #[cfg(feature = "file_io")]
```
scouten-adobe, line 1008: @magaylor this might also need an exclusion for WASI as per the test failure that just occurred. Should be other examples in this file.

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -68,7 +68,7 @@ pub fn verify_time_stamp(
 ) -> Result<TstInfo, TimeStampError> {
     // Get the signed data frorm the timestamp data
     let Ok(Some(sd)) = signed_data_from_time_stamp_response(ts) else {
-        log_item!("", "count not parse timestamp data", "verify_time_stamp")
```
scouten-adobe, line 71: Thanks for the proofreading!

sdk/src/reader.rs
```diff
@@ -995,6 +1004,19 @@ pub mod tests {
...
+    /// Test that the reader can validate a file with nested assertion errors
+    fn test_reader_detailed_json() -> Result<()> {
+        let reader = Reader::from_file("tests/fixtures/CACAE-uri-CA.jpg")?;
+        let json_length = reader.json().len();
+        let detailed_json_length = reader.detailed_json().len();
+        assert!(json_length < detailed_json_length); // Detailed JSON should contain more information
+        #[cfg(target_os = "wasi")]
+        crate::utils::io_utils::wasm_remove_dir_all(temp_dir)?;
```
gpeacock: There's no temp_dir here so this will fail under wasi

sdk/src/reader.rs
```diff
@@ -995,6 +1004,19 @@ pub mod tests {
...
     }
 
+    #[test]
+    #[cfg(feature = "file_io")]
+    /// Test that the reader can validate a file with nested assertion errors
+    fn test_reader_detailed_json() -> Result<()> {
+        let reader = Reader::from_file("tests/fixtures/CACAE-uri-CA.jpg")?;
+        let json_length = reader.json().len();
+        let detailed_json_length = reader.detailed_json().len();
+        assert!(json_length < detailed_json_length); // Detailed JSON should contain more information
```
gpeacock: probably better to check for the existence of a field that only exists in a detailed report here.

sdk/src/reader.rs
magaylor: Can you link me to the test run for my own knowledge - I ran `make test` on my localhost but that likely isnt enough.

magaylor: oh nevermind, I found the tests linked in this PR

sdk/src/reader.rs
```diff
@@ -441,14 +441,23 @@ impl Reader {
...
     pub fn json(&self) -> String {
         match self.to_json_formatted() {
             Ok(value) => serde_json::to_string_pretty(&value).unwrap_or_default(),
             Err(_) => "{}".to_string(),
         }
     }
 
+    /// Get the Reader as a detailed JSON string
+    /// This just calls to_json_detailed_formatted
+    pub fn detailed_json(&self) -> String {
```
gpeacock, line 454: I'm ok with adding this, but note that you can get the same result with the debug output from reader
println!("{:?}", reader)

magaylor: How would I do something like that using a binding API? I dont see a way through c_api.rs to currently do that, and that was why I wanted to make this change

sdk/src/reader.rs
magaylor: I believe this was fixed in a previous commit - if that is the case, can you resolve this. If not let me know how else I can improve this function to run under WASI.

sdk/src/reader.rs
gpeacock: Yes we still need it exposed to C either way, but we didn't have to add it to Reader in the Rust code. I think it is probably clearer to have the separate function though.

gpeacock (approved):

PR #1423, opened 2025-09-16
ci: use all `rust_native_crypto` features during benchmarks

ok-nick:
Our benchmarks only enable `rust_native_crypto` which is inconsistent with our other `rust_native_crypto` CI jobs. The intended outcome is to enable all features BUT `openssl`.


gpeacock:
We decided that we should just remove the no_interleaved_io feature

PR #1425, opened 2025-09-16
feat: Expose add_action from the Builder at the C level

tmathern:
- Expose add_action at the C level, with expected JSON string describing the action as parameter.

## Checklist

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -89,6 +89,12 @@ struct CliArgs {
     #[clap(short, long)]
     remote: Option<String>,
 
+    /// Path to a binary .c2pa manifest to use for validation against the input asset.
+    ///
+    /// This field will override the input asset's embedded or remote manifest.
+    #[clap(long)]
+    external_manifest: Option<PathBuf>,
+
     /// Generate a sidecar (.c2pa) manifest
     #[clap(short, long)]
     sidecar: bool,
@@ -586,6 +592,24 @@ fn validate_cawg(reader: &mut Reader) -> Result<()> {
     }
 }
 
+fn reader_from_args(args: &CliArgs) -> Result<Reader> {
+    if let Some(external_manifest) = &args.external_manifest {
+        let c2pa_data = fs::read(external_manifest)?;
+        let format = match c2pa::format_from_path(&args.path) {
+            Some(format) => format,
+            None => {
+                bail!("Format for {:?} is unrecognized", args.path);
+            }
+        };
+        Ok(
+            Reader::from_manifest_data_and_stream(&c2pa_data, &format, File::open(&args.path)?)
+                .map_err(special_errs)?,
+        )
+    } else {
+        Ok(Reader::from_file(&args.path).map_err(special_errs)?)
+    }
+}
+
 fn main() -> Result<()> {
     let args = CliArgs::parse();
 
@@ -869,7 +893,7 @@ fn main() -> Result<()> {
             Ingredient::from_file(&args.path).map_err(special_errs)?
         )
     } else if args.detailed {
-        let mut reader = Reader::from_file(&args.path).map_err(special_errs)?;
+        let mut reader = reader_from_args(&args)?;
         validate_cawg(&mut reader)?;
         println!("{reader:#?}");
     } else if let Some(Commands::Fragment {
@@ -887,7 +911,7 @@ fn main() -> Result<()> {
             println!("{} Init manifests validated", stores.len());
         }
     } else {
-        let mut reader = Reader::from_file(&args.path).map_err(special_errs)?;
+        let mut reader = reader_from_args(&args)?;
         validate_cawg(&mut reader)?;
         println!("{reader}");
     }


c2pa_c_ffi/src/c_api.rs
```diff
@@ -895,6 +895,86 @@ pub unsafe extern "C" fn c2pa_builder_add_ingredient_from_stream(
...
+///
+/// # Errors
+/// Returns -1 if there were errors, otherwise returns 0.
+/// The error string can be retrieved by calling c2pa_error.
+///
+/// # Safety
+/// Reads from NULL-terminated C strings.
+///
+/// # Example
+/// ```C
```
tmathern, line 912: Should I keep this example here? Or is it too long a comment/description?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -895,6 +895,86 @@ pub unsafe extern "C" fn c2pa_builder_add_ingredient_from_stream(
...
+///     "        \"key1\": \"value1\",\n"
+///     "        \"key2\": \"value2\"\n"
+///     "    }\n"
+/// "}";
+///
+/// int result = c2pa_builder_add_action(builder, action_json);
+/// ```
+///
+/// This creates a manifest with an actions assertion
+/// ontaining the added action (excerpt of the full manifest):
```
ok-nick:
```suggestion
/// containing the added action (excerpt of the full manifest):
```

c2pa_c_ffi/src/c_api.rs
ok-nick: I think it should stay, this is a good example.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -895,6 +895,86 @@ pub unsafe extern "C" fn c2pa_builder_add_ingredient_from_stream(
...
 
+/// Adds an action to the manifest the Builder is constructing.
+///
+/// # Parameters
+/// * builder_ptr: pointer to a Builder.
+/// * action_json: JSON string containing the action data.
+///
+/// # Errors
+/// Returns -1 if there were errors, otherwise returns 0.
+/// The error string can be retrieved by calling c2pa_error.
```
ok-nick:
```suggestion
/// The error string can be retrieved by calling [c2pa_error].
```
Just to tell the docs/lsp that it can be followed to a function.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -895,6 +895,86 @@ pub unsafe extern "C" fn c2pa_builder_add_ingredient_from_stream(
...
+/// const char* manifest_def = "{}";
+/// C2paBuilder* builder = c2pa_builder_from_json(manifest_def);
+///
+/// const char* action_json = "{\n"
+///     "    \"action\": \"com.example.test-action\",\n"
+///     "    \"parameters\": {\n"
+///     "        \"key1\": \"value1\",\n"
+///     "        \"key2\": \"value2\"\n"
+///     "    }\n"
+/// "}";
```
ok-nick, line 922: I think we can remove the `\n` and it would still work, just for clarity in the example.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -895,6 +895,86 @@ pub unsafe extern "C" fn c2pa_builder_add_ingredient_from_stream(
...
+    // Parse the JSON into a serde Value to use with the Builder
+    let action_value: serde_json::Value = match serde_json::from_str(&action_json) {
+        Ok(value) => value,
+        Err(err) => {
+            Error::from_c2pa_error(c2pa::Error::JsonError(err)).set_last();
+            return -1;
+        }
+    };
+
+    match builder.add_action(action_value) {
```
ok-nick, line 968: You should be able to simplify this with `builder.add_action(action_json)`. `Builder::add_action` handles converting the string to a `serde_json::Value`.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1468,6 +1548,83 @@ mod tests {
...
+        let signer_info = C2paSignerInfo {
+            alg: alg.as_ptr(),
+            sign_cert: sign_cert.as_ptr(),
+            private_key: private_key.as_ptr(),
+            ta_url: std::ptr::null(),
+        };
+        let signer = unsafe { c2pa_signer_from_info(&signer_info) };
+
+        assert!(!signer.is_null());
+        let manifest_def = CString::new("{}").unwrap();
```
ok-nick, line 1570: You can also do `c"{}"`, Rust now supports raw C string literals (can also do this in other places).

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1468,6 +1548,83 @@ mod tests {
...
+        let alg = CString::new("Ed25519").unwrap();
+        let sign_cert = CString::new(certs).unwrap();
+        let private_key = CString::new(private_key).unwrap();
+        let signer_info = C2paSignerInfo {
+            alg: alg.as_ptr(),
+            sign_cert: sign_cert.as_ptr(),
+            private_key: private_key.as_ptr(),
+            ta_url: std::ptr::null(),
+        };
+        let signer = unsafe { c2pa_signer_from_info(&signer_info) };
```
ok-nick, line 1567: Hm, I wonder if it would make sense to just construct the `Builder` in Rust and we can make a pointer when calling `c2pa_builder_add_action`. I think this would simplify the test case a lot and focus more on the specific C function we are testing.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1468,6 +1548,83 @@ mod tests {
...
+
+        let reader = unsafe { c2pa_reader_from_stream(format.as_ptr(), &mut read_stream) };
+        assert!(!reader.is_null());
+
+        let json = unsafe { c2pa_reader_json(reader) };
+        assert!(!json.is_null());
+        let json_str = unsafe { CString::from_raw(json) };
+        let json_content = json_str.to_str().unwrap();
+
+        assert!(json_content.contains("manifest"));
```
ok-nick, line 1615: We should also assert the action JSON that was added or at least what we can of it.

c2pa_c_ffi/src/c_api.rs
tmathern: I used the stuff that keeps the code file consistent, because I don't want to change everything in that PR

c2pa_c_ffi/src/c_api.rs
tmathern: This is a test case. It's fine to do it that way. Keeps things isolated too, and shows full scenario.

c2pa_c_ffi/src/c_api.rs
tmathern: Actually, no. Rust panics. I suppose it's due to having to cross language barriers.

ok-nick: I think this issue stems more from how we deal with serde_json and strings in Rust for `Builder::add_action`. I've encountered a similar issue before with serde_json values and having to escape the string. I'm going to reopen this because I think we should talk about changes to the Rust API.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -895,6 +895,86 @@ pub unsafe extern "C" fn c2pa_builder_add_ingredient_from_stream(
...
+///     }
+///   }
+/// ]
+/// ```
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_builder_add_action(
+    builder_ptr: *mut C2paBuilder,
+    action_json: *const c_char,
+) -> c_int {
+    check_or_return_int!(builder_ptr);
```
gpeacock: guard_boxed_int!() includes the null check so you don't need to do this check first.

gpeacock (approved):
One line you can remove because we don't need it, but otherwise this looks good!

PR #1426, opened 2025-09-16
feat: Expose add_assertion_json from the Builder at the C level

tmathern:
- Expose add_assertion_json at the C level, with label and json content as parameters of the call.

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -1620,10 +1620,10 @@ mod tests {
         TestC2paStream::drop_c_stream(read_stream);
         unsafe {
             c2pa_manifest_bytes_free(manifest_bytes_ptr);
+            c2pa_builder_free(builder);
+            c2pa_signer_free(signer);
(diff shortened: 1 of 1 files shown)


c2pa_c_ffi/src/c_api.rs
```diff
@@ -955,18 +974,96 @@ pub unsafe extern "C" fn c2pa_builder_add_action(
...
+/// * data_json: The data for the assertion as a JSON string (null-terminated C string).
+///
+/// # Errors
+/// Returns -1 if there were errors, otherwise returns 0.
+/// The error string can be retrieved by calling c2pa_error.
+///
+/// # Safety
+/// Reads from NULL-terminated C strings.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_builder_add_assertion(
```
tmathern, line 1010: With great power comes great responsibility.
Should we even expose that here? I think it makes sense to expose it since it's a layer deep down, and defer the responsibility of the call parameters (especially the values passed in the data) to the caller.

PR #1432, opened 2025-09-18
feat: Enables generating created assertions and getting the created state

gpeacock:
The 2.2 spec includes replaces the assertions array with two separate assertions arrays, one for gathered_assertions and another for created_assertions. The idea is that signer is responsible for the created assertions and others are considered gathered. 
In order to keep the api as consistent as possible, we only have one assertions array, but individual assertions can be set to a created state. If the created flag is not set, the assertion is considered gathered.
The API defaults all v2 claim assertions as gathered except for hard bindings which are generated by the API.
Hard bindings are only reported in the detailed report where you will see them in a created_assertions, vs gathered_assertions array.
When passing in an assertion via JSON you can add a created: = true field to specify that an individual assertion should be considered created. The ManifestAssertion in the Reader will also show the created state.

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -3217,7 +3217,7 @@ mod tests {
         //std::fs::write("redaction2.jpg", output.get_ref()).unwrap();
 
         let reader = Reader::from_stream("jpeg", &mut output).expect("from_bytes");
-        println!("{reader}");
+        //println!("{reader}");
         assert_eq!(reader.validation_state(), ValidationState::Trusted);
         let m = reader.active_manifest().unwrap();
         assert_eq!(m.ingredients().len(), 1);


sdk/examples/api.rs
```diff
@@ -67,7 +67,7 @@ fn manifest_def(title: &str, format: &str) -> String {
     }).to_string()
 }
 
-/// This example demonstrates how to use the new v2 API to create a manifest store
+/// This example demonstrates how to use the API to create a manifest store
```
ok-nick, line 70: We should rename the example file too.

sdk/src/assertions/actions.rs
```diff
@@ -842,8 +842,14 @@ impl Actions {
     }
 
     /// Adds an [`Action`] to this assertion's list of actions.
+    /// OPENED and CREATED actions are inserted at the beginning of the list.
```
ok-nick, line 836: Maybe clarify here that this is required by the v2 spec.

sdk/src/assertions/labels.rs
```diff
@@ -229,14 +236,77 @@ pub static METADATA_LABEL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
+/// entity = entity-component *( "." entity-component )
+/// entity-component = 1( DIGIT / ALPHA ) *( DIGIT / ALPHA / "-" / "_" )
+/// label = 1*( "." label-component )
+/// label-component = 1( DIGIT / ALPHA ) *( DIGIT / ALPHA / "-" / "_" )
+/// version = ".v" 1*DIGIT
+/// instance = "__" 1*DIGIT
+/// ```
+pub fn parse_label(label: &str) -> (&str, usize, usize) {
+    // First, extract instance if present
+    let (without_instance, instance) = if let Some(pos) = label.find("__") {
```
ok-nick: The `label-component` is possible to have multiple underscores, we should check for the last occurrence of `__` (via `label.rfind()`).

sdk/src/assertions/labels.rs
```diff
@@ -229,14 +236,77 @@ pub static METADATA_LABEL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
+        let instance = instance_str.parse::<usize>().unwrap_or(0);
+        (&label[..pos], instance)
+    } else {
+        (label, 0)
+    };
+
+    // Then, extract version if present
+    let components: Vec<&str> = without_instance.split('.').collect();
+    if let Some(last) = components.last() {
+        if last.len() > 1 && last.starts_with('v') {
```
ok-nick: And similarly here, we can do a reverse string search with `rfind`.

sdk/src/settings/builder.rs
```diff
@@ -398,18 +399,26 @@ pub(crate) struct BuilderSettings {
...
 
-    // Whether or not existing OCSP responses should be overridden by new values.
+    /// Whether or not existing OCSP responses should be overridden by new values.
     pub certificate_status_should_override: Option<bool>,
+
+    /// Sets a default intent for builders, can be Create, Edit, or Update.
+    pub intent: Option<BuilderIntent>,
+
+    /// When adding assertions in this list, they will be marked as created. The default is gathered.
+    pub created_assertions: Option<Vec<String>>,
```
ok-nick: I was a little confused here until I read `Claim::claim_assertion_type` that this is a list of labels. I think it would be more clear as `created_assertion_labels`.

We may also want to clarify in the docs that it MUST be an exact match, including the assertion version.

I'm also concerned that it's very easy to input an incorrect label since we don't do any validation. I suppose the label can be anything but this can be a very easy mistake.

sdk/src/settings/builder.rs
```diff
@@ -398,18 +399,26 @@ pub(crate) struct BuilderSettings {
...
-    // Certificate statuses will be fetched for either all the manifest labels, or just the active manifest.
+    /// Certificate statuses will be fetched for either all the manifest labels, or just the active manifest.
     pub certificate_status_fetch: Option<OcspFetch>,
 
-    // Whether or not existing OCSP responses should be overridden by new values.
+    /// Whether or not existing OCSP responses should be overridden by new values.
     pub certificate_status_should_override: Option<bool>,
+
+    /// Sets a default intent for builders, can be Create, Edit, or Update.
+    pub intent: Option<BuilderIntent>,
```
ok-nick, line 419: Are we removing the `auto_created_action`/`auto_opened_action` in favor of `intent` in this PR as well?

sdk/src/claim.rs
```diff
@@ -1356,77 +1383,61 @@ impl Claim {
...
-                ClaimAssertionType::Gathered
-            }
-        } else {
-            ClaimAssertionType::V1
-        };
-        let ca = ClaimAssertion::new(assertion.clone(), instance, &hash, self.alg(), salt, typ);
+        // Add salt to all redactable assertions  (todo: should we just salt them all?)
+        if !labels::NON_REDACTABLE_LABELS.contains(&base_label) {
+            c2pa_assertion.add_salt(salt.clone());
+        }
```
ok-nick: I agree that we should salt them all if the spec allows it. It would be much safer in the scenario a matching hash is discovered (however unlikely that may be).

sdk/src/claim.rs
```diff
@@ -1356,77 +1383,61 @@ impl Claim {
...
+            ClaimAssertionType::Created => {
+                self.created_assertions.push(c2pa_assertion.clone());
+            }
+            ClaimAssertionType::Gathered => match &mut self.gathered_assertions {
+                Some(ga) => ga.push(c2pa_assertion.clone()),
+                None => {
+                    let new_ga = [c2pa_assertion.clone()];
+                    self.gathered_assertions = Some(new_ga.to_vec());
+                }
+            },
```
ok-nick: You may be interested in `Option::get_or_insert_default` here.

sdk/src/manifest_assertion.rs
```diff
@@ -21,8 +23,10 @@ pub enum ManifestAssertionKind {
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[cfg_attr(feature = "json_schema", derive(JsonSchema))]
 #[serde(untagged)]
-enum ManifestData {
-    Json(Value),     // { label: String, instance: usize, data: Value },
+pub(crate) enum ManifestData {
+    Json(Value), // { label: String, instance: usize, data: Value },
+    #[cfg_attr(feature = "json_schema", schemars(skip))]
+    Cbor(CborValue),
```
ok-nick, line 29: We should skip the inner `CborValue` here rather than the `Cbor` variant so that other SDKs using the JSON schema for types know that a cbor variant is possible.

sdk/src/store.rs
```diff
@@ -1101,14 +1100,7 @@ impl Store {
...
     fn check_label_version(base_version_label: &str, desired_version_label: &str) -> bool {
-        if let Some(desired_version) = labels::version(desired_version_label) {
-            if let Some(base_version) = labels::version(base_version_label) {
-                if desired_version > base_version {
-                    return false;
-                }
-            }
-        }
-        true
+        labels::version(desired_version_label) <= labels::version(base_version_label)
```
ok-nick, line 1142: We may want to consider `labels::version` returning an `Option` here in case they aren't comparable.

sdk/src/assertions/labels.rs
gpeacock: nice catch

sdk/src/assertions/labels.rs
gpeacock: I updated this to avoid a potential miss with something like:
"c2pa.ingredient.value"

sdk/src/store.rs
gpeacock: labels::version always returns an numeric value since lack of a version is treated as the default version via the spec. We also insure we only identify a version if it is in the correct format. So I think this is ok.

sdk/src/settings/builder.rs
gpeacock: Let's do that in another PR to keep this from growing bigger.

sdk/src/settings/builder.rs
gpeacock: I changed this to created_assertion labels.

gpeacock: This does not require an exact match. It matches on the base of the label, so any version or instance will match.
In theory we could check if it starts with c2pa. that is is one of the known c2pa assertions. We could add an issue to follow up on that.

I also added a comment explaining that.

sdk/src/assertion.rs
```diff
@@ -103,21 +117,46 @@ pub trait AssertionBase
...
+        Self::CREATED
+    }
+
+    /// Returns the assertion kind for this assertion
+    fn kind(&self) -> AssertionKind {
+        Self::KIND
+    }
+
+    /// Returns the content type for this assertion based on its kind
+    fn content_type(&self) -> &str {
```
mauricefisher64: What is this used for?

sdk/src/assertions/actions.rs
```diff
@@ -833,8 +833,14 @@ impl Actions {
...
+    /// Note, this does not check for duplicates since it does not return errors.
     pub fn add_action(mut self, action: Action) -> Self {
-        self.actions.push(action);
+        if action.action() == c2pa_action::OPENED || action.action() == c2pa_action::CREATED {
```
mauricefisher64, line 840: Do you want to check that there is only one as well?

tmathern: +1, right now double created can happen. Maybe we should check and warn?

sdk/src/builder.rs
```diff
@@ -3170,7 +3217,7 @@ mod tests {
         //std::fs::write("redaction2.jpg", output.get_ref()).unwrap();
 
         let reader = Reader::from_stream("jpeg", &mut output).expect("from_bytes");
-        //println!("{reader}");
+        println!("{reader}");
```
tmathern: Missed to comment it out again?

tmathern (approved):

sdk/src/assertions/actions.rs
tmathern: I think this may get addressed in another PR!

PR #1438, opened 2025-09-21
fix: Add checks at C API level

tmathern:
- Add size checks in C API when creating a buffer/array from its parts
- ~Add a loose string format check~
- ~Add null checks in context_stream.rs (needed as through Python calls in rare cases this could cause panics when context was not set and checked properly)~

## Checklist

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -2161,20 +2161,21 @@ impl Claim {
                     || action.action() == c2pa_action::REMOVED
                 {
                     // 2.b.i must have parameters
-                    let params = action.parameters().ok_or_else(||
-                        // 2.e.i
+                    let Some(params) = action.parameters() else {
                         log_item!(
                             label.clone(),
                             "opened, placed and removed items must have parameters",
                             "verify_actions"
                         )
                         .validation_status(validation_status::ASSERTION_ACTION_INGREDIENT_MISMATCH)
-                        .failure_as_err(
+                        .failure(
                             validation_log,
                             Error::ValidationRule(
                                 "opened, placed and removed items must have parameters".into(),
                             ),
-                        ))?;
+                        )?;
+                        continue; // Skip the parameter-dependent checks below
+                    };
 
                     // 2.b.ii must have ingredient or ingredients param
                     if params.ingredients.is_none() && params.ingredient.is_none() {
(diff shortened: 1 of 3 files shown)


c2pa_c_ffi/src/c2pa_stream.rs
```diff
@@ -108,6 +108,17 @@ impl C2paStream {
         let context_ptr = std::mem::replace(&mut self.context, std::ptr::null_mut());
         unsafe { Box::from_raw(context_ptr) }
     }
+
+    /// Checks the stream context.
+    fn check_context(&self) -> std::io::Result<()> {
```
gpeacock: I don't think I have a rule that all stream implementations have to set a non-null context. They don't technically need to use a context, but in practice they all might need to. If we are ok with insisting that they must supply some kind of non-null context, this is ok.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -17,6 +17,93 @@ use std::{
...
+/// * `format` - The format string to validate (mimetype or extension)
+///
+/// # Returns
+/// * `Ok(())` if the format string is valid
+/// * `Err(Error)` if the format string is invalid
+fn validate_format_string(format: &str) -> Result<(), Error> {
+    for &byte in format.as_bytes() {
+        if byte == 0 {
+            return Err(Error::Other(
+                "Format string contains null bytes".to_string(),
```
gpeacock: can from_cstr return a string with null bytes? if it can, then we may need this check there, if not, we don't need it here.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -17,6 +17,93 @@ use std::{
...
+            ));
+        }
+        if !byte.is_ascii()
```
gpeacock: arguably the rest of this check could be moved into the core rust code. Was there some known bug this is fixing?
We could add a deeper check since if this is not an extension we know about, it should be a mime_type. We really just need mime_types internally.

tmathern: Oh I thought I could be a bit restrictive here. But I can leave that check open aka remove it from here right now.

c2pa_c_ffi/src/c_api.rs
tmathern: I managed to have strings containing only null bytes, albeit in convoluted test cases. But since I removed that check now too anyway...

c2pa_c_ffi/src/c2pa_stream.rs
tmathern: Removing for now then.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -570,7 +626,16 @@ pub unsafe extern "C" fn c2pa_reader_from_manifest_data_and_stream(
 ) -> *mut C2paReader {
     check_or_return_null!(manifest_data);
     let format = from_cstr_or_return_null!(format);
-    let manifest_bytes = std::slice::from_raw_parts(manifest_data, manifest_size);
+
+    // Safe bounds validation for manifest data
+    let manifest_bytes =
```
gpeacock, line 631: I'd make this into a macro, but it it will work this way too.

PR #1440, opened 2025-09-22
fix: only enable "asm" feature on aarch64 macos

dyro:
This PR disables the “asm” feature of sha2 for AArch64 Windows. Windows aarch64 builds are failing internally because of this feature.

## Checklist

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -176,7 +176,7 @@ x509-parser = "0.16.0"
 zeroize = { version = "1.8", features = ["zeroize_derive"] }
 zip = { version = "3.0.0", default-features = false }
 
-# Use the asm feature of sha2 on aarch64 for better performance. This nearly
+# Use the asm feature of sha2 on aarch64 macOS for better performance. This nearly
 # halves hashing time for large assets (> 50gb mp4, for example).
 #
 # Note: This feature has been removed in later versions of sha2.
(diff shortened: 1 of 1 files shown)


sdk/Cargo.toml
```diff
@@ -176,18 +176,18 @@ x509-parser = "0.16.0"
...
 #
 # Note: This feature has been removed in later versions of sha2.
 # If upgrading, ensure sha2 has implemented equivalent optimizations for aarch64.
 #
 # See: https://github.com/RustCrypto/hashes/pull/542 and
 # https://github.com/RustCrypto/hashes/tree/master/sha2#backends (which hasn't been released yet).
-[target.'cfg(target_arch = "aarch64")'.dependencies]
-sha2 = { version = "0.10.6", features = ["asm"] }
-[target.'cfg(not(target_arch = "aarch64"))'.dependencies]
 sha2 = "0.10.6"
```
dyro, line 187: I hadn’t realized this before, but it makes sense: dependencies in Cargo.toml are additive. You can specify the same dependency, like sha2, multiple times with different features or platform-specific settings. Cargo will combine all these requirements. For example, on macOS, the sha2 crate will include the asm feature, but on other platforms, it won’t. So instead of doing the if / else approach like I was doing before, you can do the below to get the same effect.

PR #1441, opened 2025-09-22
fix: [CAI-9586] Some validation errors are thrown as exceptions

gpeacock:
Various validation checks have been throwing as errors when we can report them via validation status. This will allow them to be properly reported in c2patool and verify applications.

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -3037,7 +3037,7 @@ impl Claim {
             let assertion_absolute_uri = if assertion.is_relative_url() {
                 to_absolute_uri(claim.label(), &assertion.url())
             } else {
-                // match sure the assertion points to this assertion store
+                // make sure the assertion points to this assertion store
                 let Some(assertion_manifest) = manifest_label_from_uri(&assertion.url()) else {
                     log_item!(
                         assertion.url(),


tmathern (approved):

sdk/src/claim.rs
```diff
@@ -3035,21 +3038,21 @@ impl Claim {
                 to_absolute_uri(claim.label(), &assertion.url())
             } else {
                 // match sure the assertion points to this assertion store
```
mauricefisher64: can we fix the typo while you are here "match" => "make"

sdk/src/crypto/cose/verifier.rs
```diff
@@ -104,14 +104,20 @@ impl Verifier<'_> {
         }
 
         if _sync {
-            self.verify_profile(&sign1, tst_info, validation_log)?;
-            self.verify_trust(&sign1, tst_info, validation_log)?;
+            self.verify_profile(&sign1, tst_info, validation_log)
```
mauricefisher64, line 107: Why unravel these?  There are really a unit of tests that are broken into two functions.

scouten-adobe: I think it's so that he can ignore the errors rather than using the `?` operator to fail immediately.

PR #1444, opened 2025-09-24
feat: Make settings not thread-local

scouten-adobe:
Settings are now snapshotted at public API surfaces and passed through all layers below.

Note related work from @ok-nick in #1447 and further work likely in #1454.

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -174,9 +174,9 @@ impl Reader {
         }?;
 
         if _sync {
-            Self::from_store(store, &mut validation_log)
+            Self::from_store(store, &mut validation_log, &settings)
         } else {
-            Self::from_store_async(store, &mut validation_log).await
+            Self::from_store_async(store, &mut validation_log, &settings).await
         }
     }
 


sdk/src/assertions/bmff_hash.rs
```diff
@@ -1044,7 +1044,11 @@ impl BmffHash {
         local_id: usize,
         unique_id: Option<usize>,
     ) -> crate::Result<()> {
-        let max_proofs = get_settings_value::<usize>("core.merkle_tree_max_proofs")?;
+        let settings = crate::settings::get_settings().unwrap_or_default();
```
tmathern: What is the default in this case? The default from the config file?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1544,9 +1545,16 @@ impl CAIReader for BmffIO {
             if let Some(update_bytes) = c2pa_boxes.update_bytes {
                 let mut validation_log = StatusTracker::default();
 
+                let settings = Settings::default();
+                // ^^ TO REVIEW: Is there a reason we'd need non-default settings here?
```
tmathern: Is there actually a setting already that changes things in here? If so, it could be useful to investigate why that setting came to be - might answer the question as side-effect.

sdk/src/builder.rs
```diff
@@ -597,13 +593,64 @@ impl Builder {
...
+    /// # Arguments
+    /// * `ingredient_json` - A JSON string representing the [`Ingredient`].  This ingredient is merged  with the ingredient specified in the `stream` argument, and these values take precedence.
+    /// * `format` - The format of the [`Ingredient`].
+    /// * `stream` - A stream from which to read the [`Ingredient`].  This ingredient is merged  with the ingredient specified in the `ingredient_json` argument, whose values take precedence.  You can specify values here that are not specified in `ingredient_json`.
+    /// # Returns
+    /// * A mutable reference to the [`Ingredient`].
+    /// # Errors
+    /// * Returns an [`Error`] if the [`Ingredient`] is not valid
+    #[async_generic]
+    pub fn add_ingredient_from_stream_with_settings<'a, T, R>(
```
tmathern: I know it's a new API, but I like the flexibility it introduces. "Use at your own risk with settings".

sdk/src/builder.rs
```diff
@@ -1138,24 +1179,36 @@ impl Builder {
...
-                        "builder.auto_opened_action.source_type",
-                    );
+                    let source_type = &settings.builder.actions.auto_opened_action.source_type;
+                    // TO DISCUSS BEFORE MERGE: Previous code was this:
+                    // let source_type =
+                    //      settings::get_settings_value::<Option<DigitalSourceType>>(
+                    //          "builder.auto_opened_action.source_type",
+                    //      );
+                    //
+                    // ... which omits the ".actions" node after "builder"
```
tmathern: My 2 cents: It would be good to keep the settings structured. The missing actions node feels like a typo to me?

sdk/src/assertions/bmff_hash.rs
scouten-adobe: `settings::get_settings()` is parsing the config file. `default` is our own built-in default settings that we fall back to if there is no config file or we can't parse it.

sdk/src/builder.rs
scouten-adobe: I agree. What I don't know is how to repair the test in question. @gpeacock?

sdk/tests/test_builder.rs
```diff
@@ -24,6 +24,7 @@ use common::compare_stream_to_known_good;
 use common::test_signer;
 
 #[test]
+#[ignore = "See discussion in builder.rs near line 1190"]
```
scouten-adobe: @gpeacock please review this before we merge.

sdk/src/reader.rs
```diff
@@ -125,30 +125,77 @@ impl Reader {
...
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn from_stream(format: &str, mut stream: impl Read + Seek + Send) -> Result<Reader> {
-        let verify = get_settings_value::<bool>("verify.verify_after_reading")?; // defaults to true
+    pub fn from_stream_with_settings(
+        format: &str,
+        mut stream: impl Read + Seek + Send,
+        settings: &Settings,
+    ) -> Result<Reader> {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Make this the official API?
```
ok-nick: I'd defer this work to #1465.

sdk/src/reader.rs
```diff
@@ -125,30 +125,77 @@ impl Reader {
...
     #[cfg(target_arch = "wasm32")]
-    pub fn from_stream(format: &str, mut stream: impl Read + Seek) -> Result<Reader> {
-        let verify = get_settings_value::<bool>("verify.verify_after_reading")?; // defaults to true
+    pub fn from_stream_with_settings(
+        format: &str,
+        mut stream: impl Read + Seek,
+        settings: &Settings,
+    ) -> Result<Reader> {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Make this the official API?
```
ok-nick: I'd defer this work to #1465.

sdk/src/reader.rs
```diff
@@ -125,30 +125,77 @@ impl Reader {
...
+    pub fn from_stream_with_settings(
+        format: &str,
+        mut stream: impl Read + Seek + Send,
+        settings: &Settings,
+    ) -> Result<Reader> {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Make this the official API?
+
+        // TO DO BEFORE MERGE: Passing `verify` may be redundant now that we're
+        // passing settings.
```
ok-nick: TODO: internal

sdk/src/reader.rs
```diff
@@ -125,30 +125,77 @@ impl Reader {
...
+    pub fn from_stream_with_settings(
+        format: &str,
+        mut stream: impl Read + Seek,
+        settings: &Settings,
+    ) -> Result<Reader> {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Make this the official API?
+
+        // TO DO BEFORE MERGE: Passing `verify` may be redundant now that we're
+        // passing settings.
```
ok-nick: TODO: internal

sdk/src/reader.rs
```diff
@@ -226,15 +273,19 @@ impl Reader {
...
-    #[async_generic()]
+    #[async_generic]
     pub fn from_manifest_data_and_stream(
         c2pa_data: &[u8],
         format: &str,
         stream: impl Read + Seek + Send,
     ) -> Result<Reader> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: I'd defer this work to #1465.

sdk/src/reader.rs
```diff
@@ -269,12 +322,16 @@ impl Reader {
...
-    #[async_generic()]
+    #[async_generic]
     pub fn from_fragment(
         format: &str,
         mut stream: impl Read + Seek + Send,
         mut fragment: impl Read + Seek + Send,
     ) -> Result<Self> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: I'd defer this work to #1465.

sdk/src/reader.rs
```diff
@@ -305,8 +364,11 @@ impl Reader {
         path: P,
         fragments: &Vec<std::path::PathBuf>,
     ) -> Result<Reader> {
-        let verify = get_settings_value::<bool>("verify.verify_after_reading")?; // defaults to true
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: I'd defer this work to #1465.

sdk/src/reader.rs
```diff
@@ -508,11 +571,15 @@ impl Reader {
 
     /// Get the [`ValidationState`] of the manifest store.
     pub fn validation_state(&self) -> ValidationState {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: We should store settings in the `Reader`, or perhaps just the `verify_trust` setting?

sdk/src/store.rs
```diff
@@ -546,67 +554,66 @@ impl Store {
...
         }
     }
 
     /// Retrieves all manifest labels that need to fetch ocsp responses.
-    pub fn get_manifest_labels_for_ocsp(&self) -> Vec<String> {
-        let labels = match crate::settings::get_settings_value::<OcspFetch>(
-            "builder.certificate_status_fetch",
-        ) {
-            Ok(ocsp_fetch) => match ocsp_fetch {
+    pub fn get_manifest_labels_for_ocsp(&self, settings: &Settings) -> Vec<String> {
```
ok-nick, line 648: We should find places like this where we don't necessarily need to pass in the entire `Settings`. I suppose we can leave it up to another PR.

sdk/src/store.rs
```diff
@@ -546,67 +554,66 @@ impl Store {
...
+                            &self.ctp,
+                            None,
+                            &mut cose_log,
+                            &settings.verify,
+                        )
+                        .await
+                    };
+                    if let Err(err) = result {
+                        error!("Signature that was just generated does not validate: {err:#?}");
+                        return Err(err);
```
ok-nick, line 637: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399921598

sdk/src/ingredient.rs
```diff
@@ -1444,13 +1503,19 @@ impl Ingredient {
     where
         R: std::io::BufRead + std::io::Seek,
     {
-        let auto_thumbnail =
-            crate::settings::get_settings_value::<bool>("builder.thumbnail.enabled")?;
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: This wasn't intended to be part of the public API, we should be able to pass in `Settings` here.

sdk/src/ingredient.rs
```diff
@@ -1398,20 +1451,26 @@ impl Ingredient {
         format: &str,
         stream: &mut dyn CAIRead,
     ) -> Result<Self> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: I think we need a builder for `Ingredient` like in #1465 but there have been discussions about potentially removing this abstraction.

sdk/src/ingredient.rs
```diff
@@ -948,29 +969,60 @@ impl Ingredient {
...
+        Self::from_stream_async_with_settings(format, stream, &settings).await
+    }
+
+    pub(crate) async fn from_stream_async_with_settings(
+        format: &str,
+        stream: &mut dyn CAIRead,
+        settings: &Settings,
+    ) -> Result<Self> {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Make this the official signature?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399930565

sdk/src/ingredient.rs
```diff
@@ -948,29 +969,60 @@ impl Ingredient {
...
+        Self::from_stream_async_with_settings(format, &mut stream, settings).await
+    }
+
     /// Creates an `Ingredient` from a stream (async version).
     ///
     /// This does not set title or hash.
     /// Thumbnail will be set only if one can be retrieved from a previous valid manifest.
     pub async fn from_stream_async(format: &str, stream: &mut dyn CAIRead) -> Result<Self> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399930565

sdk/src/ingredient.rs
```diff
@@ -948,29 +969,60 @@ impl Ingredient {
...
+    ///
+    /// This does not set title or hash.
+    /// Thumbnail will be set only if one can be retrieved from a previous valid manifest.
+    pub async fn from_memory_async_with_settings(
+        format: &str,
+        buffer: &[u8],
+        settings: &Settings,
+    ) -> Result<Self> {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Replace `from_memory_async` with this?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399930565

sdk/src/ingredient.rs
```diff
@@ -823,9 +832,13 @@ impl Ingredient {
     /// This does not set title or hash.
     /// Thumbnail will be set only if one can be retrieved from a previous valid manifest.
     pub fn from_stream(format: &str, stream: &mut dyn CAIRead) -> Result<Self> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399930565

sdk/src/ingredient.rs
```diff
@@ -732,12 +733,20 @@ impl Ingredient {
         path: P,
         options: &dyn IngredientOptions,
     ) -> Result<Self> {
-        Self::from_file_impl(path.as_ref(), options)
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399930565

sdk/src/claim.rs
```diff
@@ -1895,7 +1901,10 @@ impl Claim {
         cert_check: bool,
         ctp: &CertificateTrustPolicy,
         validation_log: &mut StatusTracker,
+        settings: &Settings,
     ) -> Result<()> {
+        // TO DO BEFORE MERGE? Are cert_check and ctp redundant?
```
ok-nick: TODO: internal

sdk/src/claim.rs
```diff
@@ -1828,7 +1828,10 @@ impl Claim {
         cert_check: bool,
         ctp: &CertificateTrustPolicy,
         validation_log: &mut StatusTracker,
+        settings: &Settings,
     ) -> Result<()> {
+        // TO DO BEFORE MERGE? Are ctp and cert_check params redundant?
```
ok-nick: TODO: internal

sdk/src/builder.rs
```diff
@@ -1513,6 +1592,10 @@ impl Builder {
         fragment_paths: &Vec<std::path::PathBuf>,
         output_path: P,
     ) -> Result<()> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: I think we should store `Settings` in the `Builder` as a field.

sdk/src/builder.rs
```diff
@@ -1435,6 +1510,10 @@ impl Builder {
         R: Read + Seek + Send,
         W: Write + Read + Seek + Send,
     {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399936374

sdk/src/builder.rs
```diff
@@ -1394,13 +1463,19 @@ impl Builder {
         signer: &dyn Signer,
         format: &str,
     ) -> Result<Vec<u8>> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399936374

sdk/src/builder.rs
```diff
@@ -1364,12 +1427,18 @@ impl Builder {
         data_hash: &DataHash,
         format: &str,
     ) -> Result<Vec<u8>> {
-        let mut store = self.to_store()?;
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399936374

sdk/src/builder.rs
```diff
@@ -1322,6 +1381,10 @@ impl Builder {
         reserve_size: usize,
         format: &str,
     ) -> Result<Vec<u8>> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399936374

sdk/src/builder.rs
ok-nick: The source type is only necessary for the `c2pa.created` action. We should be able to safely remove this from the `test_settings.toml` that the test uses.

@gpeacock for confirmation

sdk/src/builder.rs
```diff
@@ -597,13 +593,64 @@ impl Builder {
...
+        format: &str,
+        stream: &mut R,
+        settings: &Settings,
+    ) -> Result<&'a mut Ingredient>
+    where
+        T: Into<String>,
+        R: Read + Seek + Send,
+    {
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Make this the official API?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399936374

sdk/src/builder.rs
```diff
@@ -597,13 +593,64 @@ impl Builder {
         T: Into<String>,
         R: Read + Seek + Send,
     {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399936374

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -290,6 +289,10 @@ impl IdentityAssertion {
         partial_claim: &PartialClaim,
         status_tracker: &mut StatusTracker,
     ) -> Result<serde_json::Value, ValidationError<String>> {
+        let settings = crate::settings::get_settings().unwrap_or_default();
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: TODO: internal

sdk/src/crypto/time_stamp/http_request.rs
```diff
@@ -55,11 +56,16 @@ pub fn default_rfc3161_request(
     let mut local_log = StatusTracker::default();
     let ctp = CertificateTrustPolicy::passthrough();
 
+    // TO REVIEW: I think in this case, we're doing structural validation of
+    // time stamps but not trust validation. If so, the below is correct. (I think.)
```
ok-nick: TODO: internal

sdk/src/crypto/cose/sigtst.rs
```diff
@@ -40,6 +41,9 @@ pub(crate) fn validate_cose_tst_info(
     ctp: &CertificateTrustPolicy,
     validation_log: &mut StatusTracker,
 ) -> Result<TstInfo, CoseError> {
+    let settings = crate::settings::get_settings().unwrap_or_default();
+    // TO DO BEFORE MERGE? Pass Settings in here?
```
ok-nick: TODO: internal

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1684,16 +1692,22 @@ impl CAIWriter for BmffIO {
...
-            let is_update = pc.update_manifest();
-            (Some(pc.clone()), is_update)
-        } else {
-            (None, false)
-        };
+
+        let settings = Settings::default();
+        // ^^ TO REVIEW: Is there a reason we'd need non-default settings here?
+        // If so, that means we hae to rework CAIReader and CAIWriter traits
+        // to take a &Settings arg and that is a pretty huge change.
```
ok-nick: I think `Settings` in the `Store` constructors are only used for configuring trust. In this scenario we are only using the `Store` to reconstruct the split manifest stores in BMFF, the trust is validated with a higher-level `Store`. It does feel wrong to create default settings and pass it in as a parameter here, but I don't think it'll be a problem for the time being.

cc @mauricefisher64 to confirm.

sdk/src/asset_handlers/bmff_io.rs
ok-nick: I don't think this will be a problem, see here:
https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399949181

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1044,7 +1044,11 @@ impl BmffHash {
...
+        // TO DO (https://github.com/contentauth/c2pa-rs/issues/1454):
+        // Add a Settings argument here?
```
ok-nick: TODO: internal

ok-nick (changes requested):
Leaving lots of reviews to track progress.

sdk/tests/test_builder.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2399941001

sdk/src/reader.rs
ok-nick: I think we can actually use the `validation_state` stored in the `Reader`. It seems it's always set on construction but we previously recomputed it because the thread-local settings may change.

sdk/src/ingredient.rs
ok-nick: I'm going to `pub(crate)` it, leaving this discussion open for further thoughts.

sdk/src/identity/identity_assertion/assertion.rs
ok-nick: I think here we should only pass in a `Trust` struct. Maybe we should also rename it to `TrustSettings`? However, I defer this work to a separate PR. Ideally we don't change public APIs in this one.

sdk/src/crypto/time_stamp/http_request.rs
ok-nick: Yes, following the code path we explicitly set `CertificateTrustPolicy::passthrough` which stops before trust checks occur.  A bigger question is if we need settings here at all versus passing `verify_trust` through the `passthrough` field of `CertificateTrustPolicy`.

sdk/src/crypto/cose/sigtst.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2407140373

sdk/src/reader.rs
ok-nick: For this one, I agree that passing `verify` is redundant, however, we can't distinguish in the `Store` between constructing it for reading or for signing. A potential solution would be to not validate inside `Store::from_stream`, but instead let the caller call `Store::verify_store` themselves. Any thoughts on this?

sdk/src/reader.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2408151980

sdk/src/ingredient.rs
ok-nick: For now I'm going to remove the new ingredient public APIs, we should address that in a separate PR.

sdk/src/builder.rs
ok-nick: I suspect users will often need to make small changes to `Settings` and use it for another `Builder`. The problem is that if we take ownership, callers will have to make a clone of `Settings` which can be expensive due to the size of certs. If we pass by reference, we'd have to add settings to a substantial amount of methods which also isn't ideal.

sdk/src/claim.rs
ok-nick: On a first look, it seems like `cert_check` is in `ctp` which is in `settings`. I think we can make some changes here but it would require quite some refactoring and isn't necessarily blocking the goal of this PR. We should probably create a separate issue for it. 

See a related note here: https://github.com/contentauth/c2pa-rs/pull/1447#discussion_r2408118655.

sdk/src/claim.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1444#discussion_r2408605759

sdk/src/utils/io_utils.rs
```diff
@@ -425,42 +412,6 @@ mod tests {
...
-        .unwrap();
-
-        for _ in 0..threshold {
-            stream.write_all(&large_data).unwrap();
-        }
-
-        assert!(stream.is_rolled(), "data moved to disk");
-    }
+    // REVIEW NOTE: I deleted tests here that duplicate tests found in
+    // https://github.com/Stebalien/tempfile/blob/master/tests/spooled.rs.
```
ok-nick: @dyro Any objections to removing these tests?

sdk/src/reader.rs
```diff
@@ -508,35 +545,42 @@ impl Reader {
...
+        //         } else {
+        //             ValidationState::Valid
+        //         }
+        //     }
+        // }
+
+        // REVIEW-NOTE: this field is always set on construction, it seems we previously recomputed it because
+        //              thread-local settings could change at any time
+        //              I'm not sure why we define it as an optional, perhaps that should change?
+        self.validation_state.unwrap_or(ValidationState::Invalid)
```
ok-nick: @gpeacock Any insight on this?

ok-nick (commented):
There's quite a few places in this PR we can clean up and consolidate settings, however, many of those changes would require substantial refactoring. I think we can defer that additional work to separate PRs to make this one easier to review. This PR makes settings fully pass by reference and after we address a few open questions, it should be good to go.

sdk/src/asset_handlers/bmff_io.rs
ok-nick: (confirmed out-of-band)

sdk/src/utils/io_utils.rs
dyro: No objections, @ok-nick

sdk/src/reader.rs
mauricefisher64: Why is this function commented out?

sdk/src/store.rs
mauricefisher64: It may be more flexible to leave it this way, it is only passed by reference so it is not any more expensive.

sdk/src/reader.rs
mauricefisher64: Yes, we used to have it that way and people got it wrong.  They would not validate and complain later when they found that validation failed on their final asset.

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: We are not doing Store validation only Store manipulation.  Settings don't matter here.

sdk/src/reader.rs
ok-nick: Reverted that change, I realized I didn't take into account the possibility of post validation.

sdk/src/builder.rs
ok-nick: @scouten-adobe and I agreed to ignore it for now to be addressed at a later time.

ok-nick (approved):

PR #1447, opened 2025-09-24
feat: add settings structs to the public API

ok-nick:
This PR exposes our settings structs to the public API and adds json schemas to them for better compatibility with our bindings.

**For reviewers:**
- What fields do we want exposed and what fields do we not want exposed?
- Do we want to move around any fields?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -346,8 +346,9 @@ jobs:
         run: cargo codspeed build -p c2pa --no-default-features --features rust_native_crypto
 
       - name: Run the benchmarks
-        uses: CodSpeedHQ/action@v3
+        uses: CodSpeedHQ/action@v4
         with:
+          mode: instrumentation
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
 
@@ -379,16 +380,10 @@ jobs:
           curl https://wasmtime.dev/install.sh -sSf | bash
           echo "$HOME/.wasmtime/bin" >> $GITHUB_PATH
 
-      - name: Install WASI SDK
+      - name: Install clang
         run: |
-          if [ "${RUNNER_ARCH}" = "X64" ]; then
-            ARCH="x86_64";
-          else
-            ARCH="${RUNNER_ARCH}";
-          fi
-          wget https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/wasi-sdk-25.0-${ARCH}-${RUNNER_OS}.tar.gz
-          tar xf wasi-sdk-25.0-${ARCH}-${RUNNER_OS}.tar.gz
-          mv $(echo wasi-sdk-25.0-${ARCH}-${RUNNER_OS} | tr '[:upper:]' '[:lower:]') /opt/wasi-sdk
+          sudo apt-get update
+          sudo apt-get install -y clang
 
       - name: Add wasm32-wasip2 target
         run: rustup target add --toolchain nightly-2025-08-25 wasm32-wasip2
@@ -398,9 +393,8 @@ jobs:
 
       - name: Run WASI tests (c2pa-rs)
         env:
-          CARGO_TARGET_WASM32_WASIP2_RUNNER: "wasmtime -S cli -S http --dir ."
-          CC: /opt/wasi-sdk/bin/clang
-          WASI_SDK_PATH: /opt/wasi-sdk
+          CARGO_TARGET_WASM32_WASIP2_RUNNER: "wasmtime -S cli -S http --dir . --env GITHUB_ACTIONS=${GITHUB_ACTIONS}"
+          CC: clang
           RUST_MIN_STACK: 16777216
           FEATURES: ${{needs.get-features.outputs.rust-native-features}}
         run: |
--- a/.github/workflows/library-release.yml
+++ b/.github/workflows/library-release.yml
@@ -81,7 +81,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Download build artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           pattern: release-artifacts-*
           path: .
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ For details on what you can do with the library, see [Using the Rust library](ht
 
 This is a beta release (version 0.x.x) of the project. The minor version number (0.x.0) is incremented when there are breaking API changes, which may happen frequently.
 
-**NOTE**: The library now supports [C2PA v2 claims](https://c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_claims), however development is still in progress and all features are not fully implemented yet. While you can experiment with this functionality, it is not recommended for production use at this time.  For details, see [C2PA v2 claims](https://opensource.contentauthenticity.org/docs/rust-sdk/docs/release-notes#c2pa-v2-claims).
+**NOTE**: The library supports [C2PA v2 claims](https://c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_claims) by default, and implementations should not generate deprecated v1 claims.  For details, see [C2PA v2 claims](https://opensource.contentauthenticity.org/docs/rust-sdk/docs/release-notes#c2pa-v2-claims).
 
 ### New API
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -14,11 +14,11 @@ For a simple example of calling c2patool from a Node.js server application, see 
 
 **Additional documentation**:
 
-- [Using C2PA Tool](./docs/usage.md)
-- [Manifest definition file](./docs/manifest.md)
-- [Using an X.509 certificate](./docs/x_509.md)
-- [Change log / release notes](./CHANGELOG.md)
-- [Contributing to the project](./docs/project-contributions.md)
+- [Using C2PA Tool](https://github.com/contentauth/c2pa-rs/blob/main/cli/docs/usage.md)
+- [Manifest definition file](https://github.com/contentauth/c2pa-rs/blob/main/cli/docs/manifest.md)
+- [Using an X.509 certificate](https://github.com/contentauth/c2pa-rs/blob/main/cli/docs/x_509.md)
+- [Change log / release notes](https://github.com/contentauth/c2pa-rs/blob/main/cli/CHANGELOG.md)
+- [Contributing to the project](https://github.com/contentauth/c2pa-rs/blob/main/cli/docs/project-contributions.md)
 
 </div>
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -24,15 +24,12 @@ An `AssetType` assertion is now supported.
 
 ### C2PA v2 claims
 
-**NOTE**: The library now supports [C2PA v2 claims](https://c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_claims) by default.
+**NOTE**: The library now supports [C2PA v2 claims](https://c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_claims) by default. V2 claims have many new checks and status codes.  Additionally:
 
-To generate v1 claims, set the `Builder` manifest definition `claim_version` field to `1`.
+- The `title()` and `format()` methods of both `Manifest` and `Ingredient` objects now return an `Option<String>` because in v2 claims, `title` is optional and `format` does not exist.
+- The first `action` must be `c2pa.created` or `c2pa.opened` (which requires an ingredient). 
 
-The `title()` and `format()` methods of both `Manifest` and `Ingredient` objects now return an `Option<String>` because in v2 claims, `title` is optional and `format` does not exist.
-
-In v2 claims, the first `action` must be `c2pa.created` or `c2pa.opened`. 
-
-V2 claims have many new checks and status codes.
+WARNING: Implementations should not generate deprecated v1 claims.  If needed, though, you can generate v1 claims by setting the `Builder` manifest definition `claim_version` field to `1`.
 
 ### Using the old API
 
@@ -42,7 +39,7 @@ To use the old deprecated API, enable the `v1_api` feature; for example:
 c2pa = {version="0.45.2", features=["v1_api"]}
 ```
 
-This will be the last release with the v1_api feature available.
+This will be the last release with the `v1_api` feature available.
 
 ## Language binding support
 
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -1063,6 +1063,12 @@ pub mod tests {
     #[cfg(feature = "file_io")]
     /// Test that the reader can validate a file with nested assertion errors
     fn test_reader_to_folder() -> Result<()> {
+        // Skip this test in GitHub workflow when target is WASI
+        if std::env::var("GITHUB_ACTIONS").is_ok() && cfg!(target_os = "wasi") {
+            eprintln!("Skipping test_reader_to_folder on WASI in GitHub Actions");
+            return Ok(());
+        }
+
         use crate::utils::{io_utils::tempdirectory, test::temp_dir_path};
         let reader = Reader::from_file("tests/fixtures/CACAE-uri-CA.jpg")?;
         assert_eq!(reader.validation_status(), None);
@@ -1070,8 +1076,6 @@ pub mod tests {
         reader.to_folder(temp_dir.path())?;
         let path = temp_dir_path(&temp_dir, "manifest.json");
         assert!(path.exists());
-        #[cfg(target_os = "wasi")]
-        crate::utils::io_utils::wasm_remove_dir_all(temp_dir)?;
         Ok(())
     }
 
--- a/sdk/src/utils/io_utils.rs
+++ b/sdk/src/utils/io_utils.rs
@@ -223,65 +223,6 @@ pub(crate) fn tempdirectory() -> Result<TempDir> {
     return tempdir().map_err(Error::IoError);
 }
 
-#[allow(unused)]
-#[cfg(target_os = "wasi")]
-pub fn wasm_remove_dir_all<P: AsRef<std::path::Path>>(path: P) -> Result<()> {
-    for entry in std::fs::read_dir(&path)? {
-        let entry = entry?;
-        let entry_path = entry.path();
-        // List initial entries for debugging
-        eprintln!("Initial entry: {}", entry_path.display());
-        if entry_path.is_file() || entry_path.is_symlink() {
-            eprintln!("Removing file {}", entry_path.display());
-            std::fs::remove_file(&entry_path).map_err(|e| {
-                eprintln!("Failed to remove file {}: {}", entry_path.display(), e);
-                e
-            })?;
-        } else if entry_path.is_dir() {
-            eprintln!("Removing directory: {}", entry_path.display());
-            wasm_remove_dir_all(&entry_path).map_err(|e| {
-                eprintln!("Failed to remove directory {}: {}", entry_path.display(), e);
-                e
-            })?;
-        }
-    }
-
-    // List remaining entries if the directory is still not empty
-    if let Ok(entries) = std::fs::read_dir(&path) {
-        for entry in entries {
-            if let Ok(entry) = entry {
-                eprintln!("Remaining entry before removal: {}", entry.path().display());
-            }
-        }
-    }
-
-    // Retry removing the directory if it fails
-    let mut retries = 3;
-    while retries > 0 {
-        match std::fs::remove_dir_all(&path) {
-            Ok(_) => return Ok(()),
-            Err(e) => {
-                eprintln!(
-                    "Failed to remove directory {}: {}. Retries left: {}",
-                    path.as_ref().display(),
-                    e,
-                    retries - 1
-                );
-                retries -= 1;
-                std::thread::sleep(std::time::Duration::from_millis(100));
-            }
-        }
-    }
-
-    Err(std::io::Error::new(
-        std::io::ErrorKind::Other,
-        format!(
-            "Failed to remove directory {} after retries",
-            path.as_ref().display()
-        ),
-    ))?
-}
-
 /// Convert a URI to a file path using PathBuf for better path handling.
 #[cfg(feature = "file_io")]
 pub fn uri_to_path(uri: &str, manifest_label: Option<&str>) -> PathBuf {


sdk/src/settings/builder.rs
```diff
@@ -166,7 +169,7 @@ pub(crate) struct ClaimGeneratorInfoSettings {
     pub operating_system: Option<ClaimGeneratorInfoOSSettings>,
```
gpeacock: I'd suggest we make this just an optional string and perhaps have a special string like "auto" to indicate we want it to have it detected.

sdk/src/settings/builder.rs
```diff
@@ -323,33 +326,36 @@ impl TryFrom<ActionSettings> for Action {
...
 /// The reason this setting exists only for an [Actions][crate::assertions::Actions] assertion
 /// is because of its mandations and reusable fields.
-#[allow(unused)]
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-pub(crate) struct ActionsSettings {
+pub struct ActionsSettings {
     /// Whether or not to set the [Actions::all_actions_included][crate::assertions::Actions::all_actions_included]
     /// field.
     pub all_actions_included: bool,
```
gpeacock: This should be an Option defaulting to None.

sdk/src/settings/builder.rs
```diff
@@ -129,9 +132,9 @@ pub(crate) struct AutoActionSettings {
 }
 
 /// Settings for how to specify the claim generator info's operating system.
-#[allow(unused)]
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-pub(crate) struct ClaimGeneratorInfoOSSettings {
+pub struct ClaimGeneratorInfoOSSettings {
```
gpeacock: I'd suggest we remove this as a separate structure, it's just more to process and I'd rather keep the fields in the settings as close as we can to the fields in the claim

sdk/src/settings/builder.rs
gpeacock: I'm a unclear on how we would use a resource reference in this context. We need to demonstrate how to use if if we are going to include it here. I suppose it could have a file or https url to fetch from, but currently it is unclear how to define or resolve it.

sdk/src/settings/builder.rs
```diff
@@ -323,33 +326,36 @@ impl TryFrom<ActionSettings> for Action {
...
+    #[doc(hidden)]
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub templates: Option<Vec<ActionTemplateSettings>>,
-    // TODO: should we define a new struct for "Action" too, like ActionTemplateSettings?
-    /// Actions to be added to the [Actions::actions][crate::assertions::Actions::actions] field.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub actions: Option<Vec<ActionSettings>>,
-    /// Whether to automatically generate a c2pa.created [Action][crate::assertions::Action]
-    /// assertion or error that it doesn't already exist.
+    pub(crate) templates: Option<Vec<ActionTemplateSettings>>,
```
gpeacock, line 358: I'd suggest we leave actionTemplates out of Settings, or at least public settings, until we have time to resolve issues and test it. Testing is my primary concern here since there are so may possibilities.

sdk/src/settings/builder.rs
```diff
@@ -404,18 +409,21 @@ pub(crate) struct BuilderSettings {
     ///
     /// For more information on the reasoning behind this field see [ActionsSettings].
     pub actions: ActionsSettings,
-
-    // Certificate statuses will be fetched for either all the manifest labels, or just the active manifest.
-    pub certificate_status_fetch: Option<OcspFetch>,
-
-    // Whether or not existing OCSP responses should be overridden by new values.
+    // REVIEW NOTE: should this be in builder settings or core?
+    /// Certificate statuses will be fetched for either all the manifest labels, or just the active manifest.
```
gpeacock: I don't know if we need all these options. It may be best just to enable or disable certificate_status fetch for the active manifest.

gpeacock: If we don't know the answers to how to use this, it is far less likely that anyone else will

sdk/src/settings/mod.rs
```diff
@@ -160,67 +167,105 @@ impl SettingsValidate for Trust {
...
+    /// The default value is true.
+    ///
+    /// [`Reader`]: crate::Reader
+    pub verify_after_reading: bool,
+    /// Whether to verify the manifest after signing in the [`Builder`].
+    ///
+    /// The default value is true.
+    ///
+    /// [`Builder`]: crate::Builder
+    pub verify_after_sign: bool,
```
gpeacock, line 262: Disabling this can improve signing time but validation may not pass when reading signed claims.

sdk/src/settings/mod.rs
```diff
@@ -160,67 +167,105 @@ impl SettingsValidate for Trust {
...
+    /// Whether to verify certificates against the trust lists specified in [`Trust`].
+    ///
+    /// The default value is true.
+    pub verify_trust: bool,
+    /// Whether to verify the timestamp certificates against the trust lists specified in [`Trust`].
```
gpeacock, line 272: Verifying trust is required by the C2PA spec. This should be used only for development or testing.

sdk/src/settings/mod.rs
```diff
@@ -160,67 +167,105 @@ impl SettingsValidate for Trust {
...
-    verify_after_sign: bool,
-    verify_trust: bool,
-    verify_timestamp_trust: bool,
-    ocsp_fetch: bool,
-    remote_manifest_fetch: bool,
-    check_ingredient_trust: bool,
-    skip_ingredient_conflict_resolution: bool,
-    strict_v1_validation: bool,
+pub struct Verify {
+    /// Whether to verify the manifest after reading in the [`Reader`].
```
gpeacock, line 241: We should have some information added about the consequences of enabling/disabling some of these. Some of these should never be used in production

sdk/src/settings/mod.rs
```diff
@@ -248,22 +293,34 @@ const MINOR_VERSION: usize = 0;
 ///
 /// [Settings::default] will be set thread-locally by default. Any settings set via
 /// [Settings::from_toml] or [Settings::from_file] will also be thread-local.
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
-#[allow(unused)]
 pub struct Settings {
-    version_major: usize,
-    version_minor: usize,
+    // REVIEW NOTE: do we need both a major and minor version?
```
gpeacock: I don't think we do

sdk/src/settings/builder.rs
```diff
@@ -404,18 +409,21 @@ pub(crate) struct BuilderSettings {
...
+    /// Whether or not existing OCSP responses should be overridden by new values.
     pub certificate_status_should_override: Option<bool>,
 }
 
+/// The scope of which manifests to fetch for OCSP.
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(rename_all = "lowercase")]
-pub(crate) enum OcspFetch {
+pub enum OcspFetchScope {
```
gpeacock: do we need this or can we just retrieve the active manifest ocsp?

sdk/src/settings/builder.rs
```diff
@@ -198,6 +201,7 @@ impl TryFrom<ClaimGeneratorInfoSettings> for ClaimGeneratorInfo {
 }
 
 /// Settings for an action template.
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub(crate) struct ActionTemplateSettings {
```
gpeacock, line 226: I'm tempted to leave actionTemplateSettings out or undocumented for now, since this is a lot to test and not in much use at the moment.

sdk/src/settings/builder.rs
ok-nick: I'm leaning towards removing it from settings for now since resolving these `ResourceRef`s would be builder-specific (e.g. resolving relative paths or JUMBF URIs).

sdk/src/settings/builder.rs
ok-nick: I agree that's why I `pub(crate)` it for now.

sdk/src/settings/mod.rs
ok-nick: Added to the docs 👍

sdk/src/settings/mod.rs
ok-nick: Noted in the docs 👍

sdk/src/settings/builder.rs
ok-nick: Yes I agree, I `pub(crate)` it for now.

sdk/src/settings/builder.rs
ok-nick: Hm, that's a good question. I think supporting both is a valid use case, but I'm not sure how people normally expect it to work. We should probably support OCSP fetching at a more programmatic level to cover these cases.

sdk/src/settings/mod.rs
ok-nick: Good idea, I'll see what I can do.

sdk/src/settings/builder.rs
ok-nick: How does certificate status differ from OCSP fetch? Are these settings just for configuring the behavior of stapled OCSPs? If that's the case I think we can move it to core and merge it with the `core.ocsp_fetch` setting.

sdk/src/settings/builder.rs
ok-nick: I'll defer that implementation to your PR here #1459

sdk/src/settings/builder.rs
ok-nick: I'm going to also change this to default to None, unless otherwise specified.

sdk/src/settings/builder.rs
ok-nick: Changed to an enum with "Auto" and "Other."

sdk/src/settings/mod.rs
```diff
@@ -160,67 +176,123 @@ impl SettingsValidate for Trust {
...
+    /// [`Builder`]: crate::Builder
+    pub verify_after_sign: bool,
+    /// Whether to verify certificates against the trust lists specified in [`Trust`].
+    ///
+    /// The default value is true.
+    ///
+    /// <div class="warning">
+    /// Verifying trust is REQUIRED by the C2PA spec. This option should only be used for development or testing.
+    /// </div>
+    pub verify_trust: bool,
```
ok-nick: Maybe we should have `verify_c2pa_trust` and `verify_cawg_trust` to be clear they are separate and also to clarify it doesn't encompass timestamp trust. I'll add to the docs that `verify_timestamp_trust` should be used in that case.

sdk/src/settings/builder.rs
```diff
@@ -323,33 +314,36 @@ impl TryFrom<ActionSettings> for Action {
...
-    /// assertion or error that it doesn't already exist.
+    pub(crate) templates: Option<Vec<ActionTemplateSettings>>,
+    //
+    // REVIEW NOTE: ActionSettings indirectly depends on `ActionParameters` which contains a `serde_cbor::Value`
+    //              and schemars can't generate a schema for cbor values. It also doesn't feel right to change
+    //              our API for the sake of json schemas.
+    // /// Actions to be added to the [Actions::actions][crate::assertions::Actions::actions] field.
+    // #[doc(hidden)]
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub(crate) actions: Option<Vec<ActionSettings>>,
```
ok-nick: `ActionSettings` indirectly depends on `ActionParameters` which contains a `serde_cbor::Value` and `schemars` can't generate a schema for cbor values. It also doesn't feel right to change our API for the sake of json schemas.

sdk/src/settings/mod.rs
ok-nick, line 433: If we are publicizing the structs, then it probably doesn't make sense to create the struct, serialize it, then call these methods. Should we instead have a `Settings::set` or `Settings::set_thread_local` method? With these structs publicized, callers can now handle the serialization/deserialization themselves.

sdk/src/settings/builder.rs
ok-nick: After thinking about it, these settings seem reasonable and should probably stay in builder settings. I think we can do a little bit better here with documenting them though. I'll work on that.

ok-nick: I'm not sure if these settings are useful until we support the timestamp assertion? Maybe we should private them for now.

sdk/src/settings/mod.rs
```diff
@@ -160,67 +176,126 @@ impl SettingsValidate for Trust {
...
+    ///
+    /// The default value is true.
+    ///
+    /// <div class="warning">
+    /// Disabling validation can improve signing performance, BUT it carries the risk of signing an invalid
+    /// manifest.
+    /// </div>
+    ///
+    /// [`Builder`]: crate::Builder
+    pub verify_after_sign: bool,
```
ok-nick, line 262: Should we move this to `BuilderSettings`?

sdk/src/settings/mod.rs
```diff
@@ -160,67 +176,126 @@ impl SettingsValidate for Trust {
...
+    /// The default value is false.
+    ///
+    /// [`CertificateStatus`]: crate::assertions::CertificateStatus
+    pub ocsp_fetch: bool,
+    /// Whether to fetch remote manifests during reading in the [`Reader`].
+    ///
+    /// The default value is true.
+    ///
+    /// [`Reader`]: crate::Reader
+    pub remote_manifest_fetch: bool,
```
ok-nick, line 305: Maybe we should move this and `verify.verify_after_reading` to a `ReaderSettings`? I think there's a few other settings we can move to `BuilderSettings` and `ReaderSettings` from here. Any thoughts?

sdk/src/settings/mod.rs
```diff
@@ -40,15 +44,27 @@ pub(crate) trait SettingsValidate {
...
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-#[allow(unused)]
-pub(crate) struct Trust {
-    verify_trust_list: bool,
-    user_anchors: Option<String>,
-    trust_anchors: Option<String>,
-    trust_config: Option<String>,
-    allowed_list: Option<String>,
+pub struct Trust {
+    // REVIEW NOTE: should we remove this field in favor of `verify.verify_trust`, only CAWG is using it.
```
tmathern: Is this a name change then or real removal?

sdk/src/settings/mod.rs
```diff
@@ -160,67 +176,126 @@ impl SettingsValidate for Trust {
     }
 }
 
-// TODO: all of these settings aren't implemented
-// Settings for core C2PA-RS functionality
+/// Settings to configure core features.
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
```
tmathern, line 184: Shouldn't we keep core features config SDK-internal?

sdk/src/settings/mod.rs
tmathern: `remote_manifest_fetch` feels right for me in ReaderSettings. Because it configures the behavior of the Reader, aka if it retrieves remote manifests or not.

sdk/src/settings/mod.rs
ok-nick: I think the core settings in this PR are reasonable to expose to the user.

sdk/src/settings/mod.rs
ok-nick: A name change, the functionality is still useful. We talked about possibly moving the other verify settings into the trust struct to keep it more self-contained when it's plumbed down the SDK.

sdk/src/settings/mod.rs
ok-nick: I wonder if we should have a `ReadSettings` and `SignSettings` instead?

sdk/src/settings/builder.rs
ok-nick: Ok, after looking through the code, the behavior for OCSP and these options in general are very complicated and difficult to document. We should consider reviewing it and possibly simplifying the functionality.

sdk/examples/c2pa.toml
```diff
@@ -66,9 +65,8 @@ name = "My Service"
 # A human readable string of the product's version.
 version = "1.0.0"
 # The operating system the claim generator is running on.
-operating_system.name = "macOS"
-# Or if the name isn't specified, it can be inferred automatically.
-operating_system.infer = true
+# Or specify "auto" to infer the operating system automatically.
+operating_system = "macOS"
```
gpeacock, line 69: What are the valid values here? The spec isn't very helpful but we should be better. Are we following any known naming convention for this?

sdk/src/settings/builder.rs
```diff
@@ -118,55 +121,44 @@ impl SettingsValidate for ThumbnailSettings {
...
 
-/// Settings for how to specify the claim generator info's operating system.
-#[allow(unused)]
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-pub(crate) struct ClaimGeneratorInfoOSSettings {
-    /// Whether or not to infer the operating system.
-    pub infer: bool,
+#[serde(untagged, rename_all = "lowercase")]
+pub enum ClaimGeneratorInfoOperatingSystem {
```
gpeacock, line 154: do we still need this if we are using the string "auto"?

sdk/src/settings/builder.rs
```diff
@@ -118,55 +121,44 @@ impl SettingsValidate for ThumbnailSettings {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
     /// Reference to an icon.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub icon: Option<ResourceRef>,
+    pub(crate) icon: Option<ResourceRef>,
     /// Settings for the claim generator info's operating system field.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub operating_system: Option<ClaimGeneratorInfoOSSettings>,
+    pub operating_system: Option<ClaimGeneratorInfoOperatingSystem>,
```
gpeacock, line 188: maybe we just convert this to an Option<String> ?

sdk/src/settings/builder.rs
```diff
@@ -178,11 +170,10 @@ impl TryFrom<ClaimGeneratorInfoSettings> for ClaimGeneratorInfo {
             version: value.version,
             icon: value.icon.map(UriOrResource::ResourceRef),
             operating_system: {
```
gpeacock, line 202: I think this can just be os.name

sdk/src/settings/builder.rs
```diff
@@ -323,33 +314,33 @@ impl TryFrom<ActionSettings> for Action {
...
+    // /// Actions to be added to the [Actions::actions][crate::assertions::Actions::actions] field.
+    // #[doc(hidden)]
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub(crate) actions: Option<Vec<ActionSettings>>,
+    //
+    /// Whether to automatically generate a c2pa.created [Action] assertion or error that it doesn't already exist.
     ///
     /// For more information about the mandatory conditions for a c2pa.created action assertion, see here:
     /// <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_mandatory_presence_of_at_least_one_actions_assertion>
     pub auto_created_action: AutoActionSettings,
```
gpeacock, line 369: I think we can replace auto_created_action and auto_opened_action with Builder.intent.create/open
That makes the intent clear.

sdk/src/settings/builder.rs
```diff
@@ -404,18 +394,31 @@ pub(crate) struct BuilderSettings {
...
+    ///
+    /// See more information in the spec here:
+    /// <https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#certificate_status_assertion>
+    ///
+    /// [`CertificateStatus`]: crate::assertions::CertificateStatus
+    pub certificate_status_fetch: Option<OcspFetchScope>,
+    /// Whether existing [`CertificateStatus`] assertions should be refreshed, also known as OCSP stapling. Use
+    /// [`BuilderSettings::certificate_status_fetch`] to configure which certificates are affected.
+    ///
+    /// The default value is false.
```
gpeacock, line 447: In another PR I'm adding intent: Option<BuilderIntent>

sdk/src/settings/mod.rs
gpeacock: Let's make sure we get Maurice's opinion on these

sdk/src/settings/mod.rs
gpeacock: That makes sense to me, it's probably where people would expect it.

sdk/src/settings/mod.rs
gpeacock: Well there's also remote_manifest_fetch when reading ingredients in builder.

gpeacock: But maybe that should be a separate setting?

sdk/src/settings/mod.rs
gpeacock: The c_api that most of the bindings depend on needs to have string based serialization available.
Some of the structs are open ended maps or unknown structures, so we can't map them all directly in c.

gpeacock (changes requested):

sdk/src/settings/builder.rs
ok-nick: I think it's a good idea to express all the possible cases as Rust types. If this is constructed manually it forces you to make an informed decision. It also maps better to the JSON schemas so they can document each possible variant, with the added benefit that it provides more expressive types for the other languages.

sdk/src/settings/builder.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1447#discussion_r2407455724.

sdk/src/settings/builder.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1447#discussion_r2407455724.

sdk/src/settings/builder.rs
ok-nick: This is a good idea, but we will have to keep in mind that `BuilderIntent::Create(DigitalSourceType)` takes a value. So what it will end up looking like is:
```json
{
    "intent": {
        "type": "created",
        "source_type": "...",
    }
}
```
or
```json
{
    "intent": {
        "type": "opened"
    }
}
```

sdk/src/settings/mod.rs
ok-nick: For our bindings we can still require a JSON string and internally deserialize it with serde into our `Settings` struct. For our Rust callers this is a little more flexible since they have serde available to them.

sdk/src/settings/mod.rs
ok-nick: It may be a good idea to separate ingredient settings into its own thing. This would be especially useful for configuring thumbnail generation. Maybe it should be a subset of `ReadSettings`?

sdk/examples/c2pa.toml
ok-nick: Maybe this is something we should add to the `ClaimGeneratorInfoOperatingSystem` settings, alongside `Auto`?

sdk/src/settings/mod.rs
```diff
@@ -40,15 +44,27 @@ pub(crate) trait SettingsValidate {
...
+    /// List of default trust anchor root certificates as a PEM bundle.
+    ///
+    /// Normally this option contains the official C2PA-recognized trust anchors found here:
+    /// <https://github.com/c2pa-org/conformance-public/tree/main/trust-list>
+    pub trust_anchors: Option<String>,
+    /// List of allowed extended key usage (EKU) object identifiers (OID) that
+    /// certificates must have.
+    pub trust_config: Option<String>,
+    /// List of explicitly allowed certificates as a PEM bundle.
+    pub allowed_list: Option<String>,
```
ok-nick, line 71: These cert lists get big. We should consider storing them as a `Vec<u8>`, or even better an `X509` in Rust, but deserialize them from a string with serde (can add a macro on the field). In different parts of the code we deserialize the PEM, then the DER, but we can do all of that in one call and also provide validation at the top-level. This is good to think about now before we include it in our public API.

sdk/src/settings/mod.rs
gpeacock: It seems to me that if you configure the Trust settings, you probably want to use them. So why do we need a separate set of flags to enable them?

sdk/src/settings/builder.rs
gpeacock: I've tried to avoid exposing serde_cbor values in the API for this reason.

sdk/src/settings/mod.rs
gpeacock: There can definitely be different options people want for ingredient vs claim thumbnails.

sdk/examples/c2pa.toml
gpeacock: I'd recommend something like the compiler triplet convention, but that's not just an operating system.

x86_64-unknown-linux-gnu
aarch64-apple-darwin
x86_64-pc-windows-msvc
arm-linux-gnueabihf

sdk/src/settings/builder.rs
gpeacock: I think that would work.
Another alternative would be to have a separate digitalSourceType field that is required if the intent is created.

sdk/src/settings/mod.rs
gpeacock: That's an interesting idea. But we would also need a way to do that programmatically if a user has a PEM. And we need to deserialize back to  pem in some cases. I'd check with @mauricefisher64 since i'm not sure how consisten the conversion is. But it may help performance to do the conversion once when loading the settings.

sdk/src/settings/mod.rs
ok-nick: We currently do have a separate set of flags to enable them, see `Trust::verify_trust_list` (CAWG) and `Verify::verify_trust` (C2PA).

sdk/src/settings/builder.rs
ok-nick: It is more correct to allow serde_cbor values to be specified, but if most people are going through the JSON APIs, they probably don't need the extra types it offers. I think ideally we'd expose a wrapper value type that supports JSON schemas but also serde_cbor values. Maybe this is something we should look into when we move away from serde_cbor.

sdk/examples/c2pa.toml
ok-nick: I like that. One issue however, is that we cannot get the full triple at runtime (for the "Auto" setting) without adding a `build.rs` script that saves the triple and loads that value in at runtime. Not sure if this is a concern or not.

We can use some of the [`std::env`](https://doc.rust-lang.org/std/env/consts/index.html) consts but it won't give us the vendor or ABI, so we'd only have things like: `x86_64-unknown-linux`.

sdk/src/settings/mod.rs
ok-nick: We will also have to keep in mind that openssl and rust_native_crypto have a different struct representation for the cert. If we did go this route, we'd have to either feature flag them, choose one and convert it to the other, or store it as its intermediate byte vec representation.

sdk/src/settings/mod.rs
ok-nick: Hm, there's a lot of overlap with the `Reader`/`Builder`/`Ingredient` settings, such as for `remote_manifest_fetch`. We also need to come up with a strategy to preserve backwards compatibility if we were to move these fields. I'm thinking we may want to leave that up to a separate PR.

sdk/src/settings/mod.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1447#discussion_r2414237375

sdk/src/settings/builder.rs
ok-nick: This is resolved in #1432 (unsure if we are removing the auto_... fields in that PR as well).

sdk/examples/c2pa.toml
gpeacock: It isn't critical for now, since the spec only calls it a string, but I do want to give some guidance on this.

sdk/src/settings/builder.rs
gpeacock: I didn't yet.  could do that there or follow up with it.

sdk/src/settings/mod.rs
gpeacock: Ok, but why? Trust lists are required by the spec,  I guess disabling could be useful for debugging?

sdk/src/settings/mod.rs
gpeacock: We will have backwards compatibility issues for any settings we change later, so it would be a good idea to start with only what we think we need.

sdk/examples/c2pa.toml
ok-nick: Ok, I added some guidance on this and provided a _better_ "auto" value.

sdk/src/settings/mod.rs
ok-nick: I'd like to hear @scouten-adobe input on this.

sdk/src/settings/mod.rs
ok-nick: I agree. I think we should leave changes to existing settings up to a separate PR then, at least until we have a way to properly handle incompatible changes.

sdk/src/settings/mod.rs
scouten-adobe: We have a separate `Trust` structure (see https://github.com/contentauth/c2pa-rs/pull/1447/files/aee4a005f7ebabead90b13bb5f9afe92a177f40f#diff-40b334ea0cacc187a239b5763c0403155527f660c81e8e90462bc01a3c36bf0dR50) which already occurs twice, once for CAWG and once for C2PA.

I'd like to see us move as much of this configuration into that tree and then the question of which one you are enabling/disabling will be much more obvious (IMHO).

sdk/src/settings/builder.rs
```diff
@@ -46,9 +47,10 @@ pub(crate) enum ThumbnailFormat {
     Tiff,
```
mauricefisher64, line 48: Why do we support TIFF thumbnails>

cli/tests/fixtures/trust/cawg_sign_settings.toml
```diff
@@ -418,8 +417,8 @@ name = "c2pa-rs testing"
 version = "1.0.0"
 # The operating system the claim generator is running on.
 #operating_system.name = "macOS"
-# Or if the name isn't specified, it can be inferred automatically.
-operating_system.infer = true
+# Or specify "auto" to infer the operating system automatically.
```
mauricefisher64, line 420: Why are we fetching OCSP responses for every certificate as a default!

certificate_status_fetch = "all"
certificate_status_should_override = true

sdk/src/settings/builder.rs
```diff
@@ -405,18 +426,38 @@ pub(crate) struct BuilderSettings {
...
+/// The scope of which manifests to fetch for OCSP.
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
 #[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(rename_all = "lowercase")]
-pub(crate) enum OcspFetch {
+pub(crate) enum OcspFetchScope {
+    /// Fetch OCSP for all manifests.
     All,
+    /// Fetch OCSP for the active manifest only.
     Active,
```
mauricefisher64, line 477: How is this meant to be used in the settings file?  Without an "off" choice then the presence of OCSPFetchScope will mean we are fetching responses.   I do not thing fetching should ever default to on.

sdk/src/settings/mod.rs
mauricefisher64: Or not touch it!  No time to play with this.

sdk/src/settings/mod.rs
```diff
@@ -160,67 +176,123 @@ impl SettingsValidate for Trust {
...
+    ///
+    /// <div class="warning">
+    /// Verifying timestamp trust is REQUIRED by the C2PA spec. This option should only be used for development or testing.
+    /// </div>
+    pub verify_timestamp_trust: bool,
+    // REVIEW NOTE: what's the difference between this and builder.certificate_status_fetch? does it have to do with checking staples?
+    /// Whether to fetch the certificates OCSP status during validation.
+    ///
+    /// The default value is false.
+    pub ocsp_fetch: bool,
```
mauricefisher64, line 290: Has nothing to do with Reader, the value determines whether we fetch OCSP during validation.

sdk/src/settings/mod.rs
mauricefisher64: When would we fetch an ingredient for a Reader and not for its ingredients?  We are really trying to overcomplicate this stuff for users.

sdk/src/settings/builder.rs
ok-nick: It's naturally supported by the `image` crate although I have no objection to removing it.

cli/tests/fixtures/trust/cawg_sign_settings.toml
ok-nick: Hm, that's interesting. @scouten-adobe any input on this?

sdk/src/settings/builder.rs
ok-nick: `certificate_status_fetch` is defined as `Option<OcspFetchScope>` that defaults to `None`.

sdk/src/settings/mod.rs
ok-nick: Yes I agree, these changes would be nice but not necessary. Just something to think about.

sdk/tests/fixtures/test_settings.toml
```diff
@@ -220,13 +219,13 @@ xPd7wFhjRZHfuWb2cs63xjAGjQ==
...
+//id-kp-documentSigning
 1.3.6.1.5.5.7.3.36
-//id-kp-timeStamping 
+//id-kp-timeStamping
 1.3.6.1.5.5.7.3.8
-//id-kp-OCSPSigning 
+//id-kp-OCSPSigning
 1.3.6.1.5.5.7.3.9
 // MS C2PA Signing
 1.3.6.1.4.1.311.76.59.1.9
```
mauricefisher64, line 233: Where is the C2PA Signing EKU
1.3.6.1.4.1.62558.2.1

sdk/src/settings/mod.rs
ok-nick: Yep, that's what I was trying to convey. The docs seem to match your description, which part seems unclear?

sdk/tests/fixtures/test_settings.toml
ok-nick: Nice catch, I added these to all the settings files.

sdk/src/settings/builder.rs
gpeacock: yes, let's remove that, no reason to have tiff thumbnails. We should probably use Jpeg for anything that can be expressed without alpha and webp for anything else, or maybe just use webp for all thumbnails since it is smaller.

sdk/src/settings/builder.rs
```diff
@@ -58,23 +60,36 @@ pub(crate) enum ThumbnailQuality {
...
     /// This may occur, for instance, if the thumbnail media type or color layout isn't
     /// supported.
+    ///
+    /// The default value is true.
     pub ignore_errors: bool,
     /// The size of the longest edge of the thumbnail.
     ///
     /// This function will resize the input to preserve aspect ratio.
+    ///
+    /// The default value is 1024.
```
gpeacock, line 86: I think we should change the default to 512 or something. We don't really need 1024.

sdk/src/settings/mod.rs
```diff
@@ -160,69 +173,145 @@ impl SettingsValidate for Trust {
...
-pub(crate) struct Core {
-    debug: bool,
-    hash_alg: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    soft_hash_alg: Option<String>,
-    salt_jumbf_boxes: bool,
-    prefer_box_hash: bool,
-    pub(crate) merkle_tree_chunk_size_in_kb: Option<usize>,
-    pub(crate) merkle_tree_max_proofs: usize,
-    compress_manifests: bool,
```
gpeacock, line 176: what happened to compress_manifests? Are we assuming it is always enabled?

sdk/tests/fixtures/test_settings_with_cawg_signing.toml
```diff
@@ -1,8 +1,7 @@
 # c2pa-rs Configuration File
```
gpeacock, line 1: If we are switching to json, shouldn't the toml settings files be in json?

sdk/src/settings/builder.rs
mauricefisher64: Our Verify site does not show TIFF that I know of.

sdk/src/settings/mod.rs
```diff
@@ -160,69 +173,145 @@ impl SettingsValidate for Trust {
...
+    ///
+    /// This option defaults to 512MB and can result in noticeable performance improvements.
+    pub backing_store_memory_threshold_in_mb: usize,
+    /// Whether to decode CAWG [`IdentityAssertion`]s during reading in the [`Reader`].
+    ///
+    /// This option defaults to true.
+    ///
+    /// [`IdentityAssertion`]: crate::identity::IdentityAssertion
+    /// [`Reader`]: crate::Reader
+    pub decode_identity_assertions: bool,
```
mauricefisher64, line 217: Shouldn't this be a verify setting?

sdk/src/settings/mod.rs
```diff
@@ -160,69 +173,145 @@ impl SettingsValidate for Trust {
...
+    /// </div>
+    pub verify_trust: bool,
+    /// Whether to verify the timestamp certificates against the trust lists specified in [`Trust`].
+    ///
+    /// The default value is true.
+    ///
+    /// <div class="warning">
+    /// Verifying timestamp trust is REQUIRED by the C2PA spec. This option should only be used for development or testing.
+    /// </div>
+    pub verify_timestamp_trust: bool,
```
mauricefisher64: make this pub(crate),  verifying timestamp is a spec feature no matter what we decide to do through internal paths.  Not a user setting.

sdk/src/settings/builder.rs
tmathern: Let's remove it? Tiff is heavier (usually) than JPEG/PNG anyway.

sdk/src/settings/builder.rs
```diff
@@ -118,55 +137,57 @@ impl SettingsValidate for ThumbnailSettings {
...
-    pub icon: Option<ResourceRef>,
+    pub(crate) icon: Option<ResourceRef>,
     /// Settings for the claim generator info's operating system field.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub operating_system: Option<ClaimGeneratorInfoOSSettings>,
+    pub operating_system: Option<ClaimGeneratorInfoOperatingSystem>,
     /// Any other values that are not part of the standard.
     #[serde(flatten)]
-    pub other: HashMap<String, toml::Value>,
+    pub other: HashMap<String, serde_json::Value>,
```
tmathern, line 191: Thanks for that change!

sdk/src/settings/mod.rs
tmathern: I nevertheless wonder if there isn't value leaving this a setting for people who may implement things using that crate and defer that validation to a later point.

sdk/src/settings/mod.rs
ok-nick: We should handle this in a separate PR.

sdk/src/settings/builder.rs
ok-nick: Addressing in a separate PR.

sdk/src/settings/mod.rs
ok-nick: Saving settings changes (non-doc-related) to separate PR.

cli/tests/fixtures/trust/cawg_sign_settings.toml
ok-nick: I'm going to leave this up to a separate PR, this one only documents settings.

sdk/src/settings/mod.rs
ok-nick: I hid the trust options for now from the public API after internal discussion. We can reevaluate this in the future.

sdk/tests/fixtures/test_settings_with_cawg_signing.toml
ok-nick: Both are still supported by the old API, so we can leave this for now. We can move the test settings in TOML over in a separate PR.

sdk/src/settings/mod.rs
ok-nick: The setting was unused and I don't think we have it merged yet to main, so we can add that later.

sdk/src/settings/builder.rs
ok-nick: Yes I agree, I will open separate PRs addressing these default/settings changes.

PR #1449, opened 2025-09-25
feat: Expose an additional API on the Builder to set the manifest label

tmathern:
- Expose an additional API on the Builder to set the label of a manifest.

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -1056,7 +1056,6 @@ pub unsafe extern "C" fn c2pa_builder_add_action(
 /// # Safety
 /// - The `builder_ptr` points to a valid `C2paBuilder`
 /// - The `label` is a valid null-terminated C string
-///
 pub unsafe extern "C" fn c2pa_builder_set_label(
     builder_ptr: *mut C2paBuilder,
     label: *const c_char,


c2pa_c_ffi/src/c_api.rs
```diff
@@ -1710,6 +1725,86 @@ mod tests {
         }
     }
 
+    #[test]
+    fn builder_set_label_and_sign() {
+        // uuid4 test label that passes SDK placeholder validation.
+        // The 4 in the 3rd segment indicates the uuid version (not a typo),
+        // the 9 in the 4th segment indicates the uuid variant (not a typo),
+        // and they are necessary so that the test label validates.
+        let test_label: &str = "urn:c2pa:aaaaaaaa-aaaa-4aaa-9aaa-aaaaaaaaaaaa:contentauth";
```
tmathern, line 1748: @gpeacock @mauricefisher64 @dyro Existential question: The SDK checks this format, and fails on signing with the appropriate error if it's not correct. Do we want to have the SDK check only at that level/point in time, or should I add some basic checks at this C layer for the label formatting? Right now I don't do more checks (passthrough from C to Rust) and I let the SDK do its thing on signing.

Reason: I think the rules may/can evolved, I believe it's better to keep them in the deeper SDK levels.

mauricefisher64: Is this setting the actual manifests label?  If so what is the use case.

gpeacock: You can set the label in the manifest definition, so why do you need a Builder API for it?

dyro: > Existential question: The SDK checks this format, and fails on signing with the appropriate error if it's not correct. Do we want to have the SDK check only at that level/point in time, or should I add some basic checks at this C layer for the label formatting?

I'd say leave that to the SDK. If the rules change for the label format, we don't want to have to change that in multiple places.

dyro: @gpeacock : this is for clients who are uploading cloud manifests. The label and the cloud manifest url have to match, I believe. Currently, clients have to do something like this: 

```rust
let context =
    AdobeContext::from_basic_auth(auth).expect("Failed to create adobe-cr-web context");

let signer = context.signer().expect("Failed to get signer");
let uploader = context.manifest_uploader();

let mut src = Cursor::new(include_bytes!("fixtures/A.jpg"));
let mut dest = Cursor::new(Vec::new());

let label = uploader.create_remote_manifest_label(Some("adobe".to_string()));
let url = uploader.create_remote_manifest_url(label.clone());

let json = MANIFEST_JSON.replace("<replace>", &label);
let mut builder = c2pa::Builder::from_json(&json).unwrap();

builder.set_no_embed(true);
builder.set_remote_url(url);

let result = builder.sign(&signer, "image/jpeg", &mut src, &mut dest);
assert!(result.is_ok());

let upload_result = uploader.upload(result.unwrap());
assert!(upload_result.is_ok());
```

So they have to remember to set the label and set_remote_url. Wonder if set_remote_url could also take the label?

dyro: The reason Tania submitted the PR is PS ran into this issue. They forgot to set the label.

tmathern: @dyro Is right. Some of the clients have the issue because of how they handle manifest generation. Currently they workaround by modifying json, but they'd prefer that API.

gpeacock: The problem is that this is an api that no one other than internal Adobe users should ever use. Even the existence of the label in the manifest defintion caused leica to give every manifest they created the same label at first. It is is a hack we did because our manifest storage service required the manifest label inside the manifest to be included in the url. I'm pretty sure we removed that requirement years ago an yet we persist with doing things this way. Could we possibly fix that now instead of adding an open source api and telling people not to use it? It is further complicated by the multiple label formats. This would be a great time to deprecate and get rid of this feature entirely.

PR #1450, opened 2025-09-25
fix: Avoid use of spooled temp file for small files

scouten-adobe:
(Mismatch found between SpooledTempFile crate API and our settings API.)


scouten-adobe:
Withdrawing this PR. I don't see meaningful performance gains in any of these variations.

PR #1456, opened 2025-09-27
fix: Fix ARM builds

tmathern:
I thought caching was the issue that corrupts the Windows libraries. I removed caching in this PR: https://github.com/contentauth/c2pa-rs/pull/1422. I believe that hypothesis has been proven wrong, as I observed one more corruption in the latest release.

So I continued looking into this. I wanted to check how the Windows ARM files look, since Win ARM is also in the build matrix. But there were none on the release page (for WinARM, only Intel). It turns out, that we attempt a build for Windows ARM, and the build script (make_release.ps1) detects the CPU architecture to put it in the name. For both targets, if detects x86_64 (Intel) architecture, and does not cross-compile. I now believe that those two builds running in parallel override each other's zips on the last upload (since they do have the same zip name - this is an issue that can happen; file collisions if they have the same names), creating the observed ZIP corruption.

I tested the Windows builds from the fork https://github.com/tmathern/c2pa-rs-sandbox/actions with multiple pipeline runs, and the Intel libraries for Windows seem to be OK now. 
(Sidenote: I see the serde build starting to fail again for Linux Android in the fork, but did not look further into that, as I was focusing on Windows. @scouten-adobe We may need to check the fix status of this build if it occurs here too - workaround for now is to leave the build cache OFF).

Fixing the ARM build and having the artifacts with proper names fixes the corruption, since now they don't override each other anymore.

ps: I can only assume initally deactivating the cache actually helped to get "clean" builds when running the builds and sometimes avoid the *timing* issue (since essentially, the file override each other only if they upload in parallel). If one of the two Windows builds uploaded fully before the other, there was no issue. Also it seems there is a cache somewhere that stays for an hour, so waiting an hour before rerunning only the Intel Windows is a workaround to fix the library corruption after the fact.

## Checklist

### DIFF
--- a/.github/workflows/library-release.yml
+++ b/.github/workflows/library-release.yml
@@ -26,8 +26,6 @@ jobs:
             target: x86_64-unknown-linux-gnu
           - os: windows-latest
             target: x86_64-pc-windows-msvc
-          - os: windows-latest
-            target: aarm64-pc-windows-msvc
           - os: macos-latest
             target: aarch64-apple-ios
           - os: macos-latest
@@ -49,8 +47,6 @@ jobs:
 
       - name: Set up Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          cache: false
 
       - name: Build Windows release
         if: matrix.os == 'windows-latest'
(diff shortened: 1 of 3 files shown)


ok-nick (approved):
Nice, this is much better.

PR #1465, opened 2025-10-01
feat: impl `ReaderBuilder`

ok-nick:
Implements a `ReaderBuilder` for constructing a `Reader` with settings and various other options. We may want to consider introducing a similar pattern to the `Builder`, although that may take additional design consideration.

For some background, we need methods of passing in `Settings` to a `Reader` or `Builder` that are instance-local, rather than thread-local. There are numerous reasons why this is needed  that have taken place in other discussions. The reason we decided to introduce a `ReaderBuilder` is because it's impractical and arguably unidiomatic to create or change ~9 of the `Reader` constructors. It also offers much more flexibility and enables many new permutations of constructing a `Reader`.

Blocked by:
- #1444 
- #1447

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -816,11 +816,10 @@ impl BmffHash {
         Ok(())
     }
 
-    #[cfg(feature = "file_io")]
     pub fn verify_stream_segments(
         &self,
         init_stream: &mut dyn CAIRead,
-        fragment_paths: &Vec<std::path::PathBuf>,
+        fragment_paths: &mut [Box<dyn CAIRead>],
         alg: Option<&str>,
     ) -> crate::Result<()> {
         self.verify_self()?;
@@ -849,11 +848,9 @@ impl BmffHash {
                 return Err(Error::HashMismatch("No fragment specified".to_string()));
             }
 
-            for fp in fragment_paths {
-                let mut fragment_stream = std::fs::File::open(fp)?;
-
+            for mut fragment_stream in fragment_paths {
                 // get merkle boxes from segment
-                let c2pa_boxes = read_bmff_c2pa_boxes(&mut fragment_stream)?;
+                let c2pa_boxes = read_bmff_c2pa_boxes(fragment_stream)?;
                 let bmff_merkle = c2pa_boxes.bmff_merkle;
 
                 if bmff_merkle.is_empty() {
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -32,6 +32,7 @@ use serde_with::skip_serializing_none;
 #[cfg(feature = "file_io")]
 use crate::utils::io_utils::uri_to_path;
 use crate::{
+    asset_io::CAIRead,
     crypto::base64,
     dynamic_assertion::PartialClaim,
     error::{Error, Result},
@@ -39,7 +40,7 @@ use crate::{
     jumbf_io,
     manifest::StoreOptions,
     manifest_store_report::ManifestStoreReport,
-    settings::get_settings_value,
+    settings::{get_settings_value, Settings},
     status_tracker::StatusTracker,
     store::Store,
     validation_results::{ValidationResults, ValidationState},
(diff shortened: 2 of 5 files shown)


sdk/src/assertions/bmff_hash.rs
```diff
@@ -816,11 +816,10 @@ impl BmffHash {
         Ok(())
     }
 
-    #[cfg(feature = "file_io")]
     pub fn verify_stream_segments(
         &self,
         init_stream: &mut dyn CAIRead,
-        fragment_paths: &Vec<std::path::PathBuf>,
+        fragment_paths: &mut [Box<dyn CAIRead>],
```
scouten-adobe, line 822: This formulation may be necessary, given the iteration you're doing at line 851, but it _reads_ as you'll be changing the _membership_ of `fragment_paths` rather than you need _mutable access_ to each `fragment_stream`.

I wonder if there's a way to convey that distinction in meaning.

ok-nick: We _can_ store a mutable reference to each stream in the slice although I suspect that would be much more inconvenient than passing a mutable reference to the collection. Let's say you had an owned vec of files, you would have to allocate a new vec that stores mutable references to them

ok-nick: I suppose you would still have to box them anyways, but I think ideally we'd have a generic here so they don't need to be.

sdk/src/reader.rs
```diff
@@ -72,6 +73,173 @@ pub trait AsyncPostValidator {
...
+
+    pub fn with_settings(settings: Settings) -> Self {
+        Self {
+            settings,
+            external_manifest: None,
+            fragments: None,
+        }
+    }
+
+    pub fn external_manifest(
```
gpeacock, line 99: This should be set_external_manifest(), otherwise it looks like a getter in Rust.

sdk/src/store.rs
```diff
@@ -3856,17 +3855,60 @@ impl Store {
...
     /// Load Store from a init and fragments
     /// asset_type: asset extension or mime type
     /// init_segment: reader for the file containing the initialization segments
     /// fragments: list of paths to the fragments to verify
     /// verify: if true will run verification checks when loading, all fragments must verify for Ok status
     /// validation_log: If present all found errors are logged and returned, otherwise first error causes exit and is returned
-    #[cfg(feature = "file_io")]
-    pub fn load_from_file_and_fragments(
+    #[async_generic]
+    pub fn load_fragments_from_stream(
```
gpeacock, line 3908: nice

sdk/src/assertions/bmff_hash.rs
gpeacock: We are generally moving away from CaiRead and CaiWrite as we don't need them anymore. But since this function already uses CAIRead, I think this is ok here.

sdk/src/reader.rs
```diff
@@ -72,6 +73,173 @@ pub trait AsyncPostValidator {
     ) -> Result<Option<Value>>;
 }
 
+pub struct ReaderBuilder {
+    settings: Settings,
+    external_manifest: Option<Box<dyn CAIRead>>,
```
gpeacock, line 78: I think this could be done as a ReaderAsset and include format and stream.

then we can call something like this on Reader

    pub fn from_reader_asset(
        settings: &Settings,
        asset: ReaderAsset,
    ) -> Result<Reader> {
        let format = asset.format;
        let stream = asset.stream;
        if let Some(external_manifest) = asset.external_manifest {
            if let Some(fragments) = asset.fragments {
                Self::from_fragment(
                    &format,
                    stream,
                    external_manifest,
                    fragments,
                )
            } else {
                Self::from_manifest_data_and_stream(
                    &jumbf_io::read_to_end(external_manifest)?,
                    &format,
                    stream,
                )
            }
        } else {
            Self::from_stream(&format, stream)
        }
    }

magaylor:
Are there currently plans to expose this object to the c2pa_c_ffi path? I would love to be able to access this from binding code.

PR #1483, opened 2025-10-08
fix: Turn on trust by default

mauricefisher64:
Change c2patool to never disable trust (can only enable trust) Fix unit tests to use configuration files (now get trust by default) Fix logItem to not automatically write to logger
Fix some error reporting to match the spec better
Fix default test toml to correctly initialize trust anchors

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ The **[Coalition for Content Provenance and Authenticity](https://c2pa.org)** (C
 
 For the best experience, read the docs on the [CAI Open Source SDK documentation website](https://opensource.contentauthenticity.org/docs/rust-sdk/).  
 
+Join the [Content Authenticity Initiative](https://contentauthenticity.org/membership) to connect with a global community advancing the movement for content transparency, get access to the latest ecosystem news, attend events, and access the CAI brand kit to share your involvement. 
+
+Join the [CAI Discord](https://discord.com/invite/CAI) to connect with other implementers, ask questions about the CAI open-source tools, and receive expert guidance on how to build with the C2PA standard.
+
 You can also read the documentation directly in GitHub:
 
 - [Usage](https://github.com/contentauth/c2pa-rs/blob/main/docs/usage.md)


cdmurph32:
I'm confused by the title.
If trust is enabled by default, what is the point of only allowing trust to be enabled and not disabled?

cli/src/main.rs
```diff
@@ -449,6 +449,7 @@ fn configure_sdk(args: &CliArgs) -> Result<()> {
     }
 
     // if any trust setting is provided enable the trust checks
+    // there is no disabling of default setting only the ability to enable if they were internally disabled
```
gpeacock, line 452: How would they be internally disabled? This isn't the sdk, we only build c2pa_tool with this enabled now, so maybe we just get rid or the enable_trust_checks variable?

c2pa_c_ffi/src/json_api.rs
```diff
@@ -100,6 +104,8 @@ mod tests {
 
     #[test]
     fn test_read_from_file_with_base() {
+        let _ = Settings::from_toml(include_str!("../../sdk/tests/fixtures/test_settings.toml"));
```
gpeacock, line 107: Nick said that we could just make these settings the default settings when building for test. I think that would simplify things, but we could to that in a separate PR.

cli/src/main.rs
mauricefisher64: I thought about this.  We could get rid of the enable_trust_checks but then it is unclear what happens when you select trust.  While I don't expect anyone to change the default in the SDK this ensures that it is enabled if you use the trust command.

c2pa_c_ffi/src/json_api.rs
mauricefisher64: They would still need to be loaded.  That was the primary thing I changed.

cli/src/main.rs
gpeacock: well we can fix it in another PR. i'm not sure the command line Trust settings need to stay there anyway, since you can do it with Settings.

PR #1484, opened 2025-10-09
docs: fix and update Builder examples

gpeacock:


sdk/src/builder.rs
```diff
@@ -248,53 +248,32 @@ pub enum BuilderIntent {
...
-///               "name": "c2pa_test",
-///               "version": "1.0.0"
-///           }
-///        ],
-///        "title": "Test_Manifest"
-///     }).to_string();
-///
-///     let mut builder = Builder::from_json(&manifest_json)?;
+///     Settings::from_toml(include_str!("../tests/fixtures/test_settings.toml"))?;
+///     let mut builder = Builder::from_json(r#"{"title": "Test"}"#)?;
```
ok-nick, line 275: How come you removed using the `json!` macro here?

sdk/src/builder.rs
```diff
@@ -248,53 +248,32 @@ pub enum BuilderIntent {
...
-///           {
-///               "name": "c2pa_test",
-///               "version": "1.0.0"
-///           }
-///        ],
-///        "title": "Test_Manifest"
-///     }).to_string();
-///
-///     let mut builder = Builder::from_json(&manifest_json)?;
+///     Settings::from_toml(include_str!("../tests/fixtures/test_settings.toml"))?;
```
ok-nick, line 274: Hm, I think if we want to show a simplified example, we should also show a simplified settings file. I think we should inline it in the example too so it's clear what's going on and what's actually needed.

Edit: I suppose it wouldn't be readable to embed the certs here, but I think as a baseline we should have a minimal settings file.

tmathern: Complete settings are OK here I think. Makes them easy to find if you actually look for others...

tmathern (approved):

crandmck (approved):

sdk/src/builder.rs
ok-nick: It would be helpful then to include a link to the settings file from these docs.

PR #1487, opened 2025-10-13
feat: add CAWG trust args in CLI and user anchors

### LINKED ISSUE #1467: Configure CAWG trust via CLI?
`--trust_anchors` and `--allowed_list` are available `c2patool` options for configuring C2PA trust. I think equivalent flags for configuring CAWG trust are missing, and the only way to configure CAWG trust is via the settings file.

Should we add equivalent options to `c2patool` for configuring CAWG trust?

ok-nick:
This PR introduces the ability to specify a base trust anchor list for both C2PA and CAWG (such as with the official C2PA trust list) and user-provided trust anchors (such as the interim C2PA trust list).

The following args were added:
- `--cawg_user_trust_anchors`
- `--cawg_trust_anchors`
- `--user_trust_anchors`

Open to discussion on how we should pass these args (plus eventually CAWG allowed lists and EKUs) without creating a huge surface of trust args. Note that we do support passing through a settings file, but this introduces additional complexities like how to embed cert lists and how to handle combining them.

---

* Closes #1467


cli/src/main.rs
```diff
@@ -190,15 +190,27 @@ fn parse_resource_string(s: &str) -> Result<TrustResource> {
 enum Commands {
     /// Sub-command to configure trust store options, "trust --help for more details"
     Trust {
-        /// URL or path to file containing list of trust anchors in PEM format
+        /// URL or path to file containing list of user trust anchors in PEM format for CAWG validation
+        #[arg(long = "cawg_user_trust_anchors", env="C2PATOOL_CAWG_USERTRUST_ANCHORS", value_parser = parse_resource_string)]
+        cawg_user_trust_anchors: Option<TrustResource>,
+
+        /// URL or path to file containing list of CAWG trust anchors in PEM format for CAWG validation
+        #[arg(long = "cawg_trust_anchors", env="C2PATOOL_CAWG_TRUST_ANCHORS", value_parser = parse_resource_string)]
```
mauricefisher64, line 198: @scouten-adobe  does CAWG have a distinction between trust anchors and user trust anchors?  I thought it was just a set of user supplied anchors for whatever domain was being used,

PR #1490, opened 2025-10-16
feat: add well-formed validation state

ok-nick:
Adds a well-formed validation state and updates the validation state rules to better match the spec.

I believe the spec is intentionally ambiguous about how a validation state is determined; it's up to the validator. The rules I've implemented are deliberately strict although we may want to consider an alternative design. We could define the set of errors that disqualify a manifest from being well-formed and then closely follow the spec to determine if it's valid and then trusted.

### Questions
* Should we rename `ValidationState::Invalid` to `ValidationState::Malformed`? A well-formed manifest is technically invalid, so this language would be more correct.

### DIFF
--- a/sdk/tests/known_good/XCA.json
+++ b/sdk/tests/known_good/XCA.json
@@ -170,5 +170,5 @@
       }
     ]
   },
-  "validation_state": "Invalid"
+  "validation_state": "Malformed"
 }


sdk/src/validation_results.rs
```diff
@@ -21,20 +23,40 @@ use crate::{
...
+/// See [§14.3. Validation states].
 ///
-/// The Trusted state implies the manifest store is valid and the active signature is trusted.
+/// [§14.3. Validation states]: https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_validation_states
 pub enum ValidationState {
-    /// Errors were found in the manifest store.
+    // REVIEW-NOTE: A "WellFormed" manifest is invalid, should we rename this to "Malformed?"
+    /// The manifest store fails to meet [ValidationState::WellFormed] requirements, meaning it cannot
+    /// even be parsed or its basic structure is non-compliant.
     Invalid,
```
gpeacock: Yes, I think we should change this to MalFormed

sdk/src/validation_results.rs
```diff
@@ -21,20 +23,40 @@ use crate::{
...
 #[cfg_attr(feature = "json_schema", derive(JsonSchema))]
-/// Indicates if the manifest store is valid and trusted.
+/// Represents the levels of assurance a manifest store achives when evaluated against the C2PA
+/// specifications structural, cryptographic, and trust requirements.
+///
+/// See [§14.3. Validation states].
 ///
-/// The Trusted state implies the manifest store is valid and the active signature is trusted.
+/// [§14.3. Validation states]: https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_validation_states
 pub enum ValidationState {
```
gpeacock, line 37: We need to add a NotValidated state since our SDK has an option to not validate. Currently if someone enables that setting, we end up labeling the manifest store as Trusted, but we can't state anything about it other than NotValidated if we don't validate.

sdk/src/validation_results.rs
```diff
@@ -180,29 +202,71 @@ impl ValidationResults {
...
+                    .iter()
+                    .any(|idv| !idv.validation_deltas().failure().is_empty())
             });
-        }
-        if let Some(ingredient_deltas) = self.ingredient_deltas.as_ref() {
-            for idv in ingredient_deltas.iter() {
-                if !idv.validation_deltas().failure().is_empty() {
-                    return ValidationState::Invalid;
-                }
+
```
gpeacock, line 240: I think the better method is to invert the order of these test and then use the logic set in the spec.
Something like this logic

if validation_enabled
    if is_well_formed(self)
          if is_valid(self)
                if is_trusted(self)
                     Trusted
               else
                     Valid
          else
               WellFormed
    else
         MalFormed
else
    NotValidated

ok-nick: Yes I agree, I initially had this, but it's not clear to me what is considered "well-formed." The spec describes a bunch of steps but it doesn't specify exactly which error codes could occur for every case. That's why I think we need to go over it and decide what we consider well-formed.

sdk/src/validation_results.rs
ok-nick: If you disable `verify_after_read` does that also disable reporting validation codes that can determine if the manifest is well-formed? If so, maybe we should report `ValidationState::Unknown`? Otherwise, I think we should still report well-formed and the caller can determine if validation is disabled, that way we provide them with the context that it is still well-formed.

ok-nick: Ok, after doing a little research, disabling verifying disables any indications of well-formed, so we should have another state here. I'll add it.

sdk/src/validation_results.rs
ok-nick: Ok I added this feedback and clarified how well-formed is determined.

sdk/src/builder.rs
```diff
@@ -2355,7 +2358,10 @@ mod tests {
         let manifest_store = Reader::from_file(&dest).expect("from_bytes");
 
         println!("{manifest_store}");
-        assert_ne!(manifest_store.validation_state(), ValidationState::Invalid);
+        assert_ne!(
+            manifest_store.validation_state(),
+            ValidationState::Malformed
```
gpeacock: I wrote some these tests this way before we had Trusted working for unit tests, We should change them to test for Trusted now. - the ones checking for ne!(, validationState_Invalid)

sdk/src/validation_results.rs
```diff
@@ -21,30 +23,54 @@ use crate::{
...
 #[cfg_attr(feature = "json_schema", derive(JsonSchema))]
-/// Indicates if the manifest store is valid and trusted.
-///
-/// The Trusted state implies the manifest store is valid and the active signature is trusted.
 pub enum ValidationState {
-    /// Errors were found in the manifest store.
-    Invalid,
-    /// No errors were found in validation, but the active signature is not trusted.
+    /// Validation is disabled in the SDK and the [ValidationState] is unable to be determined.
+    Unknown,
```
gpeacock, line 39: I guess Unknown is ok as a state, but I think NotValidated is clearer.
We may also use this state for an ingredient without a manifest

ok-nick: The reason I did this is because in the context of the `ValidationResults` it doesn't know if validation is enabled or not in settings, it just knows there are no results. I don't have strong opinions about changing this to NotValidated, but how do you feel about my reasoning?

PR #1502, opened 2025-10-22
fix: timestamp grace for legacy manifests

mauricefisher64:
This PR allows for 1.x manifests to skip TSA trust checks. 
Fixes various other timestamp requirements required by 2.x.  
Fix/update unit tests.  
Fix sample settings files

## Checklist

### DIFF
--- a/.github/workflows/beta-preflight.yml
+++ b/.github/workflows/beta-preflight.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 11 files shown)


sdk/src/store.rs
```diff
@@ -1575,7 +1579,7 @@ impl Store {
 
                     // allow the extra ingredient trust checks
                     // these checks are to prevent the trust spoofing
-                    let check_ingredient_trust: bool = settings.verify.check_ingredient_trust;
+                    let check_ingredient_trust: bool = settings.verify.verify_trust;
```
ok-nick, line 1582: Are we removing the `check_ingredient_trust` feature completely?

tmathern (approved):

sdk/src/store.rs
mauricefisher64: Maybe, this is for a coming change where we always validate ingredient for another reason.

PR #1513, opened 2025-10-23
refactor: Plumb HTTP resolver through internals

ok-nick:
This change builds off #1355 and #1444 to plumb http resolvers throughout all the internals. It also fixes a few issues with the settings plumbing and usage of sync APIs in async APIs.

The last change is removing the `async-recursion` dependency in favor of pinning the return of `Store::ingredient_checks_async`. The dependency had conflicts with usage of the http resolver traits thus leading to this change.


sdk/src/builder.rs
```diff
@@ -1495,6 +1495,11 @@ impl Builder {
         W: Write + Read + Seek + Send,
     {
         let settings = crate::settings::get_settings().unwrap_or_default();
+        let http_resolver = if _sync {
```
scouten-adobe, line 1558: I'd like to explore in a future PR whether the HTTP resolver can/should be configured via Settings. Please add a TO DO to that effect.

sdk/src/claim.rs
```diff
@@ -1827,13 +1828,15 @@ impl Claim {
     /// Verify claim signature, assertion store and asset hashes
     /// claim - claim to be verified
     /// asset_bytes - reference to bytes of the asset
+    #[allow(clippy::too_many_arguments)]
```
scouten-adobe, line 1831: Clippy is not wrong here. Let's explore (again in a future PR) whether we can reduce # of arguments.

sdk/src/claim.rs
```diff
@@ -3919,14 +3927,24 @@ impl Claim {
...
+    settings: &Settings,
+))]
 pub(crate) fn check_ocsp_status(
     sign1: &coset::CoseSign1,
     data: &[u8],
     ctp: &CertificateTrustPolicy,
     ocsp_responses: Option<&Vec<Vec<u8>>>,
     tst_info: Option<&TstInfo>,
     validation_log: &mut StatusTracker,
+    http_resolver: &impl SyncHttpResolver,
```
scouten-adobe, line 3947: I'd also like to explore (again, in a future PR) whether we continue having artificially-synchronous APIs such as this that are hiding the fact that network calls are essentially asynchronous.

sdk/src/reader.rs
```diff
@@ -139,6 +139,11 @@ impl Reader {
     #[cfg(not(target_arch = "wasm32"))]
     pub fn from_stream(format: &str, stream: impl Read + Seek + Send) -> Result<Reader> {
         let settings = crate::settings::get_settings().unwrap_or_default();
+        let http_resolver = if _sync {
```
scouten-adobe, line 143: Hmmm. I'm seeing this pattern a lot. Worth adding http_resolver and async_http_resolver accessors to `settings`?

Not a blocker for this PR; just a thought to explore.

sdk/src/store.rs
```diff
@@ -57,7 +59,7 @@ use crate::{
     error::{Error, Result},
     hash_utils::{hash_by_alg, vec_compare, verify_by_alg},
     hashed_uri::HashedUri,
-    http::{AsyncGenericResolver, AsyncHttpResolver, SyncGenericResolver, SyncHttpResolver},
```
scouten-adobe, line 60: I know this diff is gonna be good … 🙂

sdk/src/store.rs
```diff
@@ -4385,12 +4521,13 @@ pub mod tests {
...
     fn test_jumbf_generation() {
         let settings = Settings::default();
+        let http_resolver = SyncGenericResolver::new();
 
         let (format, mut input_stream, mut output_stream) =
             create_test_streams("earth_apollo17.jpg");
 
         // Create claims store.
-        let mut store = Store::with_settings(&Settings::default());
+        let mut store = Store::with_settings(&settings);
```
scouten-adobe, line 4530: Good catch; thanks!

sdk/src/claim.rs
ok-nick: Yes, if we removed all the artificial sync code it would reduce complexity significantly, even after this change!

sdk/src/claim.rs
ok-nick: Agreed, I think there is some moving around we can do to reduce the complexity here (and in other places we do the same thing).

sdk/src/builder.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1513#discussion_r2470954288.

sdk/src/reader.rs
ok-nick: I considered that approach in my ReaderBuilder PR (#1465), although I wasn't a fan for a few reasons:
- Settings is likely to be deserialized from something, we'd either have to mark the resolver(s) as optional or default them
    - If we default them, we may be initializing an HTTP client the user will override anyways
    - If it's optional, we'd have settings represent an intermediate state when it should actually be a final state (with the resolver initialized)
- We'd either have to add a generic or a boxed trait object to the settings struct
    - The generic would be odd when deserializing from serde
    - The boxed trait object works, but can run us into problems with Send/Sync and WASM again (and other object safety issues). Ideally we wouldn't have to heap allocate and can design an API that allows for generics (in this case separate from settings)

sdk/src/builder.rs
```diff
@@ -1578,10 +1583,10 @@ impl Builder {
 
         // sign and write our store to to the output image file
         if _sync {
-            store.save_to_stream(&format, source, dest, signer, &settings)
+            store.save_to_stream(&format, source, dest, signer, &http_resolver, &settings)
```
gpeacock, line 1586: we could create the resolvers here and and simplify the code a bit.

sdk/src/claim.rs
gpeacock: we could combine several of these parameters in a context struct

sdk/src/reader.rs
gpeacock: I'm thinking we add a C2paContext that can contain both, and it is the context that we pass everywhere.

tmathern: That makes sense, I have a feeling the context could extend in the future to hold more such objects/properties too.

sdk/src/builder.rs
tmathern: Relating to [this comment](https://github.com/contentauth/c2pa-rs/pull/1513#discussion_r2492290139), I wonder if the http_resolver and settings here could go into a context. Since they are both needed for the execution here, and both tweak the execution.

PR #1528, opened 2025-10-27
feat: check if stream matches input format via file signature/structure

ok-nick:
Adds extra validation before loading the JUMBF to confirm the input format type matches the given stream, otherwise outputting a more specific error.

Some format types such as MP3 and SVG may require a more in-depth structural check.

### DIFF
--- a/sdk/src/jumbf_io.rs
+++ b/sdk/src/jumbf_io.rs
@@ -509,12 +509,8 @@ pub mod tests {
             .unwrap();
         removed.set_position(0);
         let result = load_jumbf_from_stream(asset_type, &mut removed);
-        // For c2pa this is expected to return `Error::InvalidFormat`.
-        if asset_type != "wav"
-            && asset_type != "avi"
-            && asset_type != "mp3"
-            && asset_type != "webp"
-            && asset_type != "c2pa"
+        if (asset_type != "wav")
+            && (asset_type != "avi" && asset_type != "mp3" && asset_type != "webp")
         {
             assert!(matches!(&result.err().unwrap(), Error::JumbfNotFound));
         }


sdk/src/asset_handlers/c2pa_io.rs
```diff
@@ -128,6 +128,11 @@ impl AssetIO for C2paIO {
         &SUPPORTED_TYPES
     }
 
+    fn supports_stream(&self, _stream: &mut dyn CAIRead) -> Result<bool> {
+        // TODO: complex
+        Ok(true)
```
mauricefisher64: You can look for the C2PA JUMP header.

sdk/src/asset_handlers/svg_io.rs
```diff
@@ -159,6 +159,11 @@ impl AssetIO for SvgIO {
     fn supported_types(&self) -> &[&str] {
         &SUPPORTED_TYPES
     }
+
+    fn supports_stream(&self, _stream: &mut dyn CAIRead) -> Result<bool> {
+        // TODO: complex
+        Ok(true)
```
mauricefisher64: You could check for this "<?xml version"

ok-nick: I'm looking at a few implementations online, and it seems libraries are reading the first ~4096 bytes and looking for the `<svg` string.

sdk/src/asset_handlers/svg_io.rs
```diff
@@ -159,6 +159,15 @@ impl AssetIO for SvgIO {
     fn supported_types(&self) -> &[&str] {
         &SUPPORTED_TYPES
     }
+
+    fn supports_stream(&self, stream: &mut dyn CAIRead) -> Result<bool> {
+        stream.rewind()?;
+
+        let mut bytes = [0u8; 4096];
```
tmathern: If we think more checks are needed, I would not add it here now. Either split it out, or do the full check. I'd rather we be less restrictive to possibly avoid breaking edge cases now - or due the complete due diligence.

@ok-nick What makes you say more checks may be needed for this case?

sdk/src/asset_handlers/mp3_io.rs
```diff
@@ -344,6 +344,18 @@ impl AssetIO for Mp3IO {
...
     }
+
+    fn supports_stream(&self, stream: &mut dyn CAIRead) -> Result<bool> {
+        stream.rewind()?;
+
+        let mut header = [0u8; 3];
+        stream.read_exact(&mut header)?;
+
+        Ok(
+            (header[0] == 0x49 && header[1] == 0x44 && header[2] == 0x33)
```
tmathern: If we think more checks are needed, I would not add it here now. Either split it out, or do the fill check. I'd rather we be less restrictive to possibly avoid breaking edge cases now - or due the compelte due diligence.

@ok-nick What makes you say more checks may be needed for this case?

tmathern (changes requested):
If we are unsure about mp3 and svg and need more work/verifications, I'd rather we split it out. To avoid taking the risk of breaking (currently working) edge cases.
The corollary is to add good test coverage using real files to raise confidence we covered all cases (is it even possible with mp3?).

sdk/src/asset_handlers/mp3_io.rs
ok-nick: These functions are only used as an extra check to confirm the given stream can possibly be the specified format, if not error with a descriptive error. They are purposefully less-restrictive so that they don't block correct cases. For instance, we check [0x49, 0x44, 0x33] as the starting bytes for MP3, but that may only indicate an ID3 block which may or may not be MP3, we leave the actual parsing up to the parser.

tmathern: Ok!

sdk/src/asset_handlers/svg_io.rs
ok-nick: Hm, for SVG it would be more correct to parse the XML and check for a top-level `svg` tag, but that is quite expensive. Ideally we'd construct the parser once with a stream and we can do the check there as we parse it, then cache all the important info in one pass (this is related to Gavin's content credential proposal), but that's a much larger amount of work.

But to answer your question, this is the least-restrictive check, similar to MP3, as a pre-validation step.

ok-nick:
I will be changing this PR to perform the support checks as it's being read. This will be much more reliable since we can use our normal parsing methods (like XML for SVG) and it can be done in one pass. We also do this already for some formats, like GIF.

sdk/src/asset_handlers/pdf_io.rs
```diff
@@ -28,6 +27,21 @@ pub struct PdfIO {}
 impl CAIReader for PdfIO {
     fn read_cai(&self, asset_reader: &mut dyn CAIRead) -> crate::Result<Vec<u8>> {
         asset_reader.rewind()?;
+
+        let mut header = [0u8; 5];
+        asset_reader.read_exact(&mut header)?;
```
mauricefisher64: This is wrong.  The file is a text file so you could have had spaces or nulls before you get to the %PDF

sdk/src/error.rs
```diff
@@ -370,6 +376,34 @@ pub enum Error {
     // The string should be one of the C2PA validation codes
     #[error("C2PA Validation Error: {0}")]
     C2PAValidation(String),
+
+    #[error("error parsing BMFF: {0}")]
```
mauricefisher64, line 380: Do you plan on adding more error codes for each format specific error?  Seems odd to have made these with just a single error enum for each.

ok-nick: Yes it is odd right now, I agree. Eventually I imagine that all error types returned by (for example) BMFF would be contained in the `BmffError` enum. Any error returned would then include the `error parsing BMFF: some more specific error` context. That would make it much easier to reason about adding more granular errors to the asset handlers without appending it onto the already giant `c2pa::Error` enum.

sdk/src/asset_handlers/pdf_io.rs
ok-nick: I took this behavior from the wikipedia [here](https://en.wikipedia.org/wiki/List_of_file_signatures). Are you saying that for non-spec compliant PDFs or is that part of the spec?

mauricefisher64: I am saying that some PDFs could have space before the %PDF- and it is still valid

ok-nick: Discussed more offline. We decided that the PDF parser can determine if there is a problem and only do file signature checks if it's reliable to do an easy one.

PR #1531, opened 2025-10-28
chore: Reproduce an unexpected behavior

tmathern:
Demonstrates an issue. 

I have added a file in the fixtures that I signed using ACA (https://contentauthenticity.adobe.com/). ACA files have both embedded and remote manifests (cloud). 

I added a test that signs again that image (embedded manifest only, which should be the default behavior). When I look at the generated signed file (the one I signed again using the test), I see it still has some metadata in the XMP, like the URL to the remote cloud manifest of the original file. Especially, I see the cloud manifest URL from the original image using exiftool to read the metadata.

See screenshot below (taken from exiftool reading metadata of a file signed by the new test):
Screenshot 2025-10-27 at 20 35 01

The Reader in the test after signing again does not see an embedded URL anymore either (which I would expect, since it reads the manifest from the embedded data, but still, fyi). It's only observed/visible in the XMP metadata.

For reference, I checked in the generated files:
- in the fixtures folder: original image signed through ACA;
- in the c2pa_c_ffi folder, in tmp test folder: an example of file the added test generates by signing again the ACA-signed in the fixtures folder.

Example run:
```
c2pa_c_ffi % cargo test --features "rust_native_crypto,file_io" test_sign_again_with_info -- --nocapture
   Compiling c2pa-c-ffi v0.68.0 (.../c2pa-rs/c2pa_c_ffi)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1.54s
     Running unittests src/lib.rs (.../c2pa-rs/target/debug/deps/c2pa_c-801bbacb26864c76)

running 1 test
Signed file written to: .../c2pa-rs/c2pa_c_ffi/tmp_folder/IMG_0003_aca_signed.png
No remote URL in manifest
```

## Checklist


tmathern:
(Closing PR since this is a repro example).

PR #1538, opened 2025-10-29
feat(sdk): Introduce new API for file logging in the bindings

magaylor:
This PR sets up file logging in the bindings code so that bindings can utilize and configure the Rust logs

## Checklist
- [ X] This PR represents a single feature, fix, or change.
- [X ] All applicable changes have been documented.
- [X ] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -545,6 +545,9 @@ version = "0.72.0"
 dependencies = [
  "c2pa",
  "cbindgen",
+ "chrono",
+ "fern",
+ "log",
  "scopeguard",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


magaylor:
I like that idea too!

When I wrote this, I assumed there was a better place for it than C_API, as logging to a file (or stream) should not be restricted to the clients consuming the bindings. If you want to reject this and plan to add it to the settings, that works for me.

My team may just use a forked build in the meantime so asking me to debug the Rust isn't the only way to figure out what's happening in the .NET layer. I would be happy to make the contribution myself, but with recent changes to the Settings, I figured this may want to be more planned.

My goal with these recent PRs to the C_API has been to bring its functionality up to match the CLI in areas my team found it was behind. I believe that you can configure console logging from the CLI. That is why I added a PR for Certificate Retrieval from the reader as well, our validator tool has to consume the CLI to retrieve the certificates without a method in the C_API.

PR #1541, opened 2025-10-30
feat: Refine validation state logic

ok-nick:
This PR makes validation state checks stricter.


sdk/src/validation_results.rs
```diff
@@ -180,29 +202,43 @@ impl ValidationResults {
...
     pub fn validation_state(&self) -> ValidationState {
-        let mut is_trusted = true; // Assume the state is trusted until proven otherwise
         if let Some(active_manifest) = self.active_manifest.as_ref() {
-            if !active_manifest.failure().is_empty() {
-                return ValidationState::Invalid;
-            }
-            // There must be a trusted credential in the active manifest for the state to be trusted
-            is_trusted = active_manifest.success().iter().any(|status| {
-                status.code() == crate::validation_status::SIGNING_CREDENTIAL_TRUSTED
+            let success_codes: HashSet<&str> = active_manifest
```
gpeacock, line 213: You can call active_manifest.success() here just like we do for failure below.

sdk/src/validation_results.rs
```diff
@@ -21,30 +23,46 @@ use crate::{
...
+/// [§14.3. Validation states]: https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_validation_states
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[cfg_attr(feature = "json_schema", derive(JsonSchema))]
-/// Indicates if the manifest store is valid and trusted.
-///
-/// The Trusted state implies the manifest store is valid and the active signature is trusted.
 pub enum ValidationState {
-    /// Errors were found in the manifest store.
+    /// The manifest store fails to meet [ValidationState::WellFormed] requirements, meaning it cannot
+    /// even be parsed or its basic structure is non-compliant.
```
gpeacock, line 39: If we are returning InValid for not validated, we should add a note about that here.

sdk/src/validation_results.rs
```diff
@@ -180,29 +202,43 @@ impl ValidationResults {
...
+            let success_codes: HashSet<&str> = active_manifest
+                .success()
+                .iter()
+                .map(|status| status.code())
+                .collect();
+            let failure_codes = active_manifest.failure();
+            let ingredient_failure = self.ingredient_deltas.as_ref().is_some_and(|deltas| {
+                deltas
+                    .iter()
+                    .any(|idv| !idv.validation_deltas().failure().is_empty())
```
gpeacock, line 222: I didn't realize I'd missed this, now I understand your questions better. We had best test this using make images to ensure we don't have surprising ingredient cases.

PR #1543, opened 2025-10-30
feat: add Unknown validation state

ok-nick:
Adds a new Unknown state for when validation is disabled. Previously, the WellFormed state was added in #1490, but its definition, as well as that of asset state, is unclear. Builds off #1541.


sdk/src/validation_results.rs
```diff
@@ -21,30 +23,48 @@ use crate::{
...
-/// The Trusted state implies the manifest store is valid and the active signature is trusted.
 pub enum ValidationState {
-    /// Errors were found in the manifest store.
+    /// Validation is disabled in the SDK and the [ValidationState] is unable to be determined.
+    Unknown,
+    /// The manifest store fails to meet [ValidationState::WellFormed] requirements, meaning it cannot
+    /// even be parsed or its basic structure is non-compliant.
     Invalid,
-    /// No errors were found in validation, but the active signature is not trusted.
+    /// The manifest store is well-formed and the cryptographic integrity checks succeed.
```
tmathern, line 43: I wonder if it would be worthwhile to add "WellFormed" as standalone too? Or is it too much? (While at it...). Since `Valid` is more stringent than WellFormed.

ok-nick: The `WellFormed` state isn't clearly defined and based on its current definition does not have a clear use case. We opted to implement a strict `Valid` state to at least encompass what we know is right.

sdk/src/validation_results.rs
```diff
@@ -238,6 +238,10 @@ impl ValidationResults {
             } else if is_valid {
                 return ValidationState::Valid;
             }
+        } else {
+            // REVIEW-NOTE: is this the best way to detect that it wasn't validated? should we also check if success/failure is empty if there
```
gpeacock, line 242: Does this work? I'm not sure if there will always be no active manifest in validation_results if we did not validate.

ok-nick: > /// Validation status codes for the ingredient's active manifest. Present if ingredient is a C2PA
    /// asset. Not present if the ingredient is not a C2PA asset.
    #[serde(rename = "activeManifest", skip_serializing_if = "Option::is_none")]
    active_manifest: Option\<StatusCodes\>,

I think it's a safe assumption that if there's no active_manifest then it hasn't been validated, or at the very least it couldn't be determined. The only problem is that we populate the validation results on a `Reader::post_validate`, so it's possible that normal validation is disabled but since it was post validated the state shows as invalid (or possibly even valid?).

ok-nick: There has been discussion about separating CAWG (post-)validation results from C2PA validation results. I will put this PR on hold for now until that's addressed.

PR #1547, opened 2025-10-31
feat: Remove x509_certificate

cdmurph32:
Removing x509_certificate will allow us to remove ring as a dependency and provide Wasm and C bindings more easily.
This replaces x509_certificate functionality with RustCrypto, which is already used for the rust_native_tls feature.
The ASN1 structure remains based on the bcder crate, even though RustCrypto uses the der crate. There is a bunch of compatibility code added. Eventually we can attempt to remove bcder altogether.

## Checklist

### DIFF
--- a/sdk/src/crypto/asn1/mod.rs
+++ b/sdk/src/crypto/asn1/mod.rs
@@ -256,6 +256,9 @@ impl GeneralizedTime {
     }
 
     /// Parse from bcder Primitive (no fractional or timezone offsets)
+    ///
+    /// This is used by CMS Time structures (RFC 5652) which use strict DER encoding.
+    #[allow(dead_code)] // Used by rfc5652::Time::take_from
     pub fn from_primitive_no_fractional_or_timezone_offsets<S: bcder::decode::Source>(
         prim: &mut bcder::decode::Primitive<S>,
     ) -> Result<Self, bcder::decode::DecodeError<S::Error>> {
@@ -914,7 +917,7 @@ mod tests {
         })
         .expect("Failed to parse Extensions with SEQUENCE tag");
 
-        assert!(extensions_explicit.0.as_slice().len() > 0);
+        assert!(!extensions_explicit.0.as_slice().is_empty());
 
         // Test 2: Parse Extensions with IMPLICIT [1] tagging using from_constructed()
         // This is how RFC 3161 uses it: extensions [1] IMPLICIT Extensions
@@ -929,7 +932,7 @@ mod tests {
             "Extensions should be present"
         );
         let extensions_implicit = extensions_implicit.unwrap();
-        assert!(extensions_implicit.0.as_slice().len() > 0);
+        assert!(!extensions_implicit.0.as_slice().is_empty());
 
         // Verify the tag is CTX_1, not SEQUENCE
         let (tag, _) =
@@ -1001,7 +1004,7 @@ mod tests {
         // Test parsing as it's done in TstInfo::take_from()
         let general_name_dns =
             Constructed::decode(TSA_GENERAL_NAME_DNS, bcder::Mode::Der, |cons| {
-                cons.take_opt_constructed_if(Tag::CTX_0, |cons| GeneralName::take_from(cons))
+                cons.take_opt_constructed_if(Tag::CTX_0, GeneralName::take_from)
             })
             .expect("Failed to decode TSA with dNSName");
 
@@ -1013,7 +1016,7 @@ mod tests {
         // Test URI variant
         let general_name_uri =
             Constructed::decode(TSA_GENERAL_NAME_URI, bcder::Mode::Der, |cons| {
-                cons.take_opt_constructed_if(Tag::CTX_0, |cons| GeneralName::take_from(cons))
+                cons.take_opt_constructed_if(Tag::CTX_0, GeneralName::take_from)
             })
             .expect("Failed to decode TSA with URI");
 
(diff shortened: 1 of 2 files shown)


sdk/src/assertions/assertion_metadata.rs
```diff
@@ -53,11 +53,26 @@ impl AssertionMetadata {
...
+            // On wasm32-unknown-unknown, use js_sys::Date
+            use chrono::TimeZone;
+            let js_date = js_sys::Date::new_0();
+            let millis = js_date.get_time() as i64;
+            let dt = Utc
+                .timestamp_millis_opt(millis)
+                .single()
+                .unwrap_or_else(|| Utc.timestamp_opt(0, 0).unwrap());
+            dt.to_rfc3339_opts(SecondsFormat::Millis, true)
+        };
```
ok-nick: `chrono` has a `js-sys` feature. We may not need platform-specific behavior here (and in the other places this is done).

sdk/src/crypto/asn1/mod.rs
```diff
@@ -10,9 +10,288 @@
 // cryptographic-message-syntax crate located at:
 // https://github.com/indygreg/cryptography-rs/tree/main/cryptographic-message-syntax/src/asn1
 
-// We can not incorporate the entire crate directly because other parts of the
-// crate contain dependencies on blocking calls in reqwest. Those calls are not
-// available in WASM environment.
+// We can not incorporate the entire crate directly because it relies on x509-certificate which
+// relies on ring.
```
ok-nick, line 14: Is this comment outdated now?

sdk/src/crypto/asn1/mod.rs
```diff
@@ -10,9 +10,288 @@
...
+pub struct AlgorithmIdentifier {
+    /// The algorithm OID
+    pub algorithm: Oid,
+}
+
+impl PartialEq for AlgorithmIdentifier {
+    fn eq(&self, other: &Self) -> bool {
+        self.algorithm == other.algorithm
+    }
+}
```
ok-nick: I believe you can derive `PartialEq` on the struct since `Oid` implements it already.

sdk/src/crypto/asn1/mod.rs
```diff
@@ -10,9 +10,288 @@
...
+
+impl Eq for AlgorithmIdentifier {}
```
ok-nick: Same here.

sdk/src/crypto/asn1/mod.rs
```diff
@@ -10,9 +10,288 @@
...
+#[derive(Clone, Debug)]
+pub struct Extensions(Captured);
+
+impl PartialEq for Extensions {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.as_slice() == other.0.as_slice()
+    }
+}
+
+impl Eq for Extensions {}
```
ok-nick, line 134: Consider this case as well (and the others below).

sdk/Cargo.toml
```diff
@@ -105,8 +104,8 @@ config = { version = "0.14.0", default-features = false, features = [
 const-hex = "1.14"
 const-oid = { version = "0.9.6", optional = true }
 coset = "0.4.0"
+der = { version = "0.7.9", features = ["alloc", "derive"] }
```
ok-nick: Do we use the `derive` functionality in this PR?

sdk/tests/test_failures.rs
```diff
@@ -7,6 +7,10 @@ fn test_reader_ts_changed() -> Result<()> {
     let (format, mut stream) = fixture_stream("CA_ct.jpg")?;
     let reader = Reader::from_stream(&format, &mut stream).unwrap();
     // in the older validation statuses, this was an error, but now it is informational
+    // TODO: Make this consistent with how it was before
```
scouten-adobe: Please add a Jira ticket for this TODO item.

sdk/src/crypto/asn1/mod.rs
mauricefisher64: This is real dangerous stuff here.  We have already had to fix incorrect deconstruction in this code.  I don't feel good about this.  Please look at the AlgorithmIdendifier to make sure this is correct.

sdk/src/crypto/asn1/rfc5652.rs
```diff
@@ -20,15 +20,190 @@ use std::{
...
+            use chrono::TimeZone;
+            let duration = system_time
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap_or(std::time::Duration::from_secs(0));
+            chrono::Utc
+                .timestamp_opt(duration.as_secs() as i64, duration.subsec_nanos())
+                .single()
+                .unwrap_or_else(|| chrono::Utc.timestamp_opt(0, 0).unwrap())
+        }
+    }
```
mauricefisher64: Are there any tests to make sure you got this right?  This stuff is easy to mess up.

sdk/src/assertions/assertion_metadata.rs
cdmurph32: Good point. I need to split out the emscripten target into its own logic.

sdk/tests/test_failures.rs
cdmurph32: I made the behavior consistent, but forgot to delete the comment.

scouten-adobe (approved):

sdk/src/assertions/assertion_metadata.rs
cdmurph32: Adding the 'clock' feature to chrono for wasm32-unknown-unknown resolved the need for a different code path.

sdk/src/crypto/asn1/mod.rs
cdmurph32: Sorry. It's correct. We can't incorporate the asn1 crate because it relies on x509-certificate and ring, which we are trying to get rid of.

sdk/Cargo.toml
cdmurph32: No. It's been removed. Thanks

sdk/src/crypto/asn1/mod.rs
cdmurph32: I've generated real TSA requests with openSSL and tested with them. I don't know what more I can do.

PR #1550, opened 2025-10-31
fix: add JsonSchema to Action

emensch:

## Checklist

### DIFF
--- a/sdk/src/assertions/actions.rs
+++ b/sdk/src/assertions/actions.rs
@@ -315,7 +315,8 @@ pub struct ActionParameters {
 
     /// Anything from the common parameters.
     #[serde(flatten)]
-    #[schemars(skip)] // JsonSchema does not support CBOR values
+    // JsonSchema does not support CBOR values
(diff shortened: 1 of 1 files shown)


sdk/src/assertions/actions.rs
```diff
@@ -313,6 +315,7 @@ pub struct ActionParameters {
 
     /// Anything from the common parameters.
     #[serde(flatten)]
+    #[schemars(skip)] // JsonSchema does not support CBOR values
```
tmathern: Will this be totally fine all the time?

emensch: I've gated it behind `json_schema` (which I needed to do anyway for the code to be valid), and I believe the functionality is OK.

emensch: Additionally, I understand the "common" field on the struct to mean "all possible CBOR values" which I believe accounts for the "custom fields" as described https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_parameters. 

I believe the generated schema will still allow that, as it has the `"additionalProperties": true` field.

PR #1552, opened 2025-11-01
NO-MERGE: Investigate release-plz compilation issue

tmathern:
Some tests need file that are excluded at release compilation time, make builds fail at the release-plz step. 

This shows the current issue about the files. 
Likely a different solution is needed than using those directly in the implemented code. 

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -342,6 +342,47 @@ impl AsRef<Builder> for Builder {
 }
 
 impl Builder {
+    /// Test certificate chain used for signing working store manifests.
+    /// This is a test-only certificate and should not be used in production.
+    const WORKING_STORE_CERT_CHAIN: &'static [u8] = b"\
+-----BEGIN CERTIFICATE-----
+MIICSDCCAfqgAwIBAgIUb+aBTX1CsjJ1iuMJ9kRudz/7qEcwBQYDK2VwMIGMMQsw
+CQYDVQQGEwJVUzELMAkGA1UECAwCQ0ExEjAQBgNVBAcMCVNvbWV3aGVyZTEnMCUG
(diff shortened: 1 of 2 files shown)


sdk/src/builder.rs
```diff
@@ -342,6 +342,47 @@ impl AsRef<Builder> for Builder {
 }
 
 impl Builder {
+    /// Test certificate chain used for signing working store manifests.
+    /// This is a test-only certificate and should not be used in production.
```
tmathern, line 346: This is the impl, not a test... I wouldn't include it here. Neither embedded nor through include_bytes. Could we add them as parameters instead, and pass them in?

sdk/src/builder.rs
```diff
@@ -1755,8 +1796,8 @@ impl Builder {
 
     /// We use this signer to generate working store manifests
     pub(crate) fn working_store_signer() -> Result<Box<dyn Signer>> {
-        let cert_chain = include_bytes!("../tests/fixtures/certs/ed25519.pub");
-        let private_key = include_bytes!("../tests/fixtures/certs/ed25519.pem");
+        let cert_chain = Self::WORKING_STORE_CERT_CHAIN;
```
tmathern, line 1799: https://github.com/contentauth/c2pa-rs/pull/1476/files#diff-68d7dd6679b5f8e447131edd1d4558cd39ec3d3c94c9f706824fbd0f1642cb26R1758
(Original using include_bytes)

sdk/src/builder.rs
```diff
@@ -1755,8 +1796,8 @@ impl Builder {
...
+        let private_key = Self::WORKING_STORE_PRIVATE_KEY;
```
tmathern, line 1800: https://github.com/contentauth/c2pa-rs/pull/1476/files#diff-68d7dd6679b5f8e447131edd1d4558cd39ec3d3c94c9f706824fbd0f1642cb26R1759
(Original using include_bytes)

PR #1560, opened 2025-11-04
fix: Allow C2PA archives to be unsigned

gpeacock:
We must read without validation

Note that you can still sign a builder as application/c2pa if desired but to_archive will not sign archives.

## Checklist

### DIFF
--- a/.github/workflows/beta-preflight.yml
+++ b/.github/workflows/beta-preflight.yml
@@ -200,12 +200,26 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@beta
 
-      - name: Install wasm-pack
-        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      - name: Set up Chrome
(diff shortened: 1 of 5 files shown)


sdk/src/builder.rs
```diff
@@ -1790,8 +1791,9 @@ impl Builder {
         let mut store = Store::new();
         store.commit_claim(claim)?;
 
-        let signer = Self::working_store_signer()?;
-        store.get_box_hashed_embeddable_manifest(signer.as_ref(), settings)
+        //store.to_jumbf_internal(1000)
+        store.get_data_hashed_manifest_placeholder(100, "application/c2pa")
```
tmathern, line 1795: How did you find/pick the 100? Educated guess?

gpeacock: It doesn't matter since there's no signature we are adding. It would probably work with 0 but sometimes you run into edge cases, so I just picked a small but non-zero size. I looked into creating a well structured but fake signature, but that was too complicated.

PR #1568, opened 2025-11-06
feat(sdk): Introduce new API for adding custom TSA callbacks

magaylor:
Adding custom TSA callbacks to the Rust SDK through the CallbackSigner.

Trusted Signing Service (TSS) has implemented a custom C2PA signing and timestamping endpoint that they call their unified API. The unified API builds the correct countersignature structure and timestamps it, returning the signature of the data and its timestamp. When using this custom signer, we found that adding a Timestamp callback that fetches the cached timestamp value has significant performance improvements (~400 ms observed in testing) by avoiding the extra network round-trip.

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -270,6 +270,34 @@ pub type SignerCallback = unsafe extern "C" fn(
     signed_len: usize,
 ) -> isize;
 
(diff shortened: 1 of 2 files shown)


scouten-adobe:
We're current working to plumb HTTP handlers through to all of the places where these calls are made, including TSA, so we'll close this in favor of that effort.

PR #1569, opened 2025-11-06
fix: Better support for remote-only manifests for non-BMFF assets

tmathern:
This is a (maybe hacky?) tentative fix to include CAWG data through dynamic assertions in remote manifest stores when the store is remote-only.
The follow-up fix is to have those verify their hashes. Since signing is in two steps, the addition here is to first calculate the size eneded for the hash; then for remote manifests re-hash the whole thing (since there is no embeddings to hash around).

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -20,7 +20,6 @@ use std::{
 };
 
 use async_generic::async_generic;
-use async_recursion::async_recursion;
 use log::error;
 
 #[cfg(feature = "file_io")]
@@ -38,7 +37,10 @@ use crate::{
     asset_io::{
         CAIRead, CAIReadWrite, HashBlockObjectType, HashObjectPositions, RemoteRefEmbedType,
     },
-    claim::{check_ocsp_status, Claim, ClaimAssertion, ClaimAssetData, RemoteManifest},
+    claim::{
+        check_ocsp_status, check_ocsp_status_async, Claim, ClaimAssertion, ClaimAssetData,
+        RemoteManifest,
+    },
     cose_sign::{cose_sign, cose_sign_async},
     cose_validator::{verify_cose, verify_cose_async},
     crypto::{
@@ -57,6 +59,7 @@ use crate::{
     error::{Error, Result},
     hash_utils::{hash_by_alg, vec_compare, verify_by_alg},
     hashed_uri::HashedUri,
+    http::{AsyncHttpResolver, SyncHttpResolver},
     jumbf::{
         self,
         boxes::*,
@@ -87,7 +90,6 @@ use crate::{
 };
 
 const MANIFEST_STORE_EXT: &str = "c2pa"; // file extension for external manifests
-#[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))] // Browser manages fetch & memory
 const DEFAULT_MANIFEST_RESPONSE_SIZE: usize = 10 * 1024 * 1024; // 10 MB
 
 pub(crate) struct ManifestHashes {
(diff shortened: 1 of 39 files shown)


sdk/src/store.rs
```diff
@@ -2189,11 +2189,6 @@ impl Store {
     where
         R: Read + Seek + ?Sized,
     {
-        if block_locations.is_empty() {
```
tmathern, line 2256: Let's try actually moving on

sdk/src/store.rs
```diff
@@ -2221,10 +2216,13 @@ impl Store {
...
-        if found_jumbf {
-            // add exclusion hash for bytes before and after jumbf
-            let mut dh = DataHash::new("jumbf manifest", alg);
+        // Create a DataHash regardless of whether JUMBF is found or not...
+        // For remote manifests with no embedded JUMBF, creates a hash with no exclusions,
+        // because there is nothing to exclude from the hashing (since nothing is embedded)
+        let mut dh = DataHash::new("jumbf manifest", alg);
 
+        if found_jumbf {
+            // add exclusion for embedded jumbf
```
tmathern, line 2289: Only for embedded case...

sdk/src/store.rs
```diff
@@ -2242,26 +2240,32 @@ impl Store {
...
+        }
 
-                match alg {
-                    "sha256" => dh.set_hash([0u8; 32].to_vec()),
-                    "sha384" => dh.set_hash([0u8; 48].to_vec()),
-                    "sha512" => dh.set_hash([0u8; 64].to_vec()),
-                    _ => return Err(Error::UnsupportedType),
-                }
+        // Generate or set placeholder hash
+        if calc_hashes {
```
tmathern, line 2316: ... And we deal with hashes in more cases, even the remote one

sdk/src/store.rs
```diff
@@ -3006,7 +3011,18 @@ impl Store {
                         &jumbf_bytes,
                     )?;
                 }
-                _ => (),
+                RemoteManifest::SideCar | RemoteManifest::Remote(_) => {
+                    // we are going to handle the JUMBF like we'd embed, but we won't
+                    // eventually we won't embed it, so this is a temporary hack to get the code to work
+                    println!("save_to_stream, SideCar | Remote, updating JUMBF without embedding");
```
mauricefisher64: Make sure to remove the print statements.

mauricefisher64 (approved):
Looks good.  Nice work.  Make sure to test this against BMFF files (MP4, AVIF, etc.) since they use a different hashing path.

sdk/src/store.rs
```diff
@@ -3006,7 +3011,18 @@ impl Store {
...
+
+                    // Update the JUMBF like it would normally be done
+                    jumbf_bytes = self.to_jumbf_internal(signer.reserve_size())?;
+
+                    // Intermediate stream goes to output, but still no embedding
+                    intermediate_stream.rewind()?;
+                    std::io::copy(&mut intermediate_stream, output_stream)?;
```
mauricefisher64: It this io::copy needed?

gpeacock: We still need to output the updated asset don't we?

sdk/src/store.rs
tmathern: Sure sure. I kept them for now for testing/debugging to double-check the code paths. They'll be gone when it's ready to merge.

sdk/src/store.rs
tmathern: I think it is needed for the output? Even if we don't use it to embed?

tmathern: I've got good test coverage. Let me verify what happens without the copy, data will answer the question.

tmathern: @mauricefisher64 is right. All tests pass without the copy here.

PR #1573, opened 2025-11-07
docs: Edit settings doc

crandmck:
I still have some questions I'd like to resolve before landing this, e.g.

- How does an specify this JSON file?  I see there's a reference to the test_settings.json in [sdk/examples/api.rs](https://github.com/contentauth/c2pa-rs/blob/c1aad99903b8dfe6f19f9ba4f55889127b1e3939/sdk/examples/api.rs#L87) but what about language bindings?
- `builder.claim_generator_info` is not explained well.  It says it's an object, and I assume the properties are those shown in https://opensource.contentauthenticity.org/docs/manifest/json-ref/reader#claimgeneratorinfo but we should be explicit. Does it have a default value?  What happens when it's not specified?
- Is there a default value for `core.merkle_tree_chunk_size_in_kb` ?
- Is there any diff between specifying a value of 'null' vs. not specifying a property at all?

Changes:
- Put property reference into tables.
- Alphabetize reference sections.
- Add notes in relevant places instead of at the end.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -161,13 +161,11 @@ The `builder.claim_generator_info` specifies the default claim generator informa
 Only the `name` property is required. For example:
 
 ```json
-"claim_generator_info": [
   {
     "name": "Adobe Content Authenticity",
     "com.adobe.aca-version": "81c4a25",
     "org.cai.c2pa_rs": "0.49.3"
   }
-]
 ```
 
 ### cawg_trust


gpeacock:
> How does an specify this JSON file? I see there's a reference to the test_settings.json in [sdk/examples/api.rs](https://github.com/contentauth/c2pa-rs/blob/c1aad99903b8dfe6f19f9ba4f55889127b1e3939/sdk/examples/api.rs#L87) but what about language bindings?

you can use the now un-deprecated Settings::from_string and specify the format as "json"
Settings::from_string(include_str!("fixtures/test_settings.json"), "json")?;

> builder.claim_generator_info is not explained well. It says it's an object, and I assume the properties are those shown in https://opensource.contentauthenticity.org/docs/manifest/json-ref/reader#claimgeneratorinfo but we should be explicit. Does it have a default value? What happens when it's not specified?

This is exactly the same as claim_generator_info in the manifest definition. If you supply one here, you don't have to supply it there. 
There is a default, which just uses the sdk name and version, but I should probably remove that and make it a hard requirement to either supply  claim_generator_info here or in your manifest_definition.

> Is there a default value for core.merkle_tree_chunk_size_in_kb ?

yes, but IDK what it is. You definitely don't have to set this and you probably shouldn't unless you really know what it means!

> Is there any diff between specifying a value of 'null' vs. not specifying a property at all?

Yes, not providing a value means use the default, setting it to null means use null, and not the default.
Settings can be overlapped on each other, not specifying a value means do not change it. Setting a value to null means you want it to be null and have no value.

docs/settings.md
```diff
@@ -96,132 +96,167 @@ Here's the JSON with all default values:
...
-- **remote_manifest_fetch** (`boolean`, default: `true`): Fetch remote manifests
-- **check_ingredient_trust** (`boolean`, default: `true`): Verify ingredient certificates
-- **skip_ingredient_conflict_resolution** (`boolean`, default: `false`): Skip ingredient conflict resolution
-- **strict_v1_validation** (`boolean`, default: `false`): Use strict C2PA v1 validation
+| Property | Type | Default value | Description |
+| --- | --- | --- | --- |
+| `core.merkle_tree_chunk_size_in_kb` | Number | ? | Chunk size for BMFF hash Merkle trees in KB |
+| `core.merkle_tree_max_proofs` | Number | 5 | Maximum Merkle tree proofs when validating |
+| `core.backing_store_memory_threshold_in_mb` | Number | 512 | Memory threshold before using disk storage (MB) |
+| `core.decode_identity_assertions` | Boolean | true | Whether to decode CAWG identity assertions |
```
gpeacock: This seems like it should be a cawg setting since identity assertions are a CAWG feature.

gpeacock (approved):

docs/settings.md
crandmck: This was based on the info in the file in main.  Is this part of the doc incorrect?  
Or are you just saying you want to make that change in the implementation?

PR #1576, opened 2025-11-07
fix: stack overflow and hang due to cyclic ingredients

### LINKED ISSUE #1554: Altered Manifest Undetected or Segfaults
Active manifest is tracked by a unique identifier.
The identifier is always part of the exclusion list, so it is not cryptographically protected.
What if we change it?

For the test source, I used a picture from a Google Pixel 10. This file has one dependency and fully validates.
Image

**Test #1: Alter active manifest to random data.**

With this test file, the active manifest is "urn:c2pa:959d8fa1-fe19-41d9-adf6-489b72b5389b".
```
cp test-source.jpg test-arbitrary.jpg
sed -i 's@c2pa:959d8fa1-fe19-41d9-adf6-489b72b5389b@Neal:Wuz-here-Neal-wuz-here-Neal-wuz-here@g' test-arbitrary.jpg
c2patool-0.23.4 test-arbitrary.jpg trust --trust_anchors C2PA-TRUST-LIST.pem > test-arbitrary.jpg.c2pa
```

Result #1: No error and no alteration detected.

Problem #1: In the CAI Discord and in conversations with C2PA representatives, I've asked how to distinguish C2PA versions. The repeated advice is to look at the urn. If it be begins with "urn:c2pa:..." then it is v2.2. But in this case, I have changed the string from v2.2 to something different, and nothing noticed.

**Test #2: Hostile manifest**
With this file, there is a dependency with it's own manifest id: "urn:c2pa:76408e7a-49e6-6cff-7084-4768507ccbed"
If the active manifest can be anything, what if I change it to be the same as the dependent manifest?
```
cp test-source.jpg test-duplicate1.jpg
sed -i 's@c2pa:959d8fa1-fe19-41d9-adf6-489b72b5389b@c2pa:76408e7a-49e6-6cff-7084-4768507ccbed@g' test-duplicate1.jpg
c2patool-0.23.4 test-duplicate1.jpg trust --trust_anchors C2PA-TRUST-LIST.pem > test-duplicate1.jpg.c2pa
```

Result #2: c2patool segfaults.
(I've never seen rust segfault!)
Specifcally:
c2patool 0.11.1 says "Error: claim could not be converted from CBOR"
c2patool 0.13.0 - 0.14.0 works. "assertion.required.missing" but the manifest is valid.
c2patool 0.16.4 says "Error: claim missing: label = urn:c2pa:76408e7a-49e6-6cff-7084-4768507ccbed"
c2patool 0.16.5 - 0.18.0 hangs indefinitely. (infinite loop?)
c2patool 0.19.0 - 0.23.4 segfaults.

**Test #3: Hostile manifest**
What if I change the dependency to have the same manifest as the active manifest?
```
cp test-source.jpg test-duplicate2.jpg
sed -i 's@c2pa:76408e7a-49e6-6cff-7084-4768507ccbed@c2pa:959d8fa1-fe19-41d9-adf6-489b72b5389b@g' test-duplicate2.jpg
c2patool-0.23.4 test-duplicate2.jpg trust --trust_anchors C2PA-TRUST-LIST.pem > test-duplicate2.jpg.c2pa
```

Result #3: Same as Result #2; c2patool segfaults, older versions hang.

ok-nick:
Fixes stack overflow and hang due to maliciously modifying an ingredient's active manifest hashed URI to reference its parent manifest's hashed URI. 

* Partially fixes #1554

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -934,7 +934,6 @@ impl Reader {
             // Add ingredients to stack for processing
             for ingredient in manifest.ingredients().iter() {
                 if let Some(label) = ingredient.active_manifest() {
-                    // REVIEW-NOTE: if we already saw this manifest, do we need to compute it again here?
                     if !seen.contains(label) {
                         let ingredient_uri = crate::jumbf::labels::to_assertion_uri(
                             &current_label,
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -4040,7 +4040,6 @@ impl Store {
     ) -> Result<()> {
         let claim_label = claim.label();
 
-        // REVIEW-NOTE: added this so we don't compute the same ingredients more than once.
         if svi.manifest_map.contains_key(claim_label) {
             return Ok(());
         }


sdk/src/reader.rs
```diff
@@ -931,11 +934,14 @@ impl Reader {
...
-                    );
-                    stack.push((label.to_string(), Some(ingredient_uri)));
+                    // REVIEW-NOTE: should we error if there's a cyclic ingredient?
+                    if !seen.contains(label) {
+                        let ingredient_uri = crate::jumbf::labels::to_assertion_uri(
+                            &current_label,
+                            ingredient.label().unwrap_or("unknown"),
+                        );
+                        stack.push((label.to_string(), Some(ingredient_uri)));
+                    }
```
mauricefisher64, line 943: I don't think this is a Reader check.  It is a manifest validation check.  Loading the manifest should return an error.

ok-nick: If validation after reading is disabled and you try to post-validate the reader, it will hang due to the way we resolve dynamic assertions. On the other hand, if validation after reading is enabled, the previous behavior will cause a stack overflow in the validation logic.

I also left a note, currently if we let cyclic ingredient's pass there will be a hash mismatch, although we can do better and supply an explicit error. Should it go in the validation log or hard stop validation?

mauricefisher64: That is not what I am saying.  It should be a manifest load error.  You should never get back a manifest to process.

sdk/src/store.rs
```diff
@@ -4051,13 +4069,17 @@ impl Store {
                     .insert(claim_label.clone());
 
                 // recurse nested ingredients
-                if recurse {
-                    Store::get_claim_referenced_manifests(
+                // REVIEW-NOTE: add an error to the validation log if seen already?
+                if recurse && !seen.contains(ingredient.label()) {
+                    seen.insert(ingredient.label());
+
```
mauricefisher64: Here is where we want to fail.  This code is not quite right though.  This will tell you if have seen the value while traversing one node in a depth first search.  But it will give you false positives.  Here is how you do this:  https://www.geeksforgeeks.org/dsa/detect-cycle-in-a-graph/

mauricefisher64: This must be an error when detected!

sdk/src/reader.rs
mauricefisher64: There is nothing stopping you from calling Store::get_store_validation_info to see if there is a circular reference problem.

sdk/src/store.rs
tmathern: (Don't forget to add positive and negative tests, try to break your algorithm yourself to make sure it works!)

mauricefisher64: Here is an example of what I logged:

`log_item!(
                        jumbf::labels::to_assertion_uri(claim.label(), &i.label()),
                        "ingredient cannot reference self",
                        "ingredient_checks"
                    )
                    .validation_status(validation_status::ASSERTION_INGREDIENT_MALFORMED)
                    .failure(
                        validation_log,
                        Error::HashMismatch("ingredient cannot refernce self".to_string()),
                    )?;
`

tmathern:
Does https://github.com/contentauth/c2pa-rs/pull/1582 change anything in regards to this PR?

sdk/src/store.rs
ok-nick: Ok, I figured we didn't need to compute the same ingredient if we'd seen it already here. I'll add a separate validation step for detecting cycles and erroring then.

sdk/src/reader.rs
ok-nick: I made it hard error if a cyclic ingredient is found in the `get_claim_referenced_manifests` function so you can no longer retrieve a manifest in this scenario.

sdk/src/store.rs
ok-nick: I made it a hard error if a cyclic ingredient is found in line with what you said [here](https://github.com/contentauth/c2pa-rs/pull/1576#discussion_r2504089944).

sdk/tests/test_builder.rs
```diff
@@ -79,6 +79,94 @@ fn test_builder_riff() -> Result<()> {
...
+
+    // If they aren't the same number of bytes then we can't reliably substitute the URI.
+    assert_eq!(active_manifest_uri.len(), ingredient_uri.len());
+
+    // Replace the ingredient active manifest with the main active manifest.
+    let mut bytes = dest.into_inner();
+    let old = ingredient_uri.as_bytes();
+    let new = active_manifest_uri.as_bytes();
+
+    let mut i = 0;
```
mauricefisher64, line 135: We have functions to do this BTW.  See patch_bytes.

ok-nick: Unfortunately they aren't accessible from the integration tests.

Issue #1577, opened 2025-11-07
c2patool sends Error: hash verification to stderr


hackerfactor:
c2patool 0.26.0
(This version is getting much further through my regression suite.)

Test image:
https://fotoforensics.com/analysis.php?id=1b72197844a325a0e1cfb6590f99f318e174287b.2577237&fmt=orig

Output:
This is sent to stderr:
`Error: hash verification( ingredient hash does not match found ingredient )`
It does not generate any json output.
By comparison, c2patool-0.13.0 *does* generate a valid json.

Expected:
Nothing to stderr.
It should generate the json.
The error should be in the failure[] or informational[] fields.

PR #1582, opened 2025-11-07
fix: Restore recursive ingredient checks

mauricefisher64:
Re-enable recursive check now that we have a better validation results check
Bubble up errors even if ingredient manifest hash does not match 
Catch bad 2.x manifest labels
Remove bogus certificate status integration test

## Checklist

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -4133,6 +4133,14 @@ pub mod tests {
             1,
         );
         assert!(c4.is_err());
+
+        // bad v2
+        let c5 = Claim::new_with_user_guid(
+            "claim_generator",
+            "urn:blahblah:3fad1ead-8ed5-44d0-873b-ea5f58adea82:acme",
+            2,
+        );
+        assert!(c5.is_err());
     }
 
     #[test]


sdk/src/store.rs
```diff
@@ -1972,19 +1956,17 @@ impl Store {
                     }
 
                     // if manifest hash did not match we continue on to do a full claim validation
```
ok-nick: Is this comment outdated?

sdk/src/store.rs
```diff
@@ -1856,7 +1847,7 @@ impl Store {
 
                     // allow the extra ingredient trust checks
                     // these checks are to prevent the trust spoofing
-                    let check_ingredient_trust = settings.verify.check_ingredient_trust;
```
ok-nick, line 1859: Do we still need the `verify.check_ingredient_trust` setting then?

sdk/src/store.rs
```diff
@@ -1738,18 +1731,16 @@ impl Store {
                     }
 
                     // if manifest hash did not match we continue on to do a full claim validation
```
ok-nick: Here too.

sdk/src/store.rs
mauricefisher64: Yeah, that can be removed.

sdk/src/store.rs
mauricefisher64: I think I can remove that from settings because we are unlikely to go back to giving a choice.

mauricefisher64:
> It looks like this would fix #1577 though that might be a coincidence.

No, it was an intentional fix.

PR #1586, opened 2025-11-08
fix: Use Digitalsourcetype with Builder intents

tmathern:
Either I am using set_intent on the Builder wrong by just calling set_intent with Create and a DigitalSourceType (without changing default settings), or we need a small settings code update to pass through the picked digital source type. 

Observed issue: When using the SDK OOTB (no changes in default settings), the set_intent API with Create intent ignores the Digital Source Type passed in (it falls back to the default value in settings).

@gpeacock @ok-nick If you have working examples of passing through digital source types with the set_intent APIs, please show me/tell me what is done wrong in the C API tests. Thanks!

## Checklist

### DIFF
--- a/cli/tests/fixtures/trust/cawg_sign_settings.toml
+++ b/cli/tests/fixtures/trust/cawg_sign_settings.toml
@@ -472,7 +472,7 @@ operating_system = "auto"
 # The source type field is required for the c2pa.created action.
 #
 # For more information, see `c2pa::assertions::actions::source_type`.
-#ource_type = "http://c2pa.org/digitalsourcetype/empty"
+#source_type = "http://c2pa.org/digitalsourcetype/empty"
 
 # Settings for configuring how c2pa.opened actions are auto created.
 #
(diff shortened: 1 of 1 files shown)


sdk/src/builder.rs
```diff
@@ -1239,6 +1250,7 @@ impl Builder {
         ingredient_map: &HashMap<String, (&Relationship, HashedUri)>,
         actions: &mut Actions,
         settings: &Settings,
+        intent: Option<BuilderIntent>,
     ) -> Result<()> {
```
ok-nick, line 1247: I think it would make more sense if this function took `&self` and used `self.settings` and `self.intent`.

tmathern: Sure, the bugfix can fix and change the API while at it. 
Why was it not originally done that way? Any reason?

tmathern: Update: Seems to be historical reasons/grown organically. Updated function signature.

ok-nick: Originally it took `&self` but was changed at some point. Later we added settings parameter and now intent. Since this info is stored in the builder already, I think it makes sense to bring back the `&self` parameter.

Issue #1378, opened 2025-09-04
Use `httpmock` for WASM when it's supported


ok-nick:
Tracking https://github.com/alexliesenfeld/httpmock/issues/121

Issue #1377, opened 2025-09-04
Switch to `reqwest_blocking` for WASI when it's supported


ok-nick:
Tracking https://github.com/seanmonstar/reqwest/issues/2294

Issue #1364, opened 2025-09-02
Consolidate async networking in async tests and sync networking in sync tests


ok-nick:
Unfortunately our test cases create their own tokio runtime, but call into the sync c2pa APIs. Tokio does not allow multiple async runtimes to be created within each other. As a result, those tests cannot use `reqwest_blocking` until they are fixed and we must enable both `ureq` (sync) and `reqwest` (async) by default.

From #1355

llangsner:
this would be a nice to have but its a large effort for maybe little reward

ok-nick:
When this is fixed and #1355 is merged, we should change the default from `http_ureq` to `http_reqwest_blocking` AND ensure that if both features are enabled, `http_reqwest_blocking` is preferred.

PR #1589, opened 2025-11-10
docs: Add issue templates for bug reports and feature requests

ribhavh:

This is for adding a dialog box so we can track if its a feature or bug. 

## Checklist


c2pa-rs
ribhavh, line 1: hold on, i need to delete this

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+        - Linux
+        - macOS
+        - Windows
+        - iOS
+        - Android
+        - Other
+    validations:
+      required: true
+  - type: textarea
+    id: what-happened
```
scouten-adobe, line 45: Can we ask for two things here? What happened and what did you _expect_ to happen?

ribhavh: Done

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,60 @@
...
+body:
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component
+      description: Which tool or component are you having problems with?
+      options:
+        - c2pa-rs
+        - c2patool
+        - c2pa-c
```
ok-nick, line 13: Should we also have labels for components and auto-assign them here?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
ok-nick: Hm, I think it's still useful to have blank issues or at least a way to report general tasks/improvements that aren't necessarily features. Not a blocker but just something to think about.

For example:
* #1378
* #1377 
* #1364

tmathern: +1

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,12 @@
+name: 💡 Feature request
+description: Suggest an idea for improvement
+labels: [enhancement]
+body:
+  - type: textarea
+    id: idea
+    attributes:
+      label: Describe your idea
+      description: What would make the project better?
+      placeholder: Describe your feature request here.
```
tmathern, line 10: Do we need to add a question about what is the target platform?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,60 @@
...
+        - Other
+    validations:
+      required: true
+  - type: input
+    id: component-other
+    attributes:
+      label: If Other, please specify
+      description: Which component are you using?
+      placeholder: e.g., Python bindings, Java SDK, etc.
```
tmathern: Could we use placeholders that match what we have? Eg. C bindings, docs?
It may be confusing to list Python here (it has a dedicated repo) or Java (also not here and does not really exist?).

Issue #1532, opened 2025-10-28
Signing again a file does not clear previous XMP data, especially remote manifest URL


tmathern:
Demo is in https://github.com/contentauth/c2pa-rs/pull/1531, run from folder `c2pa_c_ffi` using `cargo test --features "rust_native_crypto,file_io" test_sign_again_with_info -- --nocapture` to see all logs.

I have signed a file using ACA (https://contentauthenticity.adobe.com/) because ACA adds both embedded and remote manifests in a file.
When I use that file and I want to sign it again, signing works, a Reader claims there is no remote manifest, but I still see a remote manifest URL in the XMP metadata.

For instance, the screenshot below is for a file I signed with embedded manifest only. I would expect therefore there is no trace of a cloud manifest URL in the XMP metadata. But looking at the file with exiftool, exiftool displays some leftover (I assume leftover?) data:

<img width="802" height="508" alt="Image" src="https://github.com/user-attachments/assets/39d28f69-2e6b-4a40-9846-e5fa69ec242f" />

Please clarify is this is either a bug, or some steps when signing again are missing.

ps: In this case, I would not expect to need to explicit call to set a remote URL back to null or a placeholder (especially since currently we defend against that in the C interface and other SDKs)?

PR #1598, opened 2025-11-13
fix: add missing remote url

mauricefisher64:
Fix missing remote_url when loading via **from_manifest_data_and_stream**

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -194,6 +194,11 @@ impl Store {
         self.remote_url.as_deref()
     }
 
+    // Set the remote url of the manifest if this Store
+    pub(crate) fn set_remote_url(&mut self, url: &str) {
+        self.remote_url = Some(url.to_string());
+    }
+
     /// Returns if the [`Store`] was created from an embedded manifest.
     pub fn is_embedded(&self) -> bool {
         self.embedded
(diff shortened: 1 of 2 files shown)


sdk/tests/test_builder.rs
```diff
@@ -79,6 +79,70 @@ fn test_builder_riff() -> Result<()> {
...
+                    "myorg:Field": "MyOrgValue"
+                }
+            }
+        ]
+    }
+    "#;
+
+    let mut builder = Builder::from_json(manifest_json)?;
+    builder.set_intent(BuilderIntent::Edit);
+    builder.definition.claim_version = Some(2); // use v1 for this test
```
tmathern, line 132: Isn't that's using v2?

sdk/src/store.rs
```diff
@@ -3756,12 +3761,19 @@ impl Store {
...
                 log_item!("asset", "error loading file", "load_from_asset")
                     .failure_no_throw(validation_log, e);
             })?;
 
+        // fetch remote manifest if available
+        if let Some(ext_ref) =
+            crate::utils::xmp_inmemory_utils::XmpInfo::from_source(&mut stream, format).provenance
+        {
+            store.set_remote_url(&ext_ref);
+        }
```
ok-nick, line 3775: The [docs for `Reader::remote_url` is defined as](https://docs.rs/c2pa/latest/c2pa/struct.Reader.html#method.remote_url):
> Returns the remote url of the manifest if this [Reader](https://docs.rs/c2pa/latest/c2pa/struct.Reader.html) obtained the manifest remotely.

Here we are changing that definition but also the behavior that exists in other parts of the code.

mauricefisher64: The problem is that no remote_url is returned when using this path and Tania reported an error. What do you want here?

ok-nick: The current behavior is to ignore remote_url if the manifest is found as embedded first. Here we are always setting the remote_url. It's an unusual situation though because we aren't actually looking for an embedded manifest anyways.

If we are going to make this change here, we should probably make that change in the other constructors and change the reader docs to state that it's always returned regardless of how the manifest was obtained.

tmathern: (For context, on the error, if it's the one I am thinking of( https://github.com/contentauth/c2pa-rs/issues/1532): The reported issue is that in certain cases you can "keep" and outdated remote URL manifest. Even if our SDK does not look at it, it still could be found. For instance, even if our SDK does the right thing, you could still "find" URLs by using exiftool. More details in https://github.com/contentauth/c2pa-rs/issues/1532).

ale-adobe:
This will likely be addressed by upcoming lower-level API work.

PR #1599, opened 2025-11-14
docs: Describe top-level architecture of SDK

scouten-adobe:


sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+    AssetIO --> AssetHandlers
+    AssetIO --> JumbfIO
+    
+    CoseValidator --> Crypto
+    Signer --> Crypto
+    
+    style Builder fill:#fff4e1
+    style Reader fill:#e1f5ff
+    style Store fill:#e8f5e9
+    style Crypto fill:#f3e5f5
```
ok-nick, line 75: The colors here can be difficult to see in dark mode.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
+# C2PA SDK architecture
+
+This document describes the outermost architectural layer of the c2pa SDK. The SDK implements the C2PA (Coalition for Content Provenance and Authenticity) specification, providing functionality to create, embed, read, and validate C2PA manifests in digital assets.
+
+## Architectural overview
+
+The SDK follows a layered architecture that separates concerns between user-facing APIs, data models, storage, asset I/O, cryptography, and validation.
```
tmathern: I would call `user-facing APIs` -> `public APIs`. It's what they are,e ventually.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+```mermaid
+graph TB
```
tmathern, line 10: I sometimes consider the signer public, because users can technically define theirs (callback signers).

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+The outermost layer provides high-level APIs for common workflows.
+
+#### `Builder` ([builder.rs](builder.rs))
+
+Creates and signs C2PA manifests. The `Builder` provides both declarative (JSON-based) and imperative (method-based) interfaces for constructing manifests.
+
+##### Key responsibilities
+
+- Assemble manifest definitions from ingredients, assertions, and resources
+- Coordinate with signers to create cryptographic signatures
```
tmathern, line 91: Here, that is why I see Signers as public. They are not totally hidden.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Embed signed manifests into assets via asset handlers
```
tmathern, line 92: asset handlers = format handlers?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- [`ResourceStore`](#resourcestore-resource_storers)
+- [`settings/`](#settings)
+
+#### `Reader` ([reader.rs](reader.rs))
+
+Reads and validates C2PA manifests from assets. The `Reader` extracts manifest stores, validates signatures and bindings, and provides access to manifest content.
+
+##### Key responsibilities
+
+- Extract JUMBF manifest stores from assets
```
tmathern, line 111: I would rather say read? Since the JUMBF is left in place in the asset, reading is not destructive

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Validate cryptographic signatures and certificate chains
```
tmathern, line 112: Does `certificate chains` mean trust?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+- [`Store`](#store-storers)
+- [`Manifest`](#manifest-manifestrs)
+- [`validation_results.rs`](#validation_resultsrs)
+- [`status_tracker/`](#status_tracker)
+- [`asset_io.rs`](#asset_iors)
+
+#### Public type exports
+
+The SDK exposes selected types from internal modules as public API:
```
tmathern, line 127: I would list Signer on top officially then, since it is here + CallbackSigner.

The others are a little bit less used publicly.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- `Error`, `Result` (error handling)
+- Various assertion types from [`assertions/`](assertions/)
+
+### Data model layer (internal)
+
+Internal representations of C2PA structures.
+
+#### `Claim` ([claim.rs](claim.rs))
+
+The core internal representation of a C2PA claim. A claim is a set of assertions about an asset, including its provenance, along with signature information.
```
tmathern, line 141: Do you want to add a note about identity/cawg here? It's a special kind of signals, after all.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+##### Depends on
+
+- [`Assertion`](#assertion-assertionrs)
+- [`assertions/`](#assertions)
+- [`hashed_uri.rs`](hashed_uri.rs)
+- [`ResourceStore`](#resourcestore-resource_storers)
+
+#### `Manifest` ([manifest.rs](manifest.rs))
+
+User-facing representation of a validated C2PA manifest. Manifests are created from claims during validation.
```
tmathern: I say above that this is "less public" because very often people rather look at JSON, especially when using SDKs built on top of the Rust one.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+##### Depends on
+
+- [`Claim`](#claim-claimrs)
+- [`Ingredient`](#ingredient-ingredientrs)
+- [`assertions/`](#assertions)
+- [`ResourceStore`](#resourcestore-resource_storers)
+
+#### `Ingredient` ([ingredient.rs](ingredient.rs))
+
+Represents a source asset (parent or component) used in creating the current asset. Ingredients can have their own C2PA provenance.
```
tmathern, line 179:
```suggestion
Represents a source asset (with relationships parent or component in the current manifest) used in creating the current asset. Ingredients may have their own C2PA provenance.
```

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+##### Key responsibilities
+
+- Capture ingredient metadata (title, format, identifiers)
+- Extract and validate ingredient manifests
+- Support relationship types (parent, component, input)
```
tmathern, line 185: Do we want to mention the actual values observed? Eg. parentOf, componentOf?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+#### [assertions/](assertions/)
+
+A collection of modules implementing specific C2PA assertion types and custom extensions.
+
+TO DO: Create [assertions/README.md](assertions/README.md) to describe assertion modules.
+
+#### `ManifestAssertion` ([manifest_assertion.rs](manifest_assertion.rs))
+
+Container for assertions within manifests, supporting multiple storage formats (CBOR, JSON, binary).
```
tmathern, line 219: Maybe elaborate a little bit on the differences between the 3? And how it relates to assertion kinds (especially since soon below you also mention URI).

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+Manages persistence and serialization of C2PA data.
+
+#### `Store` ([store.rs](store.rs))
+
+The internal manifest store that manages one or more claims and their relationships. This is the central hub for C2PA data before it's written to or after it's read from an asset.
+
+##### Key responsibilities
+
+- Manage claim graph (active claim, parent claims, update manifests)
```
tmathern, line 238: We should make it more prominent that a manifest store is a directed graph. This is usually enlightening for people who encounter this the first time (helps for traversals).

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+- [`Claim`](#claim-claimrs)
+- [`jumbf/`](#jumbf)
+- [`asset_io.rs`](#asset_iors)
+- [`cose_validator.rs`](#cose_validatorrs)
+- [`status_tracker/`](#status_tracker)
+
+#### [jumbf/](jumbf/)
+
+JUMBF (JPEG Universal Metadata Box Format) is the container format used to store C2PA data in assets. This module provides the box structure and label management.
```
tmathern, line 254: Define the acronym when it is first used, please!

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+##### Depends on
+
+- [`asset_handlers/`](#asset_handlers)
+- [`jumbf/`](#jumbf)
+
+#### [asset_handlers/](asset_handlers/)
+
+Format-specific implementations for different asset types. Each handler knows how to read and write C2PA data for its format.
+
+TO DO: Create [asset_handlers/README.md](asset_handlers/README.md) to describe asset handler modules.
```
tmathern, line 316: Do you want to link to the supported formats list here?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+#### [jumbf_io.rs](jumbf_io.rs)
+
+High-level JUMBF I/O operations that coordinate between asset handlers and JUMBF serialization.
+
+##### Key responsibilities
+
+- Load JUMBF from asset streams (via appropriate handler)
+- Save JUMBF to asset streams (via appropriate handler)
+- Support multiple asset formats through handler registry
+- Provide format detection
```
tmathern, line 327: Already? I thought that was WIP for some?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+Tracks and reports validation results across all aspects of manifest validation.
+
+#### [validation_results.rs](validation_results.rs)
+
+Structured validation results following the C2PA specification's validation states.
+
+##### Key responsibilities
+
+- Define `ValidationState` enum (Invalid, Valid, Trusted)
```
tmathern, line 436: Link to the spec maybe?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Track validation status codes (success, informational, failure)
+- Support ingredient delta validation (changes from previous validation)
+- Provide structured reporting for validation outcomes
```
tmathern, line 439: I would put that bullet point second, because the others then clarify this

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- [`log_item.rs`](status_tracker/log_item.rs) - Individual logged events
+
+##### Key responsibilities
+
+- Record validation events as they occur
+- Support filtering by label/URI
+- Provide event iteration and reporting
+- Convert to user-facing validation results
+
+### Identity layer
```
tmathern, line 478: I think that one may need to be higher up?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+sequenceDiagram
+    participant Builder
+    participant Claim
+    participant Store
+    participant Signer
+    participant AssetIO
+    
+    Builder->>Claim: Build claim structure
+    Claim->>Claim: Add assertions
+    Claim->>Claim: Add ingredients
```
tmathern, line 587: Add ingredients technically is optional (especially if you "create" an asset, which does not have a parent).

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
+# C2PA SDK architecture
+
+This document describes the outermost architectural layer of the c2pa SDK. The SDK implements the C2PA (Coalition for Content Provenance and Authenticity) specification, providing functionality to create, embed, read, and validate C2PA manifests in digital assets.
+
+## Architectural overview
+
+The SDK follows a layered architecture that separates concerns between user-facing APIs, data models, storage, asset I/O, cryptography, and validation.
+
+```mermaid
```
ok-nick, line 9: I'm not sure how accurate this diagram is. There's a few places where the arrows are incorrect, e.g. asset_io doesn't reference jumbf_io, the claim doesn't reference the resource store, etc. I think there's also quite a few missing arrows too, such as in the validation layer.

sdk/src/README.md
ok-nick: Image

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+    style Crypto fill:#f3e5f5
+```
+
+## Module organization
+
+### Public API layer
+
+The outermost layer provides high-level APIs for common workflows.
+
+#### `Builder` ([builder.rs](builder.rs))
```
ok-nick, line 84: This is really cool how it knows where to reference the file.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+Creates and signs C2PA manifests. The `Builder` provides both declarative (JSON-based) and imperative (method-based) interfaces for constructing manifests.
```
ok-nick, line 86: I like to think of the JSON API as a separate API that's above the Builder. The definitions are a bit intertwined but there's ongoing work to fix that.

sdk/src/README.md
ok-nick: Yes internally we reference them as asset handlers. I usually like to refer to them as parsers although I'm not sure what language is most clear.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Embed signed manifests into assets via asset handlers
+- Support multiple signing workflows (standard, data-hashed, box-hashed)
+- Archive and restore builder state
+
+##### Depends on
+
+- [`Claim`](#claim-claimrs)
+- [`Store`](#store-storers)
+- [`Signer`](#signer-and-asyncsigner-signerrs)
+- [`asset_io.rs`](#asset_iors)
```
ok-nick, line 101: Builder does not depend on asset_io.rs.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+#### `Reader` ([reader.rs](reader.rs))
+
+Reads and validates C2PA manifests from assets. The `Reader` extracts manifest stores, validates signatures and bindings, and provides access to manifest content.
+
+##### Key responsibilities
+
+- Extract JUMBF manifest stores from assets
+- Validate cryptographic signatures and certificate chains
+- Validate hard and soft bindings between manifest and asset
```
ok-nick, line 113: I found hard/soft bindings are a common confusion point. We may want to clarify that a bit here.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Track and report validation status
+- Support both sync and async validation workflows
```
ok-nick, line 115: Hm, I don't think we need this comment.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- [`Signer`](#signer-and-asyncsigner-signerrs)
+- [`asset_io.rs`](#asset_iors)
+- [`ResourceStore`](#resourcestore-resource_storers)
+- [`settings/`](#settings)
+
+#### `Reader` ([reader.rs](reader.rs))
+
+Reads and validates C2PA manifests from assets. The `Reader` extracts manifest stores, validates signatures and bindings, and provides access to manifest content.
+
+##### Key responsibilities
```
ok-nick, line 109: We should mention it provides an interface for reading all the properties of a manifest as well.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Track and report validation status
+- Support both sync and async validation workflows
+
+##### Depends on
+
+- [`Store`](#store-storers)
+- [`Manifest`](#manifest-manifestrs)
+- [`validation_results.rs`](#validation_resultsrs)
+- [`status_tracker/`](#status_tracker)
+- [`asset_io.rs`](#asset_iors)
```
ok-nick, line 123: Same here, Reader doesn't depend on asset_io and there's a few other things it does depend on (e.g. Ingredient).

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+#### Public type exports
```
ok-nick, line 125: I think we should remove this section, there are a lot of public types and they are always changing.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+##### Key responsibilities
+
+- Maintain claim structure (assertions, ingredients, signature)
+- Manage assertion instances and uniqueness
+- Handle redactions and assertion references
+- Support claim versioning (v1, v2)
+- Provide CBOR serialization/deserialization
+
+##### Depends on
```
ok-nick, line 151: Hm, there's quite a few more dependencies here.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+- [`Assertion`](#assertion-assertionrs)
+- [`assertions/`](#assertions)
+- [`hashed_uri.rs`](hashed_uri.rs)
+- [`ResourceStore`](#resourcestore-resource_storers)
```
ok-nick, line 156: Claim doesn't use the ResourceStore.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+- Support JSON serialization for reporting
+
+##### Depends on
+
+- [`Claim`](#claim-claimrs)
+- [`Ingredient`](#ingredient-ingredientrs)
+- [`assertions/`](#assertions)
+- [`ResourceStore`](#resourcestore-resource_storers)
+
+#### `Ingredient` ([ingredient.rs](ingredient.rs))
```
ok-nick, line 177: We should make it clear here that an "ingredient" is really just an assertion that wraps a C2PA manifest. I believe there is ongoing work to clarify this in the APIs.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+### Asset I/O layer
+
+Handles format-specific reading and writing of assets with C2PA data.
+
+#### [asset_io.rs](asset_io.rs)
+
+Defines traits for asset I/O operations. These traits abstract over different asset formats, allowing the SDK to handle JPEG, PNG, MP4, and other formats uniformly.
+
+##### Key traits
```
ok-nick, line 293: There's a few more traits we may want to describe here. There's also the CAIRead trait.

sdk/src/README.md
ok-nick, line 327: We don't necessarily detect formats.

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+A module providing core cryptography primitives using either OpenSSL and Rust-native implementations.
+
+TO DO: Create [asset_handlers/README.md](asset_handlers/README.md) to describe asset handler modules.
+
+#### `Signer` and `AsyncSigner` ([signer.rs](signer.rs))
+
+Defines the `Signer` and `AsyncSigner` traits that abstract over different signing implementations. This allows the SDK to work with various signing mechanisms (local keys, HSMs, remote services, etc.).
+
+##### Key responsibilities
```
ok-nick, line 348: Also good to mention time authorities here.

sdk/src/README.md
tmathern: It's a little bit like the Signer, they work together eventually.

sdk/src/README.md
tmathern: So we have multiple vocabulary options here ;). Maybe we need a glossary at the top..

sdk/src/README.md
tmathern: Maybe we should just keep the ones that are stable, and exposed in the other SDKs?
Reader, Builder, Signer for instance are good candidates.

sdk/src/README.md
tmathern: Isn't that `Coordinate with timestamp authorities`? Or do you mean more details?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+#### [callback_signer.rs](callback_signer.rs)
+
+A `Signer` implementation that delegates signing to user-provided callbacks. This enables integration with external signing systems.
+
+##### Key responsibilities
+
+- Wrap callback functions as `Signer` trait
+- Support custom signing logic
+- Enable HSM and remote signing workflows
```
tmathern, line 386: HSM?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+Validates COSE signatures, certificates, and timestamp tokens.
+
+##### Key responsibilities
+
+- Verify COSE Sign1 signatures
+- Validate certificate chains against trust anchors
+- Verify timestamp tokens
+- Check OCSP responses
+- Apply trust policies and EKU requirements
```
tmathern, line 418: EKU?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+### Identity layer
+
+[CAWG (Creator Assertions Working Group)](https://cawg.io) identity assertion support, including W3C Verifiable Credentials and X.509 credential binding.
+
+TO DO: Create [identity/README.md](identity/README.md) to describe identity modules.
+
+### Specialized modules
+
+#### [dynamic_assertion.rs](dynamic_assertion.rs)
```
tmathern, line 486: Isn't that a key module? Eg. to add CAWG data on signing?

sdk/src/README.md
```diff
@@ -0,0 +1,594 @@
...
+
+HTTP client abstractions supporting multiple backends for network operations (OCSP, remote manifests, timestamp authorities).
+
+TO DO: Create [http/README.md](http/README.md) to describe HTTP-related modules.
+
+### Supporting modules
+
+#### [settings/](settings/)
+
+SDK-wide configuration system using thread-local settings.
```
tmathern, line 512:
```suggestion
SDK-wide configuration system using *thread-local* settings.
```
It needs the emphasis, currently.

Issue #1590, opened 2025-11-11
Support certificate chain verification logic for TSA even when provided in Root→Leaf order.


jooseong-lee:
Currently, the certificate chain verification logic for TSA seems to assume that the provided certificate chain is ordered from Leaf → Root (i.e., end-entity first). However, in some cases — such as when the TSA certificate chain is provided from Root → Leaf — verification may fail or require manual reordering before validation. It would be helpful if the verification logic could automatically detect and handle both chain orders by matching certificates based on their issuer and subject fields, instead of relying on the array order.

Why this is useful:
Strictly speaking, according to [RFC 5652]([5.1](https://datatracker.ietf.org/doc/html/rfc5652#section-5.1).  SignedData Type), the certificate chain type is defined as SET OF rather than SEQUENCE, so the order of the chain cannot be enforced. Nonetheless, it is customary to use SEQUENCE, so there is no particular issue with the current implementation.

However, one drawback is that some external systems or APIs may present the root certificate instead of the leaf certificate as the first element in the certificate chain. Supporting both directions would make the verification process more robust and developer-friendly.

jin-gyu-kim:
We have prepared a patch for this issue - https://github.com/contentauth/c2pa-rs/pull/1600
Please share the opinion.

PR #1600, opened 2025-11-14
fix: fix the timestamp cert chain validation to be independent of the order

jin-gyu-kim:
Timestamp certificates, which are part of the SignedData, are structured differently from CBOR when forming a chain. Specifically, they are composed as a SET OF type, and the order is not considered. (Refer to RFC5652)
The existing implementation assumes that the certificate chain is structured from leaf to root, which causes some TSA to fail in processing it correctly. (example: Microsoft Timestamp Server, Java Bouncy Castle)
To address this issue, the certificate chain will be sorted into an interpretable format.

Related Issue : https://github.com/contentauth/c2pa-rs/issues/1590
Before validating the timestamp certificate chain, rearrange it from leaf to root.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
       interval: "daily"
     commit-message:
       prefix: "update"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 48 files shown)


ribhavh:
@jin-gyu-kim Is it ok if we review this next week? Gavin and Maurice are out for a conference this week.

jin-gyu-kim:
> Please run "cargo fmt" to fix the formatting issues.

I have updated the code using cargo fmt. 
Since I don't have merge permissions, I kindly request you to review it and, if everything looks fine, proceed with the merge.

PR #1601, opened 2025-11-14
fix: Attempt a cleaner forced update approach

tmathern:
I suspect the previous way of forcing releases did not properly work:
- I can't find the comment it was supposed to add
- The file the workflow tried to modify for the c2patool does not exist (and it never landed on main, which seems unlikely)?

So suggesting another "hack-fix".

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,18 +50,19 @@ jobs:
           # Ensure we're on a clean main branch
           git checkout main
           git pull origin main
-          
+
           # Extract the new version of c2pa from the release PR output JSON
           NEW_VERSION=$(echo "$PR" | jq -r '.releases[] | select(.package_name == "c2pa") | .version')
-          
-          # Add a comment to the lib.rs file to trigger a change
-          echo "// Updated to use c2pa crate version $NEW_VERSION" >> c2pa_c_ffi/src/lib.rs
-          
+
+          # Update the c2pa dependency version in c2pa_c_ffi/Cargo.toml
+          sed -i.bak "s/\(c2pa = { path = \"..\/sdk\", version = \"\)[^\"]*\"/\1$NEW_VERSION\"/" c2pa_c_ffi/Cargo.toml
+          rm c2pa_c_ffi/Cargo.toml.bak
+
           # Configure git and commit the changes
           git config user.name "CAI Open Source Builds"
           git config user.email "caiopensrc@adobe.com"
-          git add c2pa_c_ffi/src/lib.rs
-          git commit -m "fix: Pick up latest c2pa-rs build"
+          git add c2pa_c_ffi/Cargo.toml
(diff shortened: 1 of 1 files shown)


scouten-adobe (approved):

.github/workflows/release.yml
```diff
@@ -50,18 +50,19 @@ jobs:
...
           # Extract the new version of c2pa from the release PR output JSON
           NEW_VERSION=$(echo "$PR" | jq -r '.releases[] | select(.package_name == "c2pa") | .version')
-          
-          # Add a comment to the lib.rs file to trigger a change
-          echo "// Updated to use c2pa crate version $NEW_VERSION" >> c2pa_c_ffi/src/lib.rs
-          
+
+          # Update the c2pa dependency version in c2pa_c_ffi/Cargo.toml
+          sed -i.bak "s/\(c2pa = { path = \"..\/sdk\", version = \"\)[^\"]*\"/\1$NEW_VERSION\"/" c2pa_c_ffi/Cargo.toml
+          rm c2pa_c_ffi/Cargo.toml.bak
```
ok-nick, line 59: You may be able to use `cargo add --path <PATH_HERE>` instead for clarity.

.github/workflows/release.yml
```diff
@@ -50,18 +50,19 @@ jobs:
...
+
           # Extract the new version of c2pa from the release PR output JSON
           NEW_VERSION=$(echo "$PR" | jq -r '.releases[] | select(.package_name == "c2pa") | .version')
-          
-          # Add a comment to the lib.rs file to trigger a change
-          echo "// Updated to use c2pa crate version $NEW_VERSION" >> c2pa_c_ffi/src/lib.rs
-          
+
+          # Update the c2pa dependency version in c2pa_c_ffi/Cargo.toml
+          sed -i.bak "s/\(c2pa = { path = \"..\/sdk\", version = \"\)[^\"]*\"/\1$NEW_VERSION\"/" c2pa_c_ffi/Cargo.toml
```
ok-nick, line 58: I'm not sure if you can specify a path and version simultaneously, only the path should be needed here.

tmathern: You can. If main is a reference.
https://github.com/contentauth/c2pa-rs/blob/main/cli/Cargo.toml#L31

ok-nick: Ah ok interesting, I didn't know that was possible.

tmathern: I think it relates to the fact we're in a workspace. There is a local path, but cargo on publish ignores local paths. (But  in local dev it uses your path and verifies the version still, I recall having had errors about mismatches when I poked at this).

.github/workflows/release.yml
tmathern: Hmmm, I think this could work if we would not already have a path? We really just do want to hit the value change here to trick release-plz into triggering a release in all cases.

PR #1603, opened 2025-11-17
feat: Public API for timestamp assertions

### LINKED ISSUE #1210: Timestamp Assertion
Support the timestamp assertion as documented in the [2.2](https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#timestamp_assertion) spec. This will likely be automatically generated by the SDK since we already have the ability to call timestamp services. I'd expect it to be a setting for builder to enable this, rather than an assertion users write, but adding one manually should be supported. On the reading side, there are some open questions about how to integrate this into existing signature reports or deal with it as as separate assertion. We will need to figure out a standard way to handle these timestamp updates in the UI.
Validation will need to respect the existence of a valid timestamp assertion. 
This assertion will generally be added as an update assertion to that needs to be tested.

ok-nick:
Introduces new public APIs for creating timestamp assertions. This is particularly useful when timestamping manifests that were previously signed offline. Currently, we only support timestamping the current manifest in the COSE header, this PR enables timestamp assertions.

Three new settings:
- `builder.auto_timestamp_assertion.enabled`
- `builder.auto_timestamp_assertion.skip_existing`
- `builder.auto_timestamp_assertion.fetch_scope` - `all` or `parent`

New builder method:
- `Builder::add_timestamp` - with manifest label

### Related issues
* Closes #1210

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -544,9 +544,9 @@ dependencies = [
 
 [[package]]
 name = "c2pa_cbor"
-version = "0.77.0"
+version = "0.77.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68c4168a4e0be327fc46a936095a6c9a41e167f317419d250317dd408df45bbf"
+checksum = "42be240232265b2ea8c21cb97fdf700bfe2077ef6aa60186adf19a39f501ee5a"
 dependencies = [
  "half",
  "serde",
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -107,7 +107,7 @@ bytes = "1.7.2"
 byteorder = { version = "1.4.3", default-features = false }
 byteordered = "0.6.0"
 chrono = { version = "0.4.42", default-features = false, features = ["serde"] }
-c2pa_cbor = "0.77.0"
+c2pa_cbor = "0.77.1"
 config = { version = "0.14.0", default-features = false, features = [
     "json",
     "toml",
--- a/sdk/src/assertions/assertion_metadata.rs
+++ b/sdk/src/assertions/assertion_metadata.rs
@@ -17,6 +17,7 @@ use chrono::{SecondsFormat, Utc};
 #[cfg(feature = "json_schema")]
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 
 use crate::{
     assertion::{Assertion, AssertionBase, AssertionCbor},
@@ -44,6 +45,10 @@ pub struct AssertionMetadata {
     localizations: Option<Vec<HashMap<String, HashMap<String, String>>>>, // not implemented
     #[serde(rename = "regionOfInterest", skip_serializing_if = "Option::is_none")]
     region_of_interest: Option<RegionOfInterest>,
+    /// Arbitrary key/value pairs as permitted by the C2PA spec.
+    /// Uses flatten to allow these fields to be serialized at the same level as known fields.
+    #[serde(flatten, skip_serializing_if = "HashMap::is_empty")]
+    additional_fields: HashMap<String, Value>,
 }
 
 impl AssertionMetadata {
@@ -63,6 +68,7 @@ impl AssertionMetadata {
             data_source: None,
             localizations: None,
             region_of_interest: None,
+            additional_fields: HashMap::new(),
         }
     }
 
@@ -139,6 +145,57 @@ impl AssertionMetadata {
         localizations: Vec<HashMap<String, HashMap<String, String>>>,
     ) -> Self {
         self.localizations = Some(localizations);
+        self
+    }
+
+    /// Sets an arbitrary key/value pair in the metadata.
+    ///
+    /// This allows adding custom fields that are not part of the standard schema,
+    /// as permitted by the C2PA specification.
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The key for the custom field
+    /// * `value` - The value as a serde_json::Value
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use c2pa::assertions::AssertionMetadata;
+    /// use serde_json::json;
+    ///
+    /// let metadata = AssertionMetadata::new()
+    ///     .set_field("customKey", json!("customValue"))
+    ///     .set_field("nestedObject", json!({"foo": "bar", "count": 42}));
+    /// ```
+    pub fn set_field<S: Into<String>>(mut self, key: S, value: Value) -> Self {
+        self.additional_fields.insert(key.into(), value);
+        self
+    }
+
+    /// Gets an arbitrary key/value pair from the metadata.
+    ///
+    /// Returns `None` if the key doesn't exist.
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The key to look up
+    pub fn get_field(&self, key: &str) -> Option<&Value> {
+        self.additional_fields.get(key)
+    }
+
+    /// Returns a reference to all additional fields.
+    pub fn additional_fields(&self) -> &HashMap<String, Value> {
+        &self.additional_fields
+    }
+
+    /// Sets multiple arbitrary key/value pairs at once, replacing any existing ones.
+    ///
+    /// # Arguments
+    ///
+    /// * `fields` - A HashMap of key/value pairs to set
+    pub fn set_additional_fields(mut self, fields: HashMap<String, Value>) -> Self {
+        self.additional_fields = fields;
         self
     }
 }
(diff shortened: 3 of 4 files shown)


sdk/src/builder.rs
```diff
@@ -1422,6 +1418,81 @@ impl Builder {
...
+    fn maybe_add_timestamp(
+        &mut self,
+        time_authority_url: &str,
+        store: &mut Store,
+        http_resolver: &impl SyncHttpResolver,
+    ) -> Result<()> {
+        if self.settings.builder.update_manifest_timestamp_assertion {
+            let provenance_claim = store.provenance_claim().ok_or(Error::ClaimEncoding)?;
+            if provenance_claim.update_manifest() {
+                let parent_claim_id = manifest_label_from_uri(
```
gpeacock: This seems like it needs a recursive algorithm to keep searching until we find a non-update manifest.

sdk/src/assertions/timestamp.rs
```diff
@@ -20,118 +20,144 @@ use serde_bytes::ByteBuf;
...
         let headers = None;
 
         let bytes = if _sync {
             crate::crypto::time_stamp::default_rfc3161_request(
-                tsa_url,
+                time_authority_url,
                 headers,
                 &body,
-                message,
+                signature,
```
mauricefisher64: why did you name this signature?  It may be in this spec but is has changed before.  It is the message payload to be signed.  It is not acutally a signature.

sdk/src/crypto/cose/sigtst.rs
```diff
@@ -232,9 +245,11 @@ pub(crate) fn add_sigtst_header(
         if tss == TimeStampStorage::V2_sigTst2_CTT {
             // In `sigTst2`, we use only the `TimeStampToken` and not `TimeStampRsp` for
             // sigTst2
-            cts = timestamptoken_from_timestamprsp(&cts).ok_or(CoseError::CborGenerationError(
-                "unable to generate time stamp token".to_string(),
-            ))?;
+            cts = timestamptoken_from_timestamprsp(&cts).map_err(|err| {
+                CoseError::CborGenerationError(format!(
```
mauricefisher64: Why is this a Cbor error?  Nothing to do with Cbor

sdk/src/settings/builder.rs
```diff
@@ -467,10 +467,25 @@ pub struct BuilderSettings {
...
+    /// [`Signer::time_authority_url`]. If the signer is acquired from settings via [`Settings::signer`],
+    /// the URL can be set in [`SignerSettings`].
+    ///
+    /// The default value is false.
+    ///
+    /// [`TimeStamp`]: crate::assertions::TimeStamp
+    /// [`Signer::time_authority_url`]: crate::Signer::time_authority_url
+    /// [`Settings::signer`]: crate::settings::signer
+    /// [`SignerSettings`]: crate::settings::signer::SignerSettings
+    pub update_manifest_timestamp_assertion: bool,
```
mauricefisher64: You can add a timestamp to regular manifests too, not just update manifests.  Those are separate issues.

sdk/src/builder.rs
```diff
@@ -1422,6 +1418,81 @@ impl Builder {
...
+        store: &mut Store,
+        http_resolver: &impl AsyncHttpResolver,
+    ))]
+    fn maybe_add_timestamp(
+        &mut self,
+        time_authority_url: &str,
+        store: &mut Store,
+        http_resolver: &impl SyncHttpResolver,
+    ) -> Result<()> {
+        if self.settings.builder.update_manifest_timestamp_assertion {
```
mauricefisher64: adding timestamps has nothing to do with update manifests

sdk/src/store.rs
```diff
@@ -3946,22 +3906,17 @@ impl Store {
...
-                    let parent_label = manifest_label_from_uri(&parent_uri.url())?;
-                    if let Some(parent) = self.get_claim(&parent_label) {
-                        // recurse until we find
-                        if parent.update_manifest() {
-                            self.get_hash_binding_manifest(parent);
-                        } else if !parent.hash_assertions().is_empty() {
-                            return Some(parent.label().to_owned());
-                        }
-                    }
+        if let Some(parent_uri) = claim.parent_claim_uri().ok()? {
```
mauricefisher64: Where is the removed loop?  You can have any number of update manifests before you hit a normal manifes

sdk/src/assertions/timestamp.rs
ok-nick: Good point, I'll switch it for this function. For the `TimeStamp::refresh_timestamp` method, should I leave it as signature since the values of the timestamp assertion must be a token derived from the signature?

sdk/src/crypto/cose/sigtst.rs
ok-nick: Changed to `TimeStampError::DecodeError`.

sdk/src/store.rs
ok-nick: Good catch, I'll restore the original behavior here.

sdk/src/settings/builder.rs
ok-nick: Good point, this specific example was discussed as a common use case, but I don't see a problem in generalizing it to any parent ingredient. I'll change it to `add_timestamp_assertion_to_parent`.

sdk/src/assertions/timestamp.rs
mauricefisher64: I look at it as what it is.  It is the message payload as described in the time stamp spec.  It does not specify what the data is.

sdk/src/builder.rs
```diff
@@ -1422,6 +1418,77 @@ impl Builder {
         Ok(self)
     }
 
+    #[async_generic(async_signature(
+        &mut self,
+        time_authority_url: &str,
+        store: &mut Store,
+        http_resolver: &impl AsyncHttpResolver,
+    ))]
+    fn maybe_add_timestamp(
```
mauricefisher64: Since this function only adds the timestamp for parent ingredient we should rename this function.  Where is the capability to add timestamps for any manifest in the store?

sdk/src/builder.rs
```diff
@@ -1422,6 +1418,77 @@ impl Builder {
...
+        &mut self,
+        time_authority_url: &str,
+        store: &mut Store,
+        http_resolver: &impl SyncHttpResolver,
+    ) -> Result<()> {
+        if self.settings.builder.add_timestamp_assertion_to_parent {
```
mauricefisher64: This should be simplified to just return when add_timestamp_assertion_to_parent is false.  Per our code guideline we should favor that instead of a whole function if statement

sdk/src/settings/builder.rs
mauricefisher64: Not sure this name makes sense because the feature does not really correct yet.  As I pointed out just adding the timestamp to the parent manifest may not be enough.  We need the be able to add a timestamp for any manifest.  That is why I said make a function that takes a list of manifest IDs to create timestamp assertions for.  Remember that a TimeStamp assertion can contain an arbitrary number of timestamp assertions.  Then you have a different function that determines which manifests we want to include (the parent, all manifests missing timestamps, all up to first found timestamp, etc.).   That is what we did for certificate status assertions.

sdk/src/builder.rs
ok-nick: Added.

sdk/src/settings/builder.rs
```diff
@@ -467,10 +467,38 @@ pub struct BuilderSettings {
...
+
+/// The scope of manifests to fetch timestamps for.
+///
+/// See [`BuilderSettings::timestamp_assertion_fetch_scope`] for more information.
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[cfg_attr(feature = "json_schema", derive(JsonSchema))]
+#[serde(rename_all = "lowercase")]
+pub enum TimeStampFetchScope {
+    /// Fetch timestamps for only the parent manifest.
+    Parent,
```
mauricefisher64: There could/should be one more option here.  That is generate a timestamp for any manifest without a timestamp.

sdk/src/builder.rs
```diff
@@ -1431,6 +1458,111 @@ impl Builder {
...
+        }
+
+        for manifest_label in &self.timestamp_manifest_labels {
+            if let Some(claim) = store.get_claim(manifest_label) {
+                claim_uris.insert(claim.uri());
+            }
+        }
+
+        let provenance_claim = store.provenance_claim().ok_or(Error::ClaimEncoding)?;
+        let timestamp_assertions = provenance_claim.timestamp_assertions();
```
mauricefisher64, line 1797: What do you mean here?  You can have any number of timestamp assertions.  The timestamp limit applies to when adding a CoseSign1 timestamp array.  It should only contain a single item in the array.

sdk/src/builder.rs
```diff
@@ -1431,6 +1458,111 @@ impl Builder {
...
+            None => {}
+        }
+
+        for manifest_label in &self.timestamp_manifest_labels {
+            if let Some(claim) = store.get_claim(manifest_label) {
+                claim_uris.insert(claim.uri());
+            }
+        }
+
+        let provenance_claim = store.provenance_claim().ok_or(Error::ClaimEncoding)?;
```
mauricefisher64: Are you trying to get the CoseSign1 timestamp here?

sdk/src/builder.rs
ok-nick: > A time-stamp assertion shall have a label of c2pa.time-stamp, and there shall be at most one time-stamp assertion per C2PA Manifest.

Here is what the [spec says per manifest](https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_requirements_2).

sdk/src/builder.rs
ok-nick: This part of the code finds the timestamp assertion in the active manifest or creates a new one.

sdk/src/settings/builder.rs
ok-nick: Maybe I can add a separate option `overwrite_existing_timestamp_assertions: bool`. Or perhaps it would be better like this:
```
auto_timestamp_assertion: {
  enabled: bool,
  skip_existing: bool,
  fetch_scope: Scope,
}
```

sdk/src/builder.rs
```diff
@@ -1516,6 +1547,136 @@ impl Builder {
...
+
+        let mut claim_uris = HashSet::new();
+        match settings.builder.auto_timestamp_assertion.fetch_scope {
+            TimeStampFetchScope::All => {
+                for claim in store.claims() {
+                    claim_uris.insert(claim.uri());
+                }
+            }
+            TimeStampFetchScope::Parent => {
+                let provenance_claim = store.provenance_claim().ok_or(Error::ClaimEncoding)?;
```
mauricefisher64: Not sure this is right.  You are working a new provenance claim so the parent claim will be the current provenance claim not the parent of the current provenance claim.

ok-nick: This function runs after the new provenance claim is already committed:
https://github.com/contentauth/c2pa-rs/blob/906e825551906aaba56fb8752b6124496a372be1/sdk/src/builder.rs?plain=1#L1478-L1482

sdk/src/builder.rs
```diff
@@ -1690,8 +1850,17 @@ impl Builder {
         #[cfg(feature = "add_thumbnails")]
         self.maybe_add_thumbnail(&format, source)?;
 
-        // convert the manifest to a store
-        let mut store = self.to_store()?;
+        let mut claim = self.to_claim()?;
+
+        if let Some(tsa_url) = signer.time_authority_url() {
```
mauricefisher64, line 2058: Good,  this is much better.  It allow you to catch TS fetch failures before you add the Claim to the Store.

sdk/src/builder.rs
```diff
@@ -1516,6 +1555,127 @@ impl Builder {
...
+        let mut timestamp_assertion = if !timestamp_assertions.is_empty() {
+            // There can only be one timestamp assertion per the spec.
+            let timestamp_assertion =
+                TimeStamp::from_assertion(timestamp_assertions[0].assertion())?;
+            timestamp_assertion
+        } else {
+            TimeStamp::new()
+        };
+
+        // If `skip_existing` is enabled, only timestamp claims in `claim_uris` that aren't already timestampped.
```
mauricefisher64: typo timestampped -> timestamped

sdk/tests/timestamp_assertion.rs
```diff
@@ -0,0 +1,308 @@
...
+    Builder, BuilderIntent, Reader, Result, Signer,
+};
+
+mod common;
+
+const TEST_IMAGE: &[u8] = include_bytes!("fixtures/no_manifest.jpg");
+const FORMAT: &str = "image/jpeg";
+
+// Basic wrapper around a Signer to include a time authority URL.
+struct WrappedTsaSigner(Box<dyn Signer>);
```
mauricefisher64, line 28: How does this test work on the build machines?  Is it actually making the TSA call?

ok-nick: Yes. I'm not sure there's much we can do at the high-level until we introduce a local test TSA. I made this an integration test for that reason.

sdk/src/builder.rs
mauricefisher64: I don't see the fix here, it should be **timestamped**

ok-nick: I fixed a few typos in the same commit, this one included.

mauricefisher64 (approved):

sdk/src/assertions/timestamp.rs
```diff
@@ -20,62 +20,90 @@ use serde_bytes::ByteBuf;
...
+    //
+    // The `signature` is normally obtained from [`Store::get_cose_sign1_signature`].
+    //
+    // [`Store::get_cose_sign1_signature`][crate::store::Store::get_cose_sign1_structure].
+    #[async_generic(async_signature(
+        &mut self,
+        tsa_url: &str,
+        manifest_id: &str,
+        signature: &[u8],
+        http_resolver: &(impl AsyncHttpResolver + ?Sized),
```
gpeacock: If you change this to take Context, you won't need two different function signatures, then just get the resolver you want from the context.

sdk/src/settings/builder.rs
```diff
@@ -445,6 +445,65 @@ impl SettingsValidate for ActionsSettings {
...
+    /// This setting will account for both existing [`TimeStamp`] assertions and timestamps embedded
+    /// in the claim.
+    ///
+    /// The default value is true.
+    ///
+    /// [`TimeStamp`]: crate::assertions::TimeStamp
+    pub skip_existing: bool,
+    /// Which manifests to fetch timestamps for.
+    ///
+    /// The default value is [`TimeStampFetchScope::All`].
```
gpeacock: I'm not sure if this the right default, but I'm not sure it isn't

PR #1618, opened 2025-11-19
fix!: Remove `fetch_remote_manifest` feature

### LINKED ISSUE #1434: Configure fetching remote manifests with a setting versus feature flag
Currently we provide a `fetch_remote_manifest` feature flag, although it would make more sense if this were a configurable runtime setting, just like OCSP fetching.

Also related is that networking libraries will optionally be able to be excluded from compilation entirely with #1355.

ok-nick:
Removes the `fetch_remote_manifest` feature and changes the `verify.remote_manifest_fetch` setting to default to false. This is related to avoiding network requests by default in the SDK. Note this is NOT a backwards compatible change although effectively has the same behavior as if the feature wasn't enabled.

* Closes #1434


tmathern:
How does this impact the cli tools?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -2134,6 +2134,19 @@ mod tests {
 
     #[test]
     fn test_c2pa_reader_remote_url() {
+        let settings = serde_json::json!( {
```
tmathern: This would be the perfect example where we'd want settings per Reader @gpeacock @ok-nick.
So some may be allowed to retrieve remote manifests, some not.

sdk/src/ingredient.rs
```diff
@@ -1861,11 +1860,10 @@ mod tests {
         assert_eq!(ingredient.validation_status(), None);
     }
 
-    #[cfg(not(any(feature = "fetch_remote_manifests", feature = "file_io")))]
```
tmathern, line 1991: What about `feature = "file_io"` here?

tmathern (commented):
Why is it a `verify.remote_manifest_fetch` settings? Verify implies a validation for me? But this is more a "read", in the sense we allow to get remote data.

What will be the expectations for the downstream SDKs? They all get the "remote manifest fetch" turned off by default and need to set the appropriate setting if they want it back?

sdk/src/ingredient.rs
ok-nick: This test doesn't use any `file_io` functionality. I actually only caught a problem here because WASM is the only platform in our CI that runs with `file_io` disabled.

tmathern: Perfect, just wanted to make sure.

tmathern:
@ok-nick Maybe a suggestion because there is a lot of related work in-flight: Group your PRs into a feature branch (as you sit fit)? So they will be easier to integrate together (since it seems they'll be in the same release too), and maybe easier to maintain while in their rbanch(es). Wdyt?

https://github.com/contentauth/c2pa-rs/pull/1630 and https://github.com/contentauth/c2pa-rs/pull/1618 I could see released together. For the rest, developer's choice!

gpeacock (changes requested):
I think this is the right change,now that we have resolvers in the API. Presumably the default resolver doesn't use http?
Definitely a hidden breaking change so this needs to be clearly documented in the Readme as something that behaves differently and now must be enabled.
So please add something to the README. about this.

scouten-adobe:
A small residual fragment of this (a beta-preflight.yml tweak and a Cargo.toml feature-flag removal) has moved to contentauth/c2patool#306 as part of the c2patool repo split. The substantive sdk/ and c2pa_c_ffi/ changes stay here.

PR #1622, opened 2025-11-19
ci: Use `dtolnay/rust-toolchain` consistently

ok-nick:
Removes usage of `actions-rust-lang/setup-rust-toolchain` and replaces it with `dtolnay/rust-toolchain` and `Swatinem/rust-cache` that we normally use. Potentially fixes a previously reported issue with intermittent errors building the Android libraries.

### DIFF
--- a/.github/workflows/library-release.yml
+++ b/.github/workflows/library-release.yml
@@ -3,7 +3,7 @@ name: Publish c2pa-library binaries
 on:
   push:
     tags:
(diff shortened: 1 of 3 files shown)


tmathern (approved):
(I think a job from the PR checks has decided to hang, or is very, very, very slow to complete and clean up)

scouten-adobe (approved):

PR #1624, opened 2025-11-20
fix: Apply same restrictions on ingredient deltas as active manifest for validation state

ok-nick:
This PR changes our logic to apply the same conditions we do on the active manifest to the ingredients as well. If you have an untrusted ingredient, should you have a trusted active manifest? If the ingredient has an untrusted claim signature does it already affect the active manifests success codes?

This change breaks many of our test cases that do not report a trusted success code (and others) for ingredients. I'm opening this PR for further discussion on this topic.


sdk/src/validation_results.rs
```diff
@@ -210,30 +208,46 @@ impl ValidationResults {
...
+                        let deltas = idv.validation_deltas();
+                        deltas.failure().is_empty()
+                            || deltas.failure().iter().any(|status| {
+                                status.code() == validation_status::SIGNING_CREDENTIAL_UNTRUSTED
+                            })
+                    })
+                });
 
             // https://spec.c2pa.org/specifications/specifications/2.2/specs/C2PA_Specification.html#_trusted_manifest
-            let is_trusted = success_codes.contains(validation_status::SIGNING_CREDENTIAL_TRUSTED)
```
gpeacock, line 227: I think all you need to do is remove these two lines, the rest of the logic looks like it already does the same thing. It's easy to make mistakes in this code, so smaller changes are better. I think these two lines were added and we need to remove them.

ok-nick: (discussed offline; this shouldn't be an issue)

PR #1625, opened 2025-11-20
feat(sdk): Introduce new API to retrieve raw JUMBF manifest from C2paReader

magaylor:
Adding an API to the Reader that allows clients to extract the RAW JUMBF Manifest from an asset.

This has been a feature I've been talking to @mauricefisher64 about for a month or two now. It's my understanding that contentauth is tracking this item, but internal Microsoft teams are beginning to become blocked on it not existing. 

We are working on adding manifestData to ingredient assertions and have no clean way, without 3p tools, to get the Raw JUMBF manifest from the file.

Adding raw jumbf method to the reader and exposing it to the bindings code.

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -764,14 +764,14 @@ pub unsafe extern "C" fn c2pa_reader_detailed_json(reader_ptr: *mut C2paReader) 
 /// The returned value MUST be released by calling c2pa_string_free
 /// and it is no longer valid after that call.
 #[no_mangle]
-pub unsafe extern "C" fn c2pa_reader_raw_jumbf(
+pub unsafe extern "C" fn c2pa_reader_raw_manifest_jumbf(
     reader_ptr: *mut C2paReader,
     manifest_bytes_ptr: *mut *const c_uchar,
 ) -> i64 {
     check_or_return_int!(reader_ptr);
     check_or_return_int!(manifest_bytes_ptr);
     let c2pa_reader = guard_boxed!(reader_ptr);
-    let result = c2pa_reader.jumbf_manifest();
+    let result = c2pa_reader.raw_manifest_jumbf();
     ok_or_return_int!(result, |manifest_bytes: Vec<u8>| {
         let len = manifest_bytes.len() as i64;
         if !manifest_bytes_ptr.is_null() {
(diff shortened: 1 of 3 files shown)


c2pa_c_ffi/src/c_api.rs
```diff
@@ -758,6 +758,30 @@ pub unsafe extern "C" fn c2pa_reader_detailed_json(reader_ptr: *mut C2paReader)
     to_c_string(c2pa_reader.detailed_json())
 }
 
+/// Returns a raw JUMBF byte array generated from a C2paReader.
+///
+/// # Safety
+/// The returned value MUST be released by calling c2pa_string_free
+/// and it is no longer valid after that call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_reader_raw_jumbf(
```
magaylor: I'd like for this naming to be consistent with the reader method, which they currently are not.

What would you folks prefer between `raw_jumbf`, `raw_manifest`, `manifest_jumbf`, `raw_manifest_jumbf`, or something else?

tmathern: (I would stick with `raw` something in the name, but defer to @mauricefisher64 @gpeacock for the naming pick)

tmathern:
(@magaylor I merged in latest changes to re-run tests with latest main)

c2pa_c_ffi/src/c_api.rs
magaylor: Sounds good. I'll move to raw_manifest_jumbf for the time being, as I think that is most clear

gpeacock:
Please explain the workflow you want to achieve. Are you trying to save pre-validated ingredient and add them to manifests later? 

The right way is to add the ingredient to a Builder and then archive the builder until you need it. You then have two options, either reconstruct the builder using Builder::from_archive or you can add the archived builder as an ingredient using builder.add_ingredient_from_stream(). In this case, you can use the format "application/c2pa" and the parent ingredient in the archive will be added as the ingredient.

We are changing the format of archived builders to be a single application/c2pa (.c2pa) asset instead of a zipped folder archive.

gpeacock (commented):
It sounds like you want this to reconstruct an ingredient, but there are other ways to do that which are much safer and will add ingredients correctly. What exactly is the workflow you need here?

shipitfaster:
@gpeacock any other questions/thoughts regarding the PR?

scouten-adobe:
This sounds like what we're now calling ingredient archives. Could you take a look at that and see if that now fits your needs? Please open a new ticket if not.

PR #1627, opened 2025-11-21
chore: relax tempfile bound

friel-openai:

The exact version makes integrating this package into workspaces with conflicting bounds difficult, e.g.: packages with bounds such as >3.16.0.

## Checklist

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -180,7 +180,7 @@ serde-transcode = "1.1.1"
 sha1 = "0.10.6"
 spki = { version = "0.7.3", optional = true }
 static-iref = "3.0"
-tempfile = ">=3.15.0, <4.0"
+tempfile = "3.23.0"
 thiserror = "2.0.17"
 url = "2.5.3"
 uuid = { version = "1.18.0", features = ["serde", "v4"] }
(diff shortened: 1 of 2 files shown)


sdk/Cargo.toml
```diff
@@ -176,11 +180,12 @@ serde-transcode = "1.1.1"
 sha1 = "0.10.6"
 spki = { version = "0.7.3", optional = true }
 static-iref = "3.0"
-tempfile = "=3.15.0"
+tempfile = ">=3.15.0, <4.0"
```
ok-nick: It should be okay to drop the bounds here (and below). We should be able to depend on the latest version, `tempfile = "3.23.0"`, as long as CI passes.

ok-nick (changes requested):

sdk/Cargo.toml
friel-openai: @ok-nick I saw that the reason this was set to `=3.15.0` was to ensure compatibility with wasip2, older versions were not compatible

ok-nick: I believe we can upgrade the wasip2 dependency version as well. The newer releases should be compatible.

sdk/Cargo.toml
```diff
@@ -237,7 +243,7 @@ rsa = { version = "0.9.6", features = ["sha2"] }
 spki = "0.7.3"
 
 [target.'cfg(target_env = "p2")'.dependencies]
-tempfile = { version = "3.15", features = ["nightly"] }
+tempfile = { version = ">=3.15.0, <4.0", features = ["nightly"] }
```
ok-nick: It looks like the latest version of `tempfile` now supports WASI even without the `nightly` feature. If so, these lines should be safe to remove.

ok-nick (changes requested):

sdk/Cargo.toml
friel-openai: removed the conditional target_env dependencies

ok-nick (approved):

PR #1630, opened 2025-11-24
feat: Restricted HTTP resolvers

ok-nick:
Introduces a new `RestrictedResolver` struct to restrict an inner `Sync/AsyncHttpResolver` with the specified allowed list. For the SDK, this is used specifically with the `core.allowed_network_hosts` setting to constrain HTTP requests to a known list of allowed hosts.

The default behavior of the SDK is to wrap a `Sync/AsyncGenericResolver` with a `RestrictedResolver` using the  `core.allowed_network_hosts` setting. Eventually HTTP resolvers and settings will be passed in directly as structs.

The behavior for `core.allowed_network_hosts` is defined as follows:
```rust
/// List of host patterns that are allowed for outbound network requests.
///
/// Each pattern may include:
/// - A scheme (e.g. `https://` or `http://`)
/// - A hostname, which may have a single leading wildcard (e.g. `*.contentauthenticity.org`)
///
/// Matching is case-insensitive. A wildcard pattern such as `*.contentauthenticity.org` matches
/// `sub.contentauthenticity.org`, but does not match `contentauthenticity.org` or `fakecontentauthenticity.org`.
/// If a scheme is present in the pattern, only URIs using the same scheme are considered a match. If the scheme
/// is omitted, any scheme is allowed as long as the host matches.
///
/// The behavior is as follows:
/// - `None` (default) no filtering enabled.
/// - `Some(vec)` where `vec` is empty, all outbound traffic is blocked.
/// - `Some(vec)` with at least one pattern, filtering enabled for only those patterns.
///
/// These settings are consumed by [`RestrictedResolver`].
///
/// For information on when the SDK might perform an outbound network request, see "[When do outbound network requests occur?]"
///
/// [When do outbound network requests occur?]: crate::http#when-do-outbound-network-requests-occur
/// [`HostPattern`]: crate::http::restricted::HostPattern
/// [`RestrictedResolver`]: crate::http::restricted::RestrictedResolver
```

### DIFF
--- a/.github/workflows/beta-preflight.yml
+++ b/.github/workflows/beta-preflight.yml
@@ -212,8 +212,15 @@ jobs:
       - name: Install cargo-binstall
         uses: cargo-bins/cargo-binstall@main
 
+      - name: Extract wasm-bindgen version from Cargo.lock
+        id: wasm-bindgen-version
+        run: |
+          VERSION=$(cargo metadata --format-version 1 --locked | jq -r '.packages[] | select(.name == "wasm-bindgen") | .version')
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Using wasm-bindgen version: $VERSION"
+
       - name: Install wasm-bindgen-cli
-        run: cargo binstall -y wasm-bindgen-cli --version 0.2.105
+        run: cargo binstall -y wasm-bindgen-cli --version ${{ steps.wasm-bindgen-version.outputs.version }}
 
       - name: Run Wasm tests
         run: cargo test -p c2pa --no-default-features --features rust_native_crypto,fetch_remote_manifests --target wasm32-unknown-unknown
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -331,8 +331,15 @@ jobs:
       - name: Install cargo-binstall
         uses: cargo-bins/cargo-binstall@main
 
+      - name: Extract wasm-bindgen version from Cargo.lock
+        id: wasm-bindgen-version
+        run: |
+          VERSION=$(cargo metadata --format-version 1 --locked | jq -r '.packages[] | select(.name == "wasm-bindgen") | .version')
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Using wasm-bindgen version: $VERSION"
+
       - name: Install wasm-bindgen-cli
-        run: cargo binstall -y wasm-bindgen-cli --version 0.2.105
+        run: cargo binstall -y wasm-bindgen-cli --version ${{ steps.wasm-bindgen-version.outputs.version }}
 
       - name: Run Wasm tests
         run: cargo test -p c2pa --no-default-features --features rust_native_crypto,fetch_remote_manifests,http_reqwest --target wasm32-unknown-unknown
--- a/.github/workflows/tier-1b.yml
+++ b/.github/workflows/tier-1b.yml
@@ -203,7 +203,7 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2025-06-05
+          toolchain: nightly-2025-11-04
           # Pinning to specific nightly build for now. More recent versions
           # introduce a lifetime check that creates a whole slew of build
           # errors.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ As of December 2025 and until the 1.0.0 version is released, the CAI team will o
 
 ## [Unreleased]
 
+## [0.72.1](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.72.0...c2pa-v0.72.1)
+_04 December 2025_
+
+### Added
+
+* Check if stream matches input format via file signature/structure ([#1528](https://github.com/contentauth/c2pa-rs/pull/1528))
+* Use MaybeSend to simplify code ([#1639](https://github.com/contentauth/c2pa-rs/pull/1639))
+
+### Documented
+
+* Add doc for CLI settings, misc copy edits ([#1636](https://github.com/contentauth/c2pa-rs/pull/1636))
+
+### Fixed
+
+* Set wasm-bindgen-cli version in CI ([#1644](https://github.com/contentauth/c2pa-rs/pull/1644))
+* Fix the timestamp cert chain validation to be independent of the order ([#1600](https://github.com/contentauth/c2pa-rs/pull/1600))
+* RFC 3161 fractional seconds ([#1640](https://github.com/contentauth/c2pa-rs/pull/1640))
+* Stack overflow and hang due to cyclic ingredients ([#1576](https://github.com/contentauth/c2pa-rs/pull/1576))
+* Allow untrusted failure code to be a `Valid` validation state ([#1623](https://github.com/contentauth/c2pa-rs/pull/1623))
+* Remove unmaintained `json5` dependency and other unused `config` crate sub-dependencies ([#1611](https://github.com/contentauth/c2pa-rs/pull/1611))
+
+### Other
+
+* Remove direct dependency on `windows-core` ([#1616](https://github.com/contentauth/c2pa-rs/pull/1616))
+
+### Updated dependencies
+
+* Bump lopdf from 0.31.0 to 0.38.0 ([#1359](https://github.com/contentauth/c2pa-rs/pull/1359))
+* Bump quick-xml from 0.37.5 to 0.38.3 ([#1575](https://github.com/contentauth/c2pa-rs/pull/1575))
+
 ## [0.72.0](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.71.3...c2pa-v0.72.0)
 _17 November 2025_
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,7 +160,7 @@ checksum = "965c2d33e53cb6b267e148a4cb0760bc01f4904c1cd4bb4002a085bb016d1490"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.110",
+ "syn 2.0.111",
  "synstructure",
 ]
 
@@ -172,7 +172,7 @@ checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.110",
+ "syn 2.0.111",
 ]
 
 [[package]]
@@ -208,7 +208,7 @@ checksum = "ddf3728566eefa873833159754f5732fb0951d3649e6e5b891cc70d56dd41673"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.110",
+ "syn 2.0.111",
 ]
 
 [[package]]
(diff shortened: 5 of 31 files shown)


sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+            };
+
+        Self {
+            pattern,
+            scheme,
+            host,
+        }
+    }
+
+    /// Returns if the given URI matches the `HostPattern`.
```
scouten-adobe: Returns `true` if ...

sdk/src/builder.rs
```diff
@@ -617,6 +617,13 @@ impl Builder {
         R: Read + Seek + Send,
     {
         let settings = crate::settings::get_settings().unwrap_or_default();
+        let http_resolver = if _sync {
+            SyncGenericResolver::new()
+        } else {
+            AsyncGenericResolver::new()
+        };
+        let mut http_resolver = RestrictedResolver::new(http_resolver);
```
scouten-adobe, line 625: Very clever use of templating here. 😎

sdk/src/http/mod.rs
```diff
@@ -11,6 +11,36 @@
...
+//! HTTP abstraction layer.
+//!
+//! This module defines generic traits and helpers for performing HTTP requests
+//! without hard-wiring a specific HTTP client. It allows host applications to
+//! plug in their own HTTP implementation, restrict where the SDK may connect,
+//! or disable networking entirely.
+//!
+//! # When do network requests occur?
+//!
+//! The SDK may issue HTTP/S requests in the following scenarios:
```
tmathern, line 23: Shouldn't we only do HTTPS?

sdk/src/http/mod.rs
```diff
@@ -150,6 +182,14 @@ pub enum HttpResolverError {
     #[error("the async http resolver is not implemented")]
     AsyncHttpResolverNotImplemented,
 
+    /// The remote URI is blocked by the allowed list.
+    ///
+    /// The allowed list is normally set in a [`RestrictedResolver`].
+    ///
+    /// [`RestrictedResolver`]: restricted::RestrictedResolver
+    #[error("remote URI \"{uri}\" is not permitted by the allowed list")]
```
tmathern, line 192: We should document/explain where that allow list is, and who manages it.

sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+// specific language governing permissions and limitations under
+// each license.
+
+//! HTTP request restriction layer.
+//!
+//! # Why restrict network requests?
+//! In some environments, you may not want the SDK to talk to arbitrary hosts. Restricting
+//! network requests help to:
+//! - Reduce SSRF-style risks (e.g. requests to internal services).
+//! - Constrain requests to a small, trusted set of domains.
```
tmathern, line 24: Point to the allowlist please... (so it's easy to find for a developer)

sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+//!
+//! # OCSP and other dynamic endpoints
+//! Some protocols used by the SDK (like OCSP or CRLs) discover endpoints from certificate
+//! metadata at runtime. In a restricted environment, there is no way for the resolver to
+//! know that these endpoints are "special" unless you anticipate them in advance and add
+//! their hosts to the allow-list.
```
tmathern: So any user can modify the allow-list?

sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+/// A host/scheme pattern used to restrict network requests.
+///
+/// Each pattern may include:
+/// - A scheme (e.g. `https://` or `http://`)
+/// - A hostname, which may have a single leading wildcard (e.g. `*.contentauthenticity.org`)
+///
+/// Matching is case-insensitive. A wildcard pattern such as `*.contentauthenticity.org` matches
```
tmathern, line 136: This should be made somehow more prominent (albeit is is somewhat implied since it's URLs).

sdk/src/settings/mod.rs
```diff
@@ -219,6 +222,30 @@ pub struct Core {
     /// [`IdentityAssertion`]: crate::identity::IdentityAssertion
     /// [`Reader`]: crate::Reader
     pub decode_identity_assertions: bool,
+    /// List of host patterns that are allowed for network requests.
```
tmathern, line 230: Could we add tests/examples with a set of allowed patterns, and some being blocked? So we can point people to it for reference.

sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct HostPattern {
+    pattern: String,
+    scheme: Option<String>,
+    host: Option<String>,
+}
+
+impl HostPattern {
+    /// Creates a new `HostPattern` with the given pattern.
+    pub fn new(pattern: &str) -> Self {
```
tmathern, line 173: Could you add a test with these patterns: 
`let pattern = HostPattern::new("https://"); `
and
`let pattern = HostPattern::new("https:// ");`
? 

Corollary - do we need port number handling too?

sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+        let pattern = pattern.to_ascii_lowercase();
```
tmathern, line 174: Careful with internationalization here. Is it handled correctly? The cyrillic a and the ascii a look the same, but won't resolve to the same URLs. (Look at [this](https://en.wikipedia.org/wiki/IDN_homograph_attack)).

sdk/src/http/restricted.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        Self {
+            pattern,
+            scheme,
+            host,
+        }
+    }
+
+    /// Returns if the given URI matches the `HostPattern`.
+    pub fn matches(&self, uri: &Uri) -> bool {
```
tmathern, line 203: What about IPs v4? Should those be allowed? Can you add a test with `https://127.0.0.1` for instance? And `https://127.0.0.1:8080`? Same question for IPv6.

Asking because there might be IPs used directly in development settings. Or maybe we should disallow IPs. To be clarified, @scouten-adobe @gpeacock @ok-nick.

tmathern (approved):

sdk/src/http/mod.rs
scouten-adobe: I'm not _aware_ of any case where we would employ plain HTTP, but I can't say definitively that it is always prohibited. A possible answer here is to add a new config option which is disallow plain HTTP.

sdk/src/http/restricted.rs
scouten-adobe: I'm not sure how this is relevant since we aren't providing an allow-list.

sdk/src/http/restricted.rs
scouten-adobe: I'd prefer to allow explicit IP addresses. I could imagine some environments where that might be required.

sdk/src/http/restricted.rs
tmathern: Well, for a developer using it, it's nice to know where to go to configure it :). Especially to make it clear it's with all the other settings too, and not something different.

sdk/src/http/restricted.rs
ok-nick: Rust ignores non-ASCII characters with this method. In any case, I believe ASCII is disallowed in URL hostnames.

sdk/src/http/restricted.rs
ok-nick: This is implicitly allowed, I can add a note in the docs about it.

Edit: For port handling, I can add that in as well.

sdk/src/http/restricted.rs
ok-nick: ~What behavior do you expect for these two cases? Port numbers are implicitly supported, the implementation here is a basic string pattern matching, the actual URL parsing/validation is deferred to the underlying HTTP resolver.~

Edit: Correction, ports are ignored at the moment when reading the given URL, I can add support for that. I'll also add tests for only the scheme.

sdk/src/http/mod.rs
ok-nick: You can have a pattern such as `https://` to only enable HTTPS.

sdk/src/http/restricted.rs
ok-nick: > `let pattern = HostPattern::new("https:// ");`

What do you expect to happen in this case? I'd consider it undefined behavior because spaces are not allowed in URIs. Also note that the rules for these patterns are loosely defined, there are a lot of edge cases that are purposefully undefined due to scope.

sdk/src/http/restricted.rs
ok-nick: What do you suggest?

sdk/src/settings/mod.rs
ok-nick: There are a lot of unit tests demonstrating this functionality, is that what you're looking for?

sdk/src/http/restricted.rs
tmathern: Docs in the readme, with the settings?

sdk/src/settings/mod.rs
tmathern: Yes. Sometimes people add ane xample in the comment too. It's already long, so I am not afraid of a few more lines. Wdyt?

sdk/src/http/restricted.rs
tmathern: I think it should reject/ignore it, so it couldn't be used as "bypass" or catch-all. What would it even mean to allow it? Undefined behavior is not great, let's define what it does and stick to it.

sdk/src/http/restricted.rs
tmathern: Cyrillic (for instance) is allowed in hostnames through the use of Internationalized Domain Names ([IDNs](https://en.wikipedia.org/wiki/Internationalized_domain_name)).

Here is an example of what I mean, regarding visually similar strings:

```
use std::str::FromStr;

fn main() {
    let ascii = "PayPal.com";
    let homoglyph = "PаyPаl.com"; // Cyrillic 'а' (U+0430) or at least I hope copy/paste kept it

    let ascii_lower = ascii.to_ascii_lowercase();
    let glyph_lower = homoglyph.to_ascii_lowercase();

    println!("ASCII:    '{}'", ascii_lower); 
    println!("Glyph:    '{}'", glyph_lower); 
    println!("Match? {}", ascii_lower == glyph_lower); //  although visually I can't tell the difference
}
```
(https://play.rust-lang.org/?version=stable&mode=debug&edition=2024 can swiftly run it).

I would have thought because of this we need some kind of normalization? Due to the possible encoding of IDNs?

It may be too much tough for now. An angle to still keep in mind.

sdk/src/http/restricted.rs
tmathern: If we allow IPs we need ports. For dev env, we often have IPs+port.

tmathern: (Can be a follow-up PR, but close follow-up)

tmathern:
@ok-nick Maybe a suggestion because there is a lot of related work in-flight: Group your PRs into a feature branch (as you sit fit)? So they will be easier to integrate together (since it seems they'll be in the same release too), and maybe easier to maintain while in their rbanch(es). Wdyt?

https://github.com/contentauth/c2pa-rs/pull/1630 and https://github.com/contentauth/c2pa-rs/pull/1618 I could see released together. For the rest, developer's choice!

sdk/src/settings/mod.rs
ok-nick: Added.

PR #1631, opened 2025-11-25
feat: Adds `Context` to contain settings, HTTPS resolvers, and signers

gpeacock:
This adds a Context struct that contains settings, https resolvers and signers. 
Readers and Builders can be constructed in a thread safe manner using Context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ As of December 2025 and until the 1.0.0 version is released, the CAI team will o
 
 ## [Unreleased]
 
+## [0.73.2](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.73.1...c2pa-v0.73.2)
+_22 December 2025_
+
+### Fixed
+
+* Nested ingredient serializing-deserializing (fix for #1685) ([#1686](https://github.com/contentauth/c2pa-rs/pull/1686))
+* Flush stream to finish write operations before rewind for FFI layer improvements ([#1672](https://github.com/contentauth/c2pa-rs/pull/1672))
+
 ## [0.73.1](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.73.0...c2pa-v0.73.1)
 _19 December 2025_
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,9 +213,9 @@ dependencies = [
 
 [[package]]
 name = "async-lock"
-version = "3.4.1"
+version = "3.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fd03604047cee9b6ce9de9f70c6cd540a0520c813cbd49bae61f33ab80ed1dc"
+checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
 dependencies = [
  "event-listener",
  "event-listener-strategy",
@@ -435,7 +435,7 @@ dependencies = [
 
 [[package]]
 name = "c2pa"
-version = "0.73.1"
+version = "0.73.2"
 dependencies = [
  "anyhow",
  "asn1-rs",
@@ -539,7 +539,7 @@ dependencies = [
 
 [[package]]
 name = "c2pa-c-ffi"
-version = "0.73.1"
+version = "0.73.2"
 dependencies = [
  "c2pa",
  "cbindgen",
@@ -553,7 +553,7 @@ dependencies = [
 
 [[package]]
 name = "c2pa_macros"
-version = "0.73.1"
+version = "0.73.2"
 dependencies = [
  "quote",
  "syn 2.0.111",
@@ -561,7 +561,7 @@ dependencies = [
 
 [[package]]
 name = "c2patool"
-version = "0.26.8"
+version = "0.26.9"
 dependencies = [
  "anyhow",
  "assert_cmd",
@@ -1318,7 +1318,7 @@ dependencies = [
 
 [[package]]
 name = "export_schema"
-version = "0.73.1"
+version = "0.73.2"
 dependencies = [
  "anyhow",
  "c2pa",
@@ -2179,9 +2179,9 @@ dependencies = [
 
 [[package]]
 name = "itoa"
-version = "1.0.15"
+version = "1.0.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
+checksum = "7ee5b5339afb4c41626dde77b7a611bd4f2c202b897852b4bcf5d03eddc61010"
 
 [[package]]
 name = "jfifdump"
@@ -2338,7 +2338,7 @@ checksum = "112b39cec0b298b6c1999fee3e31427f74f676e4cb9879ed1a121b43661a4154"
 
 [[package]]
 name = "make_test_images"
-version = "0.73.1"
+version = "0.73.2"
 dependencies = [
  "anyhow",
  "c2pa",
@@ -2939,9 +2939,9 @@ dependencies = [
 
 [[package]]
 name = "portable-atomic"
-version = "1.11.1"
+version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f84267b20a16ea918e43c6a88433c2d54fa145c92a811b5b047ccbe153674483"
+checksum = "f59e70c4aef1e55797c2e8fd94a4f2a973fc972cfde0e0b05f683667b0cd39dd"
 
 [[package]]
 name = "portable-atomic-util"
@@ -3409,9 +3409,9 @@ checksum = "7a2d987857b319362043e95f5353c0535c1f58eec5336fdfcf626430af7def58"
 
 [[package]]
 name = "reqwest"
-version = "0.12.26"
+version = "0.12.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3b4c14b2d9afca6a60277086b0cc6a6ae0b568f6f7916c943a8cdc79f8be240f"
+checksum = "eddd3ca559203180a307f12d114c268abf583f59b03cb906fd0b3ff8646c1147"
 dependencies = [
  "base64",
  "bytes",
@@ -3585,9 +3585,9 @@ checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"
-version = "1.0.20"
+version = "1.0.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28d3b2b1366ec20994f1fd18c3c594f05c5dd4bc44d8bb0c1c632c8d6829481f"
+checksum = "62049b2877bf12821e8f9ad256ee38fdc31db7387ec2d3b3f403024de2034aea"
 
 [[package]]
 name = "same-file"
@@ -3776,9 +3776,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.145"
+version = "1.0.146"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "402a6f66d8c709116cf22f558eab210f5a50187f702eb4d7e5ef38d9a7f1c79c"
+checksum = "217ca874ae0207aac254aa02c957ded05585a90892cc8d87f9e5fa49669dadd8"
 dependencies = [
  "indexmap 2.12.1",
  "itoa",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ members = [
 
 # members in this workspace can share this version setting
 [workspace.package]
-version = "0.73.1"
+version = "0.73.2"
 
 [workspace.dependencies]
 c2pa = { path = "sdk", default-features = false }
--- a/c2pa_c_ffi/CHANGELOG.md
+++ b/c2pa_c_ffi/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.73.2](https://github.com/contentauth/c2pa-rs/compare/c2pa-c-ffi-v0.73.1...c2pa-c-ffi-v0.73.2)
+_22 December 2025_
+
 ## [0.73.1](https://github.com/contentauth/c2pa-rs/compare/c2pa-c-ffi-v0.73.0...c2pa-c-ffi-v0.73.1)
 _19 December 2025_
 
--- a/c2pa_c_ffi/Cargo.toml
+++ b/c2pa_c_ffi/Cargo.toml
@@ -23,7 +23,7 @@ rust_native_crypto = ["c2pa/rust_native_crypto"]
 pdf = ["c2pa/pdf"]
 
 [dependencies]
-c2pa = { path = "../sdk", version = "0.73.1", default-features = false, features = [
+c2pa = { path = "../sdk", version = "0.73.2", default-features = false, features = [
     "add_thumbnails",
     "fetch_remote_manifests",
     "file_io",
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.26.9](https://github.com/contentauth/c2pa-rs/compare/c2patool-v0.26.8...c2patool-v0.26.9)
+_22 December 2025_
+
 ## [0.26.8](https://github.com/contentauth/c2pa-rs/compare/c2patool-v0.26.7...c2patool-v0.26.8)
 _19 December 2025_
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "c2patool"
 default-run = "c2patool"
-version = "0.26.8"
+version = "0.26.9"
 description = "Tool for displaying and creating C2PA manifests."
 authors = [
     "Gavin Peacock <gpeacock@adobe.com>",
@@ -28,7 +28,7 @@ networking = [
 [dependencies]
 anyhow = "1.0"
 atree = "0.5.2"
-c2pa = { path = "../sdk", version = "0.73.1", features = [
+c2pa = { path = "../sdk", version = "0.73.2", features = [
     "fetch_remote_manifests",
     "file_io",
     "add_thumbnails",
--- a/make_test_images/Cargo.toml
+++ b/make_test_images/Cargo.toml
@@ -12,7 +12,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
 
 [dependencies]
 anyhow = "1.0.40"
-c2pa = { path = "../sdk", version = "0.73.1", features = [
+c2pa = { path = "../sdk", version = "0.73.2", features = [
 	"fetch_remote_manifests",
 	"file_io",
 	"add_thumbnails",


sdk/src/builder.rs
ok-nick, line 760: We should mention that the ingredient here will not inherit the builder's settings.

sdk/src/reader.rs
```diff
@@ -49,6 +48,18 @@ use crate::{
...
+/// todo: move this to a common module
+#[cfg(not(target_arch = "wasm32"))]
+pub trait MaybeSend: Send {}
+#[cfg(target_arch = "wasm32")]
+pub trait MaybeSend {}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl<T: Send> MaybeSend for T {}
+#[cfg(target_arch = "wasm32")]
+impl<T> MaybeSend for T {}
```
ok-nick, line 61: Nice, this is much cleaner.

sdk/src/store.rs
```diff
@@ -3734,107 +3657,60 @@ impl Store {
...
-        validation_log: &mut StatusTracker,
-        http_resolver: &impl AsyncHttpResolver,
-        settings: &Settings,
-    ))]
+    #[async_generic]
     #[cfg(not(target_arch = "wasm32"))]
     pub fn from_manifest_data_and_stream(
         c2pa_data: &[u8],
         format: &str,
         mut stream: impl Read + Seek + Send,
```
ok-nick: We should be able to use the new `MaybeSend` here too.

sdk/src/store.rs
```diff
@@ -3612,27 +3561,18 @@ impl Store {
...
-        verify: bool,
-        validation_log: &mut StatusTracker,
-        http_resolver: &impl AsyncHttpResolver,
-        settings: &Settings,
-    ))]
+    #[async_generic]
     #[cfg(not(target_arch = "wasm32"))]
     pub fn from_stream(
         format: &str,
         mut stream: impl Read + Seek + Send,
```
ok-nick: `MaybeSend` here too.

sdk/src/context.rs
```diff
@@ -0,0 +1,217 @@
...
+
+/// Implement for serde_json::Value
+impl IntoSettings for serde_json::Value {
+    fn into_settings(self) -> Result<Settings> {
+        let json_str = serde_json::to_string(&self).map_err(Error::JsonError)?;
+        let mut settings = Settings::default();
+        settings.update_from_str(&json_str, "json")?;
+        Ok(settings)
+    }
+}
```
ok-nick: How come `IntoSettings` versus implementing `TryFrom`?

sdk/src/context.rs
```diff
@@ -0,0 +1,217 @@
...
+
+pub struct Context {
+    settings: Settings,
+    http_resolver: OnceCell<Box<dyn SyncHttpResolver>>,
+    http_resolver_async: OnceCell<Box<dyn AsyncHttpResolver>>,
+    signer: OnceCell<Box<dyn Signer>>,
+    _signer_async: OnceCell<Box<dyn AsyncSigner>>,
```
ok-nick: `OnceCell` will remove the `Sync` impl on the `Context` but I suppose the trait objects inside aren't `Send` nor `Sync` anyways.  If this were to be optionals instead, I can see how the getters would require a mutable ref to self (the get or cache behavior), which isn't clear since we really only need to mutate it once on first access. We talked a little bit about this offline but I think with this design I agree that interior mutability makes more sense.

This is part of the reason I like to separate the "intermediate state" structs from the final state to avoid having to use optional or cells although that would likely require a different design.

sdk/src/settings/signer.rs
```diff
@@ -70,81 +70,90 @@ impl SignerSettings {
...
+                alg,
+                sign_cert,
+                private_key,
+                tsa_url,
+                referenced_assertions: _,
+            } => create_signer::from_keys(
+                sign_cert.as_bytes(),
+                private_key.as_bytes(),
+                alg,
+                tsa_url.to_owned(),
```
ok-nick: Is tsa_url already owned?

sdk/src/context.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /// Returns a reference to the signer.
+    #[allow(clippy::unwrap_used)] // switch to get_or_try_init when stable
+    pub fn signer(&self) -> Result<&dyn Signer> {
+        match self.signer.get() {
+            Some(s) => Ok(s),
+            None => {
+                if let Some(c2pa_settings) = self.settings().signer.clone() {
+                    let mut signer = c2pa_settings.c2pa_signer()?;
```
ok-nick: Not a blocker, but something to consider here is if c2pa_signer could constructed from a reference to the settings and borrow its fields.

sdk/src/content_credential.rs
```diff
@@ -0,0 +1,323 @@
...
+impl std::fmt::Debug for ContentCredential<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let value = self.reader_value().map_err(|_| std::fmt::Error)?;
+        f.write_str(
+            serde_json::to_string_pretty(&value)
+                .map_err(|_| std::fmt::Error)?
+                .as_str(),
+        )
+    }
+}
```
ok-nick: Maybe we should leave Debug as the default implementation? I liked this approach previously where the Debug impl provided a detailed "pretty" interface, although I think it makes more sense as a peek into the internal data structures for debugging.

sdk/src/content_credential.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    /// internal implementation to add ingredient from stream
+    fn add_ingredient_from_stream(
+        &mut self,
+        relationship: Relationship,
+        format: &str,
+        mut stream: impl Read + Seek + Send,
+    ) -> Result<HashedUri> {
+        let (ingredient_assertion, store) =
+            Ingredient::from_stream(relationship.clone(), format, &mut stream, self.context)?;
```
ok-nick: I wonder if we should rename our assertions with a "Assertion" suffix to differentiate in situations such as this. I think it also makes it more clear in the `crate::assertions` docs.rs to differentiate against the assertion sub-structs.

docs/content_credentials.md
```diff
@@ -78,7 +70,7 @@ A future lower level API will wrap the Claim and Store structures providing a si
 
 ## Asset objects (not directly related to the above)
 
-An Asset object creates a persistent layer over the asset_io traits. Currently we parse entire asset every time we need to access information about it. We have separate passes for XMP, JUMBF, Offset/box generation & etc..
+An Asset object creates a semi-persistent layer over the asset_io traits. Currently we parse entire asset every time we need to access information about it. We have separate passes for XMP, JUMBF, Offset/box generation & etc..
```
tmathern: Why is it *semi* persistent? Because in spirit it is a kind of cache to avoid multiple re-parsings?

sdk/src/assertions/ingredient.rs
```diff
@@ -520,6 +527,103 @@ impl Ingredient {
 
         ingredient_map.end()
     }
+
+    /// Create a new Ingredient assertion from a stream
+    /// You must specify the relationship and format.
+    /// This will return the new Ingredient and associated manifest_bytes if they exist.
+    pub(crate) fn from_stream(
+        relationship: Relationship,
```
tmathern, line 535: Is this a place where we could enforce there is only one parentOf ingredient, or should the caller do so? (The question is if we want/can do the check deep down, or move it up. I think it makes more sens eto move it up as much as we can).

sdk/src/assertions/ingredient.rs
```diff
@@ -520,6 +527,103 @@ impl Ingredient {
...
+            if thumbnail.is_some()
+                && self
+                    .validation_results
+                    .as_ref()
+                    .map(|r| r.validation_state())
+                    != Some(crate::ValidationState::Invalid)
+            {
+                self.thumbnail = thumbnail;
+            } else {
+                // todo:: add thumbnail here
```
tmathern, line 622: Do we always want one? Maybe here is a place to pull down settings and check what was wished when adding/loading this ingredient?

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -360,7 +355,7 @@ pub(crate) fn fetch_and_check_ocsp_response(
     // use supplied override time if provided
     let signing_time: Option<DateTime<Utc>> = match tst_info {
         Some(tst_info) => Some(tst_info.gen_time.clone().into()),
-        None => validate_cose_tst_info(sign1, data, ctp, validation_log, settings)
+        None => validate_cose_tst_info(sign1, data, ctp, validation_log, context.settings())
```
tmathern, line 358: I wonder, if to be even more future-proof, it wouldn't be worthwhile to pass around the context. So you could change the content of the context, without changing all the function signatures again. Wdyt?
(I see settings, http resolvers in the context, in the future maybe IO handlers? Or whatever we can come up with).

sdk/src/crypto/ocsp/fetch.rs
```diff
@@ -145,14 +142,8 @@ fn process_ocsp_responders(certs: &[Vec<u8>]) -> Option<Vec<OcspRequestData>> {
...
-#[async_generic(async_signature(
-    certs: &[Vec<u8>],
-    http_resolver: &impl AsyncHttpResolver,
-))]
-pub(crate) fn fetch_ocsp_response(
-    certs: &[Vec<u8>],
-    http_resolver: &impl SyncHttpResolver,
-) -> Option<Vec<u8>> {
+#[async_generic]
+pub(crate) fn fetch_ocsp_response(certs: &[Vec<u8>], context: &Context) -> Option<Vec<u8>> {
```
tmathern, line 146: Here for instance, we already pass in the whole context.

sdk/src/settings/mod.rs
```diff
@@ -424,6 +424,78 @@ impl Settings {
         Settings::from_string(toml, "toml").map(|_| ())
     }
 
+    /// Update this `Settings` instance from a string representation.
+    /// This overlays the provided configuration on top of the current settings
+    /// without affecting the global thread-local settings.
```
tmathern, line 477: I am not sure what this means `without affecting the global thread-local settings`. Is it going to be passed around by value*, and the thread-set settings prevail when the value isn't passed in.

*or reference, but basically a way to read these values only.

sdk/src/builder.rs
```diff
@@ -861,26 +883,26 @@ impl Builder {
...
                 &mut stream,
-                false,
                 &mut validation_log,
-                &SyncGenericResolver::new(),
-                &settings,
+                &context,
             )?;
-            let reader = Reader::from_store(store, &mut validation_log, &settings)?;
+            let mut reader = Reader::new(context);
+            reader.with_store(store, &mut validation_log)?;
```
tmathern, line 978: This reader will use the context too? Since it is created from the same context?

sdk/src/builder.rs
```diff
@@ -1380,7 +1395,7 @@ impl Builder {
                 crate::utils::thumbnail::make_thumbnail_bytes_from_stream(
                     format,
                     &mut stream,
-                    settings,
+                    self.context.settings(),
```
tmathern, line 1477: I wonder if this is an API where we could benefit from passing in the full context too.

sdk/src/builder.rs
```diff
@@ -1464,7 +1477,7 @@ impl Builder {
         }
         self.definition.format = format.to_string();
         self.definition.instance_id = format!("xmp:iid:{}", Uuid::new_v4());
-        let mut store = self.to_store(&settings)?;
+        let mut store = self.to_store()?;
```
tmathern, line 1559: Will the store here even need to also have its/a context? Wondering because it did take settings before too, so should we let it get possibly its own.new context too, rather than getting it always from self?

sdk/src/builder.rs
```diff
@@ -1597,17 +1609,17 @@ impl Builder {
...
+        self.maybe_add_thumbnail(&format, source)?;
 
         // convert the manifest to a store
-        let mut store = self.to_store(&settings)?;
+        let mut store = self.to_store()?;
 
         // sign and write our store to to the output image file
         if _sync {
-            store.save_to_stream(&format, source, dest, signer, &http_resolver, &settings)
+            store.save_to_stream(&format, source, dest, signer, &self.context)
```
tmathern, line 1698: Simpler calls are nice, looks cool. This will likely help to stabilize the API signatures too, since context could evolve in the future and hold more "things".

sdk/src/claim.rs
```diff
@@ -1824,17 +1849,16 @@ impl Claim {
...
         asset_data: &mut ClaimAssetData<'_>,
         svi: &StoreValidationInfo<'_>,
         cert_check: bool,
         ctp: &CertificateTrustPolicy,
         validation_log: &mut StatusTracker,
-        http_resolver: &impl AsyncHttpResolver,
-        settings: &Settings,
+        context: &Context,
     ) -> Result<()> {
+        let settings = context.settings();
```
tmathern, line 1861: Are the settings still used directly? (I did not see one).
If in any call, I'd consider passing the context instead.

sdk/src/content_credential.rs
```diff
@@ -0,0 +1,328 @@
...
+        &mut self,
+        format: &str,
+        source: &mut R,
+        dest: &mut W,
+    ) -> Result<Vec<u8>>
+    where
+        R: Read + Seek + Send,
+        W: Write + Read + Seek + Send,
+    {
+        let signer = self.context.signer()?;
```
tmathern: I think that's the best approach. Let's a user also define a signer and put it in the context,

sdk/src/reader.rs
```diff
@@ -362,12 +339,7 @@ impl Reader {
         mut stream: impl Read + Seek + Send,
         mut fragment: impl Read + Seek + Send,
     ) -> Result<Self> {
-        let settings = crate::settings::get_settings().unwrap_or_default();
-        let http_resolver = if _sync {
-            SyncGenericResolver::new()
-        } else {
-            AsyncGenericResolver::new()
-        };
+        let mut reader = Reader::new(Context::new());
```
tmathern: Should we offer to pass in a context to `from_fragment`?

sdk/src/reader.rs
```diff
@@ -403,10 +378,8 @@ impl Reader {
         path: P,
         fragments: &Vec<std::path::PathBuf>,
     ) -> Result<Reader> {
-        let settings = crate::settings::get_settings().unwrap_or_default();
-        let http_resolver = SyncGenericResolver::new();
+        let mut reader = Reader::new(Context::new());
```
tmathern: Should we offer to pass in a context to `from_fragmented_files`?

sdk/src/reader.rs
```diff
@@ -572,13 +546,13 @@ impl Reader {
 
     /// Get the [`ValidationState`] of the manifest store.
     pub fn validation_state(&self) -> ValidationState {
-        let settings = crate::settings::get_settings().unwrap_or_default();
+        let context = Context::new();
```
tmathern, line 569: Does it need to be a new one?

sdk/src/store.rs
```diff
@@ -2582,14 +2555,14 @@ impl Store {
...
-        settings: &Settings,
+        context: &Context,
     ) -> Result<Vec<u8>> {
         let mut jumbf_bytes =
             self.prep_embeddable_store(signer.reserve_size(), dh, asset_reader)?;
 
         // sign contents
         let pc = self.provenance_claim().ok_or(Error::ClaimEncoding)?;
-        let sig = self.sign_claim(pc, signer, signer.reserve_size(), settings)?;
+        let sig = self.sign_claim(pc, signer, signer.reserve_size(), context.settings())?;
```
tmathern, line 2611: What if we would pass in the whole context here instead?

sdk/src/store.rs
```diff
@@ -2613,15 +2586,15 @@ impl Store {
...
+        context: &Context,
     ) -> Result<Vec<u8>> {
         let mut jumbf_bytes =
             self.prep_embeddable_store(signer.reserve_size(), dh, asset_reader)?;
 
         // sign contents
         let pc = self.provenance_claim().ok_or(Error::ClaimEncoding)?;
         let sig = self
-            .sign_claim_async(pc, signer, signer.reserve_size(), settings)
+            .sign_claim_async(pc, signer, signer.reserve_size(), context.settings())
```
tmathern, line 2643: What if we would pass in the whole context here instead?

tmathern (commented):
It may feel like overkill now, but I believe it would help to stabilize the API to pass in the full context everywhere when anything (namely, settings currently) is needed. Because then the context can evolve to contain more, and we can use it in the code, without changing function signatures.

cli/src/main.rs
```diff
@@ -870,14 +870,13 @@ fn main() -> Result<()> {
...
             reader.to_folder(&output)?;
-            let report = reader.to_string();
+            let _report = reader.to_string();
             if args.detailed {
                 // for a detailed report first call the above to generate the thumbnails
                 // then call this to add the detailed report
                 let detailed = format!("{reader:#?}");
                 File::create(output.join("detailed.json"))?.write_all(&detailed.into_bytes())?;
             }
-            File::create(output.join("manifest_store.json"))?.write_all(&report.into_bytes())?;
```
scouten-adobe, line 879: This change feels unrelated to the title of this PR. Can it be broken out into a separate PR and discussed there?

docs/content_credentials.md
```diff
@@ -52,14 +52,6 @@ For Builder archives - Sign and save a manifest, either embedded or sidecar.  If
 - Add a captured .c2pa archived ingredient using add_ingredient_from_stream. This will use the parent ingredient in the archive as the ingredient added.
 
 
-
-
-### Questions
```
scouten-adobe, line 57: These changes also seem unrelated to Context.

scouten-adobe (changes requested):
I think there are two or three bodies of work here; some refactoring related to assertions and assets that aren't directly related to the Context proposal. I made a few comments and stopped reviewing; would appreciate it if you would split this into separate PRs for those separate efforts.

sdk/src/assertions/ingredient.rs
gpeacock: Can't do it here since this just constructs a single assertion, but we could when we add it to a builder

sdk/src/assertions/ingredient.rs
gpeacock: yes, we could do that if we had context in this api, and perhaps we should - that's why I had not implemented it yet.

sdk/src/crypto/cose/ocsp.rs
gpeacock: Probably in most cases that's a good idea, but when we reach a point where we have a potentially separate crate, we may want to pass only what we need. There's a tradeoff between proper modularity and sharing implementations.

sdk/src/crypto/ocsp/fetch.rs
gpeacock: I really only plumbed this through as far as I needed for this PR. but I'm certainly open to doing more in another PR.

sdk/src/reader.rs
gpeacock: good point, maybe not, let me look.

docs/content_credentials.md
gpeacock: I'd rather not add these comments back in just in order to delete them again. I have removed the content_credentials.rs and other change made to c2patool.

cli/src/main.rs
gpeacock: That might be a lot of work that will further delay this effort, but I'll take a look and give it a try.

sdk/src/store.rs
gpeacock: There are a lot of tradeoffs in where to pass context vs settings vs individual values. I'm not trying to fully answer that in this PR, just trying to inject the Context at the higher level so we can start using it.

sdk/src/context.rs
gpeacock: I tried changing this and it got too complicated for now, this is simpler

sdk/src/context.rs
gpeacock: changed this to use option and an enum

sdk/src/assertions/ingredient.rs
gpeacock: at any rate, I've kept this private to the crate so I can complete the work on this in a later PR.

sdk/src/content_credential.rs
gpeacock: Not a bad idea but a very breaking change and not going into this PR

sdk/src/content_credential.rs
gpeacock: I've removed the ContentCredential stuff from this PR, but I'll review that in a new one.

sdk/src/settings/signer.rs
gpeacock: it was and i changed this, thanks

sdk/src/builder.rs
gpeacock: I don't know what you mean by that, this will use the builder's settings. There are no ingredient settings.

sdk/src/http/restricted.rs
```diff
@@ -14,15 +14,23 @@
 //! HTTP request restriction layer.
 //!
 //! This module provides a [`RestrictedResolver`] that wraps an existing [`SyncHttpResolver`]
-//! or [`AsyncHttpResolver`]. The SDK can also manage an allowed list for you via the
-//! [`Core::allowed_network_hosts`] setting.
+//! or [`AsyncHttpResolver`] to enforce host filtering.
+//!
+//! Note: The built-in [`SyncGenericResolver`] and [`AsyncGenericResolver`] already support
+//! host filtering natively. You only need [`RestrictedResolver`] if you want to wrap a
```
ok-nick: Do they? The implementation for generic resolvers seems to allow any host?

sdk/src/http/mod.rs
```diff
@@ -67,6 +67,14 @@ pub trait SyncHttpResolver {
...
         request: Request<Vec<u8>>,
     ) -> Result<Response<Box<dyn Read>>, HttpResolverError>;
+
+    /// Returns true if the given URI is allowed by this resolver's policy.
+    ///
+    /// Default implementation allows all URIs. Override this method to implement
+    /// host filtering or other URI-based access control.
+    fn is_uri_allowed(&self, _uri: &http::Uri) -> bool {
+        true
+    }
```
ok-nick: My only concern with adding this to the trait is that it's effectively a noop unless the implementor of http_resolver explicitly calls it. That's part of the reason I left the filtering up to the implementation of the http_resolve, in this case it can be misleading if not respected.

sdk/src/http/restricted.rs
```diff
@@ -68,7 +76,7 @@ impl<T> RestrictedResolver<T> {
     }
 
     /// Creates a new `RestrictedResolver` with the specified allowed list.
-    #[allow(dead_code)] // TODO: temp until http module is public
+    #[allow(dead_code)] // Public API, not used internally
```
ok-nick, line 75: Note: when `pub(crate)`->`pub` on the `http` module is changed in `lib.rs` we will no longer need this. I think we should make that change in this PR since HTTP resolvers will be part of the public API in the `Context`.

sdk/src/reader.rs
gpeacock: yes, I just haven't gotten there yet.

sdk/src/reader.rs
gpeacock: Thanks for catching this, that was a bug and I fixed it!

sdk/src/builder.rs
```diff
@@ -1608,21 +1630,102 @@ impl Builder {
...
+    ///
+    /// let mut source = std::fs::File::open("tests/fixtures/C.jpg")?;
+    /// let mut dest = Cursor::new(Vec::new());
+    ///
+    /// // Save with automatic signer from context
+    /// builder.save_to_stream("image/jpeg", &mut source, &mut dest)?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn save_to_stream<R, W>(
```
ok-nick, line 1757: Is this the new way to sign? I think the current name can be ambiguous, maybe we should name it `Builder::sign2` if the intention is to replace the old way to do it? Perhaps we should also add a `Builder::set_signer` function as well?

sdk/src/builder.rs
```diff
@@ -1608,21 +1630,102 @@ impl Builder {
...
+        self.maybe_add_thumbnail(&format, source)?;
+
+        // convert the manifest to a store
+        let mut store = self.to_store()?;
+
+        // Get signer from context
+        let signer = self.context.signer()?;
+
+        // sign and write our store to to the output image file
+        store.save_to_stream(&format, source, dest, signer, &self.context)
```
ok-nick, line 1791: Is this the same behavior as `Builder::sign`? Can this all be replaced with a call to that function?

docs/usage.md
```diff
@@ -109,6 +109,250 @@ The default operation of C2PA signing is to embed a C2PA manifest store into an
...
+
+`Context` encapsulates all configuration needed for C2PA operations:
+
+- **Settings**: Configuration options (verification, signing, network policies, etc.)
+- **HTTP Resolvers**: Customizable sync and async HTTP clients for fetching remote manifests
+- **Signers**: Cryptographic signers used to sign manifests (optional - usually created from settings)
+
+### Creating a Context
+
+The simplest way to create a Context is with default settings:
```
tmathern, line 126: Can we point to the "default" settings? I imagine a bunch of users will want to snapshot defaults to keep them stable over multiple releases.

sdk/src/settings/mod.rs
tmathern: (The usage.md is super clear about that, thanks!)

sdk/src/builder.rs
ok-nick: I misunderstood this part, ingredients take in the settings at some point, but not here.

tmathern (approved):
LGTM!

sdk/src/assertion_input.rs
scouten-adobe, line 1: This file isn't referenced anywhere. Can we delete it?

sdk/src/http/restricted.rs
gpeacock: leftover docs from an earlier pass, i'll fix this, nice catch, you need RestrictedResolver.

sdk/src/builder.rs
gpeacock: You set the signer in the context, this function knows how to use the signer in the context. This is the term the internal store has always used and it describes what we are actually doing, sign doesn't always mean we are writing to a stream, but this does. We need another name and I think I prefer this one, even if it is different.

sdk/src/builder.rs
gpeacock: I'm not sure that will work, we could factor things out and call a common function. I actually want to change this so that builder is immutable and we can reuse it without side effects like we get in sign.

sdk/src/builder.rs
gpeacock: Also, how is it ambiguous?

docs/usage.md
gpeacock: yes, that's why I made a markdown file for that, but I think it's moving so our main docs somewhere.

sdk/src/http/restricted.rs
gpeacock: removed this

sdk/src/http/mod.rs
gpeacock: I moved the is_uri_allowed to RestrictedResolver

sdk/src/builder.rs
scouten-adobe: The latest implementation of `Builder::to_store` copies the reference to `self.context`, so this is addressed.

sdk/src/claim.rs
```diff
@@ -1824,17 +1849,16 @@ impl Claim {
     /// Verify claim signature, assertion store and asset hashes
     /// claim - claim to be verified
     /// asset_bytes - reference to bytes of the asset
-    #[allow(clippy::too_many_arguments)]
```
scouten-adobe, line 1827: 🎉

sdk/src/claim.rs
```diff
@@ -3857,6 +3878,19 @@ impl Claim {
         iter.next()
     }
 
+    // /// returns first instance of an assertion whose label and instance match
```
scouten-adobe: Can we delete this rather than adding commented-out code?

sdk/src/builder.rs
ok-nick: I like that idea. I see now that calling sign wouldn't work because it takes a mutable ref and borrows a ref as well. A common function may make more sense for now just so things don't deviate between the two functions.

sdk/src/builder.rs
ok-nick: Ok, I'm not opposed to this. The reason I say ambiguous is because I normally associate the last step in producing a manifest as the signing part, although here we are doing both signing and saving to a stream, so I suppose it can go either way.

sdk/src/builder.rs
gpeacock: I'll revise this in the next PR since I need to do some refactoring anyway.

PR #1636, opened 2025-12-01
docs: Add doc for CLI settings, misc copy edits

crandmck:

Doc for `--settings` CLI option.

https://github.com/contentauth/c2pa-rs/pull/1635 added JSON support, so when the next release goes out, we can land this.

I assume if the environment variable is not set, then the default is still ` ~/.config/c2pa/c2pa.toml` (not `.json`).

Some unrelated edits to the C API file, which is not exported to https://opensource.contentauthenticity.org/. Should it be?

## Checklist

### DIFF
--- a/cli/docs/usage.md
+++ b/cli/docs/usage.md
@@ -41,7 +41,7 @@ The following options are available with any (or no) subcommand.  Additional opt
 | `--external-manifest` | N/A | `<c2pa_file>` | Path to the binary .c2pa manifest to use for validation against the input asset. |
 | `--reserve-size` | N/A | Only valid with `--signer-path` argument. The amount of memory to reserve for signing. Default: 20000. For more information, see CLI help. |
 | `--sidecar` | `-s` | N/A | Put manifest in external "sidecar" file with `.c2pa` extension. See [Generating an external manifest](#generating-an-external-manifest). |
-| `--settings`  | N/A | Path to `<settings_file>` TOML or JSON file.<br/>Default is value of environment variable C2PATOOL_SETTINGS. If the environment variable is not set, then default is` ~/.config/c2pa/c2pa.toml`. |  Path to the config file.  See [Configuring SDK settings](https://opensource.contentauthenticity.org/docs/rust-sdk/docs/settings) | 
+| `--settings`  | N/A | Path to the settings file file.<br/>Default is value of environment variable C2PATOOL_SETTINGS. If the environment variable is not set, then default is` ~/.config/c2pa/c2pa.toml`. |  Path to the config file.  See [Configuring SDK settings](https://opensource.contentauthenticity.org/docs/rust-sdk/docs/settings) | 
 | `--signer-path` | N/A | Specify path to command-line executable for signing.  See [Signing claim bytes with your own signer](#signing-claim-bytes-with-your-own-signer). |
 | `--tree` | | N/A | Create a tree diagram of the manifest store. |
 | `--version` | `-V` | N/A | Display version information. |


docs/usage.md
```diff
@@ -39,7 +41,7 @@ You can enable any of the following features:
 - **rust_native_crypto**: Use Rust native cryptography.
 - **add_thumbnails**: Adds the [`image`](https://github.com/image-rs/image) crate to enable auto-generated thumbnails, if possible and enabled in settings.
 - **fetch_remote_manifests**: Fetches remote manifests over the network when no embedded manifest is present and that option is enabled in settings.
-- **file_io**: Enables APIs that use filesystem I/O.
+- **file_io**: Enables APIs that use files ystem I/O.
```
ok-nick: Typo

docs/c_api.md
```diff
@@ -1,16 +1,21 @@
...
+## New approach
 
-#### Tradeoffs
-The C language is not object oriented and does not natively support things like exception handling. There is no garbage collection. APIs use unsafe pointer references. We must be very careful about pointers an memory management. So the C API is not something we want developers to use directly. But it makes a very solid common way to bridge between Rust and other languages when used correctly. Higher level structures in the other languages can ensure that references to Rust structures are correctly managed and freed. 
+Binding to C APIs is a well-established and mature practice. Every language has well-documented methods for binding to C.  Rust has built in support for writing C bindings. If we provide a solid C interface, we can simply bind other languages to C and leverage all the work that has gone into those bindings.
```
tmathern:
```suggestion
Binding to C APIs is a well-established and mature practice. Every language has well-documented methods for binding to C.  Rust has built in support for writing C bindings. If we provide a solid C interface, we can bind other languages to C and leverage all the work that has gone into those bindings.  
```
I would remove the `simply`, because writing binding is not always easy (for instance, languages where you don't usually deal with pointers suddenly need to take care of very deep native thigns like pointers).

docs/c_api.md
```diff
@@ -1,16 +1,21 @@
...
+Binding to C APIs is a well-established and mature practice. Every language has well-documented methods for binding to C.  Rust has built in support for writing C bindings. If we provide a solid C interface, we can simply bind other languages to C and leverage all the work that has gone into those bindings.  
+
+Bindings must still be written for each language, but since there are so many examples of this, AI engines are very good at writing the code, resulting in well-formed, well-documented bindings, though some manual effort is required to fix some things. Instead of unreadable, incomprehensible auto-generated binding glue, the result is well-structured code bindings that can be customized as needed.
```
tmathern, line 15: There are mixed results with AI. `AI engines are very good at writing the code`. You will still very often (always?) review the code (at least) and make changes manually. Especially when it comes to design APIs that feel more native in each language.

docs/c_api.md
```diff
@@ -1,16 +1,21 @@
...
+
+## Tradeoffs
+
+The C language is not object-oriented, does not perform garbage collection, and does not natively support things like exception handling.  API may use unsafe pointer references, so care must be taken with pointers and memory management. 
+
+For these reasons, you shouldn't use the C API directly. But it makes a very solid common way to bridge between Rust and other languages when used correctly. Use higher-level structures in the other languages to ensure that references to Rust structures are correctly managed and freed.
```
tmathern:
```suggestion
For these reasons, you shouldn't use the C API directly. But it makes a very solid common way to bridge between Rust and other languages when used correctly. Use higher-level structures in the other languages to ensure that references to native structures are correctly managed and freed. 
```
We need to have a concept that Rust is different than C, and here we deal with C-level APIs.

docs/usage.md
tmathern:
```suggestion
- **file_io**: Enables APIs that use filesystem I/O.
```

tmathern (changes requested):
(There is a typo to fix, see suggested change, @ok-nick flagged it too)

docs/c_api.md
crandmck: This content was written by @gpeacock and @cdmurph32 and was added back in May.  Also, it does say 

> ... some manual effort is required to fix some things.

But I added your note about creating "...APIs that feel more native in each language."

PR #1638, opened 2025-12-02
fix: Verify after sign not executing

ok-nick:
Verify after sign is currently never ran regardless of the `verify.verify_after_sign` setting.

There exists a check to ensure we do not verify an empty stream (such as when `io::empty` is supplied), it involves ensuring the stream length is non-zero. However, in the current logic we rewind the stream beforehand resulting in the length always being set to 0. The fix is to seek to the end of the stream to properly read the length.

### DIFF
--- a/sdk/examples/builder_sample.rs
+++ b/sdk/examples/builder_sample.rs
@@ -13,7 +13,7 @@
 
 //! Example App showing how to work archive and restore Builders and ingredients.
 use std::{
-    io::{Cursor, Read, Seek},
+    io::{self, Cursor, Read, Seek},
     sync::Arc,
 };
 
@@ -71,12 +71,15 @@ where
     ))?;
 
     // sign a c2pa only manifest store by using a null input stream and application/c2pa as the format.
-    let mut null_stream = Cursor::new([]);
-    let mut output = Cursor::new(Vec::new());
     let signer = context.signer()?;
-    builder.sign(signer, "application/c2pa", &mut null_stream, &mut output)?;
+    let output = builder.sign(
+        signer,
+        "application/c2pa",
+        &mut io::empty(),
+        &mut io::empty(),
+    )?;
 
-    Ok(output.into_inner())
+    Ok(output)
 }
 
 fn main() -> Result<()> {


c2pa_c_ffi/src/c_api.rs
```diff
@@ -2393,6 +2393,15 @@ mod tests {
         let builder = unsafe { c2pa_builder_from_json(manifest_def.as_ptr()) };
         assert!(!builder.is_null());
 
+        let result = unsafe {
```
tmathern, line 2395: Why is this needed now? Due to settings changes? This is more of a breaking change than the rest, if default OOTB settings already changed.

Or is it a side-effect of changing the verify behavior?
Note that with previous settings, the manifest had the needed assertions, since they did auto-add.

sdk/src/store.rs
```diff
@@ -3092,11 +3092,9 @@ impl Store {
                 let pc_mut = self.provenance_claim_mut().ok_or(Error::ClaimEncoding)?;
                 pc_mut.set_signature_val(s);
 
-                output_stream.rewind()?;
```
tmathern, line 3110: That rewind was executed as far I know. 
(It led to some weird behaviors in the bound SDKs we had to account for).

sdk/src/store.rs
```diff
@@ -3092,11 +3092,9 @@ impl Store {
...
-
                 let verify_after_sign = settings.verify.verify_after_sign;
                 // Also catch the case where we may have written to io::empty() or similar
-                if verify_after_sign && output_stream.stream_position()? > 0 {
```
tmathern, line 3114: So the issue was that after rewind, this was 0 so the if didn't take?
I'd suggest to keep the rewind and change the condition then.

c2pa_c_ffi/src/c_api.rs
ok-nick: Automatic creation of `c2pa.created` and `c2pa.opened` it seems has been disabled by default in settings. I believe due to the new `builder.intent` setting, but I'm not sure if there's supposed to be a default or not for that @gpeacock.

sdk/src/store.rs
ok-nick: Do you have more context on that? We probably shouldn't be relying on the output of a function to rewind the stream otherwise there may be more side-effects, rewinding should be handled during entrance.

sdk/src/store.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1638#discussion_r2582293383

sdk/src/store.rs
tmathern: This rewind and reread forced SDK to open write streams in read and write (w+b in Python). So I have doubts that this wasn't ever executed, or the need for this in other SDKs would not have shown. 

Rust may not be that nitpicky, but changes in stream handling have impacts currently on the using SDKs, because we need to open the streams with the right permissions, and some languages are more closely looking at those.

c2pa_c_ffi/src/c_api.rs
tmathern: We need to know and be sure here. Because then it's a larger breaking change that is going to come in for the other open source SDKs, and the users that relied on it. Not saying we shouldn't do it, but it needs to be properly announced and advertised.

sdk/src/store.rs
ok-nick: In this specific case, the stream rewinding is being done in the low-level parts of the code regardless (during store verification). It could have been that the rewind added here disabled the store verification, which thus disabled parts of the code that affected the behavior you describe. Rust is very explicit when it comes to read/write streams so I would be surprised if rewinding again would change that. That's something we may want to test before this change goes out.

mauricefisher64: It all depends on what the called function expects.  For example if I have a function that appends to a stream, I would not expect the outside code to be rewinding unless it really knew what it was doing.

c2pa_c_ffi/src/c_api.rs
ok-nick: Here's the PR where the change occurred: #1586
Here's the commit: https://github.com/contentauth/c2pa-rs/commit/0d840ecbea66eef10358dae797975e86030cf04c#diff-ac5c57c5d002dbd8ed3a4bd07c7e7bd7a9595fde63e9952cf2f0fe22a7b4a388

sdk/src/store.rs
tmathern: I have to insist on changing the condition instead. Even if the stream is in Rust, it gets handed over potentially through language barriers. I'd prefer a more careful approach here, especially since the release this will go out with will have enough breaking changes.

tmathern (approved):

tmathern:
(I think you will still need to look into the benchmarks though, they seem quite unhappy right now).

tmathern:
> @tmathern That's expected if we are doing validation after signing. Maybe we should consider disabling validation after signing for those benchmarks since we are more concerned about the signing performance there?

I'm just pointing out they seem unhappy and we can't merge like that. How you fix is dealer's choice. Can we have benchmarks with and without validation? So we'll also see the impact of the fix, and can track that in the future if we make more changes and it varies. Also I assume most people would have validate-after-sign on (isn't it going to be default behavior?), so not having benchmarks may be a bit misleading regarding perf. Middleground is to ahve data for both cases.

sdk/examples/builder_sample.rs
```diff
@@ -67,17 +67,15 @@ where
...
-    let mut output = Cursor::new(Vec::new());
-    builder.sign(
+    // sign a c2pa only manifest store by using a null input and output stream, and application/c2pa as the format.
+    let output = builder.sign(
         &Settings::signer()?,
         "application/c2pa",
-        &mut null_stream,
-        &mut output,
+        &mut io::empty(),
+        &mut io::empty(),
```
gpeacock, line 79: This changes the behavior of output, so make sure you check that we don't actually use it anywhere, even though this aligns with the comments.

c2pa_c_ffi/src/c_api.rs
gpeacock: I haven't set a default intent, so we operate the way we used to before any of the automatic stuff. I don't think it would be correct to ever use create as the default intent since we need to supply a digitalSourceType and that should be intentional. A default Edit intent makes more sense since there is always a source asset we can work with when we are signing.

The old style default behavior was edit, but only if there was a manifest in the source image and no parent was specified.

sdk/src/builder.rs
```diff
@@ -3542,6 +3542,8 @@ mod tests {
     #[test]
     fn test_builder_set_base_path() {
         let mut builder = Builder::new();
+        builder.set_intent(BuilderIntent::Create(DigitalSourceType::Empty));
```
gpeacock, line 4050: The ingredient/ingredient.json file lists the ingredient as componentOf, so this is the correct intent here,

gpeacock (approved):
This is a good catch, if we were not validating, then the updated tests were already producing invalid claims

sdk/examples/builder_sample.rs
ok-nick: There's a problem with the current implementation. When you specify an empty input stream and an output stream for the `application/c2pa` format, it will try data hashing it. When validation comes around, it fails with "Hashes do not match."

ok-nick: #1704

PR #1640, opened 2025-12-02
fix: RFC 3161 fractional seconds

cdmurph32:
This is a short term fix. The long term fix is to complete the migration from bcder to der and use the RustCrypto x509-tsp crate.

https://github.com/contentauth/c2pa-rs/issues/1641

## Checklist

### DIFF
--- a/sdk/src/crypto/asn1/mod.rs
+++ b/sdk/src/crypto/asn1/mod.rs
@@ -15,7 +15,7 @@
 
 // Migrating fully from bcder to der will eliminate the need for much of this code.
 
-use std::io::Write;
+use std::{io::Write, str::FromStr};
 
 use bcder::{
     decode::{Constructed, DecodeError, Source},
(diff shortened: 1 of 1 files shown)


sdk/src/crypto/asn1/mod.rs
```diff
@@ -292,7 +433,20 @@ impl From<GeneralizedTime> for chrono::DateTime<chrono::Utc> {
...
+
+        // Add the fractional seconds that were parsed but not stored in der_time
+        // (der::asn1::GeneralizedTime doesn't support fractional seconds per RFC 5280)
+        if gt.nanoseconds > 0 {
+            // SAFETY: nanoseconds is validated to be < 1_000_000_000 during parsing in
+            // from_der_bytes_rfc3161 (line 350-351). If with_nanosecond returns None here,
+            // it indicates a serious internal error that should never happen.
+            dt = dt
+                .with_nanosecond(gt.nanoseconds)
+                .unwrap_or_else(|| unreachable!("nanoseconds value was validated during parsing"));
```
mauricefisher64, line 455: Should this be a fatal error or just a timestamp parsing error?

cdmurph32: I change the logic a few times here, but I went with unreachable here since it's already been validated as I wrote in the comment.

andrewhalle (approved):
I tested on my input and confirmed that I don't get the `timeStamp.malformed` error anymore!

sdk/src/crypto/asn1/mod.rs
```diff
@@ -221,15 +225,143 @@ impl GeneralizedTime {
...
+            // Pad to 9 digits with trailing zeros
+            while frac_str.len() < 9 {
+                frac_str.push('0');
+            }
+            // Truncate if longer than 9 digits
+            frac_str.truncate(9);
+
+            frac_str
+                .parse::<u32>()
+                .map_err(|_| der::Tag::GeneralizedTime.value_error())?
```
andrewhalle: Could this be replaced with parsing as a `f64`, multiplying by `1e9`, and flooring?

cdmurph32: Let me think about that.

andrewhalle: Sounds good, I _think_ it can based on what I tried in [this Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e31f252a9f399934af6de831a0f9e057).

cdmurph32: I think the best approach is to copy https://github.com/RustCrypto/formats/blob/master/x509-tsp/src/generalized_time_nanos.rs more closely and put a TODO to use u32::from_ascii when it stabilizes, if we have not completely migrated from bcder before.

andrewhalle: And using that crate is the longer term fix you mention in the PR description?

PR #1648, opened 2025-12-05
feat: Change version numbering of c2patool to match c2pa-rs crate

scouten-adobe:


scouten-adobe:
Closing for now per discussion with @gpeacock. Will reconsider in a few months.

PR #1650, opened 2025-12-06
chore: Update CI for cargo fmt to use more recent nightly

scouten-adobe:


.github/workflows/tier-1a.yml
```diff
@@ -484,11 +484,11 @@ jobs:
       - name: Install nightly toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2025-07-28
+          toolchain: nightly-2025-12-06
```
tmathern, line 487: Why are we not using latest nightly, while at it? (I mean one without a date, if it exists)

scouten-adobe: @tmathern this is intentional. Latest nightly sometimes surprises us with new behaviors that we have to debug on short notice.

PR #1639, opened 2025-12-02
feat: Use MaybeSend to simplify code

cdmurph32:
## Changes in this pull request
Use MaybeSend to remove duplicate code.

Credit to @gpeacock who discovered this pattern. https://opendal.apache.org/docs/rust/src/opendal/raw/futures_util.rs.html#41

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #1664, opened 2025-12-12
fix: add maybeSync

cdmurph32:
Follow on to maybeSend PR #1639, which broke the JS SDK

## Checklist

### DIFF
--- a/sdk/src/maybe_send_sync.rs
+++ b/sdk/src/maybe_send_sync.rs
@@ -11,7 +11,7 @@
 // specific language governing permissions and limitations under
 // each license.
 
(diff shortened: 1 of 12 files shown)


sdk/src/maybe_send_sync.rs
ok-nick, line 1: Consider renaming to `maybe_send_sync.rs` for clarity?

ok-nick (approved):
Approving this to preserve the old behavior.

A more long-term solution could be to remove the `Sync` bound entirely. It seems the reason we need it is because we refer to the `AsyncSigner` as a trait object through the SDK although we may be able to substitute it with a generic?

Issue #1685, opened 2025-12-22
Signing a Builder does not generate same manifest if it was serialized vs. if it was never serialized


tmathern:
When I create a Builder, add an ingredient and sign, the signed manifest has a certain content. For the test asset I am attaching as zip (because it contains C2PA metadata), note how we have 3 manifests in the store: https://github.com/contentauth/c2pa-rs/blob/57e9938baea0b8f90a33cbab577d2a3eac3734ea/sdk/tests/test_nested_ingredients_repro_manifest1.json

When I create another Builder. add the same ingredient, serialize (to c2pa format) and deserialize it and then sign, I do not get the same manifest as result, as I have only 2 manifest in the store:  https://github.com/contentauth/c2pa-rs/blob/57e9938baea0b8f90a33cbab577d2a3eac3734ea/sdk/tests/test_nested_ingredients_repro_manifest2.json
=> It looks like some nested ingredient data is lost during the serialization-deserialization cycle.

PR reproducing the issue (with a unit test): https://github.com/contentauth/c2pa-rs/pull/1684
(The PR uses the same test asset as I am attaching here). 

[Firefly_An image of a sundae ice cream with chocolate sauce 829529.png.zip](https://github.com/user-attachments/files/24295391/Firefly_An.image.of.a.sundae.ice.cream.with.chocolate.sauce.829529.png.zip)

PR #1686, opened 2025-12-22
fix: Nested ingredient serializing-deserializing (fix for #1685)

tmathern:
This fix attempts to restore deeply nested ingredient data when going through one or more Builder serializations-deserializations to fix https://github.com/contentauth/c2pa-rs/issues/1685. 
The fix merged into the repro branch https://github.com/contentauth/c2pa-rs/pull/1684 makes the test there pass too. In this branch, I added tests that don't need an asset like in https://github.com/contentauth/c2pa-rs/pull/1684.

If it is not the correct fix or the most elegant fix, please suggest an alternative. Thanks!

ps: Cycles detection were a bit trickier here, so refactored things (it's reused in multiple places now).

## Checklist

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -1157,7 +1157,6 @@ impl Reader {
     }
 
     /// Recursively collect all ingredient claims and add them to the primary claim
-    ///
     fn collect_ingredient_claims_recursive(
         &self,
         claim: &Claim,
@@ -1165,13 +1164,7 @@ impl Reader {
         visited: &mut std::collections::HashSet<String>,
         path: &mut Vec<String>,
     ) -> Result<()> {
-        Self::collect_ingredient_claims_impl(
-            &self.store,
-            claim,
-            active_claim,
-            visited,
-            path,
-        )
+        Self::collect_ingredient_claims_impl(&self.store, claim, active_claim, visited, path)
     }
 }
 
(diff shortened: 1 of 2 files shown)


sdk/src/builder.rs
```diff
@@ -867,7 +867,6 @@ impl Builder {
             // we should be able to call Reader::from_stream and then convert to Builder
             // but we need to disable validation since we are not signing yet
             // so we will read the store directly here
-            //crate::Reader::from_stream("application/c2pa", stream).and_then(|r| r.into_builder())
```
tmathern, line 870: Why did we keep those originally? They confused me  a bit when I was debugging, but I can restore the comments.

sdk/src/builder.rs
```diff
@@ -1814,9 +1813,7 @@ impl Builder {
         let mut store = Store::new();
         store.commit_claim(claim)?;
 
-        //store.to_jumbf_internal(1000)
```
tmathern, line 1817: Why did we keep those originally? They confused me  a bit when I was debugging, but I can restore the comments.

sdk/src/builder.rs
```diff
@@ -1814,9 +1813,7 @@ impl Builder {
...
         store.get_data_hashed_manifest_placeholder(100, "application/c2pa")
-        //store.get_box_hashed_embeddable_manifest(signer.as_ref(), settings)
```
tmathern, line 1819: Why did we keep those originally? They confused me  a bit when I was debugging, but I can restore the comments.

PR #1690, opened 2025-12-22
feat: Add feature flags to embed snapshots of the C2PA and CAWG trust lists

ok-nick:
Adds feature flags to embed snapshots of the C2PA and CAWG trust lists:
- `embed_c2pa_interim_trust_list`
- `embed_c2pa_trust_list`
- `embed_cawg_interim_trust_list`

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -49,6 +49,7 @@ rust_native_crypto = [
     "dep:spki",
 ]
 embed_cawg_interim_trust_list = ["dep:cawg-roots"]
+embed_c2pa_interim_trust_list = ["dep:c2pa-roots", "c2pa-roots/interim_roots"]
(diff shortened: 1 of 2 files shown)


hackerfactor:
This is an incredibly BAD IDEA.
Old code hangs around for a long time. Often for years.
If a previously trusted cert is no longer trusted, then it will still remain acceptable in the snapshot code.
It will continue to be accepted until someone updates their copy of c2patool.

In some business environments, every code update must go through an entire testing, regression, and deployment train -- often taking months. While it may be relatively fast to change standalone cert files (bypassing most of the release train), code with embedded certs will need to go through the entire release cycle -- putting the company at risk for months.

Worse: Some c2patool versions have changed the output JSON code. Upgrading from an old version of c2patool (with an obsolete embedded trust list) to a new version may require code changes to whatever custom tools read the c2patool's output. Updating embedded certs may not be a quick update.

PR #1694, opened 2026-01-05
fix: Preflight docs.rs build step wasn't actually checking anything

scouten-adobe:
Also: Move to Tier 1A workflow since it's likely to run quickly.

### DIFF
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -228,22 +228,10 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@master
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/tier-1a.yml
```diff
@@ -211,6 +211,24 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
+
+      - name: Build docs.rs
+        uses: dtolnay/install@cargo-docs-rs
+
+      - run: cargo docs-rs
```
scouten-adobe: This step was missing in the previous workflow. 🙃

PR #1697, opened 2026-01-06
ci: Use `cargo-docs-rs` for preflight docs.rs and fix compilation features

ok-nick:
We removed `cargo-docs-rs` in #1694 because there were issues with compilation features. The source of the problem was that we never set `openssl`/`rust_native_crypto` on `c2pa-c-ffi`. Using `c2pa-c-ffi` without any dependencies would fail by default. In fact, the docs.rs for `c2pa-c-ffi` [have been failing](https://docs.rs/crate/c2pa-c-ffi/latest/builds/2797389) due to this very issue!

Note we do not need to preflight `c2patool` docs.rs because there are no generated docs for binaries.

### DIFF
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -246,6 +246,11 @@ jobs:
           RUSTDOCFLAGS: -Dwarnings
         run: cargo docs-rs -p c2pa-c-ffi
 
+      - name: Preflight c2patool cargo docs
+        env:
+          RUSTDOCFLAGS: -Dwarnings
+        run: cargo doc -p c2patool --no-deps --all-features
+
   doc-tests:
     name: Doc tests (requires nightly Rust)
     needs: get-features


tmathern (approved):

.github/workflows/tier-1a.yml
```diff
@@ -230,8 +230,21 @@ jobs:
...
 
-      - name: Preflight docs.rs
-        run: DOCS_RS=1 cargo +nightly doc --no-deps --all-features
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@v2
+
+      - name: Install cargo-docs-rs
+        uses: dtolnay/install@cargo-docs-rs
+
+      - name: Preflight c2pa-rs docs.rs
```
scouten-adobe, line 239: We also publish docs for c2patool (https://docs.rs/crate/c2patool/latest), so please add a preflight for that as well.

ok-nick: docs.rs doesn't build docs for c2patool, it only uses the README. See my test here:
Screenshot 2026-01-06 at 4 49 32 PM

You can also see in the output on the docs.rs website it doesn't build for c2patool, see https://docs.rs/crate/c2patool/latest/builds/2797391

ok-nick: However, I can add a regular cargo doc check for c2patool if we just want normal doc linting.

PR #1701, opened 2026-01-07
fix: Make parameters optional for transcoded and repackaged actions

timmurphydev:

In the 2.2 version of the specification, we only need to check the parentOf ingredient reference on `c2pa.transcoded` and `c2pa.repackaged` actions if the ingredient reference is present.  This PR wraps the ingredient checks with a conditional so they only occurs if action parameters are present.  Then it looks for `ingredients` or `ingredient` on `params` and performs the parentOf check.  If parameters are present, but neither  `ingredients` or `ingredient` are found, it passes the check silently.

> If the action field is c2pa.transcoded or c2pa.repackaged: If the ingredients field (or ingredient field for c2pa.actions) is present, check that each element of that field is a valid hashed URI that can be resolved to an ingredient assertion in the current manifest with relationship parentOf. If not, then the claim shall be rejected with a failure code of assertion.action.ingredientMismatch.

> In previous versions of this specification, c2pa.transcoded and c2pa.repackaged actions were required to reference the parentOf ingredient assertion referenced by the preceding c2pa.opened action; claim generators can do so for compatibility with older validators.

## Checklist

### DIFF
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -230,8 +230,26 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@nightly
 
-      - name: Preflight docs.rs
-        run: DOCS_RS=1 cargo +nightly doc --no-deps --all-features
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@v2
+
+      - name: Install cargo-docs-rs
+        uses: dtolnay/install@cargo-docs-rs
+
+      - name: Preflight c2pa-rs docs.rs
+        env:
+          RUSTDOCFLAGS: -Dwarnings
+        run: cargo docs-rs -p c2pa
+
+      - name: Preflight c2pa-c-ffi docs.rs
+        env:
+          RUSTDOCFLAGS: -Dwarnings
+        run: cargo docs-rs -p c2pa-c-ffi
+
+      - name: Preflight c2patool cargo docs
+        env:
+          RUSTDOCFLAGS: -Dwarnings
+        run: cargo doc -p c2patool --no-deps --all-features
 
   doc-tests:
     name: Doc tests (requires nightly Rust)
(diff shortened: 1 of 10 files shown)


sdk/src/claim.rs
```diff
@@ -2411,44 +2411,9 @@ impl Claim {
...
-
-                        match parent_tested {
-                            Some(v) => parent_tested = Some(v),
-                            None => parent_tested = Some(false),
-                        }
-                    } else if let Some(h_vec) = &params.ingredients {
-                        for h in h_vec {
+                    if let Some(params) = action.parameters() {
+                        let mut parent_tested = None; // on exists if action actually pointed to an ingredient
+                        if let Some(h) = &params.ingredient {
```
gpeacock, line 2416: Technically since this is a claim v2 requirement we shouldn't support params.ingredient and only allow params.ingredients. A v2 action, only supports params.ingredients. So I think we should remove the params.ingredient validation here.

gpeacock (changes requested):

sdk/src/claim.rs
mauricefisher64: Why can't you have the old action?

timmurphydev: I left the `params.ingredient` check in place because the 2.2 specification says "If the assertion’s label is c2pa.actions or c2pa.actions.v2..." and then "If the ingredients field (or ingredient field for c2pa.actions) is present, check that each element...".

mauricefisher64: That is my thinking as well.  You can have old actions as far as I can tell.

mauricefisher64:
Tim, can you bring the branch up to date with main?

sdk/src/claim.rs
gpeacock: You are probably right, even though that doesn't really make sense. fyi, c2pa.actions (v1)  will be deprecated in the 2.3 spec.

PR #1702, opened 2026-01-07
build: Update `codspeed-criterion-compat` to 4.2.1

ok-nick:
Fixes license failure in CI due to unmaintained sub-dependency, `bincode`.
See: https://rustsec.org/advisories/RUSTSEC-2025-0141


scouten-adobe (commented):
–

PR #1703, opened 2026-01-08
fix: Allow actions assertions to be created assertions

ok-nick:
Allows actions assertions to be created assertions in line with v2.3 of the spec.

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -4026,4 +4026,49 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn actions_created_assertion() {
+        let mut dest = Cursor::new(Vec::new());
+        Builder::new()
+            .with_definition(
+                json!({
+                  "assertions": [
+                    {
+                      "label": "c2pa.actions",
+                      "data": {
+                        "actions": [
+                          {
+                            "action": "c2pa.created",
+                            "digitalSourceType": "http://c2pa.org/digitalsourcetype/empty"
+                          }
+                        ]
+                      },
+                      "created": true
+                    }
+                  ]
+                })
+                .to_string(),
+            )
+            .unwrap()
+            .sign(
+                &Settings::signer().unwrap(),
+                "image/jpeg",
+                &mut Cursor::new(TEST_IMAGE),
+                &mut dest,
+            )
+            .unwrap();
+
+        dest.rewind().unwrap();
+
+        let reader = Reader::from_stream("image/jpeg", &mut dest).unwrap();
+        let active_manifest = reader.active_manifest().unwrap();
+
+        let actions_assertion = active_manifest
+            .assertions()
+            .iter()
+            .find(|assertion| assertion.label().starts_with(Actions::LABEL))
+            .unwrap();
+        assert!(actions_assertion.created());
+    }
 }


sdk/src/builder.rs
```diff
@@ -1259,7 +1259,7 @@ impl Builder {
                     // are resolved to their hashed URIs.
                     self.add_actions_assertion_settings(&ingredient_map, &mut actions)?;
 
-                    claim.add_assertion(&actions)
+                    add_assertion(&mut claim, &actions, manifest_assertion.created())
```
lrosenthol, line 1262: Shouldn't this be based on the flag *per-assertion* that was specified in the JSON?

lrosenthol (changes requested):

sdk/src/builder.rs
ok-nick: Yes we do that for other assertions already [here](https://github.com/contentauth/c2pa-rs/blob/4b63b7f52229633ae82931032b75c755eac2c450/sdk/src/builder.rs#L1290-L1301). For the Actions assertion we do a little bit of preprocessing to resolve things like human-readable ingredient ids into hashed URIs, so we have custom behavior and missed passing the flag through.

ok-nick: @gpeacock if we should update the other "special" assertions to be specifiable as created, see: https://github.com/contentauth/c2pa-rs/blob/4b63b7f52229633ae82931032b75c755eac2c450/sdk/src/builder.rs#L1265-L1301

lrosenthol: For ANY assertion, the JSON allows me to specify that *THAT* assertion to be put in the created array instead of gathered.   I don't see that flag being used here...

ok-nick: That's already being done today, the case you reported was special in that we weren't doing it for Actions assertions specifically. It looks like we don't do it for hard bindings as well, but I'm not sure the reason, hence @gpeacock for clarification.

lrosenthol: hard bindings ALWAYS have to be in created

but actions, like any other assertion, can be in either.

gpeacock: hard bindings are always created.

gpeacock (approved):
Can we add a unit test to make sure this is doing what we think it is, and that it doesn't get broken?

sdk/src/builder.rs
gpeacock: @ok-nick I think the others are ok. We could allow created_work to be created even though it is deprecated.
Calling them out individually here lets us do per assertion format validation. We should do that for some others as well. For the hard bindings we could use add_assertion and force them to be true for created, just for clarity, though this will get caught at a lower level.

ok-nick: Ok, let's discuss those other changes offline and settle it in a different PR.

sdk/src/builder.rs
```diff
@@ -4026,4 +4026,49 @@ mod tests {
...
+            )
+            .unwrap();
+
+        dest.rewind().unwrap();
+
+        let reader = Reader::from_stream("image/jpeg", &mut dest).unwrap();
+        let active_manifest = reader.active_manifest().unwrap();
+
+        let actions_assertion = active_manifest
+            .assertions()
```
lrosenthol, line 4068: This, I believe, gets you ALL assertions, and you should only be looking in created - to make sure that the assertion ends up in the right place.

lrosenthol (changes requested):

sdk/src/builder.rs
ok-nick: At the top-level, the SDK distinguishes gathered/created assertions via a `created` field within each assertion. That is being asserted on the line after the definition of this variable.

lrosenthol: Oh, I see.  

We should fix that in the SDK as it puts a burden on implementers, IMO...

lrosenthol (approved):

PR #1710, opened 2026-01-12
feat: Deeper integration of Context into the SDK

gpeacock:
Context::new() is no longer connected to legacy thread-local settings.

**Builder API Enhancements**
Deprecated from_context and from_shared_context in favor of the with_ syntax.
Added multiple with_* methods to Builder for context-aware construction:
with_shared_context() - attach an Arc<Context> to builder
with_context() / with_shared_context() - create builder with context
Builder::with_archive()  allows passing context 
Builder::into_builder() preserves reader context into the generated builder
No longer need to disable verify_after_reading before calling from_archive

**Reader API Enhancements**
Added context-aware reader construction methods:
from_context() / from_shared_context() - deprecated ( should we just delete them?)
*added*
with_context() / with_shared_context() 
with_manifest_data_and_stream()
with_fragment()
with_file()
with_fragmented_files()
Changed from_archive_with_context() to with_archive()
**Settings Module Refactor**
Refactored Settings API with builder pattern support
Settings can now be constructed and passed to Context instead of being set globally
Added Settings::from_string(), Settings::new(), with_json(), with_toml() methods
Context can be created from Settings: Context::new().with_settings(settings)
**Documentation Updates**
Updated docs/settings.md and docs/usage.md to reflect new Context API
Clarified that old settings are thread-local, not truly global
Added migration guide from Settings to Context pattern
Updated all code examples to show Context-based usage
Consistent use of "thread-local" terminology throughout
**Example Updates**
All examples migrated to use new Context API:
api.rs - now creates shared context from settings
builder_sample.rs - demonstrates context passing through ingredient capture workflow
cawg_identity.rs - updated to use Context
fragmented_bmff.rs - migrated to Context, resolved anyhow::Context naming conflict
**Test Updates**
integration.rs - updated test utilities to use Context
test_builder.rs - migrated tests to Context-based API
Test helpers now accept and use Context explicitly
Internal Changes
Updated timestamp, OCSP, and signature verification code to accept to use bool 
Identity credential holders now use Context for configuration
Store and claim operations refactored to work with Context

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,59 @@
+name: 🐛 Bug report
+description: Report a problem or unexpected behavior
+labels: [bug]
+body:
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component
+      description: Which tool or component are you having problems with?
+      options:
+        - c2pa-rs
+        - c2patool
+        - c2pa-c
+        - Other
+    validations:
+      required: true
+  - type: input
+    id: component-other
+    attributes:
+      label: If Other, please specify
+      description: Which component are you using?
+  - type: input
+    id: version
+    attributes:
+      label: Version
+      description: Which version are you using?
+      placeholder: e.g., v0.71.0
+    validations:
+      required: true
+  - type: dropdown
+    id: platform
+    attributes:
+      label: Platform
+      description: Which platform are you using?
+      options:
+        - Linux
+        - macOS
+        - Windows
+        - iOS
+        - Android
+        - Other
+    validations:
+      required: true
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: Describe what went wrong. Please attach example files if possible.
+      placeholder: Describe the actual behavior you observed.
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behavior
+    attributes:
+      label: What did you expect to happen?
+      description: Describe what you expected to happen instead.
+      placeholder: Describe the expected behavior.
+    validations:
+      required: true
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,12 @@
+name: 💡 Feature request
+description: Suggest an idea for improvement
+labels: [feature]
+body:
+  - type: textarea
+    id: idea
+    attributes:
+      label: Describe your idea
+      description: What would make the project better? What is the target platform for this feature?
+      placeholder: Describe your feature request here.
+    validations:
+      required: true
--- a/.github/ISSUE_TEMPLATE/others.yml
+++ b/.github/ISSUE_TEMPLATE/others.yml
@@ -1,0 +1,12 @@
+name: 💬 Other
+description: General inquiries, questions, or discussions
+labels: [question]
+body:
+  - type: textarea
+    id: details
+    attributes:
+      label: Details
+      description: Tell us what's on your mind.
+      placeholder: Describe your inquiry, question, or feedback here.
+    validations:
+      required: true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2488,9 +2488,9 @@ dependencies = [
 
 [[package]]
 name = "non-empty-string"
-version = "0.2.4"
+version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55cf0f4060e345ae505219853da9ca1150564158a648a6aa6a528f0d5794bb33"
+checksum = "260010741d90def3ca7c4b5ec0bbe28c26d0c775d3c7d291b38642514136677a"
 dependencies = [
  "delegate",
  "serde",
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -140,7 +140,7 @@ lazy_static = "1.4.0"
 memchr = "2.7.6"
 mp4 = "0.14.0"
 nom = "7.1.3"
-non-empty-string = { version = "=0.2.4", features = ["serde"] }
+non-empty-string = { version = "0.2.6", features = ["serde"] }
 nonempty-collections = { version = "0.3.1", features = ["serde"] }
 num-bigint-dig = { version = "0.8.4", optional = true }
 p256 = { version = "0.13.2", optional = true }


sdk/src/builder.rs
```diff
@@ -379,19 +383,48 @@ impl Builder {
...
-    ///     Context::new().with_settings(r#"{"verify": {"verify_after_sign": true}}"#)?,
-    /// );
+    /// let builder = Builder::new()
+    ///     .with_context(Context::new().with_settings(r#"{"verify": {"verify_after_sign": true}}"#)?);
     /// # Ok(())
     /// # }
     /// ```
+    #[deprecated(
+        since = "0.75.0",
+        note = "Use `Builder::new().with_context(context)` instead"
```
scouten-adobe: I'm not at all comfortable with this change. `Builder::new()` re-introduces the dependency on thread-local settings. Yes, I know it subsequently gets overwritten by the `with_context` call, but I'd prefer that we not have that dependency at all.

I'd prefer that we retain the existing `from_context` call to make it clear that this is the preferred way to construct a `Builder`.

scouten-adobe (changes requested):

sdk/src/builder.rs
gpeacock: Well I can go back to using from_context for Builder and Reader. That would have no conflicts at the expense of being a bit inconsistent.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -419,8 +417,7 @@ pub unsafe extern "C" fn c2pa_load_settings(
 ) -> c_int {
     let settings = from_cstr_or_return_int!(settings);
     let format = from_cstr_or_return_int!(format);
-    #[allow(deprecated)]
-    let result = load_settings_from_str(&settings, &format);
+    let result = Settings::from_string(&settings, &format);
```
tmathern, line 422: Help me follow: How does this one behave, what does it do then: is it the per-instance one?
(So I'd know how to update things in the other SDKs which depend on previous behavior).

docs/usage.md
```diff
@@ -111,7 +111,7 @@ The default operation of C2PA signing is to embed a C2PA manifest store into an
 
 ## Using Context for configuration
 
-The C2PA library uses a `Context` structure to configure operations. Context replaces the older global Settings pattern with a more flexible, thread-safe approach.
+The C2PA library uses a `Context` structure to configure operations. Context replaces the older thread-local Settings pattern with a more flexible, thread-safe approach.
```
tmathern, line 114: I would just say older global settings pattern, because that needed some brain parsing power right now (still used to "old" wording).

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -64,7 +63,7 @@ pub fn verify_time_stamp(
     data: &[u8],
     ctp: &CertificateTrustPolicy,
     validation_log: &mut StatusTracker,
-    settings: &Settings,
+    verify_trust: bool,
```
tmathern, line 66: When I see booleans in signatures, I always wonder if we're not better off with an enum. Any chance we would have exceptions or more complicated check to do here (skip trust in some cases etc)? If so, I would go for the enum for signature stability.

sdk/src/identity/x509/x509_credential_holder.rs
```diff
@@ -98,16 +97,23 @@ mod tests {
 
     #[c2pa_test_async]
     async fn simple_case() {
-        let settings = crate::settings::get_settings().unwrap_or_default();
-        let old_decode_identity_assertions = settings.core.decode_identity_assertions;
-
-        set_settings_value("core.decode_identity_assertions", false).unwrap();
+        // Create a context with decode_identity_assertions disabled
+        let settings = crate::settings::Settings::default()
+            .with_value("core.decode_identity_assertions", false)
```
tmathern, line 102: The "with_value" API is really nice for Settings.

sdk/src/crypto/time_stamp/verify.rs
gpeacock: I know this isn't ideal, but it is way better than having the Settings dependency all the way down the api chain. There was talk about separating the validation from the rest of the code, but I'm not doing that for this PR.

c2pa_c_ffi/src/c_api.rs
gpeacock: This one is already published and it is the old way that sets globals. We need to set globals with this api and create something new for the new pattern. I can look into that.

sdk/src/builder.rs
gpeacock: I've updated this to use Builder::from_context and Reader::from_context.

docs/usage.md
gpeacock: I had it saying global settings everywhere but I could sense people telling me "they aren't global - they are thread local!' so I changed everything to say thread-local. I don't really care, but others may?

sdk/src/builder.rs
```diff
@@ -357,8 +366,12 @@ impl Builder {
     /// # Returns
     /// * A new [`Builder`].
     pub fn new() -> Self {
+        // Legacy behavior: explicitly get global settings for backward compatibility
```
scouten-adobe, line 369: OK, now that we're not depending on this for the new `from_context` implementation, can we mark this function as deprecated?

sdk/src/builder.rs
```diff
@@ -487,9 +498,13 @@ impl Builder {
     /// # Errors
     /// * Returns an [`Error`] if the JSON is malformed or incorrect.
     pub fn from_json(json: &str) -> Result<Self> {
+        // Legacy behavior: explicitly get global settings for backward compatibility
+        let settings = crate::settings::get_thread_local_settings();
```
scouten-adobe, line 502: What is our transition plan to remove this legacy behavior?

docs/usage.md
tmathern: I don't know anymore. I have to much context and I can weed through, I can't guess what others would read here. It's probably fine as-is.

sdk/src/reader.rs
```diff
@@ -149,10 +137,8 @@ impl Reader {
     /// ```
     /// # use c2pa::{Context, Reader, Result};
     /// # fn main() -> Result<()> {
-    /// // Simple single-use case - no Arc needed!
-    /// let reader = Reader::from_context(
-    ///     Context::new().with_settings(r#"{"verify": {"verify_after_sign": true}}"#)?,
-    /// );
+    /// let context = Context::new().with_settings(r#"{"verify": {"verify_after_sign": true}}"#)?;
```
scouten-adobe, line 140: :+1:

sdk/src/crypto/time_stamp/verify.rs
scouten-adobe: Agree with @tmathern's comment on avoiding `bool` in situations like this. Also agree, Gavin, that we should make that change at a later time.

sdk/examples/builder_sample.rs
```diff
@@ -87,16 +86,19 @@ fn main() -> Result<()> {
...
     // Here we capture an ingredient with its validation into a c2pa_data object.
-    let ingredient_c2pa = capture_ingredient(FORMAT, &mut ingredient_source)?;
+    let ingredient_c2pa = capture_ingredient(FORMAT, &mut ingredient_source, &context)?;
     // The ingredient_c2pa can be saved to a file, blob storage, a database, or wherever you want to keep it.
     // For this example we will just keep it in memory and add it to a new manifest
 
     // Now create a new builder and set the intent to create a new manifest store
     // We will add the ingredient as a componentOf relationship
-    let mut builder = Builder::from_json(&manifest_def("Builder Sample", FORMAT))?;
+    let mut builder = Builder::from_shared_context(&context)
```
ok-nick, line 100: IMO I think we should encourage the non-shared constructor variant in our examples if we aren't doing any async/threading. That probably warrants a separate async example.

sdk/src/builder.rs
gpeacock: There a a LOT of unit tests that rely on this, and if I mark it deprecated they will all need to either be updated or marked to ignore deprecation. I'd like to take care of that in a later pass. 
But yes, I want to mark all of the global (thread-local) settings apis as deprecated, in another PR.

sdk/src/assertions/timestamp.rs
```diff
@@ -106,26 +103,21 @@ impl TimeStamp {
...
-        settings.verify.verify_timestamp_trust = false;
-
         if _sync {
             crate::crypto::time_stamp::verify_time_stamp(
                 &bytes,
                 message,
                 &ctp,
                 &mut tracker,
-                &settings,
+                false,
```
ok-nick, line 112: This is much better for now until we separate validation (as you mentioned). I did something similar in my timestamp PR.

sdk/src/identity/x509/async_x509_credential_holder.rs
```diff
@@ -126,16 +125,23 @@ mod tests {
...
         let mut source = Cursor::new(TEST_IMAGE);
         let mut dest = Cursor::new(Vec::new());
 
-        let mut builder = Builder::from_json(&manifest_json()).unwrap();
+        // Use the context when creating the Builder
+        let mut builder = Builder::from_shared_context(&context)
+            .with_definition(manifest_json())
+            .unwrap();
         builder
             .add_ingredient_from_stream(parent_json(), format, &mut source)
```
ok-nick, line 146: Async version here?

sdk/src/identity/x509/x509_credential_holder.rs
```diff
@@ -98,16 +97,23 @@ mod tests {
...
         let mut source = Cursor::new(TEST_IMAGE);
         let mut dest = Cursor::new(Vec::new());
 
-        let mut builder = Builder::from_json(&manifest_json()).unwrap();
+        // Use the context when creating the Builder
+        let definition = serde_json::from_str(&manifest_json()).unwrap();
+        let mut builder = Builder::from_shared_context(&context);
+        builder.definition = definition;
         builder
             .add_ingredient_from_stream(parent_json(), format, &mut source)
```
ok-nick, line 118: Async here too?

sdk/src/settings/mod.rs
```diff
@@ -577,13 +580,14 @@ impl Settings {
...
-    /// Get a [Settings] value by path reference. The path is nested names of of the [Settings] objects
+    /// Get a [Settings] value by path reference from the thread-local settings.
+    /// The path is nested names of of the [Settings] objects
     /// separated by "." notation.
     ///
     /// For example "core.hash_alg" would get the settings.core.hash_alg value. The nesting can be arbitrarily
     /// deep based on the [Settings] definition.
     #[allow(unused)]
-    fn get_value<'de, T: serde::de::Deserialize<'de>>(value_path: &str) -> Result<T> {
+    fn get_thread_local_value<'de, T: serde::de::Deserialize<'de>>(value_path: &str) -> Result<T> {
```
ok-nick, line 595: I like the new name!

sdk/src/builder.rs
gpeacock: The plan is to mark these methods as deprecated. But I didn't want to add all that to this PR

sdk/src/settings/mod.rs
```diff
@@ -603,21 +607,154 @@ impl Settings {
...
     }
 
-    /// Serializes the [Settings] into a pretty (formatted) toml string.
+    /// Serializes the thread-local [Settings] into a pretty (formatted) toml string.
     pub fn to_pretty_toml() -> Result<String> {
-        let settings =
-            get_settings().ok_or(Error::OtherError("could not get current settings".into()))?;
+        let settings = get_thread_local_settings();
         Ok(toml::to_string_pretty(&settings)?)
     }
```
ok-nick, line 765: Hm, maybe we should deprecate these. It feels wrong to have it from TOML and not JSON, but they can be serialized via serde manually anyways. I suppose it's also thread-local so that may be grounds for deprecation?

sdk/src/settings/mod.rs
```diff
@@ -603,21 +607,154 @@ impl Settings {
...
+    /// Load settings from string representation (builder pattern helper).
+    ///
+    /// This overlays the parsed configuration on top of the current Settings
+    /// instance without touching thread-local state.
+    fn with_string(self, settings_str: &str, format: &str) -> Result<Self> {
+        let f = match format.to_lowercase().as_str() {
+            "json" => FileFormat::Json,
+            "toml" => FileFormat::Toml,
+            _ => return Err(Error::UnsupportedType),
+        };
```
ok-nick, line 732: Maybe pass the FileFormat enum here instead of a string?

sdk/src/settings/mod.rs
ok-nick, line 437: Are we deprecating everything thread-local? If not yet, then we should also clarify some of these functions sets the thread-local settings in the docs.

sdk/src/reader.rs
```diff
@@ -259,53 +251,122 @@ impl Reader {
...
+                } else {
+                    self.with_store_async(store, &mut validation_log).await
+                }?;
+                Ok(self)
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    #[cfg(feature = "file_io")]
```
ok-nick: Did you mean to put this under the doc comments?

sdk/src/reader.rs
```diff
@@ -259,53 +251,122 @@ impl Reader {
...
+    ///
+    /// # Note
+    /// [CAWG identity] assertions require async calls for validation.
+    ///
+    /// [CAWG identity]: https://cawg.io/identity/
+    #[async_generic]
+    pub fn from_file<P: AsRef<std::path::Path>>(path: P) -> Result<Reader> {
+        // Legacy behavior: explicitly get thread-local settings for backward compatibility
+        let settings = crate::settings::get_thread_local_settings();
+        let context = Context::new().with_settings(settings).unwrap_or_default();
```
ok-nick: I think we should propagate the error here, if there is one it'd be confusing to go to a default settings.

sdk/examples/builder_sample.rs
gpeacock: Sharing is not only for async/threading, it is also useful in our examples where we create more than one builder or we create a builder and then use a reader to verify. The arc overhead is really not much of an issue. But I agree that the examples that don't need a shared context should use the non-shared one.

sdk/src/settings/mod.rs
gpeacock: yes they should be deprecated. I just decided to move deprecation to another PR since this one was big enough already.

sdk/src/builder.rs
```diff
@@ -3806,6 +3901,29 @@ mod tests {
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
 
+    #[test]
+    fn test_with_arvhive() -> Result<()> {
```
scouten-adobe:
```suggestion
    fn test_with_archive() -> Result<()> {
```

scouten-adobe (approved):
I'd like to see Nick's comment incorporated, but otherwise, looking great!

sdk/src/builder.rs
tmathern: +1 for another PR!

sdk/src/identity/x509/async_x509_credential_holder.rs
gpeacock: This is either an assignment or a serde_json deserialize and serde_json doesn't have any async operations.
it is also very fast for any of the json we use. We don't have anything async here and I don't think we need it.

sdk/src/identity/x509/x509_credential_holder.rs
gpeacock: see comment above, though I did change this to use with_definition()

sdk/src/settings/mod.rs
gpeacock: Good, tania was confused by it, but I said others would like it!

sdk/src/reader.rs
gpeacock: thanks, fixed that

sdk/src/settings/mod.rs
gpeacock: This is a private internal method only used to implement the published ones, and it can only accept json or toml. I could define that as an enum for internal use but this works, and I'd need to to more for the from_file logic.

sdk/src/reader.rs
gpeacock: moved it

sdk/src/reader.rs
gpeacock: good catch

sdk/examples/builder_sample.rs
gpeacock: In this case I need to pass the context to capture_ingredient as well as Builder.

sdk/src/settings/mod.rs
scouten-adobe: In principle, I agree, but … as Tania and I have both said, this PR is getting big. I'd like to defer that to another PR.

sdk/src/identity/x509/async_x509_credential_holder.rs
ok-nick: OK that's fine then. The reason I bring it up is because we have a few tests that call the sync version of these methods, even when performing an async operation (the reason we don't use reqwest-blocking by default).

ok-nick (approved):

sdk/src/context.rs
```diff
@@ -191,7 +191,7 @@ pub struct Context {
 impl Default for Context {
     fn default() -> Self {
         Self {
-            settings: crate::settings::get_settings().unwrap_or_default(),
```
grinstead, line 194: Hey, not sure if this is the right place to say this, however, this change did break my code, and I'm not sure that it is the right thing to do. I'll be fixing my code call, but it was a very silent failure.

PR #1728, opened 2026-01-15
fix: Fix streams handling for TIFF

tmathern:
Some (all? I always hit the issue when working on the repro) tiff images, after signing, can have a manifest with hash mismatch error.
It seems there is an issue with coordination in streams handling and rewinds, as when logging all sizes at some point the stream doubles size. It seems this is due to the offset handling "fromEnd" that is off.
An e2e test for tiff signing has been added at the c2pa_c_ffi layer to verify the fix.

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -1972,68 +1972,6 @@ mod tests {
     }
 
     #[test]
-    fn builder_create_intent_empty_and_sign_tiff() {
-        let source_image = include_bytes!(fixture_path!("test.tiff"));
-        let mut source_stream = TestC2paStream::from_bytes(source_image.to_vec());
-        let dest_vec = Vec::new();
-        let mut dest_stream = TestC2paStream::new(dest_vec).into_c_stream();
-
-        let (signer, builder) = setup_signer_and_builder_for_signing_tests();
-
-        // Create intent will create a minimal size manifest
-        let result = unsafe {
-            c2pa_builder_set_intent(
-                builder,
-                C2paBuilderIntent::Create,
-                C2paDigitalSourceType::Empty,
-            )
-        };
-        assert_eq!(result, 0);
-
-        let format = CString::new("image/tiff").unwrap();
-        let mut manifest_bytes_ptr = std::ptr::null();
-        let _ = unsafe {
-            c2pa_builder_sign(
-                builder,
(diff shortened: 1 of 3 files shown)


sdk/src/store.rs
```diff
@@ -3093,6 +3093,7 @@ impl Store {
         let sig_placeholder = Store::sign_claim_placeholder(pc, signer.reserve_size());
 
         intermediate_stream.rewind()?;
+        output_stream.rewind()?;
```
tmathern, line 3096: It seems that one may be needed in some cases nevertheless, eg. tiff.

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -717,9 +717,6 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
                     }
                 });
 
-                // seek to end of file
```
tmathern, line 720: The offset is off, according to the debug logs while debugging, This fixes it.

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -821,9 +818,6 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
...
```
tmathern, line 824: The offset is off, according to the debug logs while debugging (when I track all stream sizes). This leads to the hash mismatch observed in the bug, since we mess up streams when signing. This fixes it.

sdk/src/asset_handlers/tiff_io.rs
gpeacock: this may not be needed if the output is rewound before we start.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1971,6 +1971,68 @@ mod tests {
         }
     }
 
+    #[test]
```
gpeacock: It would be better to have this builder.rs. It's going to be hard to maintain on the C bindings.

gpeacock (approved):
Approved, but see comments. I suspect all we need to do is ensure output is rewound.

c2pa_c_ffi/src/c_api.rs
tmathern: Ack, will move. I wanted to make sure the e2e case is coered first/while debugging it.

PR #1736, opened 2026-01-15
fix: Update tiff support to latest standard

mauricefisher64:
Updates TIFF reading to support manifest locations in first or last IFD
Updates to write manifest into unique C2PA specific IFD
Write manifest at end of file to support update manifests
Multipage TIFF support
Fix incorrect hash ranges when the manifest was external

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/tiff_io.rs
+++ b/sdk/src/asset_handlers/tiff_io.rs
@@ -400,11 +400,12 @@ where
     let tiff_tree: Arena<ImageFileDirectory> = if let Some(ifd) = ts.first_ifd.clone() {
         let (mut tiff_tree, page_0) = Arena::with_data(ifd);
         let mut current_token = page_0;
+
         // get the pages
         loop {
             tokens.push(current_token);
 
-            // look for known special IFDs on page 0
+            // look for known special IFDs
             let page_subifd = tiff_tree[current_token].data.get_tag(SUBFILE_TAG).copied();
 
             // grab SubIFDs for page (DNG)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_PLZ_TOKEN }}
           CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_SECRET }}
+          RELEASE_PLZ_LOG: DEBUG
```
tmathern: ?

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -44,14 +44,20 @@ const XMP_TAG: u16 = 0x02bc;
...
+const INTEROPERABILITY_TAG: u16 = 40965;
 const C2PA_FIELD_TYPE: u16 = 7;
 
 const STRIPBYTECOUNTS: u16 = 279;
 const STRIPOFFSETS: u16 = 273;
 const TILEBYTECOUNTS: u16 = 325;
 const TILEOFFSETS: u16 = 324;
 
-const SUBFILES: [u16; 3] = [SUBFILE_TAG, EXIFIFD_TAG, GPSIFD_TAG];
+/* support when we find a use case
```
tmathern, line 55: This feels like a challenge :) Good to know, if we find such files!

.github/workflows/release.yml
mauricefisher64: I don't know what that is.  It came when I pulled in main.  I don't want it an I am happy to remove it.

sdk/src/asset_handlers/tiff_io.rs
mauricefisher64: They are in the unit test

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -659,7 +686,40 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
                 self.writer.write_u32(cnt)?;
             }
 
-            self.writer.write_all(&entry.value_bytes)?;
+            if *tag == C2PA_TAG && is_c2pa_idf {
+                let data_bytes = if self.big_tiff { 8u64 } else { 4u64 };
```
gpeacock, line 691: big_tiff makes everything so much more fun!

gpeacock (approved):
It would be good if we could find a smaller example of subfiles.dng. But I know that may not be easy.

PR #1675, opened 2025-12-16
chore: Add c2pa_cbor as an alternative for serde_cbor and ciborium

gpeacock:
The existing crates for serde_cbor and ciborium are not maintained and thus present a security risk.
The c2pa_cbor crate is a drop in replacement for serde_cbor, but a different implemenation.
It already includes a fix for one security risk in serde_cbor
The performance and functionality of this crate essentially the same as serde_cbor. (it has been tested for that)
We still have a ciborium dependency due to coset, but not in the rest of the code.
The predicted size savings did not pan out at least in tests so far. But it is still worth it for support issues.

PR #1739, opened 2026-01-15
chore: Add c2pa_cbor as an alternative for serde_cbor and ciborium

scouten-adobe:
Re-submitting what was originally submitted in #1675, which caused a release failure.

Original description:

> The existing crates for serde_cbor and ciborium are not maintained and thus present a security risk.
> The c2pa_cbor crate is a drop in replacement for serde_cbor, but a different implemenation.
> It already includes a fix for one security risk in serde_cbor
> The performance and functionality of this crate essentially the same as serde_cbor. (it has been tested for that)
> We still have a ciborium dependency due to coset, but not in the rest of the code.
> The predicted size savings did not pan out at least in tests so far. But it is still worth it for support issues.


gpeacock:
This is now updated to use the published c2pa_cbor crate (0.77.0)

deny.toml
```diff
@@ -15,7 +15,6 @@ targets = [
 [advisories]
 yanked = "deny"
 ignore = [
-  "RUSTSEC-2021-0127", # serde_cbor
```
scouten-adobe, line 18: 🎉

PR #1743, opened 2026-01-15
fix: No-embed: context propagation

tmathern:
- Verify after sign fails when the manifest is not embedded currently. Should it even try to run if the manifest is not embedded? For now I skip (because otherwise we may need to retrieve the remote manifest first).
- Reader context also showed some oversights due to earlier exits when manifest is embedded, so updated that in order to have the manfiests properly validate when not embedded.

Fixes those failures: https://github.com/contentauth/c2pa-python/actions/runs/21048169702/job/60527882202?pr=192

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,5 @@
 name: 🐛 Bug report
 description: Report a problem or unexpected behavior
-labels: [bug]
 body:
   - type: dropdown
     id: component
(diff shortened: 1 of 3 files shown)


ok-nick (approved):

sdk/src/reader.rs
```diff
@@ -444,11 +444,13 @@ impl Reader {
         format: &str,
         stream: impl Read + Seek + MaybeSend,
     ) -> Result<Reader> {
+        // Get thread-local settings if any for backward compatibility
```
gpeacock: this was a good catch, I just missed updating this one.

tmathern: The tests caught it ;) (In the Python SDK I have tests that are isolated per thread, verifying that).

PR #1770, opened 2026-01-22
docs: Remove outdated ref to CreativeWork, other small edits

crandmck:

Update C2PA Tool docs.

Move command reference into separate page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ As of December 2025 and until the 1.0.0 version is released, the CAI team will o
 
 ## [Unreleased]
 
+## [0.75.6](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.75.5...c2pa-v0.75.6)
+_22 January 2026_
+
+### Added
+
+* Trigger a new release to allow generated binaries to be published
+
 ## [0.75.5](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.75.4...c2pa-v0.75.5)
 _21 January 2026_
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -426,7 +426,7 @@ dependencies = [
 
 [[package]]
 name = "c2pa"
-version = "0.75.5"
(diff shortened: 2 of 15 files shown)


cli/docs/usage.md
```diff
@@ -20,31 +20,6 @@ The tool supports the following subcommands:
 - `fragment` [adds a manifest to fragmented BMFF content](#adding-a-manifest-to-fragmented-bmff-content).  With this subcommand, one additional option is available.
 - `help` displays command line help information.
 
-## Options
```
gpeacock, line 23: It seems odd to have no list of subcommands. Perhaps we can just show the --info output here?

cli/docs/usage.md
```diff
@@ -234,7 +209,9 @@ c2patool sample/C.jpg trust \
 
 ### Using the Verify known certificate list
 
-**IMPORTANT:** The C2PA intends to publish an official trust list. Until that time, the [C2PA Verify tool uses a temporary known certificate list](https://opensource.contentauthenticity.org/docs/verify-known-cert-list). These lists are subject to change, and will be deprecated when C2PA publishes its trust list.
+**IMPORTANT:** C2PA has an official trust list. 
+
+Currently, the C2PA Verify tool uses the interim trust list which is frozen; no new certificates can be added to this list.
```
tmathern: C2PA Verify  -> can we link it here? (It is a bit confusing, this is the Verify site I think?)

cli/docs/usage.md
crandmck: Yeah, I decided to put the options table back here.  

I thought it would be cleaner to have it separately and make this page a bit shorter, but I just think it makes it harder to find, so I've deleted the `command-ref.md` file and put all the content back here as it was.

cli/docs/usage.md
crandmck: Yes, I wasn't finished editing this section. I've updated it to have that link and a bunch more general info.

PR #1774, opened 2026-01-23
feat: make necessary changes to support arbitrary fields in assertion metadata, such as on ingredients

lrosenthol:
Adding support for arbitrary key/value pairs in assertionmetadata

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -14,17 +14,25 @@ For a simple example of calling c2patool from a Node.js server application, see 
 
 **Additional documentation**:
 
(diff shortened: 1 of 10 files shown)


ASSERTION_METADATA_ARBITRARY_FIELDS.md
```diff
@@ -0,0 +1,234 @@
+# AssertionMetadata Arbitrary Key/Value Pairs Implementation
```
tmathern, line 1: We probably don't want to keep that here, at least in a docs folder, if at all?

lrosenthol: Agreed - the docs can probably go (or at least get moved)

PR #1782, opened 2026-01-27
feat: Add support for arbitrary key/value pairs in assertion metadata

lrosenthol:

Added support, both internally and via APIs, for getting and setting arbitrary key/value pairs in the assertion metadata assertion (or the `metadata` key in supported assertions).

Added a number of tests to validate the functionality.

## Checklist

### DIFF
--- a/sdk/tests/test_builder.rs
+++ b/sdk/tests/test_builder.rs
@@ -734,8 +734,9 @@ fn test_ingredient_arbitrary_metadata_fields() -> Result<()> {
     let ingredient = &ingredients[0];
 
     // Verify standard metadata field
+    assert_eq!(ingredient["metadata"]["dateTime"]["tag"].as_i64(), Some(0));
     assert_eq!(
-        ingredient["metadata"]["dateTime"].as_str(),
+        ingredient["metadata"]["dateTime"]["value"].as_str(),
         Some("2024-01-23T10:00:00Z")
     );
 


ASSERTION_METADATA_ARBITRARY_FIELDS.md
```diff
@@ -0,0 +1,234 @@
+# AssertionMetadata Arbitrary Key/Value Pairs Implementation
```
tmathern: @lrosenthol Your call to include or not. I would not (aka remove them).

sdk/src/assertions/assertion_metadata.rs
```diff
@@ -44,6 +45,10 @@ pub struct AssertionMetadata {
     localizations: Option<Vec<HashMap<String, HashMap<String, String>>>>, // not implemented
     #[serde(rename = "regionOfInterest", skip_serializing_if = "Option::is_none")]
     region_of_interest: Option<RegionOfInterest>,
+    /// Arbitrary key/value pairs as permitted by the C2PA spec.
+    /// Uses flatten to allow these fields to be serialized at the same level as known fields.
+    #[serde(flatten, skip_serializing_if = "HashMap::is_empty")]
+    additional_fields: HashMap<String, Value>,
```
gpeacock, line 51: This will add an additional_fields field to the assertion and that is not allowed in the spec. We have a different mechanism to handle extensible objects, but the spec doesn't allow us to extend this.

gpeacock: The right way to add extra fields is with serde_flatten , like we do for standard metadata assertions
```

    /// Metadata fields with namespace prefixes.
    #[serde(flatten)]
    pub value: HashMap<String, Value>,
    
```
 
 You can then add get and insert methods like this for rust access.
 
     /// get values by key as an instance of type `T`.
    /// This return T is owned, not a reference
    /// # Errors
    ///
    /// This conversion can fail if the structure of the field at key does not match the
    /// structure expected by `T`
    pub fn get<T: DeserializeOwned>(&self, key: &str) -> Option<T> {
        self.value
            .get(key)
            .and_then(|v| serde_json::from_value(v.clone()).ok())
    }

    /// insert key / value pair of instance of type `T`
    /// # Errors
    ///
    /// This conversion can fail if `T`'s implementation of `Serialize` decides to
    /// fail, or if `T` contains a map with non-string keys.
    pub fn insert<T: Serialize>(mut self, key: String, value: T) -> Result<Self> {
        self.value.insert(key, serde_json::to_value(value)?);
        Ok(self)
    }

lrosenthol: Yes, that would work on a key-by-key, but as you can see in my PR, I also added support for batch/multiple key insertion & retrieval, which optimizes my use case for them.  

Also, you get method assumes that I know what they are - but in my use case, I don't.  I need to look at all keys to find ones that I don't recognize.

ASSERTION_METADATA_ARBITRARY_FIELDS.md
scouten-adobe, line 1: This should not be a top-level documentation file. Please move this to either the `docs` folder or as RustDoc comments on the `sdk/src/assertions/assertion_metadata.rs` file.

scouten-adobe: Also it reads as an AI sales pitch for this feature and that's a bit over the top. A simple how-to guide should be sufficient.

sdk/src/assertions/assertion_metadata.rs
scouten-adobe: Looks like `serde_flatten` has been added. :+1:

scouten-adobe (changes requested):

ASSERTION_METADATA_ARBITRARY_FIELDS.md
lrosenthol: Already removed it!

PR #1783, opened 2026-01-27
feat: Adds thread safe Settings and Context support to c_ffi_api

gpeacock:
New C2paSettings and C2paContext along with related Reader and Builder updates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,9 +417,9 @@ dependencies = [
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 dependencies = [
  "serde",
 ]
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -103,7 +103,7 @@ async-trait = "0.1.78"
 atree = "0.5.2"
 base64 = "0.22.1"
 bcder = "0.7.3"
-bytes = "1.7.2"
+bytes = "1.11.1"
 byteorder = { version = "1.4.3", default-features = false }
 byteordered = "0.6.0"
 chrono = { version = "0.4.42", default-features = false, features = ["serde"] }


c2pa_c_ffi/src/c_api.rs
```diff
@@ -383,7 +362,7 @@ pub unsafe extern "C" fn c2pa_version() -> *mut c_char {
 /// and it is no longer valid after that call.
 #[no_mangle]
 pub unsafe extern "C" fn c2pa_error() -> *mut c_char {
-    to_c_string(Error::last_message().unwrap_or_default())
+    to_c_string(Error::last_message())
```
tmathern, line 254: No default anymore?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -521,39 +652,49 @@ pub unsafe extern "C" fn c2pa_sign_file(
...
+/// - and any other objects created by this library
 ///
 /// # Safety
-/// Reads from NULL-terminated C strings.
-/// The string must not have been modified in C.
-/// The string can only be freed once and is invalid after this call.
+///
+/// * The pointer must have been allocated by this library (e.g., from c2pa_context_new(),
+///   c2pa_settings_new(), c2pa_builder_from_json(), etc.)
+/// * The pointer must not have been modified in C.
```
tmathern, line 645: What does this mean, "not modified in C", for this case?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -612,21 +751,38 @@ fn post_validate(result: Result<C2paReader, c2pa::Error>) -> Result<C2paReader,
...
+/// This function is safe to call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_reader_new() -> *mut C2paReader {
+    box_tracked!(C2paReader::from_context(Context::default()))
+}
+
+// TODO: Reader::from_context takes ownership of Context, but Context doesn't implement Clone
+// and we can't take ownership from a *const pointer. Need to decide on the correct approach:
+// 1. Change signature to *mut and actually take ownership from C
+// 2. Use from_shared_context instead
```
tmathern: Isn't that what would make the most sense? Since I assume context may be passed around potentially in  this layer?

c2pa_c_ffi/src/error.rs
```diff
@@ -10,19 +10,11 @@
...
-
-// LAST_ERROR handling borrowed from Copyright (c) 2018 Michael Bryan
-thread_local! {
-    static LAST_ERROR: RefCell<Option<Error>> = const { RefCell::new(None) };
-}
 
 #[derive(Error, Debug)]
 /// Defines all possible errors that can occur in this library
-pub enum Error {
+pub enum C2paError {
```
tmathern, line 17: Type name change for namespacing?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -416,16 +395,178 @@ pub unsafe extern "C" fn c2pa_load_settings(
...
+/// * `value` must be a valid pointer to a null-terminated UTF-8 string containing
+///   a JSON value (e.g., "true", "\"ps256\"", "42").
+/// * The pointers must remain valid for the duration of this call.
+/// * This function is not thread-safe - do not call concurrently on the same `settings`.
+///
+/// # Returns
+///
+/// 0 on success, negative value on error.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_settings_set_value(
```
tmathern, line 355: That is nice, but agreed a settings builder pattern would be nicer! (Albeit the params & ref handling can be tricky, as I am discovering too).

c2pa_c_ffi/src/c_api.rs
```diff
@@ -416,16 +281,243 @@ pub unsafe extern "C" fn c2pa_load_settings(
...
+    0 // returns 0 on success
 }
 
-/// Returns a ManifestStore JSON string from a file path.
+/// Creates a new C2PA settings object with default values.
 ///
-/// Any thumbnails or other binary resources will be written to data_dir if provided.
+/// # Safety
+///
+/// This function is safe to call. The returned pointer must be freed
```
tmathern:
```suggestion
/// The returned pointer must be freed
```

c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -0,0 +1,714 @@
...
+#[macro_export]
+macro_rules! some_or_return_other_false {
+    ($option:expr, $msg:expr) => {
+        $crate::some_or_return!($option, $msg, false)
+    };
+}
+
+/// If the expression is null, set the last error and return null.
+#[macro_export]
+macro_rules! ptr_or_return_null {
```
tmathern, line 594: Is this the same as `cstr_or_return_null`?

c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -0,0 +1,714 @@
...
+#[macro_export]
+macro_rules! cstr_or_return_null {
+    ($ptr : expr) => {
+        $crate::cstr_or_return!($ptr, std::ptr::null_mut())
+    };
+}
+
+// Internal routine to convert a *const c_char to a rust String or return a -1 int error.
+#[macro_export]
+macro_rules! cstr_or_return_int {
```
tmathern, line 618: Same as `cstr_or_return_neg`?

c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -0,0 +1,714 @@
...
+#[allow(unused_macros)]
+macro_rules! some_or_return_other_null {
+    ($option:expr, $msg:expr) => {
+        $crate::some_or_return!($option, $msg, std::ptr::null_mut())
+    };
+}
+
+/// Convenience macro: Handle Option with Error::other message, return -1
+#[macro_export]
+macro_rules! some_or_return_other_int {
```
tmathern: Same as `some_or_return_other_false`?

c2pa_c_ffi/src/c_api.rs
gpeacock: C should not be directly modifying opaque pointers to rust memory. The only modifications should happen in Rust.

c2pa_c_ffi/src/error.rs
gpeacock: Pretty much, neither of these are directly exported, but this helps keep us from clashing with other Error instances.

c2pa_c_ffi/src/cimpl/macros.rs
gpeacock: no, cstr validates it is a correctly null terminated string and converts it to CString. This just validates that this is a tracked pointer and not null.

c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -0,0 +1,714 @@
...
+#[macro_export]
+macro_rules! some_or_return_other {
+    ($option:expr, $msg:expr, $err_val:expr) => {
+        $crate::some_or_return!($option, $crate::CimplError::other($msg), $err_val)
+    };
+}
+
+// Unused - kept for potential future use
+#[allow(unused_macros)]
+macro_rules! some_or_return_other_null {
```
scouten-adobe: Naming request: `some_or_return_other` makes sense when you're providing the "other" value.

For macros such as this where the macro name specifies the `None` case, can I suggest naming as `some_or_return_null` (omitting the `other` part)?

scouten-adobe: Applies to several macros in this file.

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+    fn new() -> Self {
+        Self {
+            tracked: Mutex::new(HashMap::new()),
+        }
+    }
+
+    /// Track a pointer with its type and cleanup function
+    fn track(&self, ptr: usize, type_id: TypeId, cleanup: CleanupFn) {
+        if ptr != 0 {
+            self.tracked.lock().unwrap().insert(ptr, (type_id, cleanup));
```
scouten-adobe: Please add a comment here explaining why the `unwrap()` should be considered safe.

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+
+        Ok(())
+    }
+}
+
+impl Drop for PointerRegistry {
+    fn drop(&mut self) {
+        let tracked = self.tracked.lock().unwrap();
+        if !tracked.is_empty() {
+            eprintln!(
```
scouten-adobe, line 117: Very nice!

c2pa_c_ffi/src/cimpl/macros.rs
tmathern: Oh I see now, the cstr one calls `cstr_or_return` vs that one calls `ptr_or_return`! 
(Just 2 chars of difference, but 2 important ones!)

c2pa_c_ffi/src/cimpl/macros.rs
tmathern: Not same, one returns -1, the other false.

c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -0,0 +1,714 @@
...
+        let ptr = $ptr;
+        if ptr.is_null() {
+            $crate::CimplError::null_parameter(stringify!($ptr)).set_last();
+            return $err_val;
+        } else {
+            // SAFETY: We create a bounded slice up to MAX_CSTRING_LEN.
+            // Caller must ensure ptr is valid for reading and points to a
+            // null-terminated string within MAX_CSTRING_LEN bytes.
+            let bytes = unsafe {
+                std::slice::from_raw_parts(ptr as *const u8, $crate::macros::MAX_CSTRING_LEN)
```
tmathern, line 370: Isn't that always exactly the size `MAX_CSTRING_LEN`, even if our string is less, like if the expected size let's say is MAX_CSTRING_LEN/2? Then aren't we over-allocating? Is it fine or undefined behavior then?

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+    pub unsafe fn apply_to_box<T, F>(&self, ptr: *mut T, f: F) -> Result<(), Error>
+    where
+        T: 'static + Default,
+        F: FnOnce(T) -> T,
+    {
+        // Validate pointer is tracked with correct type
+        self.validate(ptr as usize, TypeId::of::<T>())?;
+
+        // Temporarily swap out the value (no clone needed!)
+        let value = std::ptr::replace(ptr, T::default());
```
tmathern: Do we want thread safety? We may need a guard/lock here so another thread couldn't be in this code with a pointer in parallel

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+        F: FnOnce(T) -> T,
+    {
+        // Validate pointer is tracked with correct type
+        self.validate(ptr as usize, TypeId::of::<Mutex<T>>())?;
+
+        // Get reference to mutex (pointer is valid per contract)
+        let mutex = &*ptr;
+
+        // Lock and swap - all atomic under the lock
+        let mut guard = mutex.lock().unwrap();
```
tmathern: Poisoned mutex would panic, should we return an error instead?

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+/// The pointer will be freed with `Box::from_raw()` when `cimpl_free()` is called.
+///
+/// # Returns
+/// Returns the same pointer for convenient chaining
+///
+/// # Example
+/// ```ignore
+/// let ptr = track_box(Box::into_raw(Box::new(value)));
+/// ```
+pub fn track_box<T: 'static>(ptr: *mut T) -> *mut T {
```
tmathern:
```suggestion
pub fn track_box<T: 'static + Send>(ptr: *mut T) -> *mut T {
```

Do we want thread safety?
This captures the usize, which I think is Send. If T is !Send, because the closure using usize is Send, we could end up still running it. So we could end up freeing a !Send on another thread. Are we even worried about this case?

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+/// # Returns
+/// * `true` if the size is safe to use
+/// * `false` if the size would cause integer overflow
+///
+/// # Safety
+/// Caller must ensure that `ptr` points to valid memory if not null.
+/// This function performs pointer arithmetic with `ptr.add(size)` which requires
+/// that the pointer and size are valid for the memory region being checked.
+pub unsafe fn is_safe_buffer_size(size: usize, ptr: *const c_uchar) -> bool {
+    // Combined checks for early return - improves branch prediction
```
tmathern, line 302:
```suggestion
    // Combined checks for early return - improves branch prediction
    // 0-length buffer is considered unsafe, since used by C FFI layer
```

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -0,0 +1,519 @@
...
+/// * `bytes` - The byte vector to convert
+///
+/// # Returns
+/// * `*const c_uchar` - Pointer to the byte array
+///
+/// # Safety
+/// The returned pointer must be freed exactly once by calling `free_c_bytes`
+pub fn to_c_bytes(bytes: Vec<u8>) -> *const c_uchar {
+    let len = bytes.len();
+    let ptr = Box::into_raw(bytes.into_boxed_slice()) as *const c_uchar;
```
tmathern, line 405: In theory, bytes could be 0 sized. Since we deal with the C level here, this may create dangling pointers then (0-sized allocation). In another place we said o-sized things are not allowed, we should stay consistent.

c2pa_c_ffi/src/cimpl/utils.rs
tmathern: (Or add a panic handler? TO turn panics into errors for that FFI layer?)

tmathern (commented):
(Please disregard thread-safety related comments if thread safety is not a topic here).

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1220,30 +1496,28 @@ pub unsafe extern "C" fn c2pa_builder_sign(
     signer_ptr: *mut C2paSigner,
     manifest_bytes_ptr: *mut *const c_uchar,
 ) -> i64 {
-    check_or_return_int!(builder_ptr);
-    let format = from_cstr_or_return_int!(format);
-    check_or_return_int!(source);
-    check_or_return_int!(dest);
-    check_or_return_int!(signer_ptr);
-    check_or_return_int!(manifest_bytes_ptr);
+    ptr_or_return_int!(builder_ptr);
```
tmathern: Do we need to check if these dereferenced pointers are some we track, or we allow any?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1456,7 +1723,7 @@ pub unsafe extern "C" fn c2pa_signer_create(
     certs: *const c_char,
     tsa_url: *const c_char,
 ) -> *mut C2paSigner {
-    let certs = from_cstr_or_return_null!(certs);
+    let certs = cstr_or_return_null!(certs);
```
tmathern, line 1735: Isn't tracking missing here somewhere? I'd expect tracking starts here for a signer pointer, possibly?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -916,9 +1131,78 @@ pub unsafe extern "C" fn c2pa_builder_supported_mime_types(
...
+/// // builder is now invalid, use new_builder
+/// ```
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_builder_with_definition(
+    builder: *mut C2paBuilder,
+    manifest_json: *const c_char,
+) -> *mut C2paBuilder {
+    // Take ownership of the builder (consumes it)
+    let builder = Box::from_raw(builder);
+    let manifest_json = cstr_or_return_null!(manifest_json);
```
tmathern, line 1194: I think if we return/error early before the box_tracked call, we leak the builder pointer

tmathern (approved):

c2pa_c_ffi/src/cimpl/macros.rs
gpeacock: I decided to use *_or_return_int instead of or_return_neg since they both did the same thing

c2pa_c_ffi/src/cimpl/macros.rs
gpeacock: The "other" here provides a way to specify a Error return value since this is an Option and does not provide an error of its own. The null vs int & etc are the kind of C result the function will return. So they are providing two separate pieces of information. All that being said, it turns out I'm not currently using any of the some_or_return_other_ variants so I can just remove them.

We have to specify the kind of C return value for these macros, or use the base version and provide your own return value.

c2pa_c_ffi/src/cimpl/macros.rs
gpeacock: This doesn't allocate anything, it just creates a bounded slice. We won't ever read more than MAX_CSTRING_LEN even if the passed string has no terminator. It's an extra safely check with no significant overhead.

c2pa_c_ffi/src/c_api.rs
gpeacock: last_message() now does the unwrap_or_default() internally, so you don't need to do it here.

c2pa_c_ffi/src/c_api.rs
gpeacock: This was a good catch. Thank you!

c2pa_c_ffi/src/c_api.rs
gpeacock: I've added full tracking for everything I can track now

c2pa_c_ffi/src/c_api.rs
gpeacock: I cleaned this up and I don't see any tracking issues now.

c2pa_c_ffi/src/cimpl/utils.rs
gpeacock: Very good find, I updated the track_box methods to require Send bounds. That should be fine since only WASM doesn't support Send and WASM won't use ffi

c2pa_c_ffi/src/cimpl/utils.rs
gpeacock: I removed both these methods. It was part of a different approach I considered for builder patterns, but we aren't using them and don't need the complexity.

c2pa_c_ffi/src/cimpl/utils.rs
gpeacock: fixed this

c2pa_c_ffi/src/cimpl/macros.rs
gpeacock: I removed all the "other" macros, so no more issue.

c2pa_c_ffi/src/cimpl/utils.rs
gpeacock: The unwraps have all been removed and return errors or are handled with defaults.

PR #1784, opened 2026-01-27
fix: Propagate thumbnail settings to ingredients (more places where context is needed)

tmathern:
- Thumbnails settings need to be verified at all places where a thumbnail may get added.
- Handles also edge cases when ingredients don't have claims

## Checklist

### DIFF
--- a/.github/workflows/beta-preflight.yml
+++ b/.github/workflows/beta-preflight.yml
@@ -116,7 +116,7 @@ jobs:
       - name: Run doc tests (COVERAGE DISABLED)
         run:
           cargo test --workspace --all-features --doc
-      
+
       # - name: Generate code coverage
       #   env:
       #     RUST_BACKTRACE: "1"
@@ -203,7 +203,7 @@ jobs:
       - name: Set up Chrome
         uses: browser-actions/setup-chrome@v2
         with:
-          chrome-version: 142
+          chrome-version: 144
           install-chromedriver: true
 
       - name: Add wasm web target
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -362,7 +362,7 @@ jobs:
       - name: Set up Chrome
         uses: browser-actions/setup-chrome@v2
         with:
-          chrome-version: 142
+          chrome-version: 144
           install-chromedriver: true
 
       - name: Add wasm web target


sdk/src/ingredient.rs
```diff
@@ -1279,35 +1288,39 @@ impl Ingredient {
...
-                        )?
-                    } else {
-                        // add EmbeddedData thumbnail for v3 assertions in v2 claims
-                        let thumbnail = EmbeddedData::new(
-                            labels::INGREDIENT_THUMBNAIL,
-                            format_to_mime(&thumb_ref.format),
-                            data.into_owned(),
-                        );
-                        claim.add_assertion(&thumbnail)?
+            if auto_thumbnail {
```
ok-nick: I think here we should only check `auto_thumbnail` in the `None` case. That way users manually specifying thumbnails don't need `auto_thumbnail` enabled.

tmathern: Done as long as I understood the intended behavior right!

sdk/src/ingredient.rs
```diff
@@ -1246,8 +1251,12 @@ impl Ingredient {
                 let signature_uri = jumbf::labels::to_signature_uri(manifest_label);
 
                 // if there are validations and they have all passed, then use the parent claim thumbnail if available
+                // ... but only if thumbnail generation is enabled in settings
+                let auto_thumbnail = context.settings().builder.thumbnail.enabled;
                 if let Some(validation_results) = self.validation_results() {
-                    if validation_results.validation_state() != crate::ValidationState::Invalid {
+                    if validation_results.validation_state() != crate::ValidationState::Invalid
```
gpeacock, line 1257: this is different than auto-thumbnail generation and I don't think it should be gated by this setting. Here we are not generating a thumbnail, we are just referencing an existing one that is already in the the ingredient's active manifest. There is no generation here, we are just referencing.

sdk/src/ingredient.rs
```diff
@@ -1279,18 +1288,26 @@ impl Ingredient {
...
-                    HashedUri::new(thumb_ref.identifier.clone(), thumb_ref.alg.clone(), &hash)
+                    Some(HashedUri::new(
+                        thumb_ref.identifier.clone(),
+                        thumb_ref.alg.clone(),
+                        &hash,
+                    ))
                 }
-                None => {
+                None if auto_thumbnail => {
+                    // only process thumbnails without hash if auto_thumbnail is enabled
```
gpeacock, line 1307: This isn't generating a thumbnail either, it is adding one that the user has put into the resource store and referenced. What are we really trying to fix here? Is this trying redefine auto_thumbnail to mean don't add any thumbnails even if we tell you to?

sdk/src/ingredient.rs
tmathern: But should it be there if the setting says we don't want it?

sdk/src/ingredient.rs
tmathern: ack, correct, auto_thumbnail should continue having its current behavior here.

PR #1787, opened 2026-01-27
fix: Pull down context to read them when dealing with ingredients (no thumbnails cases)

tmathern:
- When `{"builder": { "thumbnail": {"enabled": false}}}` is enabled, I *believe* we should not have any thumbnail (since ingredients and manifest thumbnails are currently the same setting). This pulls down context to check that.
- Symptom/Repro: Add an ingredient with a thumbnail (in its own content credentials' manifest store), set the setting to disable the thumbnails, sign => there still will be a thumbnail in the manifest.

## Checklist


sdk/src/ingredient.rs
```diff
@@ -1246,8 +1251,12 @@ impl Ingredient {
                 let signature_uri = jumbf::labels::to_signature_uri(manifest_label);
 
                 // if there are validations and they have all passed, then use the parent claim thumbnail if available
+                // ... but only if thumbnail generation is enabled in settings
+                let auto_thumbnail = context.settings().builder.thumbnail.enabled;
                 if let Some(validation_results) = self.validation_results() {
-                    if validation_results.validation_state() != crate::ValidationState::Invalid {
+                    if validation_results.validation_state() != crate::ValidationState::Invalid
```
tmathern, line 1257: Shouldn't the setting still be checked here?

sdk/src/ingredient.rs
```diff
@@ -1279,35 +1288,39 @@ impl Ingredient {
...
-                        )?
-                    } else {
-                        // add EmbeddedData thumbnail for v3 assertions in v2 claims
-                        let thumbnail = EmbeddedData::new(
-                            labels::INGREDIENT_THUMBNAIL,
-                            format_to_mime(&thumb_ref.format),
-                            data.into_owned(),
-                        );
-                        claim.add_assertion(&thumbnail)?
+            if auto_thumbnail {
```
tmathern, line 1294: Shouldn't the setting still be checked here?

sdk/src/ingredient.rs
tmathern: (Even if there would be a thumbnail in the ingredient, we may not want it at this level? Does this make sense, @gpeacock ?)

sdk/src/ingredient.rs
tmathern: (Even if there would be a thumbnail in the ingredient, we may not want it at this level? Does this make sense, @gpeacock ?)

PR #1795, opened 2026-01-29
fix: Making sure CBOR kind is set, JSON/CBOR deserialization changes (debug edition)

tmathern:

## Checklist

### DIFF
--- a/sdk/src/manifest_assertion.rs
+++ b/sdk/src/manifest_assertion.rs
@@ -185,7 +185,7 @@ impl ManifestAssertion {
         match self.kind() {
             ManifestAssertionKind::Cbor => {
                 // For CBOR assertions, transcode from JSON back to CBOR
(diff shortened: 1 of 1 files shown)


scouten-adobe (approved):

sdk/src/manifest_assertion.rs
```diff
@@ -181,14 +181,42 @@ impl ManifestAssertion {
...
-            Error::AssertionDecoding(AssertionDecodeError::from_json_err(
-                self.label.to_owned(),
-                None,
-                "application/json".to_owned(),
-                e,
-            ))
-        })
+        // Check the kind field to determine how to deserialize
+        match self.kind() {
+            ManifestAssertionKind::Cbor => {
```
gpeacock, line 186: This will be lossy and won't work for some Assertions that use custom cbor - i.e. the Timestamp assertion.
But it should never be coming in this way.

PR #1796, opened 2026-01-29
fix: Making sure CBOR kind is set, JSON/CBOR deserialization changes

tmathern:
- Set kind on assertion
- Use kind when deserializing

Debug PR using know good-then-failing tests: https://github.com/contentauth/c2pa-rs/pull/1795 (same fix, different branch targets).

## Checklist

### DIFF
--- a/sdk/src/manifest_assertion.rs
+++ b/sdk/src/manifest_assertion.rs
@@ -248,4 +248,100 @@ pub(crate) mod tests {
         let ma2 = ManifestAssertion::from_assertion(&actions).expect("from_assertion");
         let _actions2: Actions = ma2.to_assertion().expect("to_assertion");
     }
+
+    /// Test that CBOR newtype structs (like TimeStamp) are deserialized as CBORwhen kind=Cbor is set.
+    ///
+    /// If CBOR type isn't checked and CBOR isn't used, the test will fail with error:
+    /// "invalid type: sequence, expected a map" (likely because somewhere it fell back to JSON deserialization).
+    #[test]
+    fn test_cbor_newtype_struct_with_hashmap() {
+        use crate::assertions::TimeStamp;
+
+        // Create an assertion that needs to be CBOR
+        let mut timestamp = TimeStamp::new();
+        timestamp.add_timestamp("manifest1", &[1, 2, 3, 4]);
+        timestamp.add_timestamp("manifest2", &[5, 6, 7, 8]);
+
+        let cbor_bytes = c2pa_cbor::to_vec(&timestamp).expect("serialize to CBOR");
+
+        // Transcode CBOR to JSON to mimic a roundtrip
+        let cbor_value: c2pa_cbor::Value =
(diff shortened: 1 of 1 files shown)


sdk/src/manifest_assertion.rs
```diff
@@ -181,14 +181,41 @@ impl ManifestAssertion {
...
-        serde_json::from_value(self.value()?.to_owned()).map_err(|e| {
-            Error::AssertionDecoding(AssertionDecodeError::from_json_err(
-                self.label.to_owned(),
-                None,
-                "application/json".to_owned(),
-                e,
-            ))
-        })
+        // Check the kind field to determine how to deserialize
+        match self.kind() {
```
ok-nick, line 185: Should we be matching on `self.data` here instead, since that's how the value is stored? I'm not sure what the purpose of `kind` is, it doesn't look like it's used anywhere.

sdk/src/manifest_assertion.rs
```diff
@@ -181,14 +181,41 @@ impl ManifestAssertion {
...
+            ManifestAssertionKind::Cbor => {
+                // For CBOR assertions, transcode from JSON back to CBOR
+                // It seems everything is stored in JSON?!?
```
ok-nick, line 188: It looks like in other parts of the code, there's an assumption that all the inputs will be JSON (see `ManifestAssertion::new`). Maybe we have to fix other parts of the code that call the constructor too?

sdk/src/manifest_assertion.rs
```diff
@@ -181,14 +181,41 @@ impl ManifestAssertion {
...
+                // For JSON and other types, use JSON deserialization
+                serde_json::from_value(self.value()?.to_owned()).map_err(|e| {
+                    Error::AssertionDecoding(AssertionDecodeError::from_json_err(
+                        self.label.to_owned(),
+                        None,
+                        "application/json".to_owned(),
+                        e,
+                    ))
+                })
+            }
```
ok-nick, line 217: We should error that we don't support decoding non-CBOR/non-JSON since that will always fail. Otherwise we'd always get back a JSON error even if it's stored as binary.

sdk/src/manifest_assertion.rs
tmathern: That, I think, is a lot to fix - and not something I'd do in a bugfix PR. This should be at least a feat PR then (when this fix happens). Let's be clever to reduce blast radius (more pushes may be incoming).

sdk/src/manifest_assertion.rs
tmathern: Do we already have an appropriate error type?

sdk/src/manifest_assertion.rs
tmathern: (TBD, looking at blast radius)

sdk/src/manifest_assertion.rs
ok-nick: Usually we lean on `Error::UnsupportedType`, but in my opinion a new error with more context would be better.

PR #1799, opened 2026-01-29
fix: CBOR/JSON handling when encoding/deserializing in some code paths

tmathern:
- Symptom: After the c2pa_cbor update, it became apparent some tests started to fail. The reported error is: `AssertionDecoding(could not decode assertion c2pa.time-stamp (version (no version), content type application/json): invalid type: sequence, expected a map)`. This can happen too in some cases wher `c2pa.actions.v2` with test assets from various models providers. We likely want the backwards compatibility here at least to ensure we can parse the data in manifests predating that change.
- There is a default encoding to JSON that seems to be causing the above symptom.

## Checklist

### DIFF
--- a/sdk/src/manifest_assertion.rs
+++ b/sdk/src/manifest_assertion.rs
@@ -149,7 +149,8 @@ impl ManifestAssertion {
         self
     }
 
-    /// Creates a ManifestAssertion from an AssertionBase object
+    /// Creates a ManifestAssertion from an AssertionBase object,
+    /// with the data parameter encoded as JSON (ManifestData::Json)
     ///
     /// # Example: Creating a custom assertion an Action assertion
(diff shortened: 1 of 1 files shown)


sdk/src/manifest_assertion.rs
```diff
@@ -181,14 +202,58 @@ impl ManifestAssertion {
...
+                            Error::AssertionDecoding(AssertionDecodeError::from_json_err(
+                                self.label.to_owned(),
+                                None,
+                                "application/json".to_owned(),
+                                e,
+                            ))
+                        })
+                    }
+                }
+            }
```
ok-nick, line 254: For this case, don't we already know the value will be JSON since we match on `self.data`?

PR #1807, opened 2026-02-02
build: Only use `reqwest` for `c2pa-c-ffi` networking

ok-nick:
Removes `ureq` dependency from `c2pa-c-ffi`, which should slightly lower the build size. `ureq` is enabled by default in the SDK due to: https://github.com/contentauth/c2pa-rs/issues/1364.

### DIFF
--- a/c2pa_c_ffi/Cargo.toml
+++ b/c2pa_c_ffi/Cargo.toml
@@ -24,7 +24,7 @@ rust_native_crypto = ["c2pa/rust_native_crypto"]
 pdf = ["c2pa/pdf"]
 
 [dependencies]
-c2pa = { path = "../sdk", version = "0.75.9", default-features = false, features = [
+c2pa = { path = "../sdk", version = "0.75.10", default-features = false, features = [
     "add_thumbnails",
     "fetch_remote_manifests",
     "file_io",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ As of December 2025 and until the 1.0.0 version is released, the CAI team will o
 
 ## [Unreleased]
 
+## [0.75.10](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.75.9...c2pa-v0.75.10)
+_02 February 2026_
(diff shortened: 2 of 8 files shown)


sdk/Cargo.toml
```diff
@@ -50,17 +50,11 @@ rust_native_crypto = [
 ]
 # TODO: make our async tests always use the async API so we don't need ureq + reqwest
 #       see https://github.com/contentauth/c2pa-rs/issues/1364
-# Enables default http features for sync and async http resolvers.
```
gpeacock, line 53: why remove the comments if we aren't removing the features yet?

ok-nick: I thought it would show up as docs on docs.rs but it doesn't. I'm removing it to stay consistent with the other features since we already document them in the lib.rs.

c2pa_c_ffi/Cargo.toml
```diff
@@ -29,7 +29,10 @@ c2pa = { path = "../sdk", version = "0.75.10", default-features = false, feature
     "fetch_remote_manifests",
     "file_io",
     "pdf",
-    "default_http",
+    # We don't use the `default_http` feature because it also includes `http_ureq`.
+    # TODO: use `default_http` pending https://github.com/contentauth/c2pa-rs/issues/1364
```
tmathern, line 33: This comment is more confusing for me than helping...
So until we have `default_http`, we use those 2 instead?

ok-nick: We have a `default_http` feature but it includes `http_ureq` because our SDK tests can't use the blocking reqwest API due to #1364.

tmathern: To clarify my question: For other crates that use c2pa-rs, what does it mean?
Is the expectations they also remove default_http to use the features you used here?

ok-nick: They don't have to change anything, but they can to reduce dependencies. When #1364 is fixed we can remove `ureq` as the default.

PR #1813, opened 2026-02-04
feat: Add archive marking and detection plus Reader into builder options

gpeacock:
NOTE: This is NOT ready to Merge and is just a draft of ideas in progress:

Archive Type Detection Using Custom Metadata Assertion
Added org.contentauth.archive.metadata custom assertion to mark and detect archive types, replacing the format-based approach that failed with v2 claims.
Key Changes:
Archive Creation (builder.rs):
working_store_sign() now adds org.contentauth.archive.metadata assertion with archive:type field
Archive type determined by calling method: to_archive() → "builder", to_ingredient_archive() → "ingredient"
Fixed double hash binding (BoxHash + DataHash): now conditionally uses BoxHash or DataHash based on signer availability - this will sign with the current context signer if it is available, making a valid box hashed signed archive.
Archive Detection (reader.rs):
manifest_type() checks for archive metadata assertion first using Metadata type
Falls back to heuristics for old archives (double binding, presence of ingredients)
manifest_into_builder() strips the archive metadata assertion when restoring (no longer an archive)
Three-Way Conversion Logic (reader.rs, builder.rs):
Reader.into_builder(): ArchivedBuilder → restores builder state, ArchivedIngredient → extracts ingredient, SignedManifest → converts to parent ingredient
Builder.add_ingredient_from_reader(): handles all three types appropriately
Archive metadata naturally removed when converting to ingredient (ingredient archives only contain the ingredient, not the parent manifest)
New API (builder.rs):
Added Builder.to_ingredient_archive() for creating ingredient archives with validation

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -2437,36 +2437,53 @@ impl Builder {
     }
 
     /// This creates a working store from the builder
-    /// The working store is signed with a BoxHash over an empty string
+    /// If a signer is available in the context, it uses box hash signing
+    /// Otherwise, it falls back to a DataHash placeholder
     /// And is returned as a Vec<u8> of the c2pa_manifest bytes
     /// This works as an archive of the store that can be read back to restore the Builder state
-    fn working_store_sign(&self) -> Result<Vec<u8>> {
-        // first we need to generate a BoxHash over an empty string
-        let mut empty_asset = std::io::Cursor::new("");
-        let boxes = jumbf_io::get_assetio_handler("application/c2pa")
-            .ok_or(Error::UnsupportedType)?
-            .asset_box_hash_ref()
-            .ok_or(Error::UnsupportedType)?
-            .get_box_map(&mut empty_asset)?;
-        let box_hash = BoxHash { boxes };
-
-        // then convert the builder to a claim and add the box hash assertion
+    fn working_store_sign(&mut self) -> Result<Vec<u8>> {
+        // Convert the builder to a claim
         let mut claim = self.to_claim()?;
         
         // Mark this as a builder archive in the claim generator info
         if let Some(cgi) = claim.claim_generator_info.as_mut() {
             if let Some(first) = cgi.first_mut() {
-                first.insert("com.contentauth.archive_type", "builder");
+                first.insert(
+                    "org.contentauth.archive_type",
+                    serde_json::Value::String("builder".to_string()),
+                );
             }
         }
-        
-        claim.add_assertion(&box_hash)?;
 
-        // now commit and sign it. The signing will allow us to detect tampering.
-        let mut store = Store::new();
-        store.commit_claim(claim)?;
-
-        store.get_data_hashed_manifest_placeholder(100, "application/c2pa")
+        // Check if we have a signer in the context
+        if let Ok(signer) = self.context.signer() {
+            // With signer: add BoxHash and sign
+            // Generate a BoxHash over an empty asset (for application/c2pa format)
+            let mut empty_asset = std::io::Cursor::new("");
+            let boxes = jumbf_io::get_assetio_handler("application/c2pa")
+                .ok_or(Error::UnsupportedType)?
+                .asset_box_hash_ref()
+                .ok_or(Error::UnsupportedType)?
+                .get_box_map(&mut empty_asset)?;
+            let box_hash = BoxHash { boxes };
+            
+            claim.add_assertion(&box_hash)?;
+            
+            // Commit the claim to a store
+            let mut store = Store::new();
+            store.commit_claim(claim)?;
+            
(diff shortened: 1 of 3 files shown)


sdk/src/builder.rs
```diff
@@ -1115,14 +1115,44 @@ impl Builder {
...
+    /// Ingredient archives use a special format (`application/x-c2pa-ingredient`) to distinguish
+    /// them from builder archives. When read back, they can be converted directly to an ingredient.
+    ///
+    /// # Arguments
+    /// * `stream` - A stream to write the archive into.
+    ///
+    /// # Errors
+    /// * Returns an [`Error`] if the builder doesn't have exactly one ingredient.
+    /// * Returns an [`Error`] if the archive cannot be written.
+    pub fn to_ingredient_archive(&mut self, stream: impl Write + Seek) -> Result<()> {
```
ok-nick, line 1137: What about the information in the enclosed `Builder`? I think this would make more sense if it was `Builder::to_ingredient_archive(ingredient_json, stream)`, and it archived every ingredient already inside the builder.

sdk/src/builder.rs
```diff
@@ -2386,7 +2416,54 @@ impl Builder {
...
+    /// This method handles different types of readers:
+    /// - **Builder Archive**: Extracts all ingredients from the archived builder
+    /// - **Archived Ingredient**: Extracts the single archived ingredient
+    /// - **Regular C2PA Manifest**: Converts the manifest into a parent ingredient
+    ///
+    /// # Arguments
+    /// * `reader` - The Reader to get the ingredient from (consumed).
+    /// # Returns
+    /// * A reference to the added ingredient.
+    fn add_ingredient_from_reader_owned(
```
ok-nick, line 2430: Does this need to take an owned `Reader` since it borrows all its fields?

sdk/src/builder.rs
```diff
@@ -2395,37 +2472,75 @@ impl Builder {
...
-            .asset_box_hash_ref()
-            .ok_or(Error::UnsupportedType)?
-            .get_box_map(&mut empty_asset)?;
-        let box_hash = BoxHash { boxes };
-
-        // then convert the builder to a claim and add the box hash assertion
+    fn working_store_sign(&mut self, archive_type: &str) -> Result<Vec<u8>> {
+        // Add archive metadata assertion to mark this as an archive
+        let archive_metadata = serde_json::json!({
+            "@context": { "archive": "http://contentauth.org/archive/1.0/" },
```
ok-nick, line 2507: Is this TBD?

sdk/src/builder.rs
```diff
@@ -2395,37 +2472,75 @@ impl Builder {
...
         self.definition
             .ingredients
             .last_mut()
             .ok_or(Error::IngredientNotFound)
     }
 
     /// This creates a working store from the builder
-    /// The working store is signed with a BoxHash over an empty string
+    /// If a signer is available in the context, it uses box hash signing
+    /// Otherwise, it falls back to a DataHash placeholder
```
ok-nick, line 2501: Why box/data hash depending on signer availability? I thought working stores are unsigned? Also, is there anything special that's done for box hash with `application/c2pa`?

sdk/src/builder.rs
```diff
@@ -4236,6 +4351,167 @@ mod tests {
         Ok(())
     }
 
+    // TODO: These tests require additional network-free test infrastructure
```
ok-nick, line 4354: Can we disable timestamping for this test?

sdk/src/reader.rs
```diff
@@ -1232,13 +1232,111 @@ impl Reader {
...
+                Ok(builder)
+            }
+            ManifestType::SignedManifest => {
+                // Regular manifest: convert the entire reader to a parent ingredient
+                let ingredient = self.reader_to_parent_ingredient()?;
+                let mut builder = crate::Builder::from_shared_context(&self.context);
+                builder.add_ingredient(ingredient);
+                Ok(builder)
+            }
+        }
```
ok-nick, line 1268: Why do we convert the parent to a signed manifest and not the entire thing as a a new ingredient? The latter is what I'd naturally expect.

sdk/src/reader.rs
```diff
@@ -1232,13 +1232,111 @@ impl Reader {
...
+    }
+
+    /// Converts the entire Reader into a single parent Ingredient.
+    /// The entire manifest store is embedded as manifest_data.
+    pub(crate) fn reader_to_parent_ingredient(&self) -> Result<Ingredient> {
```
ok-nick, line 1273: I think it would make more sense to implement `From<Reader> for Ingredient` and consume an owned `Reader` since we are cloning all its fields anyways.

gpeacock:
I wrote this a different way, so closing this one

PR #1839, opened 2026-02-07
feat: Use ephemeral self-signed cert to sign working store / archive

scouten-adobe:


sdk/src/builder.rs
```diff
@@ -1176,7 +1181,7 @@ impl Builder {
             let mut no_verify_settings = self.context.settings().clone();
             no_verify_settings.verify.verify_after_reading = false;
 
-            let temp_context = Context::new().with_settings(no_verify_settings)?;
+            let temp_context = Arc::new(Context::new().with_settings(no_verify_settings)?);
```
gpeacock: Context::new().with_settings(no_verfiy_settings)?.into_shared();

sdk/src/builder.rs
```diff
@@ -1186,10 +1191,15 @@ impl Builder {
...
-            // Now use the user's original context for the Reader and Builder
-            let mut reader = Reader::from_shared_context(&self.context);
+            // Use the no-verify context to avoid validating archive signatures,
+            // then restore the caller's context on the resulting `Builder`.
+            let mut reader = Reader::from_shared_context(&temp_context);
             reader.with_store(store, &mut validation_log)?;
-            reader.into_builder()
+
+            let mut builder = reader.into_builder()?;
+            builder.context = Arc::clone(&self.context);
```
gpeacock: or self.context.clone()

sdk/src/builder.rs
```diff
@@ -1186,10 +1191,15 @@ impl Builder {
                 &temp_context,
             )?;
 
-            // Now use the user's original context for the Reader and Builder
-            let mut reader = Reader::from_shared_context(&self.context);
+            // Use the no-verify context to avoid validating archive signatures,
+            // then restore the caller's context on the resulting `Builder`.
+            let mut reader = Reader::from_shared_context(&temp_context);
             reader.with_store(store, &mut validation_log)?;
-            reader.into_builder()
```
tmathern, line 1192: I don't think this line propagated context before... did it? So Maybe we should keep the change made here no matter what. I'd expect contexts to propagate everywhere if set (but maybe here, in this very special case, it wasn't by design??).

sdk/src/builder.rs
```diff
@@ -2414,29 +2424,81 @@ impl Builder {
...
 
-        // now commit and sign it. The signing will allow us to detect tampering.
+        // Now commit and sign it. The signing will allow us to detect tampering.
         let mut store = Store::new();
         store.commit_claim(claim)?;
 
-        store.get_data_hashed_manifest_placeholder(100, "application/c2pa")
+        #[cfg(feature = "self_signed_certs")]
+        {
+            let mut params = rcgen::CertificateParams::new(vec!["c2pa-archive.local".into()])
```
tmathern: What is `c2pa-archive.local`? What does that magic value do?

sdk/src/builder.rs
```diff
@@ -2414,29 +2424,81 @@ impl Builder {
...
+                cert_pem.as_bytes(),
+                key_pem.as_bytes(),
+                SigningAlg::Ed25519,
+                None,
+            )?;
+
+            let signer = RawSignerWrapper(raw_signer);
+
+            let mut archive_settings = self.context.settings().clone();
+            archive_settings.verify.verify_after_sign = false;
```
tmathern: Should we instead have a guard for this? It's easy to miss to set/reset it.

sdk/src/builder.rs
scouten-adobe: Yes, it did. See https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/reader.rs#L1236-L1237 (unchanged in this PR).

The additional logic is required because the reader is built using `temp_context` which didn't exist before this PR. Without the new logic, we'd propagate _that_ context which would have the unwanted consequence of disabling `verify_after_reading` for anything touched by that result `Builder`.

sdk/src/builder.rs
scouten-adobe: This name is passed as a "subject alternative name" when creating the ephemeral self-signed cert. (See https://docs.rs/rcgen/latest/rcgen/struct.CertificateParams.html#method.new.)

It's mostly intended to provide a hint to a human reader about the purpose for this certificate.

sdk/src/builder.rs
scouten-adobe: I don't understand the purpose of the guard you're proposing. We create a temporary settings instance and context (`archive_settings` / `archive_context`), use it to get the signed manifest bytes at line 2495. The modified settings and context fall out of scope and are thus discarded at line 2496.

The return value of this function is a `Result<Vec<u8>>` and does not have any memory of the settings or context used to create it.

tmathern: Let me rephrase: I. The future, if this code is kept, we may mistakenly remove one of the lines. Wrapper is probably rhe better term.
Or tests, if any would start tod ail, that could help us check this stuck around.

tmathern: Actually, tripped over my own feet. The setting was here already before this change, so it's a question for before already. Nevermind for this PR.

sdk/src/builder.rs
```diff
@@ -1186,10 +1191,15 @@ impl Builder {
...
 
-            // Now use the user's original context for the Reader and Builder
-            let mut reader = Reader::from_shared_context(&self.context);
+            // Use the no-verify context to avoid validating archive signatures,
+            // then restore the caller's context on the resulting `Builder`.
+            let mut reader = Reader::from_shared_context(&temp_context);
             reader.with_store(store, &mut validation_log)?;
-            reader.into_builder()
+
+            let mut builder = reader.into_builder()?;
```
tmathern: Note for future changes: `into_builder` could use context propagation (given how it got used here).

sdk/src/builder.rs
tmathern: I have a feeling it will end up in a constant/enum.

sdk/src/builder.rs
```diff
@@ -2414,29 +2424,81 @@ impl Builder {
...
         let mut claim = self.to_claim()?;
         claim.add_assertion(&box_hash)?;
 
-        // now commit and sign it. The signing will allow us to detect tampering.
+        // Now commit and sign it. The signing will allow us to detect tampering.
         let mut store = Store::new();
         store.commit_claim(claim)?;
 
-        store.get_data_hashed_manifest_placeholder(100, "application/c2pa")
+        #[cfg(feature = "self_signed_certs")]
```
gpeacock: generating this signer feels like it belongs somewhere else in crypto or signing or something. If we want it here in builder, I'd at least move it to it's own method. I'd also vote that it not be a feature, since the placeholder logic is invalid at best.

sdk/src/builder.rs
```diff
@@ -2414,29 +2424,81 @@ impl Builder {
...
+            let raw_signer = signer_from_cert_chain_and_private_key(
+                cert_pem.as_bytes(),
+                key_pem.as_bytes(),
+                SigningAlg::Ed25519,
+                None,
+            )?;
+
+            let signer = RawSignerWrapper(raw_signer);
+
+            let mut archive_settings = self.context.settings().clone();
```
gpeacock: We should not be modifying settings on the fly like this. If they don't want verify after sign, they should disable it it settings for the builder. But this should verify now, so we don't need to disable verify, right?

sdk/src/builder.rs
```diff
@@ -4247,6 +4309,44 @@ mod tests {
         Ok(())
     }
 
+    #[cfg(feature = "self_signed_certs")]
+    #[test]
+    fn test_archive_self_signed_ed25519_signature() -> Result<()> {
+        let settings = Settings::new()
+            .with_value("builder.generate_c2pa_archive", true)?
```
gpeacock: You should not need either of these settings now. - generate_c2pa_archive is now the default and verify_after_reading should work.

gpeacock: You could disable verify_after_sign here if we don't want that overhead, but it should be required.

sdk/src/builder.rs
```diff
@@ -4247,6 +4309,44 @@ mod tests {
...
+            Builder::from_context(context).with_definition(r#"{"title": "Test Self Signed"}"#)?;
+
+        let mut archive = Cursor::new(Vec::new());
+        builder.to_archive(&mut archive)?;
+        archive.rewind()?;
+
+        let mut validation_log = crate::status_tracker::StatusTracker::default();
+        let read_context = Context::new().with_settings(settings)?;
+
+        let store = Store::from_stream(
```
gpeacock: why not use reader here? One of the good things about this change is that we can use Reader now.

sdk/src/builder.rs
tmathern: (I think I prefer self.context.clone() too)

sdk/src/builder.rs
tmathern: That is what I worried about too. It may be side-effect prone (remember the comment I didn't know how to formulate right, @scouten-adobe ? About the wrapper).

sdk/src/utils/ephemeral_signer.rs
```diff
@@ -0,0 +1,399 @@
...
+    crypto::raw_signature::{signer_from_cert_chain_and_private_key, RawSigner, SigningAlg},
+    Error, Result, Signer,
+};
+
+/// A [`Signer`](crate::Signer) that holds an ephemeral CA + end-entity certificate chain.
+///
+/// The full certificate chain (end-entity first, then CA) is stored as DER and returned from
+/// [`Signer::certs`](crate::Signer::certs). Signing is performed by the end-entity key.
+///
+/// **WARNING:** These certificates are for testing and local use only and will not
```
tmathern: Given it is already in multiple places, you could put a warning on the very top too, as a general warning?

sdk/src/builder.rs
```diff
@@ -1170,25 +1170,18 @@ impl Builder {
...
             let store = Store::from_stream(
                 "application/c2pa",
                 &mut stream,
                 &mut validation_log,
-                &temp_context,
+                &self.context,
             )?;
 
-            // Now use the user's original context for the Reader and Builder
+            // Use the no-verify context to avoid validating archive signatures,
```
tmathern: Is the comment obsolete? I don't see the temp context?

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
+# Working stores and C2PA archives
+
+This document explains how **working stores** and **C2PA archives** work in the C2PA SDK, how they relate to each other, and how to save and restore builder state.
+
+## Working stores vs C2PA archives
+
+**Working store** and **C2PA archive** refer to the same underlying concept; the difference is one of emphasis:
```
tmathern: `the same underlying concept; the difference is one of emphasis:`.

I would word that differently: they are the same concept/object, but in different parts of their lifecycle.

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+
+| Term | Emphasis |
+|------|----------|
+| **Working store** | The *content*: an editable C2PA manifest state (claims, ingredients, assertions) that has not yet been bound to a final asset. Often used when describing the in-memory or serialized form of “work in progress” manifest data. |
```
tmathern: `or serialized form` can easily be confused with an archived one, but I don't think it's what is meant here.

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+| **C2PA archive** | The *artifact*: the saved bytes (e.g. in a `.c2pa` file or stream) that you get from saving a working store and read back to restore a `Builder`. |
+
+Both use the **same on-disk/wire format**: standard JUMBF `application/c2pa`. That is, a C2PA archive *is* a working store serialized as a normal C2PA manifest store. The spec does not define a separate “archive” format; the SDK reuses the standard manifest store format so that:
+
+- The same format is used for **signed manifests** (bound to an asset), **working stores** (saved for later editing), and **saved ingredients** (e.g. validated once, reused in other manifests).
+- Archives can be embedded in files, stored in the cloud, or saved as sidecar `.c2pa` files.
```
tmathern:
```suggestion
- An archive can be embedded in files, stored in the cloud, or saved as sidecar `.c2pa` files.
```
(Otherwise it could be read as you can embed multiple different archives/working stores.We should distinguish adding multiple ingredients through an archived ingredient, and adding 1 archived builder/working store).

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Distinction in practice:**
+
+- When you **save** a `Builder` with `to_archive()`, you produce a **working store** (serialized as JUMBF `application/c2pa`). That serialized form is what we call a **C2PA archive**.
+- When you **restore** with `from_archive()` or `with_archive()`, you read that **C2PA archive** back into a `Builder` — i.e. you restore the **working store** so you can continue editing.
```
tmathern: Add a precision that adding another working store as ingredient won't merge them please.

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+
+So: one concept, two names. “Working store” stresses the editable state; “C2PA archive” stresses the saved, portable representation.
```
tmathern: Put this on top!

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+
+---
+
+## How a working store is saved (C2PA format)
+
+When using the C2PA archive format, saving a `Builder` does the following:
+
+1. The `Builder`’s manifest data (assertions, ingredients, etc.) is prepared for signing.
+2. A **BoxHash** assertion is generated over an **empty asset** (placeholder), so the manifest is not bound to any real content.
+3. The manifest is signed with an **ephemeral self-signed certificate** (e.g. `c2pa-archive.local`). This signature is only for tamper detection in a private context; it is not intended for public trust.
```
tmathern: Not sure what the eg. means here? Cert name?

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+---
+
+## How a working store is restored
+
+Restoring from a C2PA archive does the following:
+
+1. The archive stream is read and parsed as JUMBF `application/c2pa`.
+2. A **Reader** is created and populated from that stream. NOTE: validation is performed with trust validation disabled so the ephemeral archive signature can be accepted.
```
tmathern, line 68: In the future, we should consider people using trusted certificate (in the future = not this PR), especially if this approach sticks around.

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+    end
+```
+
+---
+
+## API summary
+
+| Operation | API | Description |
+|-----------|-----|-------------|
+| Save | `builder.to_archive(&mut stream)` | Writes the working store (C2PA or legacy ZIP) to `stream`. |
```
tmathern: Since it's a quick ref, also indicate which setting does what

docs/working-stores.md
```diff
@@ -0,0 +1,110 @@
...
+| Restore (new Builder) | `Builder::from_archive(stream)` | Creates a default-context `Builder` and loads the archive into it. |
```
tmathern:
```suggestion
| Restore as Builder with default context | `Builder::from_archive(stream)` | Creates a default-context `Builder` and loads the archive into it. |
```
"Restore" and "new Builder" is a bit hard to process... Just restoring or reloading?

tmathern (approved):
Approving, contingent to a second approval here.

sdk/src/utils/ephemeral_signer.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub fn new(ee_cert_name: impl Into<String>) -> Result<Self> {
+        let ee_cert_name = ee_cert_name.into();
+
+        // Create a self-signed CA using the same approach as the EE (Ed25519, etc.).
+        let mut ca_params = CertificateParams::new(vec!["c2pa-ephemeral-ca.local".into()])
+            .map_err(|e| Error::OtherError(Box::new(e)))?;
+
+        ca_params.distinguished_name.push(
+            rcgen::DnType::OrganizationName,
+            "c2pa-rs self-signed ephemeral CA",
```
gpeacock: Given that this may be visible, we should make sure we include it in the documentation and maybe use contentauth.org or something in the name that might be more approachable than c2pa-rs. But I like that it calls it out as self signed or ephemeral. We may want to get some UI feedback on the naming.

sdk/src/utils/ephemeral_signer.rs
```diff
@@ -0,0 +1,399 @@
...
+        let issuer = rcgen::Issuer::new(ca_params, ca_keypair);
+
+        // End-entity certificate parameters (aligned with working_store_sign).
+        let mut ee_params = CertificateParams::new(vec![ee_cert_name])
+            .map_err(|e| Error::OtherError(Box::new(e)))?;
+
+        // Verifier expects an Organization in the subject for issuer_org in CertificateInfo.
+        ee_params
+            .distinguished_name
+            .push(rcgen::DnType::OrganizationName, "C2PA Ephemeral");
```
gpeacock: We probably want contentauth.org here rather than any claim that this is C2PA.

sdk/src/utils/ephemeral_signer.rs
```diff
@@ -0,0 +1,399 @@
...
+
+    fn timestamp_request_body(&self, message: &[u8]) -> Result<Vec<u8>> {
+        self.raw_signer
+            .time_stamp_request_body(message)
+            .map_err(|e| e.into())
+    }
+
+    fn send_timestamp_request(&self, message: &[u8]) -> Option<Result<Vec<u8>>> {
+        self.raw_signer
+            .send_time_stamp_request(message)
```
gpeacock, line 235: make sure we can leverage the context resolver with this.

sdk/src/builder.rs
```diff
@@ -1170,25 +1170,18 @@ impl Builder {
...
+            // then restore the caller's context on the resulting `Builder`.
             let mut reader = Reader::from_shared_context(&self.context);
-            reader.with_store(store, &mut validation_log)?;
+            reader.with_store(store, &mut validation_log).unwrap();
```
gpeacock: This need to handle the error, no unwraps!

sdk/src/builder.rs
```diff
@@ -2414,29 +2407,37 @@ impl Builder {
...
         let mut claim = self.to_claim()?;
         claim.add_assertion(&box_hash)?;
 
-        // now commit and sign it. The signing will allow us to detect tampering.
+        // Now commit and sign it. The signing will allow us to detect tampering.
         let mut store = Store::new();
         store.commit_claim(claim)?;
 
-        store.get_data_hashed_manifest_placeholder(100, "application/c2pa")
+        let signer = EphemeralSigner::new("c2pa-archive.local")?;
```
gpeacock, line 2460: hurray!

sdk/src/builder.rs
```diff
@@ -4247,6 +4248,43 @@ mod tests {
...
+            Builder::from_context(context).with_definition(r#"{"title": "Test Self Signed"}"#)?;
+
+        let mut archive = Cursor::new(Vec::new());
+        builder.to_archive(&mut archive)?;
+        archive.rewind()?;
+
+        let mut validation_log = crate::status_tracker::StatusTracker::default();
+        let read_context = Context::new().with_settings(settings)?;
+
+        let store = Store::from_stream(
```
gpeacock: I'd like to see this validated at the Reader level just to make sure. But I would not stop this PR for it.

sdk/src/builder.rs
scouten-adobe: Clippy didn't like that either. Fixed just now.

sdk/src/builder.rs
scouten-adobe: It ~~is~~ _was_ obsolete. Removed.

sdk/src/builder.rs
scouten-adobe: We no longer need to disable verify-after-sign, so this is no longer relevant.

sdk/src/builder.rs
scouten-adobe: Decided to leave it in.

sdk/src/builder.rs
```diff
@@ -1118,19 +1120,34 @@ impl Builder {
...
     pub fn to_archive(&mut self, mut stream: impl Write + Seek) -> Result<()> {
         if let Some(true) = self.context.settings().builder.generate_c2pa_archive {
-            let c2pa_data = self.working_store_sign()?;
-            stream.write_all(&c2pa_data)?;
-        } else {
-            return self.old_to_archive(stream);
+            #[cfg(target_arch = "wasm32")]
+            // Generating a C2PA archive is not supported on Wasm32 because the rcgen
+            // crate has dependencies that can not be used.
+            return Err(Error::WasmFeatureUnsupported);
```
ok-nick: This seems like a major limitation for the `c2pa-js` SDK that they will not be able to use the new archives? Does anyone there depend on this right now?

docs/working-stores.md
scouten-adobe: Removing that part of the description.

docs/working-stores.md
scouten-adobe: Adapted this in my (pending) update.)

docs/working-stores.md
scouten-adobe: Removing it.

sdk/src/utils/ephemeral_signer.rs
```diff
@@ -0,0 +1,417 @@
...
+
+        // End-entity certificate parameters (aligned with working_store_sign).
+        let mut ee_params = CertificateParams::new(vec![ee_cert_name])
+            .map_err(|e| Error::OtherError(Box::new(e)))?;
+
+        // Verifier expects an Organization in the subject for issuer_org in
+        // CertificateInfo.
+        ee_params.distinguished_name.push(
+            rcgen::DnType::OrganizationName,
+            "Self-signed ephemeral certificate (Content Authenticity SDK) -- LOCAL USE ONLY",
```
gpeacock, line 116: This is probably too much for the UI to display. Did we get feedback on this?

gpeacock, line 116: But I'm not sure what get's displayed, we should test this.

sdk/src/builder.rs
```diff
@@ -1118,19 +1120,34 @@ impl Builder {
...
     pub fn to_archive(&mut self, mut stream: impl Write + Seek) -> Result<()> {
         if let Some(true) = self.context.settings().builder.generate_c2pa_archive {
-            let c2pa_data = self.working_store_sign()?;
-            stream.write_all(&c2pa_data)?;
-        } else {
-            return self.old_to_archive(stream);
+            #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
+            // Generating a C2PA archive is not supported on browser Wasm because the rcgen
+            // crate has dependencies that can not be used. It is supported on WASI.
+            return Err(Error::WasmFeatureUnsupported);
```
gpeacock, line 1140: I suspect this could be a problem. Do we have any ideas for wasm?

scouten-adobe: I'll add a ticket for that; there is a problem with random number generation on Wasm that isn't easily worked around.

scouten-adobe: Split Wasm issue to https://jira.corp.adobe.com/browse/CAI-11001.

PR #1846, opened 2026-02-10
fix!: Allow `Builder::sign_async` future to be spawnable on Tokio runtime

### LINKED ISSUE #1815: Builder.sign_async cannot be called in a web server context since AsyncResolver doesn't implement Sync trait
### Component

c2pa-rs

### If Other, please specify

_No response_

### Version

v0.75.13

### Platform

Linux

### What happened?

http::AsyncResolver/SyncResolver do not implement Sync Trait so cannot be used in a web server context easily. In a typical tokio runtime context you would need Send + Sync

Builder.sign can be called in a web server context (or a tokio context)

ok-nick:
For a future to be spawned on a Tokio runtime (or most runtimes with multithreaded support for that matter), it must implement `Send`, [see bounds here](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.spawn). Currently the future returned by `Builder::sign_async` is not `Send` because it captures `&dyn AsyncHttpResolver` across an await (from `Context::resolver_async`'s return). `&T` is `Send` only if `T: Sync`, and thus, adding `+ Sync` to the bound on the latter function's return allows `Send` to be implicitly derived.

* Closes #1815

### DIFF
--- a/sdk/src/http/mod.rs
+++ b/sdk/src/http/mod.rs
@@ -64,12 +64,12 @@ pub mod restricted;
 pub use http;
 
 /// A resolver for sync (blocking) HTTP requests.
-///
-/// This trait is a supertrait of [`MaybeSend`] and [`MaybeSync`] for consistency with the
-/// [`AsyncHttpResolver`]. For more information on the rationale, see [`AsyncHttpResolver`].
-///
-/// [`MaybeSend`]: crate::maybe_send_sync::MaybeSend
-/// [`MaybeSync`]: crate::maybe_send_sync::MaybeSync
+//
+// This trait is a supertrait of [`MaybeSend`] and [`MaybeSync`] for consistency with the
+// [`AsyncHttpResolver`]. For more information on the rationale, see [`AsyncHttpResolver`].
+//
+// [`MaybeSend`]: crate::maybe_send_sync::MaybeSend
+// [`MaybeSync`]: crate::maybe_send_sync::MaybeSync
 pub trait SyncHttpResolver: MaybeSend + MaybeSync {
     /// Resolve a [`Request`] into a [`Response`] with a streaming body.
     ///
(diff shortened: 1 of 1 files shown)


sdk/src/context.rs
```diff
@@ -331,15 +331,15 @@ impl Context {
     /// # Arguments
     ///
     /// * `resolver` - Any type implementing `AsyncHttpResolver`
-    pub fn with_resolver_async<T: AsyncHttpResolver + MaybeSend + MaybeSync + 'static>(
+    pub fn with_resolver_async<T: AsyncHttpResolver + MaybeSend + Sync + 'static>(
```
tmathern: Wasn't it MaybeSync to some WASM weirdness? (May be obsolete by now, just asking if you know why it was MaybeSync to start with?).

ok-nick: The reason for `MaybeSync` is because our JS SDK implements a `Signer` that doesn’t implement `Sync`. It’s not necessarily a WASM problem but a limitation of our implementation. In general  I don’t think we need `MaybeSync` in all of these places besides the `Signer`.

The exact line in the JS SDK where the `Signer` is `!Sync` is here:
https://github.com/contentauth/c2pa-js/blob/db669f86ab37864a4b24e67c683e023308a9c0d8/packages/c2pa-wasm/src/wasm_signer.rs#L38

ok-nick: Thought about this a little bit more. Since the only reason we have `Sync` is so that `&T: Send` which is only required in the case we have of a multi-threaded Tokio runtime, then there is no point in forcing that bound on WASM which doesn't support `Send` anyways. So, `Context::resolver_async` should probably return `dyn AsyncHttpResolver + MaybeSync` instead or return an owned reference to it via `Arc`.

tmathern: Does this also mean that this change would be compatible with the JS SDKs? We wouldn't want to break them, or at the very least, warn them.

ok-nick: Yes, I took a step back and realized we can relax the type bounds, so this wouldn't affect the JS SDK.

sdk/src/http/interop.rs
```diff
@@ -0,0 +1,50 @@
...
+pub type AsyncResolver = dyn AsyncHttpResolver;
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<T: AsyncHttpResolver + Send + Sync + ?Sized> AsyncHttpResolver for Arc<T> {
+    async fn http_resolve_async(
+        &self,
+        request: Request<Vec<u8>>,
+    ) -> Result<Response<Box<dyn Read>>, HttpResolverError> {
+        (**self).http_resolve_async(request).await
```
gpeacock: how is this not recursive?

sdk/src/http/interop.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[cfg(not(target_arch = "wasm32"))]
+pub type AsyncResolver = dyn AsyncHttpResolver + Send + Sync;
+
+#[cfg(target_arch = "wasm32")]
+pub type AsyncResolver = dyn AsyncHttpResolver;
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+impl<T: AsyncHttpResolver + Send + Sync + ?Sized> AsyncHttpResolver for Arc<T> {
```
gpeacock: I get confused with having all these definitions removed from the traits they apply to. Why do we need to move them to a separate file?

ok-nick: I refactored this out in latest change.

sdk/src/http/mod.rs
```diff
@@ -79,7 +64,13 @@ pub mod restricted;
 pub use http;
 
 /// A resolver for sync (blocking) HTTP requests.
-pub trait SyncHttpResolver {
+///
+/// This trait is a supertrait of [`MaybeSend`] and [`MaybeSync`] for consistency with the
+/// [`AsyncHttpResolver`]. For more information on the rationale, see [`AsyncHttpResolver`].
+///
+/// [`MaybeSend`]: crate::maybe_send_sync::MaybeSend
```
gpeacock: not sure why we need these comments here

sdk/src/http/mod.rs
```diff
@@ -90,10 +81,32 @@ pub trait SyncHttpResolver {
     ) -> Result<Response<Box<dyn Read>>, HttpResolverError>;
 }
 
+/// This implementation is particularly useful for compatibility with the return
+/// type of [`Context::resolver`].
+///
+/// [`Context::resolver`]: crate::Context::resolver
```
gpeacock, line 87: Are these references being generated by something?

ok-nick: They are generated in the docs under blanket implementations for HTTP resolvers. I figured it's good to  document here so we know why it's necessary.

sdk/src/http/mod.rs
ok-nick: It doesn't really make sense why they would be there without context, and I went through a couple of revisions before I settled on this solution, so I think it's good to include the reasoning here.

ok-nick: I changed these to comments rather than public docs.

PR #1847, opened 2026-02-11
feat: Embeddable manifest API with Context/Settings, CAWG and BMFF.v3 support.

gpeacock:
## Summary

This PR started as a targeted fix to make the existing `data_hashed_embeddable` API work with CAWG/dynamic assertions. But supporting CAWG dynamic assertions required new apis that could carry the signer context through for dynamic assertions. The embeddable APIs had not been updated for the new Context / Settings changes. Since we had to introduce new APIs and we also needed to add new support for embeddable BMFF signing, we combined the three different hash models into one set of APIs. 

## New API

| Method | Purpose |
|---|---|
| `needs_placeholder(format)` | Tells the caller whether a pre-embed placeholder step is required |
| `placeholder(format)` | Returns composed manifest bytes to embed (Mode 1); no-ops for BoxHash formats |
| `set_data_hash_exclusions(ranges)` | Registers the byte ranges the placeholder occupies |
| `set_bmff_mdat_hashes(leaf_hashes)` | Supplies externally-computed mdat Merkle hashes |
| `update_hash_from_stream(format, stream)` | Computes and stores the asset hash (DataHash, BmffHash, or BoxHash) |
| `sign_embeddable(format)` | Signs and returns final manifest bytes; pads to placeholder size in Mode 1, returns direct bytes in Mode 2 |

## Hard-binding paths unified

- **DataHash** — original target; placeholder → embed → hash → sign workflow, now Context/CAWG-aware.
- **BmffHash v3** — added Merkle tree support (`add_merkle_placeholder`, `set_bmff_mdat_hashes`), Merkle root-only verification, and zero-padding of signed JUMBF to placeholder size per the C2PA spec.
- **BoxHash** — implements the `builder.prefer_box_hash` setting in these new apis - defaulted to false.  This paves the way to supporting Box hash by default

## Effect on existing APIs

The older embeddable APIs (`get_data_hashed_manifest_placeholder`, `get_data_hashed_embeddable_manifest`, etc.) are effectively superseded and will be deprecated soon. The new APIs are format-agnostic from the caller's perspective — the same pattern works regardless of whether the asset ends up using DataHash, BmffHash, or BoxHash:

1. Check `needs_placeholder(format)` — skip steps 2–3 for BoxHash formats
2. Call `placeholder(format)` if needed and embed the bytes into the asset
3. Register exclusion ranges with `set_data_hash_exclusions` (DataHash) or supply mdat hashes with `set_bmff_mdat_hashes` (BMFF)
4. Call `update_hash_from_stream(format, stream)` to compute and store the asset hash if hashes are not supplied.
5. Call `sign_embeddable(format)` to get the final signed manifest bytes in the given composed format

## Limitations

The embeddable api has limitations and is intended only for advanced use cases:
- All asset I/O other than hashing must be done by the client.
- The hashes may be supplied by the client if they manually add hard binding assertions.
- It does not support Intents
- it will not generate thumbnails
- It will not auto add parent ingredients
- It does not load remote manifests

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -462,6 +462,39 @@ pub unsafe extern "C" fn c2pa_context_builder_set_settings(
     0
 }
 
+/// Set a Signer into the Builder's context.
+/// (The context will own the Signer from that point on).
+/// The signer will be available via `context.signer()` after
+/// building the context. If a signer is also configured in settings,
+/// the programmatic signer takes priority regardless of call order.
+///
+/// Works with any C2paSigner pointer, whether created by
+/// `c2pa_signer_from_info` or `c2pa_signer_create`.
+///
+/// # Safety
+///
+/// * `builder` must be a valid C2paContextBuilder pointer (not yet built).
+/// * `signer_ptr` must be a valid C2paSigner pointer. It is consumed by this
+///   call and must not be used or freed afterward.
+///
+/// # Returns
+///
+/// 0 on success, negative value on error.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_context_builder_set_signer(
+    builder: *mut C2paContextBuilder,
+    signer_ptr: *mut C2paSigner,
+) -> c_int {
+    let builder = deref_mut_or_return_int!(builder, C2paContextBuilder);
+    // Untrack the signer before taking ownership via Box::from_raw.
+    // This prevents double-free if C code later calls c2pa_signer_free().
+    untrack_or_return_int!(signer_ptr, C2paSigner);
+    let c2pa_signer = Box::from_raw(signer_ptr);
+    let result = builder.set_signer(c2pa_signer.signer);
+    ok_or_return_int!(result);
+    0
+}
+
 /// Builds an immutable, shareable context from the builder.
 ///
 /// The builder is consumed by this operation and becomes invalid.
@@ -481,6 +514,7 @@ pub unsafe extern "C" fn c2pa_context_builder_set_settings(
 pub unsafe extern "C" fn c2pa_context_builder_build(
     builder: *mut C2paContextBuilder,
 ) -> *mut C2paContext {
+    untrack_or_return_null!(builder, C2paContextBuilder);
     let context = Box::from_raw(builder);
     box_tracked!((*context).into_shared())
 }
@@ -820,6 +854,7 @@ pub unsafe extern "C" fn c2pa_reader_with_stream(
     let stream = deref_mut_or_return_null!(stream, C2paStream);
 
     // Now safe to take ownership - all validations passed
+    untrack_or_return_null!(reader, C2paReader);
     let reader = Box::from_raw(reader);
     let result = (*reader).with_stream(&format, stream);
     let result = ok_or_return_null!(post_validate(result));
@@ -864,6 +899,7 @@ pub unsafe extern "C" fn c2pa_reader_with_fragment(
     let fragment = deref_mut_or_return_null!(fragment, C2paStream);
 
     // Now safe to take ownership - all validations passed
+    untrack_or_return_null!(reader, C2paReader);
     let reader = Box::from_raw(reader);
     let result = (*reader).with_fragment(&format, stream, fragment);
     let result = ok_or_return_null!(post_validate(result));
@@ -1204,6 +1240,7 @@ pub unsafe extern "C" fn c2pa_builder_with_definition(
     let manifest_json = cstr_or_return_null!(manifest_json);
 
     // Now safe to take ownership - all validations passed
+    untrack_or_return_null!(builder, C2paBuilder);
     let builder = Box::from_raw(builder);
     let result = (*builder).with_definition(manifest_json);
     box_tracked!(ok_or_return_null!(result))
@@ -1240,6 +1277,7 @@ pub unsafe extern "C" fn c2pa_builder_with_archive(
     let stream = deref_mut_or_return_null!(stream, C2paStream);
 
     // Now safe to take ownership - stream is valid
+    untrack_or_return_null!(builder, C2paBuilder);
     let builder = Box::from_raw(builder);
     let result = (*builder).with_archive(stream);
     box_tracked!(ok_or_return_null!(result))
@@ -1547,6 +1585,54 @@ pub unsafe extern "C" fn c2pa_builder_sign(
         &mut *source,
         &mut *dest,
     );
+    let manifest_bytes = ok_or_return_int!(result);
+    let len = manifest_bytes.len() as i64;
+    if !manifest_bytes_ptr.is_null() {
+        *manifest_bytes_ptr = to_c_bytes(manifest_bytes);
+    }
+    len
+}
+
+/// Sign using the Signer from the Context.
+///
+/// Equivalent to `c2pa_builder_sign` but the signer comes from the Builder's
+/// context instead of being passed explicitly.
+///
+/// If the context has no signer (neither programmatic via
+/// `c2pa_context_builder_set_signer` nor from settings), an error
+/// will be returned.
+///
+/// # Parameters
+///
+/// * `builder_ptr` - pointer to a Builder whose context has a signer set.
+/// * `format` - MIME type or file extension (null-terminated C string).
+/// * `source` - pointer to a readable C2paStream.
+/// * `dest` - pointer to a read+write+seek C2paStream.
+/// * `manifest_bytes_ptr` - out-pointer for the manifest bytes.
+///
+/// # Safety
+///
+/// Reads from NULL-terminated C strings.
+/// The returned bytes MUST be released by calling `c2pa_free`.
+///
+/// # Returns
+///
+/// The length of the manifest bytes on success, or -1 on error.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_builder_sign_context(
+    builder_ptr: *mut C2paBuilder,
+    format: *const c_char,
+    source: *mut C2paStream,
+    dest: *mut C2paStream,
+    manifest_bytes_ptr: *mut *const c_uchar,
+) -> i64 {
+    let builder = deref_mut_or_return_int!(builder_ptr, C2paBuilder);
+    let format = cstr_or_return_int!(format);
+    let source = deref_mut_or_return_int!(source, C2paStream);
+    let dest = deref_mut_or_return_int!(dest, C2paStream);
+    ptr_or_return_int!(manifest_bytes_ptr);
+
+    let result = builder.save_to_stream(&format, &mut *source, &mut *dest);
     let manifest_bytes = ok_or_return_int!(result);
     let len = manifest_bytes.len() as i64;
     if !manifest_bytes_ptr.is_null() {
@@ -2825,6 +2911,10 @@ mod tests {
             unsafe { c2pa_reader_with_stream(reader, format.as_ptr(), stream.as_ptr()) };
         assert!(!configured_reader.is_null());
 
+        // Verify consumed reader is no longer tracked
+        let free_result = unsafe { c2pa_free(reader as *const c_void) };
+        assert_eq!(free_result, -1);
+
         // Verify we can read the manifest
         let json = unsafe { c2pa_reader_json(configured_reader) };
         assert!(!json.is_null());
@@ -2840,7 +2930,6 @@ mod tests {
             c2pa_free(settings as *mut c_void);
             c2pa_free(context as *mut c_void);
             c2pa_free(configured_reader as *mut c_void);
-            // Original reader was consumed by with_stream, don't free it
         };
     }
 
@@ -3314,10 +3403,13 @@ verify_after_sign = true
         let context = unsafe { c2pa_context_builder_build(builder) };
         assert!(!context.is_null());
 
+        // Verify consumed builder is no longer tracked
+        let free_result = unsafe { c2pa_free(builder as *const c_void) };
+        assert_eq!(free_result, -1);
+
         unsafe {
             c2pa_free(settings as *mut c_void);
             c2pa_free(context as *mut c_void);
-            // builder is now invalid - don't free it
         };
     }
 
@@ -3565,7 +3657,10 @@ verify_after_sign = true
         let new_builder = unsafe { c2pa_builder_with_definition(builder, new_manifest.as_ptr()) };
         assert!(!new_builder.is_null(), "Should return new builder");
 
-        // Original builder pointer is now invalid, use new_builder
+        // Verify consumed builder is no longer tracked
+        let free_result = unsafe { c2pa_free(builder as *const c_void) };
+        assert_eq!(free_result, -1);
+
         unsafe {
             c2pa_free(new_builder as *mut c_void);
         }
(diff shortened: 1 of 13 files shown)


sdk/examples/data_hash.rs
```diff
@@ -11,207 +11,120 @@
...
     dh.set_hash(hash);
 
-    // tell SDK to fill in the hash and sign to complete the manifest
-    let final_manifest = builder.sign_data_hashed_embeddable(signer.as_ref(), &dh, "image/jpeg")?;
+    // Remove the old placeholder DataHash and add the updated one
+    builder
+        .definition
+        .assertions
+        .retain(|a| !a.label.starts_with(DataHash::LABEL));
+    builder.add_assertion(DataHash::LABEL, &dh)?;
```
ok-nick: I'm a little bit concerned about this approach here, it feels like we are trying to fit a workflow in `Builder` that doesn't really need all the capabilities of a `Builder`.

sdk/src/builder.rs
```diff
@@ -1935,6 +1938,93 @@ impl Builder {
...
+    /// # Returns
+    /// * The raw JUMBF bytes of the `c2pa_manifest` placeholder.
+    ///
+    /// # Errors
+    /// * Returns an [`Error`] if the placeholder cannot be created or if no hash assertion exists.
+    pub fn placeholder(&self, format: &str) -> Result<Vec<u8>> {
+        // Check that a hash assertion exists before proceeding
+        let has_data_hash = self.find_assertion::<DataHash>(DataHash::LABEL).is_ok();
+        let has_bmff_hash = self.find_assertion::<BmffHash>(BmffHash::LABEL).is_ok();
+
```
ok-nick: Can this function take a placeholder hash binding assertion as a parameter so we force them to think about it at compile time?

sdk/src/builder.rs
```diff
@@ -1935,6 +1938,93 @@ impl Builder {
...
+    /// This returns raw JUMBF bytes for the placeholder manifest.
+    /// Use [`Builder::composed_manifest`] to compose for a specific format if needed.
+    /// The signer is obtained from the Builder's context.
+    ///
+    /// **Important:** You must add a hash assertion (DataHash or BmffHash) to the Builder
+    /// before calling this method. The hash doesn't need final values, but should be sized
+    /// large enough to hold the final hash (e.g., with enough exclusions).
+    ///
+    /// # Workflow
+    /// 1. Add a placeholder hash assertion to the Builder with proper sizing
```
ok-nick: Maybe we don't need to do this for them right now, but it would be very helpful if we had methods on our hash bindings that can calculate the placeholder hash sizes given a stream.

sdk/src/builder.rs
```diff
@@ -1935,6 +1938,93 @@ impl Builder {
...
+    /// This method takes a placeholder JUMBF created by [`Builder::placeholder`], updates
+    /// any hash assertions (DataHash, BmffHash) from the Builder's current state, and signs
+    /// the manifest. This supports dynamic assertions if configured in the signer.
+    ///
+    /// # Workflow
+    /// 1. Call [`Builder::placeholder`] to create a placeholder manifest
+    /// 2. Embed the placeholder into your asset
+    /// 3. Calculate the hash of the asset (excluding the placeholder)
+    /// 4. Update the hash assertion in the Builder: `builder.add_assertion(DataHash::LABEL, &updated_hash)`
+    /// 5. Call this method to sign: `builder.sign_placeholder(&placeholder, signer, format)`
```
ok-nick: I've been thinking about this approach and maybe we can change the API to "force" users into doing the right thing at compile time.

Here's what I'm thinking: what if we made `Builder::placeholder` return a `Placeholder` struct and the context gets passed along to it from the builder. Then, we can have a method like `Placeholder::sign(&self, format, updated_hard_binding_assertion)`. In this approach, users are forced to do the correct thing and there isn't much room for mistakes.

There can also be a method such as `Placeholder::as_bytes` so it can be embedded in the asset for calculating offsets. Although if it's a box hash, you can completely skip that step.

sdk/src/store.rs
```diff
@@ -2470,6 +2472,115 @@ impl Store {
...
+    /// # Arguments
+    /// * `signer` - The signer to use.
+    /// * `settings` - The settings to use.
+    /// # Returns
+    /// * The signed manifest bytes.
+    /// # Errors
+    /// * Returns an [`Error`] if the placeholder cannot be signed.
+    pub fn sign_manifest(&mut self, signer: &dyn Signer, settings: &Settings) -> Result<Vec<u8>> {
+        let pc = self.provenance_claim_mut().ok_or(Error::ClaimEncoding)?;
+
```
ok-nick, line 2382: Is this function used anywhere?

sdk/examples/data_hash.rs
gpeacock: where else would you put it?

sdk/src/builder.rs
gpeacock: I think I see what you mean. Let me think about that. I'm not eager to introduce more structures to the C_api but this  may make more sense than extending the builder.

sdk/src/builder.rs
gpeacock: Yes we could, but that means another pass through the asset

gpeacock: If we move to the asset_io model, we can preserve the asset structure and reuse it, avoiding extra passes. This is more or less a stopgap until we get there.

sdk/src/builder.rs
gpeacock: That would require the user to know how to generate a hash binding placeholder, and know which one to create. That's most of what this method is for.

ok-nick: In this implementation, doesn't the user create the hash binding placeholder anyways and add it to the builder?

sdk/examples/data_hash.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1847#discussion_r2799632769

tmathern:
Ask: `They change the model`: does it mean this will be a breaking change compared to previous way of doing things?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -27,6 +27,10 @@ use serde::{
 use serde_bytes::ByteBuf;
 use sha2::{Digest, Sha256, Sha384, Sha512};
 
+/// Maximum number of mdat boxes supported in the Merkle placeholder.
+/// Practical BMFF files rarely have more than 4 mdat boxes.
+const MAX_MDAT_BOXES: usize = 4;
```
tmathern, line 32: Wasn't there a setting relating to that? (Maybe I dreamt it, or I am confusing it with the other settings)

sdk/src/builder.rs
gpeacock: I realized that all i really needed to track was the size of the placeholder once calculated, so we no longer need to use the placeholder as a parameter for signing. I think I've covered all the variations in workflow now. Please let me know if the updated design addresses your issues.

sdk/src/store.rs
gpeacock: this has been replaced with other methods.

sdk/src/builder.rs
gpeacock: no, we can add it for them, it is still an option to provide your own.

sdk/src/builder.rs
tmathern: (Maybe for the next API layer, we should go with state machines; the state transitions could enforce similar, albeit only at runtime)

docs/embeddable-api.md
```diff
@@ -0,0 +1,195 @@
+# Embeddable signing API
+
+The embeddable signing API gives callers explicit control over how a C2PA manifest is embedded into an asset. Instead of handing both the source and destination streams to `Builder::sign()` and letting the SDK manage everything, you drive each step: create a placeholder, embed it yourself, hash the asset, sign, then patch the manifest in place.
+
+This is a new, more generic api that replaces the older `Builder` methods. These will soon be deprecated:
```
tmathern, line 5: Ack'ed. 
For C++ bindings, they had dedicated tests exercising the APIs here: https://github.com/contentauth/c2pa-c/blob/vNext/tests/embeddable.test.cpp

docs/embeddable-api.md
```diff
@@ -0,0 +1,195 @@
...
+## Concepts
+
+### Hard-binding modes
+
+The embeddable API supports three hard-binding strategies, selected automatically based on format and settings:
+
+| Mode | Assertion | Formats | Requires placeholder |
+|------|-----------|---------|----------------------|
+| DataHash | `DataHash` | JPEG, PNG, GIF, WebP, and others | Yes |
+| BmffHash | `BmffHash` | MP4, video (BMFF containers) | Yes |
```
tmathern:
```suggestion
| BmffHash | `BmffHash` | MP4, video (BMFF containers), AVIF, HEIF/HEIC | Yes |
```
(I believe they're based on BMFF, so are they in this category too?)

docs/embeddable-api.md
```diff
@@ -0,0 +1,195 @@
...
+
+// No placeholder step. Hash the original asset directly.
+let mut source = std::fs::File::open("input.jpg")?;
+builder.update_hash_from_stream("image/jpeg", &mut source)?;
+
+// Sign — returns composed bytes at their natural size.
+let manifest_bytes = builder.sign_embeddable("image/jpeg")?;
+
+// Append manifest_bytes as a new independent chunk in the asset.
+// The exact mechanism depends on the format handler used by your embedding code.
```
tmathern, line 192: That needs to be made super obvious in the docs! (Warning on top?)

sdk/src/builder.rs
```diff
@@ -1933,6 +1962,513 @@ impl Builder {
...
+    /// The chunk size is read from the Builder's [`Context`]
+    /// (`core.merkle_tree_chunk_size_in_kb` must be set).
+    ///
+    /// # Arguments
+    /// * `leaf_hashes` - Per-mdat leaf hashes: `leaf_hashes[mdat][chunk] = hash_bytes`
+    ///
+    /// # Errors
+    /// * Returns an [`Error`] if no BmffHash assertion exists (call [`Builder::placeholder`] first)
+    /// * Returns an [`Error`] if `core.merkle_tree_chunk_size_in_kb` is not set in settings
+    pub fn set_bmff_mdat_hashes(&mut self, leaf_hashes: Vec<Vec<Vec<u8>>>) -> Result<()> {
```
tmathern, line 2120: leaf_hashes: Vec<Vec<Vec<u8>>>, can we make this a type?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1659,6 +1659,253 @@ pub unsafe extern "C" fn c2pa_builder_sign_data_hashed_embeddable(
...
+/// # Errors
+/// Returns -1 on error (call c2pa_error() for the message).
+/// In direct mode, also returns -1 if no valid hard binding assertion exists.
+/// On success, returns the byte length of the signed manifest.
+///
+/// # Safety
+/// Reads from NULL-terminated C strings.
+/// The returned bytes MUST be released by calling c2pa_free.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_builder_sign_embeddable(
```
tmathern, line 1844: The signature implies the signer now is in the context.
(Just a note, because that changes compared to how related APIs worked before, with a signer explicitly passed in).

Sidenote: Context is on purpose a flexible API. Should we make sure a signer is in the context here?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1659,6 +1659,253 @@ pub unsafe extern "C" fn c2pa_builder_sign_data_hashed_embeddable(
...
+/// Returns 0 on success, -1 on error (call c2pa_error() for the message).
+/// Fails if `core.merkle_tree_chunk_size_in_kb` is not set or if no BmffHash assertion exists.
+///
+/// # Safety
+/// All pointers must remain valid for the duration of the call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_builder_set_bmff_mdat_hashes(
+    builder_ptr: *mut C2paBuilder,
+    hashes_ptr: *const c_uchar,
+    hash_size: usize,
```
tmathern, line 1983: If possible, I'd suggest avoiding usize in favor of fixed-size types, so that we don't have variations on 32bits vs 64 bits platforms. (Flagging here because there are 3 of those types in this signature, but suggestion is valid for the whole C API file).

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1659,6 +1659,253 @@ pub unsafe extern "C" fn c2pa_builder_sign_data_hashed_embeddable(
...
+    let chunk_counts_slice = std::slice::from_raw_parts(chunk_counts, mdat_count);
+    let total_chunks: usize = chunk_counts_slice.iter().sum();
+
+    if hash_size == 0 || total_chunks == 0 {
+        use crate::CimplError;
+        CimplError::null_parameter("hash_size or total chunk count is zero").set_last();
+        return -1;
+    }
+
+    let all_hashes = std::slice::from_raw_parts(hashes_ptr, total_chunks * hash_size);
```
tmathern: Concern due to usize: Can this overflow especially on 32 bits platforms? (Because of the multiplciation)

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1659,6 +1659,253 @@ pub unsafe extern "C" fn c2pa_builder_sign_data_hashed_embeddable(
...
+    format: *const c_char,
+    manifest_bytes_ptr: *mut *const c_uchar,
+) -> i64 {
+    ptr_or_return_int!(manifest_bytes_ptr);
+    let builder = deref_mut_or_return_int!(builder_ptr, C2paBuilder);
+    let format = cstr_or_return_int!(format);
+    let result = builder.placeholder(&format);
+    let manifest_bytes = ok_or_return_int!(result);
+    let len = manifest_bytes.len() as i64;
+    if !manifest_bytes_ptr.is_null() {
```
tmathern, line 1809: How could this be null if ptr_or_return_int!(manifest_bytes_ptr); already runs above?

sdk/src/builder.rs
```diff
@@ -1933,6 +1962,513 @@ impl Builder {
...
+    /// placeholder step entirely and use the direct-sign workflow
+    /// ([`Builder::sign_embeddable`] Mode 2) instead.
+    ///
+    /// The method also returns `false` when the builder already contains a [`BoxHash`]
+    /// assertion (regardless of `prefer_box_hash`), since the caller has explicitly
+    /// opted into the box-hash path.
+    ///
+    /// # Arguments
+    /// * `format` — MIME type or file extension of the target asset.
+    pub fn needs_placeholder(&self, format: &str) -> bool {
```
tmathern, line 1983: I think a function returning true/false for this is also neede in the C FFI so the API can be fully leveraged by end users to make decisions, but either it is not there, or I didn't see it. Wdyt?

c2pa_c_ffi/src/c_api.rs
gpeacock: The call will fail with an error if no signer is in the context.

c2pa_c_ffi/src/c_api.rs
gpeacock: Actually I think the right thing here is to allow manifest_bytes_ptr to be null and just not return the bytes in that case. In many cases all that is needed is the size to reserve. It is extra overhead to write bytes that are going to be replaced.

sdk/src/builder.rs
gpeacock: I added one, but it returns an int with 0 - false, 1 - true, and -1 for errors. I think this is the better way in the C api to deal with boolean results that can have errors.

sdk/src/assertions/bmff_hash.rs
gpeacock: There's a core.merkle_tree_chunk_size_in_kb,  but that's a different beast.

c2pa_c_ffi/src/c_api.rs
gpeacock: There are usize values throughout the c_ffi_api. I agree we should probably change them, but all at the same time. There are different points of view on this.

c2pa_c_ffi/src/c_api.rs
gpeacock: good point, I put in a safely check for overflow on multiplication and it will return an error.

sdk/src/builder.rs
gpeacock: this will likely change with the work that Maurice is doing so I'll wait on that.

PR #1852, opened 2026-02-12
feat!: Pass HTTP resolvers to `Signer::send_timestamp_request`

### LINKED ISSUE #1816: Singer/AsyncSigner send_timestamp_request() method is not called for timestamping.
### Component

c2pa-rs

### If Other, please specify

_No response_

### Version

v0.75.13

### Platform

Linux

### What happened?

In Builder::sign() function
```Rust
        if let Some(tsa_url) = signer.time_authority_url() {
            if _sync {
                self.maybe_add_timestamp(&tsa_url, &mut claim)?;
            } else {
                self.maybe_add_timestamp_async(&tsa_url, &mut claim).await?
            }
        }

``` 
But this code bypasses the siger.send_timestamp_request() so even if someone implements it. It wont be called.  

The maybe_add_timestamp function of Builder should use  signer.send_timestamp_reqeust()

ok-nick:
This PR builds off the changes in #1846 to pass HTTP resolvers to the `Signer::send_timestamp_request` function. Our current code creates a new HTTP resolver on the fly for timestamping and doesn't respect the one contained in the resolver, this PR fixes that. It is step one to solving #1816, where we change the timestamp assertion code to call the signer function.

Note, some of the changes in #1846 are done here as well, so let's get that PR through first.

* Partially resolves #1816

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -58,7 +58,7 @@ use crate::{
     error::{Error, Result},
     hash_utils::{hash_by_alg, vec_compare, verify_by_alg},
     hashed_uri::HashedUri,
-    http::{AsyncHttpResolver, SyncHttpResolver},
+    http::{AsyncHttpResolver, SyncGenericResolver, SyncHttpResolver},
     jumbf::{
         self,
         boxes::*,
@@ -2425,7 +2425,13 @@ impl Store {
 
                 // Get pc again
                 let pc = self.provenance_claim().ok_or(Error::ClaimEncoding)?;
-                let sig = self.sign_claim(pc, signer, signer.reserve_size(), settings)?;
+                let sig = self.sign_claim(
+                    pc,
+                    signer,
+                    signer.reserve_size(),
+                    settings,
+                    &SyncGenericResolver::new(),
+                )?;
                 let sig_placeholder = Store::sign_claim_placeholder(pc, signer.reserve_size());
 
                 if sig_placeholder.len() != sig.len() {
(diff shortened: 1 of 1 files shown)


sdk/src/crypto/raw_signature/signer.rs
```diff
@@ -268,12 +268,16 @@ impl AsyncTimeStampProvider for AsyncRawSignerWrapper {
...
+    // TODO: this feels hacky but the problem is that we can't pass an AsyncHttpResolver as a SyncHttpResolver
+    //       We create an AsyncRawSignerWrapper from the actual crypto signers which is never supposed to
+    //       implement its own timestamping code anyways(?) because of that I thikn we can remove most of this trait impl
+    // async fn send_time_stamp_request(
+    //     &self,
+    //     http_resolver: &dyn AsyncHttpResolver,
+    //     message: &[u8],
+    // ) -> Option<Result<Vec<u8>, TimeStampError>> {
+    //     self.0.send_time_stamp_request(message)
+    // }
```
ok-nick: TODO: confirm this

sdk/src/crypto/raw_signature/signer.rs
```diff
@@ -268,12 +268,16 @@ impl AsyncTimeStampProvider for AsyncRawSignerWrapper {
         self.0.time_stamp_request_body(message)
     }
 
-    async fn send_time_stamp_request(
-        &self,
-        message: &[u8],
-    ) -> Option<Result<Vec<u8>, TimeStampError>> {
-        self.0.send_time_stamp_request(message)
-    }
+    // TODO: this feels hacky but the problem is that we can't pass an AsyncHttpResolver as a SyncHttpResolver
```
gpeacock: can you just delete all this now?

ok-nick: Yes I think so, I left another review comment to make sure I resolve this before merge. I want to understand more of the problem.

sdk/src/crypto/raw_signature/signer.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1852#discussion_r2801019198

sdk/src/crypto/raw_signature/signer.rs
ok-nick: Ok, it is not needed in this scenario and I left a note explaining the behavior and reasoning.

PR #1855, opened 2026-02-12
feat!: call `Signer::send_timestamp_request` for timestamp assertions

### LINKED ISSUE #1816: Singer/AsyncSigner send_timestamp_request() method is not called for timestamping.

ok-nick:
Builds off the work in #1846 and #1852 to call `Signer::send_timestamp_request` for timestamp assertions using the HTTP resolver supplied in the context.

Note #1846 and #1852 should be merged before merging this PR.

* Closes #1816


sdk/src/signer.rs
```diff
@@ -351,29 +339,27 @@ impl RawSigner for Box<dyn Signer> {
...
-        self.as_ref().timestamp_request_headers()
+        self.timestamp_request_headers()
     }
 
     fn time_stamp_request_body(
         &self,
         message: &[u8],
     ) -> std::result::Result<Vec<u8>, TimeStampError> {
-        Ok(self.as_ref().sign(message)?)
+        Ok(self.timestamp_request_body(message)?)
```
ok-nick, line 360: Why was `time_stamp_request_body` calling `Signer::sign`?

sdk/src/utils/ephemeral_signer.rs
```diff
@@ -299,8 +314,14 @@ mod tests {
         let mut settings = Settings::default();
         settings.verify.verify_trust = false;
 
-        let cose_bytes = sign_claim(&claim_bytes, &signer, signer.reserve_size(), &settings)
-            .expect("sign_claim with EphemeralSigner");
+        let cose_bytes = sign_claim(
+            &claim_bytes,
+            &signer,
+            signer.reserve_size(),
+            &settings,
```
gpeacock: At this point it feels like we should be passing context here since it includes the signer (and reserve size, settings and resolvers.

sdk/src/cose_sign.rs
```diff
@@ -117,13 +120,15 @@ pub fn sign_claim(
     box_size: usize,
     time_stamp_storage: TimeStampStorage,
     settings: &Settings,
+    http_resolver: &impl AsyncHttpResolver,
 ))]
 pub(crate) fn cose_sign(
     signer: &dyn Signer,
```
gpeacock, line 125: I think cose sign just needs to take context at this point, we are just passing the same parameters seprately.

gpeacock (changes requested):
Not critical but I think we should be passing context instead of adding more parameters. I also have a fuzzy feeling that there is a better way to handle this, but I know we already went over that.

gpeacock (changes requested):
Not critical but I think we should be passing context instead of adding more parameters. I also have a fuzzy feeling that there is a better way to handle this, but I know we already went over that.

sdk/src/cose_sign.rs
ok-nick: This isn't feasible in many cases since we may adjust settings on the fly and we can't do that with an immutable context (like those wrapped in an Arc). For example:
https://github.com/contentauth/c2pa-rs/blob/b3787364c67856a571c198dc9b7951af3613fbe0/sdk/src/store.rs?plain=1#L564-L585

sdk/src/utils/ephemeral_signer.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1855#discussion_r2976720332

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -411,6 +414,7 @@ pub(crate) fn fetch_and_check_ocsp_response(
     ctp: &CertificateTrustPolicy,
     tst_info: Option<&TstInfo>,
     validation_log: &mut StatusTracker,
+    verify_timestamp_trust: bool,
```
mauricefisher64, line 417: I we have the context why do we need to pass this in?

ok-nick: @mauricefisher64 In some parts of the code we dynamically modify settings to pass down the stack. This fixes an issue with the existing plumbing where we weren't respecting that modification since we can't mutate `&Context`.

sdk/src/assertions/timestamp.rs
```diff
@@ -60,17 +65,19 @@ impl TimeStamp {
...
+    // [`CoseSign1::signature`][coset::CoseSign1::signature].
     #[async_generic(async_signature(
         &mut self,
         tsa_url: &str,
         manifest_id: &str,
         signature: &[u8],
         http_resolver: &impl AsyncHttpResolver,
     ))]
-    pub(crate) fn refresh_timestamp(
+    pub fn refresh_timestamp(
```
scouten-adobe, line 81: Would we want to consider deprecating the non-`Signer` versions here?

sdk/src/assertions/timestamp.rs
```diff
@@ -99,7 +106,7 @@ impl TimeStamp {
         message: &[u8],
         http_resolver: &impl AsyncHttpResolver,
     ))]
-    pub(crate) fn send_timestamp_token_request(
+    pub fn send_timestamp_token_request(
```
scouten-adobe, line 110: Same question: Deprecate?

sdk/src/crypto/cose/ocsp.rs
scouten-adobe: Blocker: Seems to me that should be addressed in the upstream code by cloning the context and doing the mutation there instead.

sdk/src/crypto/time_stamp/provider.rs
```diff
@@ -47,32 +54,25 @@ pub trait TimeStampProvider {
...
 
     /// Request a [RFC 3161] time stamp over an arbitrary data packet.
     ///
-    /// The default implementation will send the request to the URL
-    /// provided by [`Self::time_stamp_service_url()`], if any.
+    /// Implement this function to provide custom networking for timestamp
+    /// requests. The default implementation returns `None`.
+    ///
+    /// If this method returns `None` and [`TimeStampProvider::time_stamp_service_url`] is
+    /// set, the SDK falls back to its built-in networking implementation.
```
scouten-adobe: Would you document where that fallback occurs? From reading this code, it isn't obvious.

sdk/src/crypto/time_stamp/provider.rs
```diff
@@ -103,42 +109,81 @@ pub trait AsyncTimeStampProvider: MaybeSync {
...
     /// Request a [RFC 3161] time stamp over an arbitrary data packet.
     ///
-    /// The default implementation will send the request to the URL
-    /// provided by [`Self::time_stamp_service_url()`], if any.
+    /// `message` is a preliminary hash of the claim.
+    ///
+    /// Implement this function to provide custom networking for timestamp
+    /// requests. The default implementation returns `None`.
+    ///
+    /// If this method returns `None` and [`AsyncTimeStampProvider::time_stamp_service_url`] is
```
scouten-adobe: Same comment about fallback as at line 70.

sdk/src/crypto/time_stamp/provider.rs
scouten-adobe: (Later: I see that it follows later in this file, but for those of us with short attention spans … 😆 )

sdk/src/signer.rs
scouten-adobe: Yeah, that's weird.

scouten-adobe (changes requested):
Looks really good. I have the same question that Maurice does about passing `verify_timestamp_trust` through and also a few questions about documentation; otherwise 👍.

sdk/src/crypto/cose/ocsp.rs
ok-nick: @scouten-adobe The context is wrapped in an Arc at the higher level layers of our code so it isn't clonable and mutatable as a ref. I'm not a fan of boolean parameters, so I think splitting this function into a separate verify timestamp makes more sense, but that will require quite a bit of refactoring which I don't think belongs in this PR.

sdk/src/assertions/timestamp.rs
ok-nick: We should actually consider removing the `Signer`-variants rather than the other way around. I discussed this briefly with @gpeacock, that it doesn't seem necessary to pass a `Signer` when you only want to fetch a timestamp. This was before contexts and builders, so I suspect there is a different design we can do that doesn't involve the signer, though that would require a much larger change.

sdk/src/assertions/timestamp.rs
ok-nick: See https://github.com/contentauth/c2pa-rs/pull/1855#discussion_r3024751354

sdk/src/signer.rs
ok-nick: Ok I'm going to keep this change then since it makes the most sense.

scouten-adobe (approved):

PR #1858, opened 2026-02-13
feat: Enable self-signed ephemeral certs on Wasm build

scouten-adobe:
(Removes dependency on rcgen crate.)

### DIFF
--- a/sdk/src/utils/ephemeral_cert.rs
+++ b/sdk/src/utils/ephemeral_cert.rs
@@ -65,6 +65,11 @@
 //! we don't know which part of our cert triggers it. The cross-check test
 //! treats these as a known platform quirk and skips the verify assertion
 //! instead of failing.
+//!
+//! On some environments (e.g. older or minimal OpenSSL/LibreSSL, or aarch64
+//! Linux with a build that omits Ed25519), `openssl verify` fails with
+//! "unsupported algorithm" or "unable to get certs public key" because our
+//! certs use Ed25519. The test skips the verify assertion in that case.
 
 use chrono::Utc;
 use ed25519_dalek::{Signer, SigningKey, VerifyingKey};
@@ -576,7 +581,8 @@ mod tests {
     }
 
     /// True when `openssl verify` failed due to a known platform/version quirk so we skip
-    /// instead of failing the test (e.g. strict extension handling, ASN.1 decoding differences).
+    /// instead of failing the test (e.g. strict extension handling, ASN.1 decoding differences,
+    /// or OpenSSL/LibreSSL build without Ed25519 support).
     #[cfg(any(target_os = "linux", target_os = "macos"))]
     fn is_openssl_verify_skip(stderr: &str) -> bool {
         let s = stderr;
@@ -586,6 +592,8 @@ mod tests {
             || s.contains("nested asn1")
             || s.contains("header too long")
             || s.contains("bad object header")
+            || s.contains("unsupported algorithm")
+            || s.contains("unable to get certs public key")
     }
 
     /// Cross-check generated CA and EE certificates with OpenSSL.


sdk/src/crypto/cose/certificate_trust_policy.rs
```diff
@@ -456,20 +453,14 @@ pub enum CertificateTrustError {
     InternalError(String),
 }
 
-#[cfg(all(
-    feature = "openssl",
-    not(all(feature = "rust_native_crypto", target_arch = "wasm32"))
-))]
```
ok-nick, line 462: Btw, the reason I included `rust_native_crypto` here is so that when `rust_native_crypto` and `openssl` are both enabled, it only compiles the `rust_native_crypto` code. I'm not sure if it's a problem here but we should make sure that still works. I typically have `--all-features` set by default in rust-analyzer and also use it for testing which is a common pattern.

sdk/Cargo.toml
```diff
@@ -111,6 +111,7 @@ const-oid = { version = "0.9.6", optional = true }
 coset = "0.4.0"
 der = { version = "0.7.9", features = ["alloc"] }
 extfmt = "0.2.0"
+getrandom = { version = "0.3", default-features = false, features = ["std"] }
```
ok-nick: WASM currently depends on `0.3.4`, should we match that up here as well?

sdk/src/builder.rs
```diff
@@ -1134,17 +1132,9 @@ impl Builder {
...
-
-            #[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
-            {
-                let c2pa_data = self.working_store_sign()?;
-                stream.write_all(&c2pa_data)?;
-                return Ok(());
-            }
+            let c2pa_data = self.working_store_sign()?;
+            stream.write_all(&c2pa_data)?;
+            return Ok(());
```
ok-nick, line 1137: Nice!

sdk/src/crypto/cose/certificate_trust_policy.rs
scouten-adobe: FWIW my typical "is it good" script is this …

```
at 09:31:14 ❯ alias cft
cft='cargo +nightly fmt --all && cargo test --all-features'
```

… and it passes with the build as it stands. Can you show me a different invocation I should use that would demonstrate the problem you're positing?

sdk/Cargo.toml
tmathern: I think you should pin to exact versions for WASM, to avoid surprises

scouten-adobe: Done.

sdk/src/crypto/cose/certificate_trust_policy.rs
ok-nick: At a higher-level we have code that conditionally uses the `openssl` or `rust_native_crypto` path which is where I was concerned, but it looks like that isn't affected here.

I don't have any strong opinions here although I lean towards leaving the flag in to avoid compiling the extra code when it's not used.

sdk/src/utils/ephemeral_cert.rs
```diff
@@ -0,0 +1,377 @@
...
+    let not_before = now - chrono::Duration::days(1);
+    let not_after = now + chrono::Duration::days(365);
+
+    Validity {
+        not_before: Time::Utc(not_before),
+        not_after: Time::Utc(not_after),
+    }
+}
+
+fn subject_key_identifier_ext(public_key_der: &[u8]) -> Result<Extension> {
```
tmathern, line 190: Is this right, to be the whole key and not a part of it? Would this give the same result as what rcgen did?

sdk/src/utils/ephemeral_cert.rs
```diff
@@ -0,0 +1,377 @@
...
+    let hash = Sha1::digest(public_key_der);
+
+    Ok(Extension {
+        extn_id: oid(SUBJECT_KEY_ID_OID)?,
+        critical: false,
+        extn_value: OctetString::from(hash.to_vec()),
```
tmathern, line 196: I am confused by encoding. Why is it different than other locations (der::encode vs OctetString )?
```
extn_value: rasn::der::encode(&BasicConstraints {
                ca: true,
                path_len_constraint: None,
            })
```
Why one only OctetString, and the other DER?

sdk/src/utils/ephemeral_cert.rs
```diff
@@ -0,0 +1,377 @@
...
+                path_len_constraint: None,
+            })
+            .map_err(|e| Error::OtherError(Box::new(e)))?
+            .into(),
+        },
+        Extension {
+            extn_id: oid(KEY_USAGE_OID)?,
+            critical: true,
+            // keyCertSign (5), cRLSign (6), digitalSignature (0)
+            extn_value: rasn::der::encode(&BitString::from_slice(&[0x86]))
```
tmathern, line 250: A favorite topic of mine: Edge cases: Does this properly handle null-bytes correctly? (Just checking).

sdk/src/utils/ephemeral_cert.rs
```diff
@@ -0,0 +1,377 @@
...
+
+fn fill_random(buf: &mut [u8]) -> Result<()> {
+    getrandom::fill(buf).map_err(|e| Error::OtherError(Box::new(e)))
+}
+
+/// Generate a new Ed25519 keypair using OS/browser RNG (works on Wasm via
+/// getrandom).
+fn generate_ed25519_keypair() -> Result<SigningKey> {
+    let mut seed = [0u8; 32];
+    fill_random(&mut seed)?;
```
tmathern, line 133: Does this need zeroization so the seed couldn't leak?
Since ephemeral I think it's acceptable not to.

sdk/src/utils/ephemeral_cert.rs
```diff
@@ -0,0 +1,377 @@
...
+    Ok(EphemeralCertChain {
+        ee_der,
+        ca_der,
+        ee_private_key_pem,
+    })
+}
+
+/// Encode a single certificate DER as PEM (CERTIFICATE block).
+pub fn der_to_pem(der: &[u8]) -> String {
+    pem::Pem::new("CERTIFICATE", der.to_vec()).to_string()
```
tmathern, line 525: Defensive here given the nature of the thing: Do we want to validate before encoding?

sdk/src/utils/ephemeral_cert.rs
```diff
@@ -0,0 +1,377 @@
...
+fn generate_ed25519_keypair() -> Result<SigningKey> {
+    let mut seed = [0u8; 32];
+    fill_random(&mut seed)?;
+    Ok(SigningKey::from_bytes(&seed))
+}
+
+fn dn_attr(oid_components: &[u64], value: &str) -> Result<AttributeTypeAndValue> {
+    let ps =
+        PrintableString::try_from(value.to_string()).map_err(|e| Error::OtherError(Box::new(e)))?;
+    let value_der = rasn::der::encode(&ps).map_err(|e| Error::OtherError(Box::new(e)))?;
```
tmathern, line 140: I see multiple variations of encoding use, I am not sure it they all match together, or match the spec, without confronting them to another implementation to verify it. (II thought I'd point out an idea for testing and verifying...). Which is why I'd suggest to test with rcgen side-by-side. We don't need to keep these tests (since they will trouble WASM), but we should have them at least now for supported platform to make sure we match against a somewhat reference implementation.

sdk/src/crypto/cose/certificate_trust_policy.rs
scouten-adobe: Reverted this change.

sdk/src/utils/ephemeral_cert.rs
scouten-adobe: Now cross-checking with openssl, which did point up one encoding issue. Believe that is now resolved.

tmathern (approved):
Pre-approving, pending certificate linting check.
(Also comments resolution, because I am curious).

sdk/src/utils/ephemeral_cert.rs
scouten-adobe: I think it doesn't matter here since the input is hard-coded, but I'll file a ticket for follow-up once I'm back. https://jira.corp.adobe.com/browse/CAI-11072

PR #1863, opened 2026-02-17
fix: correct validation of OCSP signature

mauricefisher64:
Fix selecting the incorrect handler for RSA signatures
Enforce explicit EKUs for timestamp and ocsp certificates
Check OCSP certs against trust list
Correctly ignore OCSP responses that have errors (these are a noop)

## Checklist


sdk/src/crypto/cose/ocsp.rs
```diff
@@ -315,20 +358,36 @@ fn check_stapled_ocsp_response(
...
     // If we get a valid response, validate the certs.
-    if ocsp_data.revoked_at.is_none() {
-        if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
-            // if the OCSP signing cert cannot be validated do not use this response
-            if check_end_entity_certificate_profile(
+    if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
+        // make sure this is an OCSP signing EKU
+        let mut new_ctp = ctp.clone();
+        new_ctp.clear_ekus();
+        new_ctp.add_valid_ekus("1.3.6 .1 .5 .5 .7 .3 .9".as_bytes()); // ocsp signing EKU
```
ok-nick: Did you mean to put spaces in the EKU?

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -387,10 +454,22 @@ pub(crate) fn fetch_and_check_ocsp_response(
...
 
     // If we get a valid response validate the certs.
-    if ocsp_data.revoked_at.is_none() {
-        if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
-            check_end_entity_certificate_profile(&ocsp_certs[0], ctp, validation_log, None)?;
+    if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
+        // make sure this is an OCSP signing EKU
+        let mut new_ctp = ctp.clone();
+        new_ctp.clear_ekus();
+        new_ctp.add_valid_ekus("1.3.6 .1 .5 .5 .7 .3 .9".as_bytes()); // ocsp signing EKU
```
ok-nick: Here too.

sdk/src/crypto/cose/ocsp.rs
ok-nick: It looks like the OIDs in the `oid!` macro are formatted by rustfmt with spaces and that got carried over here.

sdk/src/crypto/cose/ocsp.rs
mauricefisher64: Good catch

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -107,11 +113,49 @@ pub fn check_ocsp_status(
...
                     ctp,
                     tst_info,
-                    validation_log,
+                    &mut ocsp_log,
                     context.settings(),
                 )
                 .await
+            };
+
+            // we only care about OCSP value log info the result is OK
```
tmathern:
```suggestion
            // we only care about OCSP value log info if the result is OK
```
I think that's how I understand it?

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -41,6 +41,8 @@ use crate::{
     },
 };
 
+const TIMESTAMP_OID_STR: &str = "1.3.6.1.5.5.7.3.8";
```
ok-nick, line 44: Maybe we can define this like the other oids using the `oid!` macro and then use `::as_bytes` when adding it to the `ctp`?

PR #1867, opened 2026-02-19
fix: Use cose_sign1 signature field for timestamp assertion verification

timmurphydev:
Timestamp assertions are displaying `timeStamp.mismatch` errors in the informational array on validation.
```
{
  "code": "timeStamp.mismatch",
  "url": "",
  "explanation": "timestamp message digest did not match: DigiCert SHA256 RSA4096 Timestamp Responder 2025 1"
}
```

It appears that verify_time_stamp() is getting passed the CoseSign1 structure bytes instead of just the CoseSign1 raw signature.

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -2074,15 +2074,17 @@ impl Store {
                 // we only use valid timestamps, otherwise just ignore
                 for (referenced_claim, time_stamp_token) in timestamp_assertion.as_ref() {
                     if let Some(rc) = svi.manifest_map.get(referenced_claim) {
-                        if let Ok(tst_info) = verify_time_stamp(
-                            time_stamp_token,
-                            rc.signature_val(),
-                            &self.ctp,
-                            validation_log,
-                            // no trust checks for leagacy timestamps
-                            rc.version() != 1,
-                        ) {
-                            svi.timestamps.insert(rc.label().to_owned(), tst_info);
+                        if let Ok(sign1) = rc.cose_sign1() {
+                            if let Ok(tst_info) = verify_time_stamp(
+                                time_stamp_token,
+                                &sign1.signature,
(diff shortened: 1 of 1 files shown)


sdk/src/store.rs
```diff
@@ -2074,15 +2074,17 @@ impl Store {
...
-                            &self.ctp,
-                            validation_log,
-                            // no trust checks for leagacy timestamps
-                            rc.version() != 1,
-                        ) {
-                            svi.timestamps.insert(rc.label().to_owned(), tst_info);
+                        if let Ok(sign1) = rc.cose_sign1() {
+                            if let Ok(tst_info) = verify_time_stamp(
+                                time_stamp_token,
+                                &sign1.signature,
```
mauricefisher64, line 2080: This is not right either.  I think the issue is because the payload to verify_time_stamp is different between v1 and v2 claims.  So the code need to pass the correct value based on that.  There is a function to do just that:  

fn validate_cose_tst_info(
    sign1: &coset::CoseSign1,
    data: &[u8],
    ctp: &CertificateTrustPolicy,
    verify_trust: bool,
) -> Result<TstInfo, CoseError>

The data param should be the same code from the original implementation rc.signature_val();

mauricefisher64: Good catch BTW

timmurphydev: Don't timestamp assertions always use the raw signature for the payload?

```
let signature = claim.cose_sign1()?.signature;
if _sync {
    timestamp_assertion.refresh_timestamp(
        tsa_url,
        &manifest_label,
        &signature,
        &self.context().resolver(),
    )?;
```

mauricefisher64: You are correct.

PR #1878, opened 2026-02-24
fix: Allow fine-grained control over thumbnails when signing

tmathern:
Allow fine-grained control over thumbnails when signing by using the none flags.
I think this worked before, some code changes must have broken/updated it.

Archives roundtrip support is in https://github.com/contentauth/c2pa-rs/pull/1879, because behavior is debatable when round-triping archives(and should be debated there, if/as needed).

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -2429,8 +2429,8 @@ impl Builder {
             let mut bmff_hash = self.find_assertion::<BmffHash>(BmffHash::LABEL)?;
             bmff_hash.set_bmff_version(stored_version);
 
-            // add in the Merkle leaf hashes that were collected via hash_bmff_mdat_bytes().
-            // we add any remainders (partially filled fixed-size leaves that are unhashed) as the last leaf of the Merkle leaves for that mdat_id
+            // Add in the Merkle leaf hashes that were collected via hash_bmff_mdat_bytes().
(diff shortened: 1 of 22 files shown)


sdk/src/ingredient.rs
```diff
@@ -1277,37 +1277,40 @@ impl Ingredient {
...
-                    } else {
-                        // add EmbeddedData thumbnail for v3 assertions in v2 claims
-                        let thumbnail = EmbeddedData::new(
-                            labels::INGREDIENT_THUMBNAIL,
-                            format_to_mime(&thumb_ref.format),
-                            data.into_owned(),
-                        );
-                        claim.add_assertion(&thumbnail)?
+            // Setting the format to "none" suppresses the ingredient thumbnail
+            if thumb_ref.format == "none" {
```
tmathern, line 1289: Here is the real change, that check

sdk/src/builder.rs
```diff
@@ -4178,6 +4178,135 @@ mod tests {
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
 
+    #[test]
```
tmathern, line 5352: Those are all "e2e" test to verify thumbnails behaviors

tmathern:
Deferred (for now).

PR #1879, opened 2026-02-24
fix: Thumbnails handling in archives roundtrips

tmathern:

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -6060,6 +6060,195 @@ mod tests {
         assert!(reader.active_manifest().is_some());
     }
 
+    #[test]
+    fn test_thumbnail_data_accessible_after_archive_roundtrip() {
+        // Verify thumbnail() data accessor (not just thumbnail_ref()) works
+        // correctly after archive round-trip for manifest and ingredients.
+        let mut builder = Builder::from_json(&simple_manifest_json()).unwrap();
+        let signer = test_signer(SigningAlg::Ps256);
+
+        builder
+            .add_ingredient_from_stream(
+                r#"{"title": "with_thumb"}"#,
+                "image/jpeg",
+                &mut Cursor::new(TEST_IMAGE),
+            )
+            .unwrap();
+
+        builder
+            .add_ingredient_from_stream(
+                r#"{"title": "no_thumb", "thumbnail": { "format": "none", "identifier": "none" }}"#,
+                "image/jpeg",
+                &mut Cursor::new(TEST_IMAGE_CLEAN),
+            )
+            .unwrap();
+
+        let mut builder = archive_round_trip(&mut builder);
+
+        let mut output = Cursor::new(Vec::new());
+        builder
+            .sign(
+                &signer,
+                "image/jpeg",
+                &mut Cursor::new(TEST_IMAGE_CLEAN),
+                &mut output,
+            )
+            .unwrap();
+
+        let reader = Reader::from_stream("image/jpeg", &mut output).unwrap();
+        let manifest = reader.active_manifest().unwrap();
+
+        // Manifest thumbnail depends on add_thumbnails feature; if present,
+        // verify it contains real data.
+        if let Some((fmt, data)) = manifest.thumbnail() {
+            assert_ne!(fmt, "none");
+            assert!(
+                !data.is_empty(),
+                "manifest thumbnail data should not be empty"
+            );
+        }
+
+        // Thumbnailed ingredient should have accessible data
+        let with = manifest
+            .ingredients()
+            .iter()
+            .find(|i| i.title() == Some("with_thumb"))
+            .unwrap();
+        assert!(with.thumbnail_ref().is_some());
+        let (fmt, data) = with
+            .thumbnail()
+            .expect("thumbnailed ingredient should have data");
+        assert_ne!(fmt, "none");
+        assert!(!data.is_empty());
+
+        // Suppressed ingredient should return None from both accessors
+        let without = manifest
+            .ingredients()
+            .iter()
+            .find(|i| i.title() == Some("no_thumb"))
+            .unwrap();
+        assert!(without.thumbnail_ref().is_none());
+        assert!(without.thumbnail().is_none());
+    }
+
+    #[test]
+    fn test_double_archive_roundtrip_preserves_thumbnail_state() {
+        // Two consecutive archive round-trips should preserve thumbnail state.
+        let mut builder = Builder::from_json(&simple_manifest_json()).unwrap();
+        let signer = test_signer(SigningAlg::Ps256);
+
+        builder
+            .add_ingredient_from_stream(
+                r#"{"title": "has_thumb"}"#,
+                "image/jpeg",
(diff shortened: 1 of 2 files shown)


sdk/src/builder.rs
```diff
@@ -2440,6 +2451,18 @@ impl Builder {
         let mut claim = self.to_claim()?;
         claim.add_assertion(&box_hash)?;
 
+        // Preserve "none" thumbnail intent so it survives the archive round-trip.
+        // to_claim() skips adding a thumbnail assertion when format is "none",
+        // so we add a zero-byte marker here that from_store() can detect,
+        // otherwise we loose the "suppressive" info after serialization roundtrip
+        if let Some(thumb_ref) = self.definition.thumbnail.as_ref() {
```
tmathern, line 2983: I don't know how else we can have the thumbnails suppression info survive? Somehow we need to know why there is/was none?

sdk/src/builder.rs
```diff
@@ -1187,7 +1187,18 @@ impl Builder {
...
             let mut reader = Reader::from_shared_context(&self.context);
             reader.with_store(store, &mut validation_log)?;
-            reader.into_builder()
+            let mut builder = reader.into_builder()?;
+
+            // Keep ingredient as configured during initial add:
+            // if an ingredient has no_thumbnail here, it was very likely
+            // intentionally skipped!
+            for ingredient in &mut builder.definition.ingredients {
+                if ingredient.thumbnail_ref().is_none() {
```
tmathern: This is a marker.

I don't know how else we can have the thumbnails suppression info survive? Somehow we need to know why there is/was none?

sdk/src/manifest.rs
```diff
@@ -555,13 +562,17 @@ impl Manifest {
...
                 label if label.starts_with(labels::CLAIM_THUMBNAIL) => {
                     let thumbnail = EmbeddedData::from_assertion(assertion)?;
-                    let id = to_assertion_uri(claim.label(), label);
-                    //let id = jumbf::labels::to_relative_uri(&id);
-                    manifest.thumbnail = Some(manifest.resources.add_uri(
-                        &id,
-                        &thumbnail.content_type,
-                        thumbnail.data,
-                    )?);
+                    if thumbnail.content_type == "none" {
```
tmathern, line 565: Here we reuse the marker

sdk/src/builder.rs
```diff
@@ -1187,7 +1187,18 @@ impl Builder {
 
             let mut reader = Reader::from_shared_context(&self.context);
             reader.with_store(store, &mut validation_log)?;
-            reader.into_builder()
+            let mut builder = reader.into_builder()?;
+
+            // Keep ingredient as configured during initial add:
+            // if an ingredient has no_thumbnail here, it was very likely
+            // intentionally skipped!
+            for ingredient in &mut builder.definition.ingredients {
```
gpeacock: Why aren't we doing this inside into_builder?

sdk/src/manifest.rs
```diff
@@ -87,7 +87,7 @@ pub struct Manifest {
     #[serde(default = "default_instance_id")]
     instance_id: String,
 
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "is_none_or_suppressed")]
```
gpeacock, line 90: doesn't this effectively strip out the "none" information? Is that desired here?

sdk/src/ingredient.rs
```diff
@@ -80,7 +80,7 @@ pub struct Ingredient {
     /// A thumbnail image capturing the visual state at the time of import.
     ///
     /// A tuple of thumbnail MIME format (for example `image/jpeg`) and binary bits of the image.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "is_none_or_suppressed")]
```
gpeacock, line 83: why do we want to strip the "none" information here?

sdk/src/builder.rs
gpeacock: I just meant that this logic could move into the reader.into_builder() method so it will always happen when we need it to happen.

sdk/src/builder.rs
```diff
@@ -2440,6 +2451,18 @@ impl Builder {
...
         claim.add_assertion(&box_hash)?;
 
+        // Preserve "none" thumbnail intent so it survives the archive round-trip.
+        // to_claim() skips adding a thumbnail assertion when format is "none",
+        // so we add a zero-byte marker here that from_store() can detect,
+        // otherwise we loose the "suppressive" info after serialization roundtrip
+        if let Some(thumb_ref) = self.definition.thumbnail.as_ref() {
+            if thumb_ref.format == "none" {
+                let marker =
+                    EmbeddedData::new(labels::CLAIM_THUMBNAIL, "none".to_string(), Vec::new());
```
gpeacock, line 2986: No matter how careful we are about this. I'm pretty sure we will see someone in the wild asking where the "none" thumbnail came from. But I'm at a loss for an alternative at the moment.

sdk/src/manifest.rs
gpeacock, line 90: One caution on this is that if one of these does bleed into a claim, we may not know it was there. I guess the detailed report will still show it.

sdk/src/manifest.rs
gpeacock, line 565: Something that occurs to me is that technically all we we use to flag these is the format. The id value of "none" isn't important. So could let it have a real id that points to a filler resource and it will a have the same effect. In other words, I don't think we need special handling here since it should work either way. But this is fine for consistency.

sdk/src/builder.rs
tmathern: That is a good catch. I can look into fixing this.

sdk/src/manifest.rs
tmathern: Yes! I don't want it in manifests if we just reload from the archive. THe none is a marker so we know about thumbnail wishes.
(A little bit like https://github.com/contentauth/c2pa-rs/pull/1878)

sdk/src/ingredient.rs
tmathern: Yes! I don't want it in manifests if we just reload from the archive. The none is a marker so we know about thumbnail wishes. I think it would make sense to keep track specifically of the wish for a thumbnail vs no thumbnail in the archive too.
(A little bit like https://github.com/contentauth/c2pa-rs/pull/1878)

sdk/src/builder.rs
tmathern: Yes, same. It would be cool if we could define settings per ingredient, I think that is where somehow we ened to go. I think if we'd serialize settings with the archives, maybe we could pull it off, since ingredient archives only contain the ingredient? But then the question becomes how do we handle it when we sign. That is where https://github.com/contentauth/c2pa-rs/pull/1878 comes handy (we broke the none trick in another update btw, when we did other settings I think.).

sdk/src/manifest.rs
tmathern: A "filter resource"?

PR #1882, opened 2026-02-24
docs: Separate docs on intents and archives

crandmck:

Per [Slack discussion](https://contentauthenticity.slack.com/archives/C0396MCDRFZ/p1771947699659409):

I split the existing `intents-and-archives.md` into two files: `intents.md` and `archives.md`.

From @DaveStein :
> People do not need to use intents to capture an archive. It feels like we'd want a section on intents, a section on archives, and maybe one that shows how to use both together

### DIFF
--- a/docs/archives.md
+++ b/docs/archives.md
@@ -172,12 +172,6 @@ builder.add_action(json!({
 }))?;
 ```
 
-### Ingredient libraries
-
-**Scenario:** Content library with pre-validated assets.
-
-**Solution:** ??
-
 ## FAQs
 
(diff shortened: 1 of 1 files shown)


docs/archives.md
```diff
@@ -271,6 +172,12 @@ builder.add_action(json!({
 }))?;
 ```
 
+### Ingredient libraries
+
+**Scenario:** Content library with pre-validated assets.
+
+**Solution:** ??
```
crandmck: Need content here, or delete this section.

Issue #1885, opened 2026-02-25
High memory and time for verifying specific images


dbellavista:
### Component

c2patool

### Version

v0.26.30

### Platform

macOS

### What happened?

When analyzing a particular image, the execution time and memory usage is unreasonably high (I had to interrupt it because of OOM)

```
$ /usr/bin/time -l -h -p c2patool evil.png -o manifest
^C^C^C^Ctime: command terminated abnormally
real 40.50
user 3.00
sys 7.74
         12265586688  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
             3367127  page reclaims
                 266  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
               56539  voluntary context switches
               81112  involuntary context switches
         68823663820  instructions retired
         37702768769  cycles elapsed
         34199658736  peak memory footprint
```

I cannot share the original image, but I can share the `manifest.json` extracted with an earlier version (**0.19.0**) that is able to extract the JSON using 50MB or RAM and 45second of processing:

[manifest.json.gz](https://github.com/user-attachments/files/25545158/manifest.json.gz)

### What did you expect to happen?

Memory usage should be kept at bay and also processing time shjoudl not be that high. We risk DoS attacks against validator platforms

bgon:
I can  reproduce the problem using this public test image:
https://www.androidauthority.com/wp-content/uploads/2025/12/AI-image-4-test.png

Versions 0.26.5–0.26.8 showed a dramatic improvement in processing time, but starting with v0.26.9, c2patool consistently runs into Out‑Of‑Memory conditions (even on a machine with 32 GB RAM). 

| Linux Version            | Real Time (s) | Error / OOM      |
|--------------------|----------------|------------------|
| c2patool-v0.19.0   | 23.645         | No               |
| c2patool-v0.20.3   | 22.544         | Error            |
| c2patool-v0.21.0   | 22.595         | Error            |
| c2patool-v0.22.0   | 23.029         | Error            |
| c2patool-v0.23.1   | 23.054         | Error            |
| c2patool-v0.24.0   | 23.961         | No               |
| c2patool-v0.26.1   | 24.395         | No               |
| c2patool-v0.26.5   | 24.658         | No               |
| c2patool-v0.26.6   | 0.314          | No               |
| c2patool-v0.26.7   | 0.314          | No               |
[...]

PR #1887, opened 2026-02-26
fix: remove exponential memory growth from nested claim reconstruction (#1885)

dcondrey:

Commit 5f04441e (PR #1686, v0.26.9) introduced a nested claim reconstruction loop in `Store::from_jumbf_impl` and an eager `populate_ingredient_manifest_data` call in `Reader::with_store` that together cause exponential memory growth for manifest chains with diamond topology. For a 2 MB JUMBF payload with 22 diamond-topology manifests, this produces 5+ GB RSS and OOM.

This PR removes both amplification sources and replaces the nested cloning approach with a flat ingredient store builder that walks ingredient assertions (the same pattern validation already uses), cloning each claim exactly once.

**store.rs**: Delete the "Reconstruct nested claim relationships" block in `from_jumbf_impl` that deep-cloned entire `Claim` trees into each parent's `ingredients_store`. After this change, deserialized claims keep empty `ingredients_store`, matching pre-v0.26.9 behavior.

**reader.rs**: Remove the eager `populate_ingredient_manifest_data` call from the read path. Add a `build_ingredient_store` helper that collects transitive ingredient claims in depth-first post-order and adds each to a new Store via `insert_restored_claim`. Use this in `into_builder()` and `to_ingredient()` to generate `manifest_data` on demand instead of eagerly for every manifest. Delete four now-unused recursive collection methods.

 **tests**: Add diamond topology tests (`test_diamond_topology_read`, `test_diamond_topology_into_builder_round_trip`) that verify the exact shape that triggers exponential growth.

## Checklist

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -1148,28 +1148,15 @@ impl Reader {
             path.push(claim_label.clone());
 
             for ing_assertion in claim.ingredient_assertions() {
-                match IngredientAssertion::from_assertion(ing_assertion.assertion()) {
-                    Ok(ingredient) => {
-                        let manifest_uri = ingredient
-                            .active_manifest
-                            .as_ref()
-                            .or(ingredient.c2pa_manifest.as_ref());
-                        if let Some(manifest_uri) = manifest_uri {
-                            let ingredient_label =
-                                Store::manifest_label_from_path(&manifest_uri.url());
-                            if let Some(ingredient_claim) = store.get_claim(&ingredient_label) {
-                                collect_flat(
-                                    store,
-                                    ingredient_claim,
-                                    ingredient_store,
-                                    visited,
-                                    path,
-                                )?;
-                            }
-                        }
-                    }
-                    Err(e) => {
-                        log::warn!("Failed to deserialize ingredient assertion: {e}");
+                let ingredient = IngredientAssertion::from_assertion(ing_assertion.assertion())?;
+                let manifest_uri = ingredient
+                    .active_manifest
+                    .as_ref()
+                    .or(ingredient.c2pa_manifest.as_ref());
+                if let Some(manifest_uri) = manifest_uri {
+                    let ingredient_label = Store::manifest_label_from_path(&manifest_uri.url());
+                    if let Some(ingredient_claim) = store.get_claim(&ingredient_label) {
+                        collect_flat(store, ingredient_claim, ingredient_store, visited, path)?;
                     }
                 }
             }


sdk/src/reader.rs
```diff
@@ -1228,6 +1106,62 @@ impl Reader {
...
+
+            // Cycle detection
+            if path.iter().any(|p| p == &claim_label) {
+                return Ok(());
+            }
+
+            path.push(claim_label.clone());
+
+            for ing_assertion in claim.ingredient_assertions() {
+                if let Ok(ingredient) =
```
tmathern: warn on the error branch? This could silently skip ignredients.

sdk/src/reader.rs
```diff
@@ -1228,6 +1106,62 @@ impl Reader {
...
+        let mut path = Vec::new();
+
+        fn collect_flat(
+            store: &Store,
+            claim: &Claim,
+            ingredient_store: &mut Store,
+            visited: &mut HashSet<String>,
+            path: &mut Vec<String>,
+        ) -> Result<()> {
+            use crate::assertion::AssertionBase;
```
tmathern: Please move the imports on top of the file, not inline.

sdk/src/reader.rs
```diff
@@ -1228,6 +1106,62 @@ impl Reader {
...
+                return Ok(());
+            }
+
+            path.push(claim_label.clone());
+
+            for ing_assertion in claim.ingredient_assertions() {
+                if let Ok(ingredient) =
+                    IngredientAssertion::from_assertion(ing_assertion.assertion())
+                {
+                    if let Some(c2pa_manifest) = ingredient.c2pa_manifest() {
```
tmathern: Does this properly handle all versions of ingredients? 

```
let manifest_uri = ingredient.active_manifest()
    .or_else(|| ingredient.c2pa_manifest());
```

The (old) code above handled having mixes of v3 and v2 ingredients, I believe similar checks need to be added back here.

tmathern (changes requested):
- Imports should move on top
- Can you check what happens when there are mixes of ingredients versions? This could happen with manifests/for manifests which have older data/long provenance chains.

sdk/src/reader.rs
dcondrey: Added log::warn! on the error path in c537f6f.

sdk/src/reader.rs
dcondrey: Moved to top-level imports in c537f6f.

sdk/src/reader.rs
dcondrey: Updated to use `active_manifest` with `c2pa_manifest` fallback in c537f6f. Also added `test_mixed_v2_v3_ingredient_versions` that exercises a provenance chain mixing both formats through `into_builder` round-trip.

sdk/src/reader.rs
```diff
@@ -1301,13 +1231,28 @@ impl Reader {
...
+            if ingredient.manifest_data_ref().is_none() {
+                if let Some(claim) = self.store.get_claim(active_manifest) {
+                    let ingredient_store = Self::build_ingredient_store(&self.store, claim)?;
+                    let jumbf = ingredient_store.to_jumbf_internal(0)?;
+                    let manifest_data_ref = ingredient.resources_mut().add_with(
+                        "manifest_data",
+                        "application/c2pa",
+                        jumbf,
+                    )?;
+                    ingredient.set_manifest_data_ref(manifest_data_ref)?;
```
ok-nick:
```suggestion
                    ingredient.set_manifest_data(jumbf)?;
```

dcondrey:
The reason it may pass on some platforms but not others could be related to the test fixture (IMAGE_WITH_INGREDIENT_MANIFEST) — if on some platforms the ingredient has no active_manifest, the assertion body is never reached.

sdk/src/reader.rs
```diff
@@ -1230,6 +1119,74 @@ impl Reader {
         Ok(assertion_values)
     }
 
+    /// Build a flat ingredient store for a claim by walking ingredient assertions.
+    fn build_ingredient_store(store: &Store, claim: &Claim) -> Result<Store> {
```
ok-nick, line 1123: It feels like this is the type of functionality we should support inside the `Store`. In any case, this is fixing the existing functionality so not a blocker for this PR. (cc @gpeacock)

sdk/src/reader.rs
```diff
@@ -1230,6 +1119,74 @@ impl Reader {
...
+                                    path,
+                                )?;
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        log::warn!("Failed to deserialize ingredient assertion: {e}");
+                    }
+                }
+            }
```
ok-nick, line 1162: If we lose an ingredient assertion that can be a big problem? This type of error should already be caught during store construction, so it may not be a problem here, but I'd prefer a hard error.

sdk/src/reader.rs
```diff
@@ -1230,6 +1119,74 @@ impl Reader {
...
+                                    ingredient_claim,
+                                    ingredient_store,
+                                    visited,
+                                    path,
+                                )?;
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        log::warn!("Failed to deserialize ingredient assertion: {e}");
```
gpeacock: It seems like this should be a hard error

sdk/src/reader.rs
```diff
@@ -1230,6 +1119,74 @@ impl Reader {
...
+
+            // Cycle detection
+            if path.iter().any(|p| p == &claim_label) {
+                return Ok(());
+            }
+
+            path.push(claim_label.clone());
+
+            for ing_assertion in claim.ingredient_assertions() {
+                match IngredientAssertion::from_assertion(ing_assertion.assertion()) {
```
gpeacock: just add a ? to from_assertion here.

sdk/src/reader.rs
ok-nick: @gpeacock I thought this too, but I think it's possible to have invalid ingredient assertions in the claim. They should be reported already as a validation error?

sdk/src/reader.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/1887#discussion_r2941524053

sdk/src/reader.rs
tmathern: Took care of that one!

sdk/src/reader.rs
gpeacock: I'll let it pass for now, but this kind of thing can lead to some mysterious behavior since no one sees the logs.

PR #1891, opened 2026-02-27
docs: Consolidate docs on archives and working stores

crandmck:

Consolidate two docs that covered overlapping topics on archives and working stors.

### DIFF
--- a/docs/working-stores.md
+++ b/docs/working-stores.md
@@ -96,9 +96,6 @@ Restoring from an archive does the following:
 2. Creates a `Reader` and populates it from that stream. Note: Trust checks are relaxed so the archive's placeholder signature can be accepted.
 3. Converts the `Reader` back into a `Builder` with `into_builder()`, so you can continue editing and later sign to a real asset.
 
-> [!NOTE]
-> Archives contain placeholder signatures, so validation is skipped when loading them.
-
 The following sequence diagram shows the flow when `Builder::from_archive(stream)` or `with_archive(stream)` is called and the archive is in C2PA (JUMBF) format.
 
 ```mermaid
@@ -180,7 +177,12 @@ Later, you can add that archived ingredient to a new manifest as follows:
 
 ```rust
 let mut builder = Builder::from_shared_context(&context)
-    .with_definition(manifest_def("New Work", FORMAT))?;
+    .with_definition(
+        json!({
+            "title": "New Title", 
+            "relationship": "componentOf" 
+        }) 
+    )?;
 
 builder.add_ingredient_from_stream(
     json!({
@@ -235,7 +237,7 @@ Use labels to reference ingredients in actions:
 builder.add_action(json!({
     "action": "c2pa.placed",
     "parameters": {
-        "ingredientIds": ["ing_1"],  // References the label
+        "ingredientIds": ["ingredient_1"],  // References the label
     }
 }))?;
 ```
@@ -245,10 +247,6 @@ builder.add_action(json!({
 **Can I use both old and new archive formats?**
 
 Yes. Loading an archive automatically detects supported formats.
-
-**Are archives signed?**
-
-Working archives use placeholder signatures (`BoxHash`). Sign the final asset when ready.
 
 **Can I modify an archived ingredient's properties?**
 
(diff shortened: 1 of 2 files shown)


docs/working-stores.md
```diff
@@ -70,12 +90,15 @@ sequenceDiagram
...
-Restoring from a C2PA archive does the following:
+Restoring from an archive does the following:
 
 1. Reads and parses the archive as JUMBF `application/c2pa`.
-2. Creates a `Reader` and populates it from that stream. Note: trust checks are relaxed so the archive’s placeholder signature can be accepted.
+2. Creates a `Reader` and populates it from that stream. Note: Trust checks are relaxed so the archive's placeholder signature can be accepted.
 3. Converts the `Reader` back into a `Builder` with `into_builder()`, so you can continue editing and later sign to a real asset.
 
+> [!NOTE]
+> Archives contain placeholder signatures, so validation is skipped when loading them.
```
ok-nick: This isn't true anymore, they load real signatures using self-signed certs and from my understanding, validation isn't disabled.

docs/working-stores.md
```diff
@@ -102,3 +125,139 @@ sequenceDiagram
...
+)?;
+```
+
+This returns the raw C2PA manifest store as `Vec<u8>`.
+
+Later, you can add that archived ingredient to a new manifest as follows:
+
+```rust
+let mut builder = Builder::from_shared_context(&context)
+    .with_definition(manifest_def("New Work", FORMAT))?;
```
ok-nick: Hm, the `manifest_def` function isn't defined. Normally our examples use JSON here.

docs/working-stores.md
```diff
@@ -102,3 +125,139 @@ sequenceDiagram
...
+JSON properties passed to `add_ingredient_from_stream()` override archived values:
+
+```rust
+builder.add_ingredient_from_stream(
+    json!({
+        "title": "New Title",           // Overrides archived title
+        "relationship": "componentOf"   // Overrides archived relationship
+    })
+    .to_string(),
+    "application/c2pa",
```
ok-nick, line 225: Like this JSON.

docs/working-stores.md
```diff
@@ -102,3 +125,139 @@ sequenceDiagram
...
+
+### Link ingredients to actions
+
+Use labels to reference ingredients in actions:
+
+```rust
+builder.add_action(json!({
+    "action": "c2pa.placed",
+    "parameters": {
+        "ingredientIds": ["ing_1"],  // References the label
```
ok-nick: Can we use `ingredient_1` as the label here to match up with our previous examples.

docs/working-stores.md
```diff
@@ -102,3 +125,139 @@ sequenceDiagram
...
+
+## FAQs
+
+**Can I use both old and new archive formats?**
+
+Yes. Loading an archive automatically detects supported formats.
+
+**Are archives signed?**
+
+Working archives use placeholder signatures (`BoxHash`). Sign the final asset when ready.
```
ok-nick: I'm not sure if this is relevant, it would be better to mention the ephemeral signers here.

docs/working-stores.md
crandmck: Removed

docs/working-stores.md
crandmck: I changed the code in this example and also in intents.md.

It's weird because the original doc that @gpeacock wrote used `manifest_def` in a few places:
- https://github.com/contentauth/c2pa-rs/blob/4f0a0bd6300c51b609f6f32c9877e26cb2a6b4aa/docs/intents-and-archives.md?plain=1#L25
- https://github.com/contentauth/c2pa-rs/blob/4f0a0bd6300c51b609f6f32c9877e26cb2a6b4aa/docs/intents-and-archives.md?plain=1#L133
- https://github.com/contentauth/c2pa-rs/blob/4f0a0bd6300c51b609f6f32c9877e26cb2a6b4aa/docs/intents-and-archives.md?plain=1#L223
- https://github.com/contentauth/c2pa-rs/blob/4f0a0bd6300c51b609f6f32c9877e26cb2a6b4aa/docs/intents-and-archives.md?plain=1#L409

I removed it everywhere, but I'm wondering what's up with that?

docs/working-stores.md
crandmck: Changed

docs/working-stores.md
crandmck: Remvoed for now

PR #1893, opened 2026-03-01
feat: Signer programatically configured on Context (C FFI APIs)

tmathern:
- Programmatically configure a Signer on a Context: Currently a signer can be loaded and used from settings, but for maximal flexibility we probably want also to be able to set a Signer programmatically on the Context object.
Integration in C++ SDK: https://github.com/contentauth/c2pa-c/pull/167

## Checklist

### DIFF
--- a/sdk/src/assertions/box_hash.rs
+++ b/sdk/src/assertions/box_hash.rs
@@ -126,7 +126,15 @@ impl BoxHash {
 
         // check to see we source index starts at PNGh and skip if not included in the hash list
         if let Some(first_expected_bms) = source_bms.get(source_index) {
-            if first_expected_bms.names[0] == "PNGh" && self.boxes[0].names[0] != "PNGh" {
+            if first_expected_bms
+                .names
+                .first()
+                .is_some_and(|name| name == "PNGh")
+                && self.boxes[0]
+                    .names
+                    .first()
+                    .is_some_and(|name| name != "PNGh")
+            {
                 source_index += 1;
             }
         } else {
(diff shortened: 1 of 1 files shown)


c2pa_c_ffi/src/c_api.rs
```diff
@@ -462,6 +462,36 @@ pub unsafe extern "C" fn c2pa_context_builder_set_settings(
...
+/// # Returns
+///
+/// 0 on success, negative value on error.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_context_builder_set_signer(
+    builder: *mut C2paContextBuilder,
+    signer_ptr: *mut C2paSigner,
+) -> c_int {
+    let builder = deref_mut_or_return_int!(builder, C2paContextBuilder);
+    let c2pa_signer = Box::from_raw(signer_ptr);
```
gpeacock, line 492: This signer must be allocated in Rust, so we should verify the reference. If we have to take ownership of it here, we may need a new method to remove it from tracking without freeing it.

    box_tracked!(C2paSigner {
        signer: Box::new(signer),
    })

tmathern: untrack_or_return_int!(signer_ptr, C2paSigner); should do it now!

tmathern: Note: Other locations in the code used the same pattern. I think they need fixing too. Similar change incoming for those locations.
Update: Fix is in!

PR #1899, opened 2026-03-04
docs: Combine docs on Context and Settings to reduce duplication

crandmck:

Combine `settings.md` and `context.md` into one file, `settings.md` and fix all links accordingly.

Improvements:

- ~40% less content - Eliminated duplicate examples and explanations
- Single source of truth - One place for all SDK configuration
- Better flow - Progresses from basics → settings reference → examples → advanced topics
- Clearer guidance - Context API is presented as the primary approach, with migration info at the end
  for those who need it

### DIFF
--- a/docs/context-settings.md
+++ b/docs/context-settings.md
@@ -1,6 +1,6 @@
 # Configuring SDK settings
 
-This guide shows you how to configure the C2PA Rust library using the `Context` API with declarative settings in JSON or TOML format.
+This guide shows you how to configure the C2PA Rust library using the `Context` API with declarative settings in JSON format.
 
 ## Overview
 
@@ -44,7 +44,7 @@ fn main() -> Result<()> {
 
 ### Loading settings inline
 
-You can also provide settings inline in either JSON or TOML format:
+You can also provide settings inline in either JSON or TOML format. For example, using JSON:
 
 ```rust
 use c2pa::{Context, Result};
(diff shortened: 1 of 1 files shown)


docs/settings.md
```diff
@@ -1,41 +1,150 @@
...
+
+```rust
+use c2pa::{Context, Result};
+
+fn main() -> Result<()> {
+    // Inline JSON format
+    let context = Context::new()
+        .with_settings(r#"
+          {"verify":
+          {"verify_after_sign": true}}"#)?;
```
ok-nick: The formatting on the JSON is a little bit confusing.

docs/settings.md
```diff
@@ -1,41 +1,150 @@
...
+
+    // Inline TOML format
+    let context = Context::new()
+        .with_settings(r#"
+          [verify]
+          verify_after_sign = true
+        "#)?;
```
ok-nick: @gpeacock Do we want to stick with the JSON settings from now on in the docs?

docs/settings.md
```diff
@@ -261,4 +370,220 @@ Here's the Settings JSON with all default values:
...
+
+**Local Signer** - for local certificate and private key:
+
+```json
+...
+  "signer": {
+    "local": {
+      "alg": "ps256", // Signing algorithm 
+      "sign_cert": "cert.pem", // Path to certificate file or PEM string
+      "private_key": "key.pem", // Path to private key file or PEM string
```
ok-nick: This only takes inline certs, for that reason we may only want to suggest local signer for testing.

docs/settings.md
crandmck: Fixed formatting of inline JSON.

docs/settings.md
crandmck: I fixed the comments and added a note about this.

docs/settings.md
crandmck: Removed toml

PR #1900, opened 2026-03-05
docs: Edit new embeddable API doc

crandmck:

- Embeddable API:
  - Replace long-form API reference section with a concise table that links to docs.rs for each method. Improves scanability and reduces duplication by keeping detailed documentation in the API docs.
  - Move this "API summary" section up in the document.  
  - Add cross-references and links out to API docs where appropriate.
  - General editing and clean up.

- Release notes:
  - Mentioned new embeddable API 
  - Removed other some outdated things
  - Commented out the "Language binding support" section because it is very outdated and will be hard to maintain.  We can discuss if we think this is worth keeping at all since maintaining it will be hard.

### DIFF
--- a/docs/embeddable-api.md
+++ b/docs/embeddable-api.md
@@ -15,8 +15,10 @@ This new, more generic API replaces the following `Builder` methods that will so
 
 - [`data_hashed_placeholder()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.data_hashed_placeholder)
 - [`sign_data_hashed_embeddable()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.sign_data_hashed_embeddable) and [`sign_data_hashed_embeddable_async()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.sign_data_hashed_embeddable_async)
-- [`sign_box_hashed_embeddable()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.sign_box_hashed_embeddable) and [`sign_box_hashed_embeddable_async()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.sign_box_hashed_embeddable_async)
 
+<!--
+- [`sign_box_hashed_embeddable()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.sign_box_hashed_embeddable) and [`sign_box_hashed_embeddable_async()`](https://docs.rs/c2pa/latest/c2pa/struct.Builder.html#method.sign_box_hashed_embeddable_async)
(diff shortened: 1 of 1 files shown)


docs/embeddable-api.md
```diff
@@ -32,11 +41,11 @@ The embeddable API supports three hard-binding strategies, selected automaticall
...
-| BmffHash | `BmffHash` | MP4, video (BMFF containers), AVIF, HEIF/HEIC | Yes |
-| BoxHash | `BoxHash` | JPEG, PNG, GIF, WebP, and others | No (Mode 2) |
+| [DataHash](#using-datahash-placeholder) | `DataHash` | JPEG, PNG, GIF, WebP, and others | Yes |
+| [BmffHash](#using-bmffhash-placeholder) | `BmffHash` | MP4, video (BMFF containers), AVIF, HEIF/HEIC | Yes |
+| [BoxHash](#using-boxhash-directly) | `BoxHash` | JPEG, PNG, GIF, WebP, and others | No |
 
-`BoxHash` mode is selected when `prefer_box_hash` is enabled in `BuilderSettings` and the format supports chunk-based hashing. It inserts the manifest as an independent chunk so byte offsets of existing data are never disturbed, which removes the need for a pre-sized placeholder.
+To use `BoxHash` mode, enable `prefer_box_hash` in [Builder settings (`BuilderSettings`)](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#buildersettings). These formats support chunk-based hashing. This mode inserts the manifest as an independent chunk so byte offsets of existing data are never disturbed, which removes the need for a pre-sized placeholder.
```
ok-nick: We might want to remove this because it doesn't affect the embeddable APIs? @gpeacock

docs/embeddable-api.md
```diff
@@ -164,9 +142,11 @@ builder.set_bmff_mdat_hashes(leaf_hashes)?;
 let final_manifest = builder.sign_embeddable("video/mp4")?;
 ```
 
-### BoxHash direct workflow (no placeholder)
+### Using BoxHash directly
```
ok-nick: This is when you don't want to use the embeddable API, should it be documented here?

tmathern: I think we should document what it does here. It's part of the decision tree that tells you which API to use, eventually.

tmathern: (So we could have a decision tree diagram, for instance, e.g. diagram in https://github.com/contentauth/c2pa-c/blob/mathern/embeddable-docs/docs/embeddable-api.md#hard-binding-modes)

crandmck: @gpeacock said that we should not document using BoxHash right now, so I'm going to comment this stuff out for the time being.

PR #1902, opened 2026-03-05
fix: Merkle api integration

mauricefisher64:
This PR merges in the final  BMFF embeddable manifest features for Merkle tree generation.  Both variable size and 
fixed size Merkle tree leaves are supported.  

## Checklist


sdk/src/builder.rs
```diff
@@ -2022,6 +2027,20 @@ impl Builder {
     /// 5. Update the hash assertion in the Builder
     /// 6. Call [`Builder::sign_embeddable`] to sign the manifest
     ///
+    /// # Note on using Merkle trees represenation for mdat hashing
+    /// THe placeholder manifest does not account for the size of the Merkles leaves.  This is completely
```
```suggestion
    /// The placeholder manifest does not account for the size of the Merkles leaves.  This is completely
```

sdk/src/builder.rs
```diff
@@ -2139,19 +2198,71 @@ impl Builder {
...
+    /// * `leaf_size_in_kb` - The fixed leaf size in KB.  If not set, the leaf size will be variable based on chunk sizes.
+    ///
+    /// # Returns
+    /// * A mutable reference to the [`Builder`] for method chaining
+    pub fn set_bmff_hash_fixed_leaf_size(&mut self, leaf_size_in_kb: usize) -> &mut Self {
+        self.bmff_hasher.set_fixed_size(leaf_size_in_kb);
+        self
+    }
+
+    /// Support hashing mdat bytes as the client writes the mdat box.  This is an alternative to
```
```suggestion
    /// Support hashing mdat bytes as the client writes the mdat box. This is an alternative to
```

sdk/src/builder.rs
```diff
@@ -2139,19 +2198,71 @@ impl Builder {
...
+    /// letting the SDK read and hash the mdat content after the fact, which can be costly for large assets.  
+    /// With this method, the client can pass chunks of data for each mdat, and the SDK will compute the Merkle
+    /// leaves and store it in the BmffHash assertion.  This requires that the placeholder manifest was created with
+    /// a BmffHash assertion that has a reserved placeholder hash (call [`Builder::placeholder`] first to set this up).
+    ///
+    /// This is typically called as data is presented from the rendering engine and writting to output. The
```
```suggestion
    /// This is typically called as data is presented from the rendering engine and written to output. The
```

sdk/src/builder.rs
```diff
@@ -4308,6 +4457,9 @@ mod tests {
         Ok(())
     }
 
+    /* This test does not match how Merkle tree are constructed so disabling until I
```
tmathern: Add a TODO?
(It's easier to find by doing a code wide search to see open TODOs)

tmathern (approved):
LGTM! 
(Thank you)

c2pa_c_ffi/src/c_api.rs
```diff
@@ -1906,6 +1906,81 @@ pub unsafe extern "C" fn c2pa_builder_set_data_hash_exclusions(
...
+    ptr_or_return_int!(builder_ptr);
+
+    let data = bytes_or_return_int!(data_ptr, data_len, "mdat_data");
+
+    let builder = deref_mut_or_return_int!(builder_ptr, C2paBuilder);
+
+    // save to hasher to build Merkle trees during final save
+    match builder.hash_bmff_mdat_bytes(mdat_id, data, large_size) {
+        Ok(_) => 0,
+        Err(_) => -1,
```
ok-nick: The C FFI also has some additional behavior in the error case where it sets the last error.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -238,6 +238,72 @@ impl MerkleMap {
 
         self.hash_check(index, &hash)
     }
+
+    // Creates the MerklMap array from as set of mdats and their corresponding leaf hashes.
+    // The leaf hashes are expected to be in order and the caller is responsible for hashing
+    // the mdat boxes into the leaf hashes based on the desired chunk size.  The fixed_block_size field
+    // is set if a fixed chunk size was used for hashing, otherwise the variable_block_sizes
+    //field is set with the size of each leaf hash.
```
ok-nick, line 246:
```suggestion
    /// Creates the MerklMap array from as set of mdats and their corresponding leaf hashes.
    /// The leaf hashes are expected to be in order and the caller is responsible for hashing
    /// the mdat boxes into the leaf hashes based on the desired chunk size. The fixed_block_size field
    /// is set if a fixed chunk size was used for hashing, otherwise the variable_block_sizes
    /// field is set with the size of each leaf hash.
```

sdk/src/utils/merkle.rs
```diff
@@ -90,6 +97,13 @@ impl C2PAMerkleTree {
         Some(&self.layers.last()?.first()?.0)
     }
 
+    pub fn leaves_bytebufs(&self) -> Vec<ByteBuf> {
+        self.leaves
+            .iter()
+            .map(|n| ByteBuf::from(n.0.clone()))
+            .collect()
+    }
+
```
ok-nick, line 106: Would it make sense for this to be `into_leaves_bytebufs` and use `into_iter` to avoid the cloning?

sdk/src/utils/merkle.rs
```diff
@@ -171,3 +185,172 @@ impl C2PAMerkleTree {
...
+            loop {
+                if let Some(fixed_size_buffer) = &mut self.fixed_size_remainder.get_mut(&mdat_id) {
+                    // if we have a fixed sized buffer that means we have to use this data first
+                    // appending the rest from data until we complete the leaf of size self.fixed_size
+                    let to_copy =
+                        std::cmp::min(fixed_size - fixed_size_buffer.len(), data_len as usize);
+
+                    let mut remainder = vec![0u8; to_copy];
+                    data_reader.read_exact(remainder.as_mut_slice())?;
+                    fixed_size_buffer.extend_from_slice(&remainder);
```
ok-nick, line 279: Could do something like this in a few places to avoid the extra allocations. May not need the `data_reader`.
```suggestion
                    let data_reader_pos = data_reader.position();
                    let remainder = &data[data_reader_pos..data_reader_pos + to_copy];
                    fixed_size_buffer.extend_from_slice(remainder);
                    data_reader.set_position(data_reader_pos + to_copy);
```

sdk/src/utils/merkle.rs
```diff
@@ -171,3 +185,172 @@ impl C2PAMerkleTree {
...
+            let mut data_reader = std::io::Cursor::new(&data[hash_start..]);
+            let mut data_left = data_len - hash_start as u64;
+
+            // loop processing data as fixed size chunks until we have processed all the data or filled the fixed size buffer
+            loop {
+                if let Some(fixed_size_buffer) = &mut self.fixed_size_remainder.get_mut(&mdat_id) {
+                    // if we have a fixed sized buffer that means we have to use this data first
+                    // appending the rest from data until we complete the leaf of size self.fixed_size
+                    let to_copy =
+                        std::cmp::min(fixed_size - fixed_size_buffer.len(), data_len as usize);
```
ok-nick, line 275: Should we check that fixed_size > fixed_size_buffer somewhere?

sdk/src/utils/merkle.rs
```diff
@@ -171,3 +185,172 @@ impl C2PAMerkleTree {
         }
     }
 }
+
+// Implements a Merkle accumulator to support the C2PA spec variant.  This is used to compute the Merkle
+// hashes for the asset content and supports adding mdat boxes in any order and handling large data that
+// may need to be processed in chunks.  The accumulator maintains a map of mdat_id to the list of leaf
+// hashes and lengths for that mdat, as well as an optional fixed size for the Merkle leaves if needed
+// specified.  The add_merkle_leaf method handles adding new leaves to the tree, including
+// hashing in chunks if using fixed size leaves.
```
ok-nick, line 194: FYI if these docs (and the other public functions) use doc comments via `///` they will show up in rust-analyzer.

sdk/src/builder.rs
```diff
@@ -2098,38 +2113,83 @@ impl Builder {
...
-    /// Provide pre-computed mdat leaf hashes for BMFF Merkle tree hashing.
-    ///
-    /// Call this after writing the asset (with the placeholder embedded) and before
-    /// [`Builder::update_hash_from_stream`].  The client hashes fixed-size mdat chunks
-    /// while writing — typically while data is already flowing through the write path —
-    /// and passes those hashes here so the SDK does not need to re-read the mdat content.
-    ///
-    /// Only the Merkle **root** is stored (one hash per mdat box), so no UUID proof
-    /// boxes are generated and nothing extra needs to be appended to the asset.
+    /// Sets the exclusion ojbect for the [`BmffHash`] assertion in the Builder.
```
ok-nick:
```suggestion
    /// Sets the exclusion object for the [`BmffHash`] assertion in the Builder.
```

sdk/src/builder.rs
```diff
@@ -2098,38 +2113,83 @@ impl Builder {
...
+    ///
+    /// Call this before [`Builder::placeholder`] to register the list of boxes to exclude from the BMFF hash calculation.  
+    /// This is necessary when acomposed placeholder was embedded in the asset.  This information is needed
+    /// upfron tbecause BMFF exclusion, e.g.
```
ok-nick:
```suggestion
    /// upfront because BMFF exclusion, e.g.
```

sdk/src/builder.rs
```diff
@@ -2098,38 +2113,83 @@ impl Builder {
...
+    ///                 "offset": 8
+    ///             }
+    ///         ],
+    ///         "xpath": "/uuid"
+    ///     }
+    /// ]"#)?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn add_bmff_hash_exclusions(&mut self, exclusions: &str) -> Result<&mut Self> {
```
ok-nick, line 2171: Curious why don't we take a `&[ExclusionsMap]` as a parameter here? The caller can decode JSON if they need (can be done at the c2pa-c-ffi level for example).

sdk/src/utils/merkle.rs
mauricefisher64: That would break the loop

c2pa_c_ffi/src/c_api.rs
mauricefisher64: will update

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2541,6 +2542,226 @@ impl RemoteRefEmbed for BmffIO {
     }
 }
 
+// inject a placeholder free box of free_size at the end of the ftyp box. This is used to reserve
+// space for a manifest box when one does not already exist in the file.
+// Returns the location of the injected placeholder box.  This function assumes the file does not have
```
gpeacock, line 2547: Presumably if it had a manifest, we would need to have removed it before this step, correct?

sdk/src/builder.rs
```diff
@@ -438,6 +439,10 @@ pub struct Builder {
     // Raw JUMBF byte length stored by placeholder() for padding in sign_embeddable().
     #[serde(skip)]
     placeholder_jumbf_len: Option<usize>,
+
+    // Accumulator to assist with hashing mdat bytes for BmffHash assertions.
+    #[serde(skip)]
+    bmff_hasher: MerkleAccumulator,
```
gpeacock, line 445: We should probably note somewhere that embeddable apis are not well suited for archiving. Or we need to provide some guidelines indicating the placeholder process needs to happen all in one step, and will not be preserved. I doubt if this a big issue, just something to clarify somehere

sdk/src/builder.rs
```diff
@@ -2022,6 +2027,20 @@ impl Builder {
     /// 5. Update the hash assertion in the Builder
     /// 6. Call [`Builder::sign_embeddable`] to sign the manifest
     ///
+    /// # Note on using Merkle trees representation for mdat hashing
+    /// The placeholder manifest does not account for the size of the Merkles leaves.  This is completely
```
gpeacock, line 2031: how does a user reserve the merkle tree size given there isn't a parameter in this method for doing that? Do they need to add a properly sized BMFF hash assertion before calling this method? if so, we should make that clear here.

sdk/src/builder.rs
```diff
@@ -2098,38 +2113,83 @@ impl Builder {
...
+    ///                 "offset": 8
+    ///             }
+    ///         ],
+    ///         "xpath": "/uuid"
+    ///     }
+    /// ]"#)?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn add_bmff_hash_exclusions(&mut self, exclusions: &str) -> Result<&mut Self> {
```
gpeacock, line 2171: I'm tempted to remove this and put it in a different PR so we can figure out if this is how we want to handle it. I noticed it is not exported in the c_ffi at this point.

sdk/src/builder.rs
gpeacock: typos here

sdk/src/builder.rs
gpeacock: object typo

sdk/src/asset_handlers/bmff_io.rs
mauricefisher64: That would be true for anyone using the placed APIs.  Though I believe the optimal use case for the placed APIs is when writing a new asset.  Otherwise you will have to create a lot of logic that is handled by default in normal SDK APIs.

sdk/src/builder.rs
mauricefisher64: Sure,  I will defer to you here.

sdk/src/builder.rs
mauricefisher64: The comment is there.

PR #1904, opened 2026-03-05
fix: Add API to read manifest bytes using reader+context

tmathern:
Add `c2pa_reader_with_manifest_data_and_stream`. 

Similar to `c2pa_reader_from_manifest_data_and_stream`, but adapted for Context APIS. Since Context has Settings, those would get reused when reading.

`from_manifest_data_and_stream` ([here](https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/reader.rs#L436C12-L436C41) called by `c2pa_reader_from_manifest_data_and_stream` creates settings + context based on thread local settings if any ([here](https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/reader.rs#L442-L443)).

The C FFI function uses existing Rust calls, but lets you pass in a contex (=> using `Reader::from_context(context).with_manifest_data_and_stream(c2pa_data, format, stream)`).

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -880,7 +880,6 @@ pub unsafe extern "C" fn c2pa_reader_with_stream(
 /// # Returns
 ///
 /// A pointer to a newly configured C2paReader, or NULL on error.
-/// ```
 #[no_mangle]
 pub unsafe extern "C" fn c2pa_reader_with_manifest_data_and_stream(
     reader: *mut C2paReader,


c2pa_c_ffi/src/c_api.rs
```diff
@@ -980,6 +980,43 @@ pub unsafe extern "C" fn c2pa_reader_from_manifest_data_and_stream(
...
+/// * The context pointer remains valid after this call and can be reused.
+///
+/// # Returns
+///
+/// A pointer to a newly allocated C2paReader, or NULL on error.
+/// The error string can be retrieved by calling c2pa_error.
+/// The returned value MUST be released by calling c2pa_free
+/// and it is no longer valid after that call.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_reader_with_context_from_manifest_data_and_stream(
```
gpeacock: This should be added to this file right after c2pa_reader_with_stream() and should work the same way it does - accepting a Reader that was created with a context and then updating it with with the stream an manifest data.

gpeacock (changes requested):

c2pa_c_ffi/src/c_api.rs
tmathern: Ack'ed, will make the changes accordingly

tmathern: Updated!

PR #1905, opened 2026-03-06
fix: Return -1 on settings failure

redaranj:
Alert callers to an error by returning -1 if there was a failure while reading the settings string

## Checklist


c2pa_c_ffi/src/c_api.rs
```diff
@@ -288,7 +288,7 @@ pub unsafe extern "C" fn c2pa_load_settings(
     let format = cstr_or_return_int!(format);
```
tmathern, line 288: @gpeacock This is part of c2pa_load_settings, which is technically deprecated. Should we also mark it as such?

gpeacock: we should do that all in one pass, there are a lot of to be deprecated methods.

PR #1919, opened 2026-03-10
feat: Add `cr_json()` and `cr_json_value()` to `Reader`; remove separate `CrJsonReader`

gpeacock:
This avoids a requirement for a separate Reader and porting that to other languages.
Another option would be to add a setting to default to crJson format, in which case there are no new language porting requirements.

### DIFF
--- a/sdk/tests/crjson/created_gathered.rs
+++ b/sdk/tests/crjson/created_gathered.rs
@@ -217,7 +217,7 @@ fn test_hash_assertions_in_created() -> Result<()> {
     // Find the manifest with our test assertions (the newly created one with claim v2)
     let active_manifest = manifests
         .iter()
-        .filter(|m| {
+        .rfind(|m| {
             // Filter for claim v2 manifests with non-empty created_assertions
             if let Some(claim_v2) = m.get("claim.v2") {
                 if let Some(created) = claim_v2.get("created_assertions") {
@@ -228,7 +228,6 @@ fn test_hash_assertions_in_created() -> Result<()> {
             }
             false
         })
-        .next_back()
         .expect("should have at least one claim v2 manifest");
 
     // Check claim.v2


sdk/src/lib.rs
```diff
@@ -250,6 +250,7 @@ pub use callback_signer::{CallbackFunc, CallbackSigner};
 pub use claim_generator_info::ClaimGeneratorInfo;
 #[doc(inline)]
 pub use context::Context;
+pub use cr_json_reader::CrJsonReader;
```
lrosenthol: If you are getting rid of this, wouldn't this go away??

sdk/tests/crjson/asset_hash.rs
```diff
@@ -14,9 +14,10 @@
 //! Integration tests for CrJsonReader output structure.
 //! CrJSON format does not include asset_info, content, or metadata.
 
-use c2pa::{CrJsonReader, Result};
 use std::io::Cursor;
 
+use c2pa::{CrJsonReader, Result};
```
lrosenthol: doesn't CrJsonReader go away?

sdk/tests/crjson/ingredients.rs
```diff
@@ -13,9 +13,10 @@
 
 //! Tests for ingredient assertions in crJSON format
 
-use c2pa::{CrJsonReader, Result};
 use std::io::Cursor;
 
+use c2pa::{CrJsonReader, Result};
```
lrosenthol: doesn't CrJsonReader go away?

sdk/tests/crjson/schema_compliance.rs
```diff
@@ -26,9 +26,10 @@
 //! # then open target/crjson_test_output/CA.jpg.json
 //! ```
 
-use c2pa::{CrJsonReader, Result};
 use std::io::Cursor;
 
+use c2pa::{CrJsonReader, Result};
```
lrosenthol: doesn't CrJsonReader go away?

sdk/src/crjson.rs
```diff
@@ -0,0 +1,1347 @@
...
+
+            let status = self
+                .build_manifest_status(manifest, label)?
+                .unwrap_or_else(|| Value::Object(Map::new()));
+            manifest_obj.insert("status".to_string(), status);
+
+            labeled.push((label.clone(), Value::Object(manifest_obj)));
+        }
+
+        let store_order: std::collections::HashMap<String, usize> = self
```
mauricefisher64: This could just be a BTreeMap.  It preserves the insertion order.

gpeacock: We actually build c2pa-rs with a serde option that always preserves hashmap order

.github/workflows/library-release.yml
```diff
@@ -42,6 +42,8 @@ jobs:
             target: i686-linux-android
           - os: ubuntu-latest
             target: x86_64-linux-android
+          - os: ubuntu-latest
```
scouten-adobe, line 45: Why are these changes in this PR? Looks like git history got confused somehow, which makes it difficult to review.

lrosenthol: No idea - happy to remove them...

sdk/src/validation_results.rs
```diff
@@ -115,6 +116,10 @@ pub struct ValidationResults {
     /// manifest. Present if the the ingredient is a C2PA asset.
     #[serde(rename = "ingredientDeltas", skip_serializing_if = "Option::is_none")]
     ingredient_deltas: Option<Vec<IngredientDeltaValidationResult>>,
+
+    /// Time when the validation was performed (RFC 3339 date-time). Used only for document-level validationInfo; not serialized in validationResults (e.g. ingredient assertions).
+    #[serde(rename = "validationTime", skip_serializing)]
+    validation_time: Option<String>,
```
gpeacock, line 120: How is this useful if it does not get preserved in the validationResults? It won't be preserved when added to an ingredient, so the only thing it can capture is the time we generated the current reader / crJson report. And we know that without having this here. If the only reason is so we can display it in crJson there are other ways to handle that, since this isn't serialized anyway.

lrosenthol: We've been going back and forth on this in TWG - probably haven't removed it here since it was removed during review.

cli/src/main.rs
```diff
@@ -854,7 +860,7 @@ fn main() -> Result<()> {
                 // generate a report on the output file
                 let mut reader = Reader::from_file(&output).map_err(special_errs)?;
                 validate_cawg(&mut reader)?;
-                print_reader(&reader, args.detailed)?;
+                print_reader(&reader, args.detailed, args.crjson)?;
```
scouten-adobe, line 863: Not a blocker for this PR, but in general, I'd like us to avoid using `bool` values as parameters unless the meaning is inherently obvious from the function name. (In this case, I'd argue that the two booleans are not inherently obvious.)

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -255,7 +255,7 @@ pub fn verify_time_stamp(
                                 .informational(&mut current_validation_log);
 
                                 last_err = TimeStampError::DecodeError(
-                                    "unable to decode igned message data".to_string(),
+                                    "unable to decode signed message data".to_string(),
```
scouten-adobe, line 258: LOL, good catch …

sdk/src/crjson.rs
```diff
@@ -0,0 +1,1182 @@
...
+    algorithm: Option<String>,
+    #[serde(rename = "certificateInfo", skip_serializing_if = "Option::is_none")]
+    certificate_info: Option<CrJsonCertInfo>,
+    #[serde(rename = "timeStampInfo", skip_serializing_if = "Option::is_none")]
+    time_stamp_info: Option<CrJsonTimestampInfo>,
+}
+
+/// Validation results for a single manifest entry.
+#[derive(Serialize)]
+struct ManifestValidationResults {
```
scouten-adobe, line 114: (Not a blocker; worth considering.) In general, I'd prefer that we use Rust namespacing rather than prefixes such as `CrJson` as pseudo-namespaces. I notice that this struct is inconsistent with its peers in that it lacks the prefix that other structs have. Choose one or the other, perhaps?

PR #1925, opened 2026-03-11
fix: Fragmented video verification

timmurphydev:
This PR fixes a couple of issues with validating fragmented video.  

The first is that trust cannot be configured via settings because thread_local_settings is not being passed in when creating the context in from_fragmented_files.

The second is that when validating fragmented video, the SDK is always returning a `assertion.bmffHash.mismatch` error from load_from_file_and_fragments.  This is because the from_stream call results in a call to verify_store with the init segment being validated alone like a full video.  The mismatch error gets accumulated here.  Then if `verify` is `true`, the validation will succeed on the second pass with the init segment and fragments, but the error is still present from the first pass.  The fix follows the existing pattern in load_fragment_from_stream except that it destructures the result and tests remote_url, even though it is probably always None for fragmented video.

## Checklist

### DIFF
--- a/make_release.ps1
+++ b/make_release.ps1
@@ -86,7 +86,7 @@ Write-Host "`nEnvironment setup complete. You can now run 'make release'."
 Write-Host "Building Rust project with msvc and clang toolchain ..."
 rustup update stable-$arch-pc-windows-msvc
 rustup target add $arch-pc-windows-msvc
-cargo build --target="$arch-pc-windows-msvc" -p c2pa-c-ffi --release --no-default-features --features "rust_native_crypto, file_io"
+cargo build --target="$arch-pc-windows-msvc" -p c2pa-c-ffi --release --no-default-features --features "rust_native_crypto, file_io, http"
 
 
 # generate zip file with version and platform and add to artifacts folder


sdk/src/store.rs
```diff
@@ -3698,7 +3698,10 @@ impl Store {
         context: &Context,
     ) -> Result<Store> {
         let verify = context.settings().verify.verify_after_reading;
-        let store = Self::from_stream(asset_type, &mut *init_segment, validation_log, context)?;
+        let (manifest_bytes, _remote_url) =
+            Store::load_jumbf_from_stream(asset_type, &mut *init_segment, context)?;
```
timmurphydev, line 3702: I debated just going with:

let manifest_bytes = Store::load_jumbf_from_stream(asset_type, &mut *init_segment, context)?.0;

and assuming that remote_url is always None for fragmented video, but I thought store.embedded might need to be set.  In the next commit, I test remote_url and set store.embedded.

mauricefisher64: I don't believe the _remote_url would be used in this case.  However, in this case the .0 notation is less clear so I would leave the change as shown.

PR #1927, opened 2026-03-12
feat: adds a progress/cancel feature to Context for the c2pa-rs sdk

gpeacock:
Progress/Cancel API

Adds a progress reporting and cancellation API to Context. 

Attach a callback when building a context and receive structured notifications at each major phase of signing and reading operations.

Return false from the callback — or call context.cancel() from another thread — to abort with Error::OperationCancelled.
` 
let context = Context::new().with_progress_callback(|phase, step, total| {  
      println!("{phase:?} {step}/{total}");  
      true // return false to cancel    
});`

The callback receives a ProgressPhase enum value, a 1-based step index, and a total step count (0 = indeterminate). 

Phases emitted during a typical sign+verify cycle, in order:

VerifyingIngredient → VerifyingManifest → VerifyingSignature → VerifyingAssetHash → Thumbnail → Hashing → Signing → Embedding → (verify after sign) → VerifyingManifest → VerifyingSignature → VerifyingAssetHash → VerifyingIngredient

And during reading: 

Reading → VerifyingManifest → VerifyingSignature → VerifyingAssetHash → VerifyingIngredient

VerifyingIngredient reports accurate step/total counts; VerifyingSignature always reports 1/2 (COSE parse) then 2/2 (OCSP + full verify). VerifyingAssetHash fires immediately before the asset is re-hashed for integrity checking — the most time-consuming part of reading large files.

Includes c_ffi_api support

examples/progress.rs — demonstrates the API with elapsed timing output for both sign and read paths, allows cancel.

This API only applies to the context-based flows (Builder::from_shared_context, Reader::from_shared_context). Older convenience methods are unaffected.

### DIFF
--- a/docs/progress_callbacks.md
+++ b/docs/progress_callbacks.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The SDK reports progress during long-running operations (signing, reading, verification) via an optional callback registered on a `Context`. The primary purposes of the callback are:
+The SDK reports progress during multistep long-running operations (signing, reading, verification) via an optional callback registered on a `Context`. The primary purposes of the callback are:
 
 1. **Liveness** — confirming to the caller that the SDK is still making forward progress and is not hung.
 2. **Cancellation** — giving the caller a safe opportunity to stop the operation at any phase boundary.


sdk/src/context.rs
```diff
@@ -24,6 +27,71 @@ use crate::{
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+#[repr(u8)]
+pub enum ProgressPhase {
+    /// Parsing and extracting JUMBF manifest data from an asset stream (I/O phase).
+    Reading = 0,
+    /// Verifying the structure and integrity of a manifest store entry.
+    VerifyingManifest = 1,
+    /// Verifying a COSE cryptographic signature and certificate chain for a claim.
+    /// Fires twice per claim: once before COSE parse (`step=1`) and once after
```
mauricefisher64, line 45: Do we need to say how many times it fires.  I don't think anyone cares and I doubt they will even see it since it is so fast.  I think what they really care about is that things keep moving (i.e. that the value changes) and they have a way to cancel.  I think adding the count to docs will confuse things since the number is actually variable depending on the contents of the manifest.

sdk/src/context.rs
```diff
@@ -24,6 +27,71 @@ use crate::{
...
+///
+/// `#[repr(u8)]` ensures stable integer values for FFI and WASM consumers.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+#[repr(u8)]
+pub enum ProgressPhase {
+    /// Parsing and extracting JUMBF manifest data from an asset stream (I/O phase).
+    Reading = 0,
+    /// Verifying the structure and integrity of a manifest store entry.
+    VerifyingManifest = 1,
```
mauricefisher64, line 43: Some of these values will be nested for example VeryifyManifest contains most of the phases below.  So how do I represent validating ingredient 2 of 5 and hashing  35%.   That is the way the current structure reads but there is no way to represent that.  The way I would have to do this is all the outer layers would use the indeterminant count case.  I don't think it will matter much because all of the phases will fly by except hashing, I/O and network.

gpeacock: The phases are the important indicator of progress, and I know most of them will happen rapidly. I don't think we need a lot of them, but some will take some time. The 2 of 5 part is really only there for things like hashing where we would want to show progress through a phase. Most phases will always just be 1/1. If there isn't a practical way to get the total up front, I can remove that the total and we can indicat a step count. But if you have an idea of how many fragments we will hash, that could be valuable. I changed this to not report percents since I don't think we would ever get that one right.

sdk/src/context.rs
gpeacock: I think we can make this one just one step. There's no need to add an extra where it won't ever matter.

docs/progress_callbacks.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `c2pa_context_builder_set_progress_callback(builder, user_data, callback)` – attach callback and user data
+- `c2pa_context_cancel(ctx)` – request cancellation
+
+**C callback:** `int (*)(void* context, uint8_t phase, float pct)`
+- Return non-zero to continue, zero to cancel
+- `phase` matches `ProgressPhase` discriminants (0–6)
+- `pct` is 0.0–1.0
+
+## When It Fires
```
tmathern:
```suggestion
## When it fires
```

sdk/src/context.rs
```diff
@@ -24,6 +27,71 @@ use crate::{
...
+/// Phases reported by the progress callback.
+///
+/// Passed to the progress callback registered on [`Context`] so callers can
+/// display progress indicators or make phase-specific cancellation decisions.
+///
+/// `#[repr(u8)]` ensures stable integer values for FFI and WASM consumers.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+#[repr(u8)]
+pub enum ProgressPhase {
```
tmathern, line 39: This is the outline of a state machine too! Could become a nice state diagram doc to show exactly where we are.

sdk/src/context.rs
```diff
@@ -192,6 +260,10 @@ pub struct Context {
     async_resolver: AsyncResolverState,
     signer: SignerState,
     async_signer: AsyncSignerState,
+    progress_callback: Option<Box<ProgressCallbackFunc>>,
+    /// Embedded cancellation flag.  Any thread holding an `Arc<Context>` can call
+    /// [`cancel()`](Context::cancel) without needing a separate token object.
+    cancel_flag: AtomicBool,
```
tmathern, line 282: 👍

sdk/src/context.rs
```diff
@@ -617,6 +691,89 @@ impl Context {
...
+    /// let ctx2 = ctx.clone();
+    /// // std::thread::spawn(move || ctx2.cancel());
+    /// ```
+    pub fn cancel(&self) {
+        self.cancel_flag.store(true, Ordering::Relaxed);
+    }
+
+    /// Returns `true` if [`cancel()`](Context::cancel) has been called.
+    pub fn is_cancelled(&self) -> bool {
+        self.cancel_flag.load(Ordering::Relaxed)
```
tmathern: I think this is too relaxed if multithreaded? I'd use `Ordering::Acquire` to avoid have a Release/Acquire model. More strict, more overhead, but probably the most reliable also. And I assume for long-running operations we want to make sure they get cancelled when they should!

sdk/src/context.rs
```diff
@@ -617,6 +691,89 @@ impl Context {
...
+    /// # use c2pa::Context;
+    /// # use std::sync::Arc;
+    /// let ctx = Arc::new(Context::new());
+    ///
+    /// // Hand a clone to a background thread; call cancel() to abort.
+    /// let ctx2 = ctx.clone();
+    /// // std::thread::spawn(move || ctx2.cancel());
+    /// ```
+    pub fn cancel(&self) {
+        self.cancel_flag.store(true, Ordering::Relaxed);
```
tmathern: I think this is too relaxed. I'd use `Ordering::Release` if the overhead may be acceptable.

sdk/src/reader.rs
```diff
@@ -200,6 +200,9 @@ impl Reader {
     ) -> Result<Self> {
         let mut validation_log = StatusTracker::default();
         stream.rewind()?; // Ensure stream is at the start
+
+        self.context.check_progress(ProgressPhase::Reading, 1, 1)?;
```
tmathern, line 205: I wonder if for check_progress, it wouldn't be wiser to replace the two last params with a type like `PhaseDetails` or `ProgressDetails`, holding those two.
1. In case you need to add more details (no signature changes)
2. It makes the signature a tiny little bit easier to read if typed (I had to check which param is what here)

gpeacock: I did the integers to keep the API easy for C, if I make it a struct, we have to export that struct and figure out how to represent it. I don't really want to go with JSON serializing for something like progess info unless there's a need.

tmathern: You are right. Sounds good, I hope it won't ever ebcome more "verbose". We'll cross that bridge then.

sdk/src/context.rs
gpeacock: You can also specify a 0 total for indeterminate steps

sdk/src/context.rs
gpeacock: yes it works very well for that!

sdk/src/context.rs
gpeacock: Good catch! I updated this.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -93,6 +94,23 @@ mod cbindgen_fix {
...
 
+/// Wraps a `*const c_void` (C `void*`) so it can be moved into a `Send + Sync` closure.
+///
+/// Accessing through [`as_ptr()`](SendPtr::as_ptr) rather than the raw `.0` field
+/// ensures the Rust 2021 disjoint-capture analysis captures the whole `SendPtr`
+/// (which is `Send + Sync`) rather than the inner `*const c_void` (which is not).
+///
+/// # Safety
+/// The caller must guarantee the pointer is valid for the closure's lifetime.
+struct SendPtr(*const c_void);
```
tmathern, line 105: Interesting

docs/progress_callbacks.md
```diff
@@ -0,0 +1,383 @@
+# Progress and Cancellation API
+
+## Overview
+
+The SDK reports progress during long-running operations (signing, reading, verification) via an optional callback registered on a `Context`. The primary purposes of the callback are:
```
tmathern:
```suggestion
The SDK reports progress during multistep long-running operations (signing, reading, verification) via an optional callback registered on a `Context`. The primary purposes of the callback are:
```

tmathern (approved):

PR #1936, opened 2026-03-16
feat: Allow emscripten users to define HTTP handler

cdmurph32:
Follow-up to the Emscripten build support added in c2pa-rs.

Emscripten does not support reqwest or ureq, the standard http handlers in the Rust SDK. The Rust SDK supports users defining their own http handlers through the context. We need to expose this in the C Bindings.

## Checklist


c2pa_c_ffi/src/c_api.rs
```diff
@@ -88,6 +88,10 @@ mod cbindgen_fix {
     #[repr(C)]
     #[allow(dead_code)]
     pub struct C2paSettings;
+
+    #[repr(C)]
+    #[allow(dead_code)]
```
ok-nick, line 94: Why are these `#[allow(dead_code)]`?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,128 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+    context: *mut c_void,
+    url: *const c_char,
+    method: *const c_char,
+    headers: *const c_char,
+    request_body: *const c_uchar,
+    request_body_len: usize,
+    response_data: *mut c_uchar,
+    response_capacity: usize,
+    response_status: *mut i32,
+) -> i32;
```
ok-nick: I think it would make more sense to have a Request/Response struct with these fields like the Rust SDK. This is a lot of parameters.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,128 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+        let body = request.into_body();
+        let (body_ptr, body_len) = if body.is_empty() {
+            (std::ptr::null(), 0usize)
+        } else {
+            (body.as_ptr(), body.len())
+        };
+
+        let mut capacity: usize = 64 * 1024;
+        loop {
+            let mut buf: Vec<u8> = vec![0u8; capacity];
```
ok-nick: Hm, would it make more sense to return a C FFI stream? Seems like we can run into issues setting a hard limit like this.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,128 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+
+            if written < 0 {
+                let needed = (-written) as usize;
+                if needed > capacity {
+                    capacity = needed;
+                    continue;
+                }
+                let msg = CimplError::last_message()
+                    .unwrap_or_else(|| "HTTP callback returned error".to_string());
+                return Err(HttpResolverError::Io(std::io::Error::other(msg)));
```
ok-nick: Can we return `HttpResolverError::Other` instead.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,128 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+            }
```
ok-nick: This behavior seems confusing, I'm not fully understanding how it works, can you explain it?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,128 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+        }
+    }
+}
+
+/// Opaque handle for a C-callback-based HTTP resolver.
+/// Created by `c2pa_http_resolver_create()`. Either consumed by
+/// `c2pa_context_builder_set_http_resolver()` or freed via `c2pa_free()`.
+pub struct C2paHttpResolver {
+    inner: CHttpResolver,
+}
```
ok-nick, line 422: How come we have a wrapper struct for this? Wouldn't it look like the same opaque handle to the caller if they use `CHttpResolver` directly?

c2pa_c_ffi/examples/emscripten_example.cpp
ok-nick, line 1: Do we want to add a cpp examples in this repo? This feels more like a c2pa-c example? (cc @gpeacock)

c2pa_c_ffi/src/c_api.rs
cdmurph32: The structs in cbindgen_fix exist solely for cbindgen. They aren't used by the Rust code

c2pa_c_ffi/src/c_api.rs
cdmurph32: For me, it was easier conceptually to have this defined on the C side. I'll push a change with it the other way and we can see.

c2pa_c_ffi/src/c_api.rs
cdmurph32: This is a good reason for having a Request/Response struct. This part is much easier.

c2pa_c_ffi/src/c_api.rs
cdmurph32: It's completely different with the Request/Response struct

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,147 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+pub struct C2paHttpResponse {
+    /// HTTP status code (e.g. 200, 404)
+    pub status: i32,
+    /// Response body bytes, allocated by the callback.
+    pub body: *mut c_uchar,
+    /// Length of `body` in bytes
+    pub body_len: usize,
+    /// Deallocator for `body`. Called by Rust after copying the data.
+    /// May be NULL if `body` does not need to be freed.
+    pub body_free: Option<unsafe extern "C" fn(*mut c_uchar)>,
```
ok-nick: Is it safe to just call `libc::free`?

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,147 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+    fn http_resolve(
+        &self,
+        request: c2pa::http::http::Request<Vec<u8>>,
+    ) -> Result<c2pa::http::http::Response<Box<dyn std::io::Read>>, c2pa::http::HttpResolverError>
+    {
+        use std::ffi::CString;
+
+        use c2pa::http::HttpResolverError;
+
+        let uri_cstr = CString::new(request.uri().to_string())
```
ok-nick: Not a blocker here, but it would be clean if `impl TryFrom<Request> for C2paHttpRequest` and same for response.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,147 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+                .unwrap_or_else(|| "HTTP callback returned error".to_string());
+            return Err(HttpResolverError::Other(msg.into()));
+        }
+
+        let response = c2pa::http::http::Response::builder()
+            .status(c_response.status as u16)
+            .body(Box::new(std::io::Cursor::new(response_body)) as Box<dyn std::io::Read>)
+            .map_err(HttpResolverError::Http)?;
+        Ok(response)
+    }
```
ok-nick: Would it be possible to return a `Cursor` over `c_response.body` rather than cloning it into a vec? We can free it through `impl Drop for C2paHttpResponse`.

c2pa_c_ffi/src/c_api.rs
cdmurph32: Yes. I'll add libc.

c2pa_c_ffi/src/c_api.rs
cdmurph32: I'll give it a shot.

c2pa_c_ffi/src/c_api.rs
cdmurph32: Yes.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,186 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+}
+
+impl C2paHttpResponse {
+    /// Consumes the C-allocated response body and converts it into an
+    /// `http::Response`. Calls `free()` on the body pointer.
+    ///
+    /// # Safety
+    /// `self.body` must have been allocated with `malloc()` (or be null).
+    unsafe fn into_http_response(
```
ok-nick: Can also implement TryFrom here too.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -217,6 +221,186 @@ pub type SignerCallback = unsafe extern "C" fn(
...
+    pub body: *mut c_uchar,
+    /// Length of `body` in bytes
+    pub body_len: usize,
+}
+
+/// Owns the backing storage for a [`C2paHttpRequest`].
+///
+/// Use [`as_ffi`](OwnedC2paHttpRequest::as_ffi) to obtain the `#[repr(C)]` view
+/// whose pointers borrow from this struct.
+struct OwnedC2paHttpRequest {
```
ok-nick, line 280: How come you split OwnedC2paHttpRequest and C2paHttpRequest into separate structs?

c2pa_c_ffi/examples/emscripten_example.cpp
ok-nick: It seems like we might also be able to reuse a lot of structs defined there.

c2pa_c_ffi/src/c_api.rs
cdmurph32: C2paHttpRequest needs `*const c_char fields` so that cbindgen can generate the bindings.
OwnedC2paHttpRequest owns the CString data (Box<[u8]>).

c2pa_c_ffi/examples/emscripten_example.cpp
tmathern: (That is currently more C in spirit than C++)

c2pa_c_ffi/examples/emscripten_example.cpp
```diff
@@ -0,0 +1,237 @@
+// Example: Using c2pa library in an Emscripten C++ project
```
tmathern: c2pa.h is the C FFI directly, so I'd say it shows more how to use the C FFI library.

```suggestion
// Example: Using c2pa C library in an Emscripten C++ project
```

tmathern: We could make the case to move this to c2pa-c which is the C++ repo is people want example on eg. how to use the C++ classes, and add the additional includes. (This is similar to the note @ok-nick left about reusing C++ wrapper from c2pa-c, I believe. It makes it more C++ like than C like)

cdmurph32: Sounds goot

PR #1951, opened 2026-03-18
fix: Limit the maximum number of assertions allowed for C2PA Manifest

ssanthosh:
# Security Fix: Unbounded Assertion Count Allows Resource Exhaustion

## Issue

No limit exists on the number of assertions that can be added to a manifest across three distinct code paths:

1. **`Builder::add_assertion` / `add_assertion_json`** (`builder.rs`) — the public API for constructing manifests programmatically.
2. **`Claim::add_assertion`** (`claim.rs`) — the lower-level API callable directly, bypassing `Builder`.
3. **`Store::from_jumbf_impl`** (`store.rs`) — the parsing path that processes assertions from untrusted, existing C2PA files.

An attacker can exploit paths 1 and 2 by constructing a manifest with an arbitrarily large number of assertions, and path 3 by supplying a crafted C2PA file with excessive assertions. All three cause uncontrolled memory and CPU consumption — a DoS vector for any application that creates or reads C2PA content.

---

### New error variant (`sdk/src/error.rs`)

```
TooManyAssertions { max: usize }
```

Returned by all three enforcement points when the limit is exceeded.

### Shared constant

`pub(crate) const MAX_ASSERTIONS: usize = 100_000` is declared once in `sdk/src/settings/mod.rs` and imported directly by all three enforcement points. There are no configurable settings fields or thread-local accessors — all modules use the constant directly for consistency.

### Limit enforcement

| File | Where | Limit source |
|---|---|---|
| `sdk/src/builder.rs` | `check_assertion_limit()` helper called by `add_assertion` and `add_assertion_json` | `MAX_ASSERTIONS` constant |
| `sdk/src/claim.rs` | `add_assertion_impl` | `MAX_ASSERTIONS` constant |
| `sdk/src/store.rs` | `from_jumbf_impl`, before the assertion parsing loop | `MAX_ASSERTIONS` constant |

Each point returns `Err(TooManyAssertions { max: MAX_ASSERTIONS })` before processing begins when the count meets or exceeds the limit.

---

### `builder.rs` — `test_add_assertion_limit`
- Adds assertions up to `MAX_ASSERTIONS` — all succeed.
- Verifies the next one returns `TooManyAssertions { max: MAX_ASSERTIONS }`.
- Completes in ~0.02 seconds.

---
## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -48,7 +48,7 @@ use crate::{
     jumbf_io,
     maybe_send_sync::MaybeSend,
     resource_store::{ResourceRef, ResourceResolver, ResourceStore},
-    settings::builder::TimeStampFetchScope,
+    settings::{builder::TimeStampFetchScope, MAX_ASSERTIONS},
     store::Store,
     utils::{hash_utils::hash_to_b64, merkle::MerkleAccumulator, mime::format_to_mime},
     AsyncSigner, ClaimGeneratorInfo, EphemeralSigner, HashRange, HashedUri, Ingredient,
@@ -803,9 +803,10 @@ impl Builder {
     /// Adds a CBOR assertion to the manifest.
     /// In most cases, use this function instead of `add_assertion_json`, unless the assertion must be stored in JSON format.
     fn check_assertion_limit(&self) -> Result<()> {
-        let max = self.context.settings().builder.max_assertions;
-        if self.definition.assertions.len() >= max {
-            return Err(Error::TooManyAssertions { max });
+        if self.definition.assertions.len() >= MAX_ASSERTIONS {
+            return Err(Error::TooManyAssertions {
+                max: MAX_ASSERTIONS,
+            });
         }
         Ok(())
     }
@@ -3396,7 +3397,7 @@ mod tests {
         crypto::raw_signature::SigningAlg,
         hash_stream_by_alg,
         maybe_send_sync::MaybeSend,
-        settings::{Settings, MAX_ASSERTIONS},
+        settings::Settings,
         utils::{
             hash_utils::HashRange,
             test::{test_context, write_bmff_placeholder_stream, write_jpeg_placeholder_stream},
@@ -7491,32 +7492,6 @@ mod tests {
                 max: MAX_ASSERTIONS
             }
         ));
-    }
-
-    #[test]
-    fn test_add_assertion_limit_is_configurable() {
-        let data = serde_json::json!({"value": 1});
-        let settings = Settings {
-            builder: crate::settings::builder::BuilderSettings {
-                max_assertions: 2,
-                ..Default::default()
-            },
-            ..Default::default()
-        };
-        let context = Context::new()
-            .with_settings(settings)
-            .expect("valid settings");
-        let mut builder = Builder::from_context(context);
-        builder
-            .add_assertion_json("org.test.a1", &data)
-            .expect("first assertion should succeed");
-        builder
-            .add_assertion_json("org.test.a2", &data)
-            .expect("second assertion should succeed");
-        let err = builder
-            .add_assertion_json("org.test.a3", &data)
-            .expect_err("third assertion should exceed limit of 2");
-        assert!(matches!(err, Error::TooManyAssertions { max: 2 }));
     }
 
     // Ensures that the future returned by `Builder::sign_async` implements `Send`, thus making it
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -74,7 +74,7 @@ use crate::{
     log_item,
     resource_store::UriOrResource,
     salt::{DefaultSalt, SaltGenerator},
-    settings::Settings,
+    settings::{Settings, MAX_ASSERTIONS},
     status_tracker::{ErrorBehavior, StatusTracker},
     store::StoreValidationInfo,
     utils::hash_utils::{hash_by_alg, vec_compare},
@@ -251,8 +251,6 @@ const ALG_SOFT_F: &str = "alg_soft";
 const METADATA_F: &str = "metadata";
 const CREATED_ASSERTIONS_F: &str = "created_assertions";
 const GATHERED_ASSERTIONS_F: &str = "gathered_assertions";
-
-use crate::settings::max_assertions;
 
 /// A `Claim` gathers together all the `Assertion`s about an asset
 /// from an actor at a given time, and may also include one or more
@@ -1431,10 +1429,9 @@ impl Claim {
     ) -> Result<C2PAAssertion> {
         // Enforce the per-manifest assertion limit to prevent resource exhaustion
         // regardless of how the claim is constructed.
-        let max_assertions = max_assertions();
-        if self.assertion_store.len() >= max_assertions {
+        if self.assertion_store.len() >= MAX_ASSERTIONS {
             return Err(Error::TooManyAssertions {
-                max: max_assertions,
+                max: MAX_ASSERTIONS,
             });
         }
 
@@ -4107,12 +4104,7 @@ pub mod tests {
     #![allow(clippy::unwrap_used)]
 
     use super::*;
-    use crate::{
-        assertions::{Action, Actions},
-        resource_store::UriOrResource,
-        settings,
-        utils::test::create_test_claim,
-    };
+    use crate::{resource_store::UriOrResource, utils::test::create_test_claim};
 
     #[test]
     fn test_build_claim() {
@@ -4291,33 +4283,5 @@ pub mod tests {
         }
 
         Ok(())
-    }
-
-    #[test]
-    fn test_add_assertion_default_limit_in_claim() {
-        // Use a small limit so the test runs quickly regardless of the default.
-        const TEST_LIMIT: usize = 5;
-        settings::set_settings_value("verify.max_assertions", TEST_LIMIT as u64).unwrap();
-
-        let mut claim = Claim::new("test_default_assertion_limit", Some("test"), 2);
-        let actions = Actions::new().add_action(Action::new("c2pa.created"));
-
-        // TEST_LIMIT assertions must succeed within the configured limit.
-        for _ in 0..TEST_LIMIT {
-            claim
-                .add_assertion(&actions)
-                .expect("assertion should succeed within the configured limit");
-        }
-
-        // The next one must be rejected.
-        let err = claim
-            .add_assertion(&actions)
-            .expect_err("assertion beyond limit should fail with TooManyAssertions");
-        assert!(
-            matches!(err, Error::TooManyAssertions { max: TEST_LIMIT }),
-            "expected TooManyAssertions {{ max: {TEST_LIMIT} }}, got {err:?}"
-        );
-
-        settings::reset_default_settings().unwrap();
     }
 }
--- a/sdk/src/error.rs
+++ b/sdk/src/error.rs
@@ -120,7 +120,7 @@ pub enum Error {
     #[error("more than one manifest store detected")]
     TooManyManifestStores,
 
-    #[error("assertion limit exceeded: maximum allowed is {max}; adjust verify.max_assertions in Settings to change this limit")]
+    #[error("assertion limit exceeded: maximum allowed is {max}")]
     TooManyAssertions { max: usize },
 
     #[error("manifest is not refernced by any ingredient")]
--- a/sdk/src/settings/builder.rs
+++ b/sdk/src/settings/builder.rs
@@ -583,18 +583,6 @@ pub struct BuilderSettings {
     ///
     /// [`TimeStamp`]: crate::assertions::TimeStamp
     pub auto_timestamp_assertion: TimeStampSettings,
-    /// Maximum number of assertions that can be added to a single manifest.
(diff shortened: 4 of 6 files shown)


sdk/src/settings/builder.rs
```diff
@@ -583,6 +583,18 @@ pub struct BuilderSettings {
     ///
     /// [`TimeStamp`]: crate::assertions::TimeStamp
     pub auto_timestamp_assertion: TimeStampSettings,
+    /// Maximum number of assertions that can be added to a single manifest.
+    ///
+    /// Limits resource consumption when processing untrusted manifests with a large number
+    /// of assertions. Calls to [`Builder::add_assertion`] or [`Builder::add_assertion_json`]
+    /// that would exceed this limit return [`Error::TooManyAssertions`].
+    ///
+    /// The default value is 50.
```
tmathern: Why 50?

sdk/src/builder.rs
```diff
@@ -779,6 +779,12 @@ impl Builder {
         S: Into<String>,
         T: Serialize,
     {
+        let max_assertions = self.context.settings().builder.max_assertions;
```
tmathern: Optional: If doing twice exactly the same check, maybe it could be refactored?

sdk/src/builder.rs
```diff
@@ -5773,6 +5785,47 @@ mod tests {
         assert!(reader.active_manifest().is_some());
     }
 
+    #[test]
+    fn test_add_assertion_limit() {
+        use crate::settings::Settings;
```
tmathern: I prefer all exports on top at the test suite

sdk/src/builder.rs
```diff
@@ -5773,6 +5785,47 @@ mod tests {
...
+                .add_assertion_json(format!("org.test.assertion.{i}"), &data)
+                .expect("should succeed within limit");
+        }
+        let err = builder
+            .add_assertion_json("org.test.assertion.overflow", &data)
+            .expect_err("51st assertion should be rejected");
+        assert!(matches!(err, Error::TooManyAssertions { max: 50 }));
+
+        // Verify the limit is configurable: set max_assertions=2 via settings.
+        let settings = Settings {
```
tmathern: This should be another test, test_add_assertion_limit_is_configurable

sdk/src/claim.rs
```diff
@@ -1427,6 +1427,15 @@ impl Claim {
         salt_generator: &impl SaltGenerator,
         add_as_created_assertion: bool,
     ) -> Result<C2PAAssertion> {
+        // Enforce the per-manifest assertion limit during signing to prevent
+        // resource exhaustion regardless of how the claim is constructed.
+        let max_assertions = get_thread_local_settings().builder.max_assertions;
```
tmathern: I get the reason for this is that there is no context here or whatsoever to read the setting. But thread_local settings are another settings API that is deprecated. This likely needs to be weaved through differently.

sdk/src/store.rs
```diff
@@ -1377,6 +1377,15 @@ impl Store {
 
             let num_assertions = assertion_store_box.data_box_count();
 
+            // Reject manifests that embed more assertions than the configured limit to
+            // prevent unbounded memory and CPU consumption on untrusted input.
+            let max_assertions = get_thread_local_settings().verify.max_assertions;
```
tmathern: I get the reason for this is that there is no context here or whatsoever to read the setting. But thread_local settings are another settings API that is deprecated. This likely needs to be weaved through differently.

sdk/src/builder.rs
gpeacock: or at the top of the test module in this case

sdk/src/claim.rs
gpeacock: If we are using a setting it should only be one from the context. Perhaps the answer is we just set a large fixed constant for this. If we really need to configure it in claim, then we may want to consider adding a way to set a max assertions value in claim from the current settings. But I think a constant may be fine. The other approach is to add context or settings to Claim but that feels like overkill for this.

sdk/src/settings/builder.rs
ssanthosh: Value was suggested by @mauricefisher64 as appropriate limit

sdk/src/builder.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/1951/commits/db0dc521847ecb84f8830a1b9fe774d57ac50791

sdk/src/builder.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/db0dc521847ecb84f8830a1b9fe774d57ac50791

sdk/src/builder.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/db0dc521847ecb84f8830a1b9fe774d57ac50791

sdk/src/claim.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/db0dc521847ecb84f8830a1b9fe774d57ac50791

sdk/src/store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/db0dc521847ecb84f8830a1b9fe774d57ac50791

sdk/src/settings/mod.rs
```diff
@@ -392,6 +403,7 @@ impl Default for Verify {
             remote_manifest_fetch: true,
             skip_ingredient_conflict_resolution: false,
             strict_v1_validation: false,
+            max_assertions: 50,
```
gpeacock: Declare the MAX_ASSERTIONS const once in this settings mod.rs file, and make it pub(crate) Then use it here and in Store and Claim, so we have the same default constant used everywhere.

sdk/src/store.rs
```diff
@@ -110,6 +110,12 @@ pub(crate) struct StoreValidationInfo<'a> {
     pub certificate_statuses: HashMap<String, Vec<Vec<u8>>>, // list of certificate status assertions for each serial
 }
 
+/// Maximum number of assertions allowed when parsing an untrusted manifest.
+/// `from_jumbf_impl` does not receive a [`Context`][crate::Context], so a fixed constant is used
+/// rather than `Verify::max_assertions` from settings.  The settings field is kept for future use
+/// when context is properly propagated through the parsing path.
+const MAX_ASSERTIONS: usize = 50;
```
gpeacock: This declaration should be shared.

sdk/src/store.rs
```diff
@@ -1377,6 +1383,14 @@ impl Store {
 
             let num_assertions = assertion_store_box.data_box_count();
 
+            // Reject manifests that embed more assertions than the configured limit to
+            // prevent unbounded memory and CPU consumption on untrusted input.
+            if num_assertions > MAX_ASSERTIONS {
```
gpeacock: use a shared default max from Settings.

sdk/src/settings/mod.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/1951/commits/8929f9b0c6c2f0ac50d3b40173b41dab21ac2637

sdk/src/store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/8929f9b0c6c2f0ac50d3b40173b41dab21ac2637

sdk/src/store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/8929f9b0c6c2f0ac50d3b40173b41dab21ac2637

ssanthosh:
@gpeacock : Performed the requested changes. Please review

sdk/src/settings/mod.rs
```diff
@@ -34,6 +34,11 @@ use crate::{
 
 const VERSION: u32 = 1;
 
+/// Default maximum number of assertions allowed per manifest.
+/// Shared by [`BuilderSettings`], [`Verify`], [`crate::Claim`], and [`crate::Store`] so that
+/// all enforcement points use the same value.
+pub(crate) const MAX_ASSERTIONS: usize = 50;
```
ok-nick: Do we need a const here if it's defined in settings?

sdk/src/settings/builder.rs
```diff
@@ -583,6 +583,18 @@ pub struct BuilderSettings {
...
+    ///
+    /// [`Builder::add_assertion`]: crate::Builder::add_assertion
+    /// [`Builder::add_assertion_json`]: crate::Builder::add_assertion_json
+    /// [`Error::TooManyAssertions`]: crate::Error::TooManyAssertions
+    pub max_assertions: usize,
```
ok-nick: I'm still not sure if we need to be constraining the number of assertions that can be added to a manifest. That is defined by the developer already. As long as we limit the number of manifests read (and furthermore validated), such as when reading an ingredient to be added to the builder, it would address the security vulnerability. Not a blocker but opening for further discussion.

sdk/src/store.rs
```diff
@@ -1377,6 +1379,14 @@ impl Store {
...
+                return Err(Error::TooManyAssertions {
+                    max: MAX_ASSERTIONS,
+                });
+            }
+
```
ok-nick, line 1387: This can be a big problem because it's not configurable. The max assertions should be read from settings.

sdk/src/error.rs
```diff
@@ -120,6 +120,9 @@ pub enum Error {
     #[error("more than one manifest store detected")]
     TooManyManifestStores,
 
+    #[error("assertion limit exceeded: maximum allowed is {max}")]
```
ok-nick: Can we also suggest how to fix by mentioning the respective setting.

sdk/src/claim.rs
```diff
@@ -1427,6 +1429,14 @@ impl Claim {
         salt_generator: &impl SaltGenerator,
         add_as_created_assertion: bool,
     ) -> Result<C2PAAssertion> {
+        // Enforce the per-manifest assertion limit to prevent resource exhaustion
+        // regardless of how the claim is constructed.
+        if self.assertion_store.len() >= MAX_ASSERTIONS {
+            return Err(Error::TooManyAssertions {
+                max: MAX_ASSERTIONS,
+            });
+        }
```
ok-nick, line 1436: Same problem here as in the store, this should be read from settings.

sdk/src/settings/mod.rs
ssanthosh: Still to better have the default value as const

sdk/src/store.rs
ssanthosh: Updated the logic to read from Settings

sdk/src/error.rs
ssanthosh: Done

sdk/src/claim.rs
ssanthosh: Updated the logic to read from Settings

sdk/src/settings/builder.rs
tmathern: +1, same, not a blocker.

sdk/src/builder.rs
```diff
@@ -6260,6 +6269,53 @@ mod tests {
         assert!(reader.active_manifest().is_some());
     }
 
+    #[test]
+    fn test_add_assertion_limit() {
+        // Verify all MAX_ASSERTIONS assertions succeed and the next one is rejected.
+        let mut builder = Builder::new();
+        let data = serde_json::json!({"value": 1});
+        for i in 0..MAX_ASSERTIONS {
```
gpeacock, line 7481: How long does this test take to run? Another way to do it is to allocate a assertions array of MAX_ASSERTIONS length and assign that to builder.assertions. 
If it runs fast enough that we don't care, it may not be a concern.

sdk/src/settings/mod.rs
```diff
@@ -975,6 +992,19 @@ pub(crate) fn get_thread_local_settings() -> Settings {
     })
 }
 
+/// Returns the configured maximum number of assertions allowed per manifest.
+///
+/// Reads `verify.max_assertions` from thread-local settings. Returns the
+/// user-configured value when set, or [`MAX_ASSERTIONS`] (100,000) when
+/// no limit has been configured.
+pub(crate) fn max_assertions() -> usize {
```
gpeacock: thread local settings are deprecated, We shouldn't rely on them in any new code.

sdk/src/builder.rs
ssanthosh: It takes only 0.02 seconds to run the test and hence no perf concerns

sdk/src/settings/mod.rs
ssanthosh: Refactored the code to use the global constant `MAX_ASSERTIONS`  in Settings on all modules : builder, claim and store. Removed all configurability with context and thread local settings

gpeacock (approved):
We should revisit having configurable options for this later, but this will fix the attack for now, and that's what we need.

PR #1953, opened 2026-03-18
chore: speed up make test with optional nextest support and some othe…

gpeacock:
This will use nexttest if you have it installed
cargo install next-test
and that alone will make things much faster
It also includes some alignment of feature and target options to minimize build time.
I added a make test-sdk that will only test the sdk and not all of the examples and c2patool

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,8 @@ version = "0.78.4"
 c2pa = { path = "sdk", default-features = false }
 
 [profile.dev]
-opt-level = 1          # slightly better codegen than 0, much faster than 2
-debug = 1  
+opt-level = 1          # allows test code to run faster at a small compile time cost
+debug = 1              # line tables only for debugging - switch to 2 or debug for full symbols
 
 [profile.release]
 strip = true  # Automatically strip symbols from the binary.


Cargo.toml
```diff
@@ -17,6 +17,10 @@ version = "0.78.4"
 [workspace.dependencies]
 c2pa = { path = "sdk", default-features = false }
 
+[profile.dev]
+opt-level = 1          # slightly better codegen than 0, much faster than 2
+debug = 1
```
ok-nick: Isn't this the default?

Cargo.toml
```diff
@@ -17,6 +17,10 @@ version = "0.78.4"
 [workspace.dependencies]
 c2pa = { path = "sdk", default-features = false }
 
+[profile.dev]
+opt-level = 1          # slightly better codegen than 0, much faster than 2
```
ok-nick: Hm this would increase compile times on dev builds.

Makefile
```diff
@@ -13,17 +13,30 @@ else
 	endif
 endif
 
+# Use cargo-nextest if installed (faster), otherwise fall back to cargo test.
+# Install nextest with: cargo install --locked cargo-nextest
+CARGO_TEST := $(shell cargo nextest --version > /dev/null 2>&1 && echo "cargo nextest run" || echo "cargo test")
+
+# Common feature set used by check-docs, clippy, and test-local so that all
+# three share the same incremental build cache and avoid redundant recompilation.
+FEATURES := file_io,fetch_remote_manifests,add_thumbnails
```
ok-nick, line 22: IMO I think these should always run with --all-features so as much as possible is compiled/tested.

Cargo.toml
tmathern: Is compile time an issue already though?

Makefile
tmathern: Well, the C libs do not have all features either. It makes sense to test with that limited set too.
We could test with both limited and extended set.

gpeacock: This is better than what we had since docs and clippy were only testing with file_io before. I'm going to go ahead with this change and we can debate about activating all features - there may be some things to work out with that.

Cargo.toml
gpeacock: I'm updating the comments to make the rational and tradeoffs clearer.
[profile.dev]
opt-level = 1          # allows test code to run faster at a small compile time cost
debug = 1              # line tables only for debugging - switch to 2 or debug for full symbols

Issue #1911, opened 2026-03-08
Add compatibility testing workflow


s35-lee:
### Details

It would be useful to add compatibility tests to the CI workflows for c2pa-rs.

As the SDK evolves, a dedicated workflow could help detect compatibility issues early and prevent regressions.

My idea is to introduce a new Github Action workflow that run compatibility tests to ensure that change do not break existing behavior. This would help maintain both forward compatibility and backward compatibility.

<img width="1549" height="604" alt="Image" src="https://github.com/user-attachments/assets/fbf16523-ec4c-4802-a854-1782b7c013d7" />

Here is an example script that checks out the base code and the PR code in order to compare the changes introduced by the pull request.
```yaml
  - name: Checkout PR code
    uses: actions/checkout@v6
    with:
      ref: ${{ github.event.pull_request.head.sha }}
      path: pr
  - name: Checkout base code
    uses: actions/checkout@v6
    with:
      ref: ${{ github.event.pull_request.base.sha }}
      path: base
```

I think it would be useful to run this workflow on every PR, similar to the `tier-1a` in this repo.
If anyone has suggestions or concerns about this approach, please feel free to share your thoughts.

I'm happy to work on a draft implementation of the workflow if this approach makes sense.

PR #1960, opened 2026-03-22
Add compatibility check workflow

s35-lee:
I have added a new Github Action workflow to perform conpatibility checks between the base branch and the pull request branch.

Related Issue: #1911 

I'm aware there was earlier work in #513 .
My understanding is that the previous approach aimed for a much stricter comparison, including format-level checks with multiple asset types.
This PR intentionally takes a lighter approach. Instead of enforcing exact snapshot matches, it uses `c2patool` to run practical forward and backward compatibility checks.

One possible concern with the previous PR is that the workflow may fail if compatibility is broken due to spec changes.
That is also part of the motivation for this work.
If a spec or implementation change causes compatibility to break, that should be handled as a critical event.
And this is need to be detected as early as possible, before it impacts downstream solutions built on this repository.

There are two behavior for compatibility.
- Forward compatibility: `base` signs a sample asset, `PR` reads it.
- Backward compatibility: `PR` signs a sample asset, `base` reads it.

This PR uses trigger conditions similar to `tier-1a`, but I'm happy to adjust if needed.

## Checklist

### DIFF
--- a/.github/workflows/compatibility-check.yml
+++ b/.github/workflows/compatibility-check.yml
@@ -1,0 +1,99 @@
+name: Compatibility Check
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+      - v1_api
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+  push:
+    branches:
+      - main
+      - v1_api
+
+jobs:
+  compatibility:
+    name: ${{ matrix.direction.name }}
+    if: |
+      github.event_name != 'pull_request' ||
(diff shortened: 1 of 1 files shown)


scouten-adobe:
Closing for now as we're not ready to make stability guarantees for the format of c2patool output. We will consider doing something similar as we plan for our 1.0 release.

s35-lee:
@scouten-adobe 
Thank you for the answer.
I understand the concern about making stability guarantees.
I still believe that having a way to automatically check compatibility would be valuable.
I'd happy to revisit this dicussion after 1.0 release.

PR #1967, opened 2026-03-24
perf: optimize signing passes/copies for large PNGs

ok-nick:
These benchmarks were ran on a 1GB PNG streamed from disk on an M4 Max MacBook Pro. I noticed an improved performance of ~50% for large PNGs on my device. I expect these results to be much more significant on slower disks.

Before:
Screenshot 2026-03-24 at 10 47 55 AM
After:
Screenshot 2026-03-24 at 10 48 14 AM

￼1. The first change removes a call that reads the stream to a buffer in `get_object_locations`, which takes ~4.1% of the total sign time.
image

2. ￼The second change reduces additional passes and copies of the stream in `get_object_locations` when adding a dummy C2PA chunk, which takes ~40% of the total sign time. 
image


tmathern (commented):
- assets aren't corrupted (still viewable).

tmathern (approved):
Also may overlap with https://github.com/contentauth/c2pa-rs/pull/1974

mauricefisher64 (approved):

gpeacock (approved):
LGTM

PR #1970, opened 2026-03-24
perf: optimize signing passes/copies for large RIFFs

ok-nick:
These benchmarks were ran on a 1GB WAV streamed from disk on an M4 Max MacBook Pro. I noticed an improved performance of ~23% for large WAVs on my device. I expect these results to be much more significant on slower disks.

Before:
Screenshot 2026-03-24 at 11 33 06 AM
After:
Screenshot 2026-03-24 at 11 33 16 AM

1. Remove call to buffer stream in `get_object_locations`, which takes ~6.7% of the total sign time.
image

2. Reduce passes through stream in `get_object_locations`, which takes ~17.2% of the total sign time.
image


tmathern (commented):
- assets aren't corrupted (still viewable), and all pre-existing metadata is still here

tmathern (approved):

gpeacock (approved):

mauricefisher64 (approved):

PR #1971, opened 2026-03-24
perf: optimize signing passes/copies for large JPEGs

ok-nick:
These benchmarks were ran on a 1.5GB JPEG streamed from disk on an M4 Max MacBook Pro. I noticed an improved performance of ~22% for large JPEGs on my device. I expect these results to be much more significant on slower disks.

Before:
Screenshot 2026-03-24 at 11 44 48 AM
After:
Screenshot 2026-03-24 at 11 44 21 AM

1. Reduce copies/passes in `get_object_locations`, which takes ~32% of the total sign time.
image


sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -288,6 +253,7 @@ impl CAIWriter for JpegIO {
         let mut buf = Vec::new();
         // read the whole asset
         input_stream.rewind()?;
+        // TODO: optimize this by implementing a streaming parser
```
tmathern, line 260: There is nothing that exists we could reuse?

tmathern (commented):
- assets aren't corrupted (still viewable), and all pre-existing metadata is still here

tmathern (changes requested):
What is up with the benchmarks here? `Merging this PR will not alter performance`? Due to this being multipass but benchmarks aren't? Maybe have https://github.com/contentauth/c2pa-rs/pull/1974 go in first, solves the hot paths?

sdk/src/asset_handlers/jpeg_io.rs
tmathern: https://github.com/contentauth/c2pa-rs/pull/1974 seems to change things here.

gpeacock (approved):

sdk/src/asset_handlers/jpeg_io.rs
gpeacock: The two are complementary approaches. They do overlap the same areas but I'm ok with merging these first.

gpeacock: The two are complementary approaches. They do overlap the same areas but I'm ok with merging these first.

tmathern (approved):

sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -368,94 +334,104 @@ impl CAIWriter for JpegIO {
...
+
+        // TODO: optimize this by implementing a streaming parser
+        input_stream.read_to_end(&mut buf).map_err(Error::IoError)?;
+        let jpeg = Jpeg::from_bytes(buf.into())
+            .map_err(|_err| Error::InvalidAsset("Could not parse input JPEG".to_owned()))?;
 
-        let buf: Vec<u8> = output_stream.into_inner();
+        let jpeg_with_dummy = if get_cai_segments(&jpeg)?.is_empty() {
+            // create dummy JUMBF seg
+            let mut no_bytes = vec![0; 50]; // enough bytes to be valid
```
ok-nick, line 357: This was taken from the existing code:
https://github.com/contentauth/c2pa-rs/blob/05c526b4675bdb247364b333a541bc126bdd0473/sdk/src/asset_handlers/jpeg_io.rs#L95-L97

mauricefisher64 (approved):

PR #1974, opened 2026-03-24
perf: eliminate redundant full-file rewrite pass during signing

gpeacock:
perf: eliminate redundant full-file rewrite pass during signing

The signing pipeline previously always wrote the asset to an intermediate stream, then copied it to the output stream a second time after embedding the final JUMBF (Pass 4). For large files this means reading and writing the entire asset twice.

This PR eliminates Pass 4 for JPEG, PNG, BMFF (MP4/MOV/HEIC), RIFF (AVI/WAV/WebP), and TIFF/DNG by adding an AssetPatch::patch_cai_store_from_stream method to the AssetPatch trait. When the format supports it, save_to_stream writes the placeholder JUMBF directly to the output stream and patches only the manifest bytes in place after signing — a targeted seek-and-write rather than a full re-copy.

Formats that don't implement patch_cai_store_from_stream automatically fall back to the existing intermediate-stream path. BMFF update manifests and sidecar/remote manifests are also excluded from the fast path since they have structural reasons that make in-place patching unsafe.

As a side effect, RIFF reading and hash-position scanning are rewritten to use a lean O(chunk-count) header scanner instead of loading the entire file into RAM, fixing a VLC-encoded AVI compatibility issue (null-byte padding between chunks).

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -24,15 +24,12 @@ use std::{
     io::Write,
     path::{Path, PathBuf},
     str::FromStr,
-    sync::Arc,
-    time::Instant,
 };
 
 use anyhow::{anyhow, bail, Context, Result};
 use c2pa::{
     format_from_path, identity::validator::CawgValidator, settings::Settings, Builder,
-    ClaimGeneratorInfo, Context as C2paContext, Error, Ingredient, ManifestDefinition,
-    ProgressPhase, Reader, Signer,
+    ClaimGeneratorInfo, Error, Ingredient, ManifestDefinition, Reader, Signer,
 };
 use clap::{Parser, Subcommand};
 use etcetera::BaseStrategy;
@@ -118,14 +115,9 @@ struct CliArgs {
     #[clap(long = "certs")]
     cert_chain: bool,
 
-    /// Print progress phase timings to stderr during signing and reading.
-    ///
-    /// Each line written to stderr has the form:
-    ///   [  12.345ms] VerifyingSignature 1/2
-    ///
-    /// This does not affect the JSON written to stdout.
-    #[clap(long)]
-    progress: bool,
+    /// Do not perform validation of signature after signing.
+    #[clap(long = "no_signing_verify")]
+    no_signing_verify: bool,
 
     #[command(subcommand)]
     command: Option<Commands>,
@@ -400,16 +392,9 @@ fn blocking_get(url: &str) -> Result<String> {
     }
 }
 
-/// Build a shared [`C2paContext`] from CLI arguments.
-///
-/// Applies settings from the settings file and any trust sub-command options,
-/// then optionally attaches a progress callback (when `--progress` is set) that
-/// prints phase timings to stderr.
-fn build_context(args: &CliArgs) -> Result<Arc<C2paContext>> {
-    let mut settings = Settings::new();
-
+fn configure_sdk(args: &CliArgs) -> Result<()> {
     if args.settings.exists() {
-        settings = settings.with_file(&args.settings)?;
+        Settings::from_file(&args.settings)?;
     }
 
     let mut enable_trust_checks = false;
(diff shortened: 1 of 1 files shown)


sdk/src/asset_handlers/jpeg_io.rs
```diff
@@ -579,6 +583,114 @@ impl AssetIO for JpegIO {
     }
 }
 
+/// Lean O(metadata) JPEG marker scanner that locates the contiguous C2PA APP11
```
tmathern, line 586: I think this may always be contiguous, but is not a MUST. Fine for now!

sdk/src/store.rs
```diff
@@ -2992,17 +2992,76 @@ impl Store {
...
+        // Also probe that output_stream supports reading. Callers may pass a
+        // write-only file (e.g. fs::File::create) whose read would return EBADF.
+        // The inplace path reads back from output_stream for hash computation,
+        // so it needs a readable stream. A 1-byte probe detects this cheaply;
+        // Ok(_) means readable (even if the file is currently empty), Err(_)
+        // means write-only → fall back to the intermediate-stream path.
+        let output_is_readable = {
+            let mut probe = [0u8; 1];
+            output_stream.read(&mut probe).is_ok()
+        };
```
ok-nick, line 3029: Hm, do we need this check? Is there ever a case where the output stream isn't readable? I assume in many places we are already reading from the output stream anyways.

sdk/src/store.rs
```diff
@@ -2992,17 +2992,76 @@ impl Store {
...
+            && !is_bmff_update_manifest
+            && !is_sidecar_or_remote
+            && get_assetio_handler(format)
+                .and_then(|h| h.asset_patch_ref())
+                .map(|p| p.patch_from_stream_supported())
+                .unwrap_or(false);
+
+        // For formats without in-place patching, use the traditional intermediate
+        // stream so that output_stream is only touched once on success.
         let mut intermediate_stream = io_utils::stream_with_fs_fallback(threshold);
```
ok-nick, line 3041: If in place patch is enabled then do we need to create the intermediate stream?

sdk/src/store.rs
```diff
@@ -2992,17 +2992,76 @@ impl Store {
...
+            self.start_save_stream(
+                format,
+                input_stream,
+                &mut intermediate_stream,
+                signer.reserve_size(),
+                context,
+            )?
+        };
+
+        context.check_progress(ProgressPhase::Hashing, 1, 1)?;
```
ok-nick, line 3064: Are we hashing here?

sdk/src/store.rs
```diff
@@ -3041,9 +3106,11 @@ impl Store {
                     // Update the JUMBF like it would normally be done
                     jumbf_bytes = self.to_jumbf_internal(signer.reserve_size())?;
 
-                    // Intermediate stream goes to output, but still no embedding
-                    intermediate_stream.rewind()?;
-                    //std::io::copy(&mut intermediate_stream, output_stream)?;
+                    if !supports_inplace_patch {
+                        // Intermediate stream goes to output, but still no embedding
+                        intermediate_stream.rewind()?;
+                        //std::io::copy(&mut intermediate_stream, output_stream)?;
```
ok-nick, line 3112: Did you mean to leave this as a comment?

sdk/src/store.rs
```diff
@@ -3060,16 +3127,31 @@ impl Store {
...
-            &mut intermediate_stream,
-            output_stream,
-            sig,
-            &sig_placeholder,
-        ) {
+        match if supports_inplace_patch {
+            // Pass 4 eliminated: patch the signed JUMBF directly into output_stream.
+            // Flush before seeking so buffered write data is committed (important
+            // for FFI-backed streams that require flush before seek/rewind).
+            output_stream.flush()?;
```
ok-nick, line 3134: This sounds like something we should be doing in the FFI layer for every seek/rewind.

sdk/src/store.rs
```diff
@@ -3172,13 +3263,19 @@ impl Store {
...
             input_stream.rewind()?;
             std::io::copy(input_stream, &mut intermediate_stream)?;
         }
 
-        let is_bmff = is_bmff_format(format);
+        // `source_is_intermediate` tracks whether intermediate_stream (true) or
+        // input_stream (false) should be used as the asset source for this call.
+        // When fast_path is true we skip the intermediate copy for ALL formats, not
+        // just BMFF, saving one full-file read+write pass.
+        let mut source_is_intermediate = !fast_path;
```
ok-nick, line 3278: Can we say:
```rust
let input_stream = intermediate_stream;
```
then? That may simplify the followup pieces of code.

PR #1975, opened 2026-03-25
fix: Updated crJSON Schema

lrosenthol:
Updated the crJSON schema with the version from C2PA, and fixed some implementation details & tests to match

## Checklist

### DIFF
--- a/.github/workflows/beta-preflight.yml
+++ b/.github/workflows/beta-preflight.yml
@@ -114,8 +114,7 @@ jobs:
       # doc tests.
 
       - name: Run doc tests (COVERAGE DISABLED)
-        run:
-          cargo test --workspace --all-features --doc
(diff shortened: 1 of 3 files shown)


sdk/tests/fixtures/schemas/crJSON-schema.json
```diff
@@ -208,7 +208,7 @@
                 },
                 "hash": {
                     "type": "string",
-                    "description": "Hash value (Base64-encoded)"
+                    "description": "Hash value (Base64-encoded with b64' prefix)"
```
gpeacock, line 211: does this require changing how we generate the hashes?

lrosenthol: Code is already in place - this is purely documentation update

PR #1981, opened 2026-03-26
feat: Replace atree dependency with indextree

MatrixA:
## Summary

Replace the `atree` crate (pinned at `=0.5.2`) with [`indextree`](https://crates.io/crates/indextree) v4 across the SDK and CLI.

**Motivation:**
- `atree` is unmaintained — last release in 2021, ~9K total downloads, no activity
- `atree` contains **undefined behavior**: uses `MaybeUninit::uninit().assume_init()` for non-ZST types in [`token.rs:31`](https://github.com/macthecadillac/atree/blob/master/src/token.rs#L31), which is a soundness bug
- `indextree` is actively maintained, has ~11M downloads, and provides an equivalent arena + token API

**Changes:**
- `sdk/Cargo.toml`, `cli/Cargo.toml`: `atree = "=0.5.2"` → `indextree = "4"`
- `sdk/src/asset_handlers/bmff_io.rs`: Migrate BMFF box tree operations (187 lines changed)
- `sdk/src/asset_handlers/tiff_io.rs`: Migrate TIFF IFD tree operations (73 lines changed)
- `cli/src/tree.rs`: Migrate CLI manifest tree display (31 lines changed)

**Key API translation patterns:**
| atree | indextree |
|-------|-----------|
| `Arena::with_data(d)` | `Arena::new()` + `arena.new_node(d)` |
| `token.append(&mut arena, data)` | `token.append_value(data, &mut arena)` |
| `token.append_node(&mut arena, other)` | `token.append(other, &mut arena)` |
| `node.data.field` | `node.get().field` / `node.get_mut().field` |
| `token.ancestors()` (excludes self) | `token.ancestors().skip(1)` |
| `token.children()` yields `&Node` | `token.children()` yields `NodeId` |

## Test plan

- [x] All bmff_io tests pass (17 tests)
- [x] All tiff_io tests pass (12 tests)
- [x] CLI tree tests pass (2 tests: unit + integration)
- [x] Full CI test suite


cli/Cargo.toml
```diff
@@ -27,8 +27,8 @@ networking = [
 
 [dependencies]
 anyhow = "1.0"
-atree = "=0.5.2"
```
ok-nick: Can you specify an explicit version here, `4.8.0`.

sdk/Cargo.toml
```diff
@@ -97,7 +97,7 @@ crate-type = ["lib"]
 asn1-rs = "0.7.1"
 async-generic = "1.1"
 async-trait = "0.1.78"
-atree = "=0.5.2"
```
ok-nick: Here too.

cli/Cargo.toml
```diff
@@ -27,8 +27,8 @@ networking = [
...
-c2pa = { path = "../sdk", version = "0.78.7", features = [
```
ok-nick: Why is the version downgraded?

cli/src/tree.rs
```diff
@@ -84,10 +84,11 @@ pub fn tree<P: AsRef<Path>>(path: P) -> Result<String> {
...
     Ok(if let Some(manifest_label) = reader.active_label() {
         let data = format!("Asset:{asset_name}, Manifest:{manifest_label}");
-        let (mut tree, root_token) = Arena::with_data(data);
-        populate_node(&mut tree, &reader, manifest_label, &root_token, false)?;
+        let tree = &mut Arena::new();
+        let root_token = tree.new_node(data);
+        populate_node(tree, &reader, manifest_label, root_token, false)?;
         // print tree
-        format!("Tree View:\n {}", walk_tree(&tree, &root_token))
+        format!("Tree View:\n {}", walk_tree(tree, root_token))
```
ok-nick: How does the new tree look compared to the old?

cli/src/tree.rs
```diff
@@ -84,10 +84,11 @@ pub fn tree<P: AsRef<Path>>(path: P) -> Result<String> {
     // walk through the manifests and show the contents
     Ok(if let Some(manifest_label) = reader.active_label() {
         let data = format!("Asset:{asset_name}, Manifest:{manifest_label}");
-        let (mut tree, root_token) = Arena::with_data(data);
-        populate_node(&mut tree, &reader, manifest_label, &root_token, false)?;
+        let tree = &mut Arena::new();
```
ok-nick: Why define this variable as `&mut Arena::new()` rather than `Arena::new()`?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1406,9 +1406,7 @@ pub(crate) fn build_bmff_tree<R: Read + Seek + ?Sized>(
                 };
 
                 let new_token = bmff_tree.new_node(b);
-                current_node
-                    .append_node(bmff_tree, new_token)
-                    .map_err(|_err| Error::InvalidAsset("Bad BMFF Graph".to_string()))?;
+                current_node.append(new_token, bmff_tree);
```
ok-nick: Can we use the checked variants, such as `checked_append`, etc.

ok-nick (commented):
Thank you for this change, much appreciated. I have a few comments.

cli/Cargo.toml
MatrixA: Done — pinned to `"4.8.0"` (latest stable, requires Rust 1.85.0; project MSRV is 1.88.0).

sdk/Cargo.toml
MatrixA: Done — pinned to `"4.8.0"` here as well.

cli/Cargo.toml
MatrixA: Good catch — the PR was branched before the release commit (`c1ea72d3 chore: release (#1973)`) that bumped the version to 0.78.7. Fixed now: version is back to `"0.78.7"`.

cli/src/tree.rs
MatrixA: The tree output is identical — both versions use the `treeline::Tree` crate for rendering. The only changes in `walk_tree()` are API translations (`.data` → `.get()`, `.is_leaf()` → `.first_child().is_none()`). The tree structure and traversal logic are unchanged. Existing tests verify the same assertions (`"Tree View:"`, `"Assertion:c2pa.actions"`).

cli/src/tree.rs
MatrixA: Good point — changed to `let mut tree = Arena::new()`, which is the idiomatic pattern and consistent with how Arena is used everywhere else in this PR (bmff_io.rs, tiff_io.rs).

sdk/src/asset_handlers/bmff_io.rs
MatrixA: Done — switched to `checked_append()` in all 4 locations (1 in bmff_io.rs, 3 in tiff_io.rs). The unchecked `append()` internally calls `checked_append().expect(...)` which panics on failure, whereas the old atree code returned `Result` via `.map_err()`. Using `checked_append()` restores that graceful error handling.

ok-nick (dismissed):
CI is failing, can you take a look at those errors?

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -496,17 +497,19 @@ where
...
             // move to next page
-            if let Some(next_ifd_offset) = tiff_tree[current_token].data.next_ifd_offset {
+            if let Some(next_ifd_offset) = tiff_tree[current_token].get().next_ifd_offset {
                 // move to next page
                 input.seek(SeekFrom::Start(next_ifd_offset))?;
                 let next_ifd =
                     TiffStructure::read_ifd(input, ts.byte_order, ts.big_tiff, IfdType::Page)?;
-                current_token = current_token.insert_after(&mut tiff_tree, next_ifd);
+                let next_token = tiff_tree.new_node(next_ifd);
+                current_token.insert_after(next_token, &mut tiff_tree);
```
ok-nick: Can we do a checked insert after here too?

ok-nick (approved):
Looks good pending one comment and approval from @mauricefisher64.

sdk/src/asset_handlers/tiff_io.rs
MatrixA: Done — switched to `checked_insert_after()` with the same `.map_err(|_err| Error::InvalidAsset("Bad TIFF Structure"...))` pattern used for the `checked_append()` calls.

MatrixA:
Heads up — I see #2089 has landed, which bumps atree to 0.5.4 and addresses the UB issue that motivated this PR. The "fix UB" framing in my original description is no longer accurate, so I want to re-state what this PR still offers:

**What's still on the table:**

- **Active maintenance** — atree 0.5.4 was a welcome one-off fix, but the crate has had no other releases since 2021. indextree has regular releases and an active issue tracker.
- **Safer API surface** — indextree provides `checked_append` / `checked_insert_after` variants that validate tree invariants (no cycles, no use-after-removal). Per @ok-nick's earlier review, this PR uses them throughout, restoring graceful error handling at every node insertion site.
- **Ecosystem alignment** — indextree (~11M downloads) vs atree (~9K) means future contributors are more likely to be familiar with the API.

**What's no longer a reason to merge:**
- ~~Fixing UB in atree~~ — addressed by #2089.

I'll defer to the maintainers on whether the remaining benefits justify the migration cost. Happy to update the PR description accordingly if it stays open, or close it if you'd prefer to consolidate on atree 0.5.4. Either way, glad #2089 closed the immediate risk.

scouten-adobe:
Given that `atree` crate has since been updated, we're going to continue to use it. Closing.

PR #1990, opened 2026-03-29
fix: Prevent O(2^N) exponential blowup in diamond DAG manifest validation

peekmoar:
## Summary

- Fix exponential O(2^N) manifest revisits in diamond DAG topologies by adding visited-node tracking to three recursive validation functions
- In a diamond DAG (manifest A references B and C, both referencing D), the recursive validation revisited already-processed manifests, causing 16,384 visits instead of ~28 at depth 14, leading to 30-70 minute hangs
- `get_claim_referenced_manifests_impl`: check `svi.manifest_map.contains_key()` before recursing (it already tracks visited manifests)
- `ingredient_checks` / `ingredient_checks_async`: add `visited: &mut HashSet<String>` parameter with early-exit via `visited.insert()`

## Test plan

- [x] Added regression test `test_diamond_dag_dedup.rs` that constructs a depth-8 diamond DAG and verifies it completes within 60 seconds (takes ~1.5s with the fix)
- [x] `cargo check` passes
- [x] `cargo test --test test_diamond_dag_dedup` passes


sdk/tests/test_diamond_dag_dedup.rs
```diff
@@ -0,0 +1,142 @@
+// Copyright 2024 Adobe. All rights reserved.
```
tmathern: Wrong year.

peekmoar:
Not rushing you but is there anything else that I need to do to get this change in?

sdk/tests/test_diamond_dag_dedup.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let image_data = generate_diamond_dag(depth)?;
+
+    let start = Instant::now();
+    let mut stream = Cursor::new(&image_data);
+    let reader = Reader::from_stream("image/jpeg", &mut stream)?;
+    let elapsed = start.elapsed();
+
+    let manifest_count = reader.iter_manifests().count();
+    assert!(manifest_count > 0, "should have parsed manifests");
```
tmathern: How many manifests should be parsed? Checks could be tightened here to make sure result stays the same even if code changes in future

tmathern: Also: It may be good to verify a Reader reports all expected checks in reader.validation_results()

sdk/tests/test_diamond_dag_dedup.rs
```diff
@@ -0,0 +1,142 @@
...
+    let mut dest = Cursor::new(Vec::new());
+    builder.sign(&signer, format, &mut source, &mut dest)?;
+
+    Ok(dest.into_inner())
+}
+
+/// Test that a diamond DAG at depth 8 completes in reasonable time.
+///
+/// Without the dedup fix, depth 8 would cause 2^8 = 256 manifest visits.
+/// With the fix, it should visit only ~17 unique manifests.
```
tmathern: Why the approximation? It should always be the same number at each run?

sdk/tests/test_diamond_dag_dedup.rs
```diff
@@ -0,0 +1,142 @@
...
+    let mut stream = Cursor::new(&image_data);
+    let reader = Reader::from_stream("image/jpeg", &mut stream)?;
+    let elapsed = start.elapsed();
+
+    let manifest_count = reader.iter_manifests().count();
+    assert!(manifest_count > 0, "should have parsed manifests");
+
+    // With the dedup fix, this should complete in well under 60 seconds.
+    // Without the fix at depth 8, it would take significantly longer due to 256 visits.
+    assert!(
```
mauricefisher64: I would prefer not to have timed unit test.  If we know this code works you can remove this test.

mauricefisher64:
I am OK with the changes.  Please address the PR comments and I will approve.

peekmoar:
Thanks for the reviews @tmathern @mauricefisher64. Pushed an update that reworks the test to address the feedback together:

- Removed the wall-clock assertion (per @mauricefisher64's preference for no timed tests).
- Replaced the loose `manifest_count > 0` check with an exact expected count (`1 + 2*(depth-1) + 1 = 16` at depth 8) (per @tmathern).
- Added assertions on `reader.validation_results()`: the number of `signingCredential.untrusted` entries across active + ingredient deltas must equal the unique manifest count, and no manifest URI may appear more than once. A dedup regression would either change the count or introduce duplicates, so it's caught deterministically without depending on time.
- Fixed the copyright year.

edit: There's a conflict, gimme a bit.

ale-adobe:
Hi @peekmoar, could you please update your branch to latest and resolve any conflicts? We would be happy to review again.

PR #2003, opened 2026-04-01
Update fetch.rs

yateeshbhardwaj:
At line 156, URL builder function was replacing the path of the URL because of the leading Slash in the request.str

Improved the URL builder function by using trim method

## Checklist
- [ ] This PR represents an update in the URL builder function at Line 149 in https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/crypto/ocsp/fetch.rs


yateeshbhardwaj (commented):
Please review the URL builder method as it is replacing the path in the request.url by forward slash in the request.str

ale-adobe:
Hi @yateeshbhardwaj, can you please provide a test case or example of the issue you're trying to fix with this change?

PR #2030, opened 2026-04-09
fix: Redaction follow-up fix

tmathern:
Not all thumbnails properly get redacted, as some are added differently.
This PR looks for that kind of thumbnails, and (tries) to make sure the redaction is taken into account on validation (no missing assertion error, since it is legitimately redacted).

## Checklist


sdk/src/ingredient.rs
```diff
@@ -1294,9 +1313,15 @@ impl Ingredient {
...
+            // If the identifier is a JUMBF URI, extract the manifest label from it.
+            // If it is a local resource ID (e.g. generated by add_thumbnails), fall back to
+            // the ingredient's own active manifest label so we still suppress the thumbnail
+            // when that manifest's thumbnails are being redacted.
             let thumbnail_is_redacted =
                 jumbf::labels::manifest_label_from_uri(&thumb_ref.identifier)
-                    .is_some_and(|label| thumbnail_redacted_manifests.contains(&label));
+                    .as_deref()
+                    .or(self.active_manifest.as_deref())
+                    .is_some_and(|label| thumbnail_redacted_manifests.contains(label));
```
ok-nick: It's not entirely clear to me here how we are determining if the thumbnail ID is a local resource ID?

tmathern: It is implicit through manifest_label_from_uri's return value.
If a local resource ID (a bare UUID or opaque string like "something-something-else"), it won't match the expected path structure and will return the rightful None. 
The .or(self.active_manifest.as_deref()) then kicks in, falling back to the ingredient's own active manifest label . Since a local resource ID is always relative to the manifest that owns the ingredient, this should work as fallback.

ok-nick: By "local resource ID" do you mean relative JUMBF URI? If that's the case, can we normalize it into an absolute URI then resolve it? I'd prefer if we can avoid the implicit behavior.

ok-nick (approved):

sdk/src/ingredient.rs
tmathern: Fair point. Updating, will be in next push.

PR #2031, opened 2026-04-09
fix: DLL hijacking vulnerability in c2patool (CAI-8608)

scouten-adobe:


cli/src/main.rs
```diff
@@ -630,6 +630,20 @@ fn print_reader(reader: &Reader, detailed: bool, crjson: bool) -> Result<()> {
...
+    // Harden against DLL hijacking on Windows by removing the current working
+    // directory from the DLL search path. Without this, Windows' default DLL
+    // search order includes the CWD, which allows an attacker to place a
+    // malicious DLL alongside the executable. LOAD_LIBRARY_SEARCH_DEFAULT_DIRS
+    // restricts the search to: the application directory, System32, and
+    // directories added via AddDllDirectory/SetDllDirectory — excluding the CWD.
+    #[cfg(windows)]
+    // SAFETY: no invariants to uphold; the argument is a valid constant.
+    unsafe {
+        windows_sys::Win32::System::LibraryLoader::SetDefaultDllDirectories(
```
tmathern: https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-setdefaultdlldirectories

`SetDefaultDllDirectories` has a return value, a bool indicating success. Shouldn't the result of this call be checked then?

Should not fail, but while hardening...

tmathern (approved):

gpeacock (approved):

PR #2032, opened 2026-04-10
chore: Deprecate thread-local settings APIs in favor of explicit Context

gpeacock:
Deprecate thread-local settings APIs in favor of explicit Context
This PR deprecates the legacy thread-local configuration APIs across the SDK and C FFI, replacing them with context-aware equivalents that accept an explicit Context object. No breaking changes — all deprecated methods retain their original behavior.

What's deprecated

SDK:
Builder::new(), Builder::from_json(), Builder::from_archive() → Builder::default().with_*()
Reader::from_stream(), Reader::from_file(), Reader::from_manifest_data_and_stream() → Reader::default().with_*()
Settings::from_toml(), Settings::signer() → Settings::new().with_*()
All public Ingredient constructors (none are context-aware)

C FFI:
c2pa_load_settings, c2pa_reader_from_stream, c2pa_reader_from_manifest_data_and_stream, c2pa_builder_from_json, c2pa_builder_from_archive, c2pa_signer_from_settings
c2pa_read_file, c2pa_read_ingredient_file, c2pa_sign_file
Type-specific free functions (c2pa_reader_free, c2pa_builder_free, c2pa_string_free, etc.) → c2pa_free()

Key changes
Builder::default() and Reader::default() are now the idiomatic way to construct with a default context (replacing the wordy Builder::from_context(Context::new()))
Builder::sign() replaced with save_to_stream() in tests where the signer is already configured on the context
cbindgen.toml configured to propagate Rust #[deprecated] to C/C++ headers via portable deprecation macros
Added test_context() and test_settings() test helpers to reduce boilerplate across ~50 test files
Tests updated throughout: deprecated APIs restricted to #[allow(deprecated)] blocks in their declaring modules; all other tests use the new context-based patterns

### DIFF
--- a/c2pa_c_ffi/cbindgen.toml
+++ b/c2pa_c_ffi/cbindgen.toml
@@ -73,7 +73,7 @@ after_includes = """
 [fn]
 prefix = "C2PA_API extern"
 deprecated = "C2PA_DEPRECATED"
-deprecated_with_note = "C2PA_DEPRECATED_MSG"
+deprecated_with_note = "C2PA_DEPRECATED_MSG({})"
 
 [export]
 # Exclude internal functions from C API
--- a/c2pa_c_ffi/cbindgen.toml
+++ b/c2pa_c_ffi/cbindgen.toml
@@ -52,13 +52,17 @@ after_includes = """
     #endif
 #endif
 
-/* Deprecation macros: produce compiler warnings when deprecated C2PA functions are called. */
-#if defined(__cplusplus) && __cplusplus >= 201402L
-    #define C2PA_DEPRECATED [[deprecated]]
-    #define C2PA_DEPRECATED_MSG(msg) [[deprecated(msg)]]
-#elif defined(__GNUC__) || defined(__clang__)
+/* Deprecation macros: produce compiler warnings when deprecated C2PA functions are called.
+ * GCC/Clang __attribute__ is checked first because it is position-flexible
+ * (works after 'extern'), whereas C++ [[deprecated]] is only valid at the
+ * very start of a declaration and cbindgen emits the annotation after the
+ * 'extern' keyword. */
+#if defined(__GNUC__) || defined(__clang__)
     #define C2PA_DEPRECATED __attribute__((deprecated))
     #define C2PA_DEPRECATED_MSG(msg) __attribute__((deprecated(msg)))
+#elif defined(__cplusplus) && __cplusplus >= 201402L
+    #define C2PA_DEPRECATED [[deprecated]]
+    #define C2PA_DEPRECATED_MSG(msg) [[deprecated(msg)]]
 #elif defined(_MSC_VER)
     #define C2PA_DEPRECATED __declspec(deprecated)
     #define C2PA_DEPRECATED_MSG(msg) __declspec(deprecated(msg))
--- a/c2pa_c_ffi/cbindgen.toml
+++ b/c2pa_c_ffi/cbindgen.toml
@@ -52,11 +52,28 @@ after_includes = """
     #endif
 #endif
 
+/* Deprecation macros: produce compiler warnings when deprecated C2PA functions are called. */
+#if defined(__cplusplus) && __cplusplus >= 201402L
+    #define C2PA_DEPRECATED [[deprecated]]
+    #define C2PA_DEPRECATED_MSG(msg) [[deprecated(msg)]]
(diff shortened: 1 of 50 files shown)


sdk/src/ingredient.rs
```diff
@@ -973,19 +999,30 @@ impl Ingredient {
...
     ///
     /// This does not set title or hash.
     /// Thumbnail will be set only if one can be retrieved from a previous valid manifest.
+    ///
+    /// Use [`Builder::from_context`](crate::Builder::from_context) with an explicit [`Context`](crate::Context) instead.
+    #[deprecated(
+        note = "Use `Builder::from_context(context)` with an explicit `Context` instead of relying on thread-local settings."
+    )]
+    #[allow(deprecated)]
     pub async fn from_memory_async(format: &str, buffer: &[u8]) -> Result<Self> {
```
tmathern, line 1014: Shouldn't this be deprecated alltogheter? That or the note is misleading, it points to Builder but we're in Ingredients?

sdk/src/ingredient.rs
```diff
@@ -973,19 +999,30 @@ impl Ingredient {
...
 
-    /// Creates an `Ingredient` from a stream (async version).
+    /// Creates an `Ingredient` from a stream (async version) using thread-local settings.
     ///
     /// This does not set title or hash.
     /// Thumbnail will be set only if one can be retrieved from a previous valid manifest.
+    ///
+    /// Use [`Builder::from_context`](crate::Builder::from_context) with an explicit [`Context`](crate::Context) instead.
+    #[deprecated(
+        note = "Use `Builder::from_context(context)` with an explicit `Context` instead of relying on thread-local settings."
```
tmathern, line 1026: Shouldn't this be deprecated alltogheter? That or the note is misleading, it points to Builder but we're in Ingredients?

cli/src/info.rs
```diff
@@ -23,6 +23,7 @@ pub fn info(path: &Path) -> Result<()> {
             None
         }
     }
+    #[allow(deprecated)]
```
tmathern, line 26: If we allow deprecated, when will we update to the latest API?

cli/src/main.rs
```diff
@@ -280,7 +279,9 @@ fn load_ingredient(path: &Path) -> Result<Ingredient> {
         }
         Ok(ingredient)
     } else {
-        Ok(Ingredient::from_file(path)?)
+        #[allow(deprecated)]
```
tmathern, line 285: If we allow deprecated, when will we update to the latest API?

sdk/src/builder.rs
```diff
@@ -488,9 +466,16 @@ impl AsRef<Builder> for Builder {
...
-    /// Creates a new [`Builder`] struct.
+    /// Creates a new [`Builder`] struct using thread-local settings.
+    ///
+    /// Use [`Builder::default()`](Builder::default) for a builder with default settings, or
+    /// [`Builder::from_context(context)`](Builder::from_context) to pass an explicit
+    /// [`Context`](crate::Context).
     /// # Returns
     /// * A new [`Builder`].
+    #[deprecated(
+        note = "Use `Builder::default()` for default settings, or `Builder::from_context(context)` to pass an explicit `Context`."
```
```suggestion
        note = "Use `Builder::default()` for default settings, or `Builder::from_context(context)` to pass an explicit `Context` (with potential settings)."
```

Not sure how to word it, but link the settings note to context

c2pa_c_ffi/cbindgen.toml
```diff
@@ -52,11 +52,28 @@ after_includes = """
     #endif
 #endif
 
+/* Deprecation macros: produce compiler warnings when deprecated C2PA functions are called. */
```
tmathern: Note: This may cause build failures in c2pa-cpp when we integrate, because deprecation are not allowed. I filtered some, but those will probably not be filtered OOTB.

tmathern (approved):

tmathern:
(Looks like Clippy also still has opinions on the deprecations)

tmathern:
Ah, I forgot - but you can test run builds for c2pa-c: https://github.com/contentauth/c2pa-cpp?tab=readme-ov-file#building-using-local-sources

cli/src/main.rs
gpeacock: I'm working on c2patool next and I'll see if I can clean this up
Some of this behavior is only legacy

PR #1680, opened 2025-12-19
feat!: Store `Context` as `Arc` for shared context and threading support

gpeacock:
### Overview
This PR introduces a new `Context` API that centralizes configuration management and enables thread-safe sharing of settings, HTTP resolvers, and signers across the SDK. The `Context` can be wrapped in an `Arc` for efficient sharing across threads while maintaining compatibility with WASM targets.

#### 1. New `Context` API
- **`Context` struct**: Central configuration container for settings, HTTP resolvers, and signers
- **Thread-safe sharing**: Implements `Send + Sync` (non-WASM) to enable `Arc<Context>` usage
- **Lazy initialization**: Uses `OnceLock` for default resolvers and settings-based signers
- **Conditional compilation**: Uses `cfg` attributes for WASM compatibility where `Send + Sync` aren't available

#### 2. Updated Builder API
- `Builder::from_context(context: Context)` - Create builder with a context
- `Builder::with_context(context: &Context)` - Share context reference (clones Arc internally)
- Builders can now leverage shared configuration without duplicating settings

#### 3. Updated Reader API  
- `Reader::from_context(context: Context)` - Create reader with a context
- `Reader::with_stream(context: &Context, format, stream)` - Read from stream with shared context
- Consistent API with Builder for context usage

#### 4. `BoxedSigner` Type Alias
- Public type alias for `Box<dyn Signer + Send + Sync>` (non-WASM)
- Conditional bounds via `cfg` for WASM compatibility
- Used throughout the SDK for consistent signer handling

#### 5. Settings Improvements
- `Settings::update_from_str()` - Overlay settings on existing instance
- Better integration with Context for dynamic configuration
- Support for JSON-formatted settings

Checklist
[x] All tests pass
[x] Documentation updated
[x] WASM compatibility verified
[x] No regressions in existing APIs

PR #2035, opened 2026-04-10
feat!: Store ClaimGeneratorInfo strings as Arc<str> for thread-shared identity

### LINKED ISSUE #2034: Support thread-shared ClaimGeneratorInfo via Arc<str>
Following the pattern from #1680 (Context → `Arc`), I'd like to propose converting `ClaimGeneratorInfo`'s string fields (`name`, `version`, `operating_system`) from `String` to `Arc<str>` so multi-threaded pipelines can share a single handle across workers without re-allocating per `sign()`. Today, `Builder::claim_generator_info` is cloned into each generated `Claim`, which re-allocates every string field even though the generator identity is constant for the lifetime of the application — a small but linear overhead for batch/streaming signing workloads.

Wire format is unchanged: `Arc<str>` serializes identically to `String` through serde, and deterministic re-serialization (critical for signature stability) is preserved. Scope would be `ClaimGeneratorInfo` only; per-manifest assertions like `Action` and `HashDataAssertion` don't benefit from sharing and would be left alone. Constructor would continue to accept `impl Into<Arc<str>>` so `&str`, `String`, and `Arc<str>` all work at call sites. Breaking surface is limited to direct struct-literal construction.

dcondrey:
Closes #2034. Following the pattern from #1680 (Context → `Arc`), this converts `ClaimGeneratorInfo`'s string fields (`name`, `version`, `operating_system`) from `String` to `Arc<str>` so a single generator identity can be shared across builders and workers without duplicating its backing strings on every clone. Wire format is unchanged — `Arc<str>` serializes identically to `String` for both JSON and CBOR via a visitor-based `deserialize_str` path that allocates the `Arc<str>` buffer directly (no intermediate `String`), and the byte-stable roundtrip tests assert signature stability. Scope is limited to `ClaimGeneratorInfo`; per-manifest assertions hold mostly dynamic values and wouldn't benefit. Constructors continue to accept `impl Into<Arc<str>>`, so `&str`/`String`/`Arc<str>` all work at call sites; `Default::default()` interns its name and version strings via a single `OnceLock` so repeat calls allocate zero strings. Also folds in a one-line fix for a pre-existing panic in `cli/src/main.rs` where `!is_empty()` on the `||` short-circuit path made `[0]` indexing unreachable only in the non-empty case, leaving the empty case as an unconditional out-of-bounds panic.

Microbenchmarks in `sdk/benches/sign.rs` show `ClaimGeneratorInfo::clone` drops from 43.8 ns → 12.7 ns (3.4x) and `Vec<ClaimGeneratorInfo>` clone from 90.4 ns → 36.8 ns (2.5x). Deserialization is a small trade: 132.2 ns → 136.8 ns (3.5% slower) because `Arc<str>` has an intrinsic 16-byte refcount header vs `String`'s inline layout — acceptable since typical workflows deserialize once and clone many times, and the end-to-end `sign 100kb jpeg` bench lands within noise (277.7 µs → 273.1 µs). The real benefit is enabling the shared-identity pattern without per-worker string duplication; `cargo test -p c2pa --lib` passes all 702 tests, with new coverage for `Arc::ptr_eq`-verified thread sharing, JSON and CBOR byte-stable roundtrips, `#[serde(alias)]` handling with the custom serializer, missing optional field deserialization, and `Default` memoization verified by pointer identity.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -716,8 +716,8 @@ fn main() -> Result<()> {
         // add claim_tool generator so we know this was created using this tool
         let mut tool_generator = ClaimGeneratorInfo::new(env!("CARGO_PKG_NAME"));
         tool_generator.set_version(env!("CARGO_PKG_VERSION"));
-        if !manifest.claim_generator_info.is_empty()
-            || manifest.claim_generator_info[0].name == "c2pa-rs"
+        if manifest.claim_generator_info.is_empty()
+            || &*manifest.claim_generator_info[0].name == "c2pa-rs"
         {
             manifest.claim_generator_info = vec![tool_generator];
         } else {
--- a/sdk/benches/sign.rs
+++ b/sdk/benches/sign.rs
@@ -1,7 +1,7 @@
 use std::io::Cursor;
 
-use c2pa::{Builder, CallbackSigner, SigningAlg};
-use criterion::{criterion_group, criterion_main, Criterion};
+use c2pa::{Builder, CallbackSigner, ClaimGeneratorInfo, SigningAlg};
+use criterion::{criterion_group, criterion_main, black_box, Criterion};
 
 const CERTS: &[u8] = include_bytes!("../tests/fixtures/certs/ed25519.pub");
 const PRIVATE_KEY: &[u8] = include_bytes!("../tests/fixtures/certs/ed25519.pem");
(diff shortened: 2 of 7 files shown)


dcondrey:
Closing this. The 3.5% deserialize regression is intrinsic to Arc<str>'s refcount header overhead and cannot be eliminated within this approach; the clone microbenchmark gains (3.4x) don't translate to the end-to-end sign path which stays within noise, so the breaking API change isn't justified by measurable benefit. Leaving #2034 open as a design discussion — if someone finds a concrete downstream workload where the clone:deserialize ratio makes the trade worth it, or wants to pursue Arc-at-the-container-level as an alternative, the analysis here is a starting point.

PR #2041, opened 2026-04-10
feat: add libc2pa Linux .deb and .rpm package builds

gpeacock:
Add a CI step to library-release.yml that produces libc2pa (runtime) and libc2pa-dev (headers + pkg-config) packages in both .deb and .rpm formats for x86_64 and aarch64 Linux targets using fpm.
This is pretty standard practice on linux platforms and will allow the library to be installed via apt or yum


ok-nick (approved):
Approving given the workflows were verified to work.

ZakryDat:
just wondering if this is still progressing? would be useful to us!

tmathern (approved):

.github/workflows/library-release.yml
```diff
@@ -108,6 +108,92 @@ jobs:
...
+
+          # libc2pa-dev — header, linker .so, and pkg-config; depends on runtime
+          fpm -s dir -t deb "${COMMON_ARGS[@]}" \
+            --name libc2pa-dev \
+            --version "$VERSION" \
+            --architecture "$DEB_ARCH" \
+            --description "C2PA content authenticity library - development files" \
+            --depends "libc2pa (= $VERSION)" \
+            --package "target/artifacts/libc2pa-dev_${VERSION}_${DEB_ARCH}.deb" \
+            "$LIB=/usr/lib/libc2pa_c.so" \
```
tmathern, line 172: This is dev package, but dev seem to depend on non-dev: it adds  "$LIB=/usr/lib/libc2pa_c.so" too.

This may lead to installation issues if one overwrites the other. The package manager might abort the install because it gets confused and can't overwrite.
Dev should probably only add ehaders & dev related things?

.github/workflows/library-release.yml
```diff
@@ -108,6 +108,92 @@ jobs:
...
+
+          # libc2pa-dev .rpm
+          fpm -s dir -t rpm "${COMMON_ARGS[@]}" \
+            --name libc2pa-dev \
+            --version "$VERSION" \
+            --architecture "$RPM_ARCH" \
+            --description "C2PA content authenticity library - development files" \
+            --depends "libc2pa = $VERSION" \
+            --package "target/artifacts/libc2pa-dev-${VERSION}-1.${RPM_ARCH}.rpm" \
+            "$LIB=/usr/lib/libc2pa_c.so" \
```
tmathern, line 193: Is the LIB link/setup required here too in dev package?

.github/workflows/library-release.yml
```diff
@@ -108,6 +108,92 @@ jobs:
         if: matrix.os != 'windows-latest' && matrix.os != 'windows-11-arm' && matrix.target != 'wasm32-unknown-emscripten'
         run: make release TARGET=${{ matrix.target }}
 
+      - name: Build Linux .deb and .rpm packages
+        if: contains(matrix.target, 'unknown-linux-gnu')
+        run: |
+          sudo apt-get install -y rpm ruby-dev build-essential
+          sudo gem install fpm --no-document
```
tmathern, line 115: Does Ruby/gem install support pinning versions?

PR #2044, opened 2026-04-13
chore: Ignore `rand` advisory as it doesn't apply to our feature set

ok-nick:
Our dependency `rand` is reporting an ["unsound error"](https://github.com/contentauth/c2pa-rs/actions/runs/24354271956/job/71116890612) with `cargo-deny` on all branches. We can't update it because our dependency `ed25519-dalek` still depends on an older version of `rand_core`, at least until they cut a new release (https://github.com/dalek-cryptography/curve25519-dalek/issues/876). Luckily, this error doesn't apply to us because we don't enable the `log` features of `rand`, so we can add an entry to ignore it in the `deny.toml`.


ok-nick:
Tracking issue: #2045

gpeacock (approved):

scouten-adobe (approved):

PR #2049, opened 2026-04-14
fix: Handle more thumbnail redaction edge cases

tmathern:
As in title. Some providers handle thumbnails differently, this change is to handle that.

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -1269,7 +1269,6 @@ impl Builder {
     /// Copies binary resources from `store` into this builder when the id is not already present.
     fn merge_resources_from_store(&mut self, store: &ResourceStore) -> Result<()> {
         for (id, data) in store.resources() {
-            dbg!(&id);
             let _sanitized_id = sanitize_archive_path(id)?;
             if !self.resources.exists(id) {
                 self.resources.add(id, data.clone())?;
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -1348,10 +1348,17 @@ impl Ingredient {
             // pass through `to_absolute_uri` unchanged and keep pointing at their own
             // nested manifest. Either way, the extracted manifest label is compared
             // against `thumbnail_redacted_manifests` to decide whether to drop it.
+            // A thumbnail whose resolved manifest label equals the claim being built is a
+            // self-reference to the outer archive manifest: it is always stale after re-signing
+            // and must be suppressed unconditionally.  Other thumbnails are suppressed only when
+            // the redaction list explicitly targets their source manifest.
             let thumbnail_is_redacted = jumbf::labels::manifest_label_from_uri(
                 &jumbf::labels::to_absolute_uri(claim.label(), &thumb_ref.identifier),
             )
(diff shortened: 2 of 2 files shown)


sdk/src/ingredient.rs
```diff
@@ -1337,18 +1341,19 @@ impl Ingredient {
...
+            // against `thumbnail_redacted_manifests` to decide if to be skipped.
+            // A thumbnail whose resolved manifest label equals the claim being built is a
+            // self-reference to the outer archive manifest: it is stale after re-signing
+            // and must be suppressed. Other thumbnails are suppressed only when
+            // the redaction list explicitly targets their source manifest.
+            let thumbnail_is_redacted = jumbf::labels::manifest_label_from_uri(
+                &jumbf::labels::to_absolute_uri(claim.label(), &thumb_ref.identifier),
+            )
+            .is_some_and(|label| {
+                label == claim.label() || thumbnail_redacted_manifests.contains(&label)
```
ok-nick: The claim doesn't correspond to the ingredient's claim, rather the builder's new claim. Checking with the active_manifest field as it was done before seems like the right approach to me.

ok-nick (approved):

sdk/src/ingredient.rs
tmathern: For posterity: we want to remove the outer thumbnail here, aka the one being built

sdk/src/ingredient.rs
```diff
@@ -1337,18 +1341,19 @@ impl Ingredient {
...
-                })
-                .and_then(|abs_uri| jumbf::labels::manifest_label_from_uri(&abs_uri))
-                .is_some_and(|label| thumbnail_redacted_manifests.contains(&label));
+            // The extracted manifest label is compared
+            // against `thumbnail_redacted_manifests` to decide if to be skipped.
+            // A thumbnail whose resolved manifest label equals the claim being built is a
+            // self-reference to the outer archive manifest: it is stale after re-signing
+            // and must be suppressed. Other thumbnails are suppressed only when
+            // the redaction list explicitly targets their source manifest.
+            let thumbnail_is_redacted = jumbf::labels::manifest_label_from_uri(
```
gpeacock: I feels like we are going through a lot of loops here to keep track of manifests, when all we really need to do is keep track of the redacted URIs. We just need to make sure all URIs we compare are in absolute form and there would be no need to compare manifest labels. It seems like the URIs are what we should be comparing.

tmathern: Ok, I tried being surgical, it didn't go as expected. Following that advice now, being less surgical but...

sdk/src/builder.rs
```diff
@@ -1269,7 +1269,6 @@ impl Builder {
     /// Copies binary resources from `store` into this builder when the id is not already present.
     fn merge_resources_from_store(&mut self, store: &ResourceStore) -> Result<()> {
         for (id, data) in store.resources() {
-            dbg!(&id);
```
tmathern, line 1272: Bonus clean up

gpeacock (approved):
There's probably a better way to deal with avoiding generating to be redacted thumbnails, but his is good enough for now.

PR #2052, opened 2026-04-14
docs: Separate out C docs

crandmck:

Per https://github.com/contentauth/c2pa-rs/issues/1987 this separates the docs for the C API from Rust docs.


c2pa_c_ffi/README.md
```diff
@@ -1,20 +1,305 @@
...
+
+However, binding to C APIs is a well-established and mature practice. Every language has well-documented methods for binding to C, and Rust has built-in support for it. A solid C interface enables leveraging that work to provide other language bindings.
+
+Since there are so many examples of this, AI engines are very good at constructing well-structured,  well-documented, and easily customizable bindings for other languages instead of using unreadable, incomprehensible auto-generated binding glue.
+
+### Caveats
+
+The C language is not object-oriented, does not perform garbage collection, and does not natively support things like exception handling. The API may use unsafe pointer references, so take care with pointers and memory management.
+
+For these reasons, you shouldn't use the C API directly for application code. Instead, use higher-level structures in other languages to ensure that references to native structures are correctly managed and freed.
```
gpeacock: I don't think we should say you  shouldn't use it, it should be something like we recommend using the the C++ API unless they have a specific reason to use the C api.

gpeacock (changes requested):

c2pa_c_ffi/README.md
crandmck: OK, made that change.

That text was based on text that you wrote and I reworded:

> ...the C API is not something we want developers to use directly. But it makes a very solid common way to bridge between Rust and other languages when used correctly. ...

c2pa_c_ffi/README.md
```diff
@@ -1,20 +1,305 @@
...
 
 ```sh
 make test
 ```
 
-The build will have 2 features activated: `rust_native_crypto` and `file_io`.
+The build will have two features activated: `rust_native_crypto` and `file_io`.
 Note that running the `make test` command will also check formatting of the code.
+
+## Progress and cancellation
```
gpeacock: Let's remove this progress and cancelation section entirely.

gpeacock (approved):
Approved with a request to entirely remove the progress and cancelation section in the c_ffi_docs

c2pa_c_ffi/README.md
crandmck: Deleted!

PR #2056, opened 2026-04-15
fix: identity assertions dropped in split-signing paths

### LINKED ISSUE #2055: Bug: Identity assertions silently dropped in `sign_embeddable` and `sign_data_hashed_embeddable`
### Component

c2pa-rs

### If Other, please specify

IdentityAssertionSigner

### Version

v0.79.3

### Platform

macOS

## What happened

I configured an `IdentityAssertionSigner` with a CAWG credential holder and used the split-signing APIs (`data_hashed_placeholder` + `sign_data_hashed_embeddable`). The signed manifests came out valid but with no `cawg.identity` assertion — silently missing, no error.

`sign_embeddable` has the same problem. The only path that works with identity is `Builder::sign` → `save_to_stream`, which requires the full asset to flow through the signer.

## How to reproduce (Pseudo Code)

```rust
// 1. Create signer with identity
let c2pa_raw_signer = signer_from_cert_chain_and_private_key(
    &c2pa_cert, &c2pa_key, SigningAlg::Ps256, None,
)?;
let mut signer = IdentityAssertionSigner::new(c2pa_raw_signer);

let cawg_raw_signer = signer_from_cert_chain_and_private_key(
    &cawg_cert, &cawg_key, SigningAlg::Ed25519, None,
)?;
let holder = X509CredentialHolder::from_raw_signer(cawg_raw_signer);
signer.add_identity_assertion(
    IdentityAssertionBuilder::for_credential_holder(holder),
);

// 2. Build and sign via split-signing
let mut builder = Builder::default().with_definition(manifest_json)?;

let placeholder = builder
    .data_hashed_placeholder(signer.reserve_size(), "application/c2pa")?;

let mut dh = DataHash::new("source_hash", "sha256");
dh.exclusions = Some(vec![HashRange::new(0, placeholder.len() as u64)]);
dh.gen_hash_from_stream(&mut Cursor::new(&placeholder))?;

let signed = builder
    .sign_data_hashed_embeddable(&signer, &dh, "application/c2pa")?;

// 3. Read back — cawg.identity is missing
let reader = Reader::default()
    .with_stream("application/c2pa", Cursor::new(signed))?;
let manifest = reader.active_manifest().unwrap();
let labels: Vec<_> = manifest.assertions().iter().map(|a| a.label()).collect();
// labels == ["c2pa.actions.v2"]  — no cawg.identity
```

Same result with `placeholder` + `sign_embeddable`. Using `Builder::sign` with the same signer produces `cawg.identity` correctly.

## What's going on

`IdentityAssertionSigner::dynamic_assertions()` uses `split_off(0)`, it only returns assertions on the first call. After that it returns empty.

`save_to_stream` gets this right: drains once, reuses the same vec for both `add_dynamic_assertion_placeholders` and `write_dynamic_assertions`.

The other paths don't:

- **`data_hashed_placeholder` + `sign_data_hashed_embeddable`**: Placeholder never calls `dynamic_assertions()`, so no slots are reserved. Signing drains them, checks for slots, finds none, silently discards.

- **`sign_embeddable`**: Drains at builder level to add slots, then `sign_manifest` drains again — gets empty. Slots exist but `write_dynamic_assertions` is never called.

## Consequences

- CAWG identity is unusable with any split-signing path. Only `Builder::sign` works.
- The failure is **silent**, no error, valid manifest, just missing `cawg.identity`.
- Both sync and async paths are affected (same `split_off(0)` in `AsyncIdentityAssertionSigner`).

The `cawg.identity` assertion should be present in the signed manifest regardless of which signing path is used. The signer is configured identically, the only difference is `sign_data_hashed_embeddable` vs `Builder::sign`. At the very least, if identity can't be written through a given path, it should return an error instead of silently producing a manifest without it.

stupside:
Closes #2055

## Summary

`cawg.identity` assertions are silently dropped when using `IdentityAssertionSigner` with `data_hashed_placeholder` + `sign_data_hashed_embeddable` or `placeholder` + `sign_embeddable`. `Builder::sign` works correctly with the same configuration.

Adds two tests that reproduce the issue. Both should pass once a fix is implemented.

## Test plan

- [ ] `cargo test -p c2pa --lib -- test_identity_present_in`


mauricefisher64:
Fixed.  Please reopen if issue persists.

mauricefisher64:
See link for fix in 2421

PR #2058, opened 2026-04-15
fix: Multi rendition support

mauricefisher64:
The C2PA spec requires a single common manifest across all renditions in a set.  This PR generates  a single manifest that is is inserted into all init segments of the set.

Update the fragment signing APIs to accept either a single file or a glob pattern (to specify a set if rendition init segments).

## Checklist

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -483,7 +483,7 @@ fn sign_fragmented(
     builder: &mut Builder,
     signer: &dyn Signer,
     init_pattern: &Path,
-    frag_pattern: &PathBuf,
+    frag_pattern: &Path,
     output_path: &Path,
 ) -> Result<()> {
     // search folders for init segments
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -494,9 +494,7 @@ fn sign_fragmented(
     let count = inits.count();
 
     if count > 0 {
-        builder
-            .sign_fragmented_files(signer, init_pattern, frag_pattern, output_path)
-            .map_err(anyhow::Error::from)?;
+        builder.sign_fragmented_files(signer, init_pattern, frag_pattern, output_path)?;
     } else {
         println!("No files matching pattern: {ip}");
     }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -492,37 +492,16 @@ fn sign_fragmented(
         "could not parse source pattern".into(),
     ))?;
     let inits = glob::glob(ip).context("could not process glob pattern")?;
-    let mut count = 0;
-    for init in inits {
-        match init {
-            Ok(p) => {
-                let mut fragments = Vec::new();
-                let init_dir = p.parent().context("init segment had no parent dir")?;
-                let seg_glob = init_dir.join(frag_pattern); // segment match pattern
+    let count = inits.count();
 
-                // grab the fragments that go with this init segment
-                let seg_glob_str = seg_glob.to_str().context("fragment path not valid")?;
-                let seg_paths = glob::glob(seg_glob_str).context("fragment glob not valid")?;
-                for seg in seg_paths {
-                    match seg {
-                        Ok(f) => fragments.push(f),
-                        Err(_) => return Err(anyhow!("fragment path not valid")),
-                    }
-                }
-
-                println!("Adding manifest to: {p:?}");
-                let new_output_path =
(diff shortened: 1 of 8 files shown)


gpeacock (approved):
This looks great.  Like how it simplifies things in the client side while not changing the api!

cli/src/main.rs
```diff
@@ -483,45 +483,24 @@ fn sign_fragmented(
...
-                count += 1;
-            }
-            Err(_) => bail!("bad path to init segment"),
-        }
-    }
-    if count == 0 {
+    if count > 0 {
+        builder
+            .sign_fragmented_files(signer, init_pattern, frag_pattern, output_path)
+            .map_err(anyhow::Error::from)?;
```
ok-nick: `anyhow` should automatically map the error via `?`.

sdk/src/store.rs
```diff
@@ -2836,107 +2842,192 @@ impl Store {
...
+        */
 
-        Ok(data) // return JUMBF data
+        Ok(())
     }
 
     /// Embed the claims store as jumbf into fragmented assets.
     #[cfg(feature = "file_io")]
-    pub fn save_to_bmff_fragmented(
+    pub fn save_to_bmff_fragmented<P: AsRef<Path> + Copy>(
```
ok-nick: Why `+ Copy`?

sdk/src/store.rs
```diff
@@ -2836,107 +2842,192 @@ impl Store {
...
+            output_map.insert(init_path.to_owned(), (i, i));
+        }
 
-        // get temp store as JUMBF
-        let jumbf = self.to_jumbf(signer)?;
+        // now save the manifest to each output init segment (the manifest is the same for each segment per the spec to allow related rendtions to be validated as a set)
+        let mut output_files = Vec::new();
+        let pc = self.provenance_claim().ok_or(Error::ClaimEncoding)?; // reborrow to change mutability
+        let bmff_hashes = pc.bmff_hash_assertions();
+        let mut bmff_hash = BmffHash::from_assertion(bmff_hashes[0].assertion())?;
```
ok-nick, line 2961: Can `bmff_hashes` be empty? This may panic.

cli/src/main.rs
```diff
@@ -483,45 +483,24 @@ fn sign_fragmented(
...
-                    output_path.join(init_dir.file_name().context("invalid file name")?);
-                builder.sign_fragmented_files(signer, &p, &fragments, &new_output_path)?;
-
-                count += 1;
-            }
-            Err(_) => bail!("bad path to init segment"),
-        }
-    }
-    if count == 0 {
+    if count > 0 {
```
ok-nick, line 496: Maybe we should add this check in `sign_fragmented_files` instead of here?

sdk/src/store.rs
```diff
@@ -2836,107 +2842,192 @@ impl Store {
...
+            );
 
-        // make sure there is no compression enabled since we do not support it for BMFF manifests
-        let pc = self.provenance_claim_mut().ok_or(Error::ClaimEncoding)?;
-        pc.set_compressed_manifest(false);
+            // add the Merkle tree map for this rendition
+            // creating fragments in the output location
+            self.add_merkmap_for_rendition(
+                &fragments,
+                i, // unique id for this rendition
```
ok-nick: The spec mentions the `uniqueId` should be "1-based." That doesn't seem like what's happening here, but it's also a non-normative comment.

mauricefisher64: The actual values don't really matter as long as they are self consistent.  I will start with 1 but that is really meaningless.  If you look at the example they are using track info the examples do even follow this.

cli/src/main.rs
mauricefisher64: Already checks. I left this here so the error message is the same.

sdk/src/store.rs
mauricefisher64: Can't happen,  see above.

sdk/src/store.rs
mauricefisher64: Left over from experiment.  I removed this

cli/src/main.rs
mauricefisher64: Done

sdk/src/store.rs
ok-nick: If the fragment glob path doesn't find any files, wouldn't this be empty?

Edit: I see, we error in that case.

PR #2059, opened 2026-04-15
fix: Harden tiff parser against forged IFD entry count fields

ssanthosh:
# Security Fix: TIFF/BigTIFF IFD Entry Count OOM Kill

## Vulnerability

`clone_ifd_entries` in `sdk/src/asset_handlers/tiff_io.rs` allocates a buffer sized
from the IFD entry's `value_count` field — which is raw attacker-controlled data — without
validating it against the actual file size.

`safe_vec` guards using `try_reserve_exact`, which is **insufficient** on Linux with memory
overcommit (the default in containers):

1. `try_reserve_exact(10_000_000_000)` **succeeds** — the OS commits virtual address space lazily
2. `output.resize(10_000_000_000, 0u8)` touches 10 GB of pages → OOM killer (exit 137)

A **52-byte BigTIFF** with `entry_type = Byte` and `value_count = 10_000_000_000` kills any
application running in an 8 GB container. On macOS the allocation fails immediately at
`try_reserve_exact`, so the issue is Linux-specific.

## Fix

- **`clone_ifd_entries`:** `let file_size = stream_len(asset_reader)?` once at function entry;
  `total_bytes > file_size` check before each `safe_vec` call for all 1/2/4-byte types.
- **`map_tiff`:** Renamed `_size` to `file_size`; added `num_longs_x4 as u64 > file_size`
  check before `safe_vec(subifd.value_count, Some(0u32))`.

## Test Added

| Test | Validates |
|------|-----------|
| `test_oom_clone_ifd_entries_byte_type` | `get_object_locations_from_stream` returns `Err(InvalidAsset)` on BigTIFF with Byte entry claiming 10 billion bytes |

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/tiff_io.rs
+++ b/sdk/src/asset_handlers/tiff_io.rs
@@ -384,6 +384,19 @@ fn decode_offset(offset_file_native: u64, endianness: Endianness, big_tiff: bool
     Ok(offset)
 }
 
+/// Reject forged IFD `value_count` fields whose claimed byte size exceeds the
+/// actual file size. Used before every `safe_vec`/`read_to_vec` allocation
+/// driven by attacker-controlled count fields, so a 52-byte BigTIFF can't
+/// trigger a multi-GB allocation.
+fn check_ifd_data_size(claimed_size: u64, file_size: u64) -> Result<()> {
+    if claimed_size > file_size {
+        return Err(Error::InvalidAsset(
+            "IFD entry data size exceeds file size".to_string(),
+        ));
+    }
+    Ok(())
+}
+
 // create tree of TIFF structure IFDs and IFD entries.
 fn map_tiff<R>(
     mut input: &mut R,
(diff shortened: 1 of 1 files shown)


sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -421,6 +421,13 @@ where
                         .ok_or_else(|| Error::InvalidAsset("value out of range".to_string()))?,
                 )
                 .map_err(|_err| Error::InvalidAsset("value out of range".to_string()))?;
+
+                if num_longs_x4 as u64 > file_size {
```
tmathern, line 425: Maybe make that check a small helper function? check_helper(current, max), then if current > max error?
And call it everywhere?

mauricefisher64 (approved):

sdk/src/asset_handlers/tiff_io.rs
ssanthosh: Done using https://github.com/contentauth/c2pa-rs/pull/2059/commits/f5aab0d85c3f0b7d2e913b7d217a48112f8c95ad

ssanthosh:
@tmathern  : Performed the necessary changes to address the review comments. Please check.

PR #2062, opened 2026-04-16
fix: Harden against image decompression bomb attacks in building thumbnail

ssanthosh:

# Security Fix: Thumbnail Decompression Bomb (PNG/TIFF/etc. Oversized Dimensions)

## Vulnerability

`make_thumbnail_from_stream` in `sdk/src/utils/thumbnail.rs` calls
`DynamicImage::from_decoder(decoder)` without first validating the image's declared
decoded size against any limit.

`ImageDecoder::into_decoder()` reads only the image header (e.g. PNG IHDR), which
contains attacker-controlled width/height fields. `from_decoder` then allocates a pixel
buffer of `width × height × bytes_per_pixel` bytes unconditionally:

1. A malicious PNG declares `16384 × 16384 RGBA` (1 GB decoded) in a ~65-byte file
2. `from_decoder` attempts to allocate 1 GB for the pixel buffer
3. Any application running in a container with < 1 GB available memory is killed (exit 137)
   or the process aborts with an allocation failure

The attack vector is any code path that calls thumbnail generation on untrusted input
(e.g. processing C2PA ingredients with thumbnails enabled).

## Fix

Two layered protections against decompression bombs:

1. **Apply the `image` crate's own `Limits`** via `ImageReader::limits()` — for
   decoders that honour limits internally.
2. **Explicit `decoder.total_bytes()` check** before `DynamicImage::from_decoder`
   — defense-in-depth for the path the `image` crate doesn't cover (see below).

**Why both layers are needed.** `Limits::default()` sets `max_alloc = 512 MiB` and
that budget is stored on the decoder by `set_limits()`, but `DynamicImage::from_decoder`
calls `decoder_to_image` which allocates the pixel buffer directly via
`vec![0u8; total_bytes as usize]` — that allocation does not consult
`Limits::max_alloc`. A 16384×16384 RGBA PNG (1 GiB) or an 8192×8192 RGBA32F TIFF
(1 GiB) would still allocate 1 GiB before `Limits` had a chance to act. The
explicit `total_bytes()` check closes this gap on the only path that allocates.

## Test Added

| Test | Validates |
|------|-----------|
| `test_make_thumbnail_rejects_decompression_bomb` | `make_thumbnail_bytes_from_stream` returns `Err(InvalidAsset)` for a 65-byte PNG claiming 16384×16384 RGBA (1 GB decoded). Uses `matches!(result, Err(Error::InvalidAsset(_)))`. |

## Checklist

### DIFF
--- a/sdk/src/utils/thumbnail.rs
+++ b/sdk/src/utils/thumbnail.rs
@@ -21,7 +21,7 @@ use image::{
         jpeg::JpegEncoder,
         png::{CompressionType, FilterType, PngEncoder},
     },
-    DynamicImage, ImageDecoder, ImageFormat, ImageReader,
+    DynamicImage, ImageDecoder, ImageFormat, ImageReader, Limits,
 };
 
 use crate::{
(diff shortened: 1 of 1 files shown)


sdk/src/utils/thumbnail.rs
```diff
@@ -145,6 +145,18 @@ where
     let mut decoder = ImageReader::with_format(input, input_format.into()).into_decoder()?;
     let orientation = decoder.orientation()?;
 
+    // Guard against decompression bombs: reject images whose raw decoded size would
+    // exceed 512 MB before allocating any pixel buffer. 512 MB matches the default
+    // `Limits::max_alloc` in the `image` crate (image-0.25, src/io/limits.rs line 54),
+    // covering all legitimate inputs including 16-bit 50 MP professional images (~302 MB).
+    const MAX_IMAGE_BYTES: u64 = 512 * 1024 * 1024;
+    let total_bytes = decoder.total_bytes();
+    if total_bytes > MAX_IMAGE_BYTES {
```
tmathern, line 154: This uses the image crate, can't we use its own https://docs.rs/image/0.25.10/image/struct.Limits.html ?

tmathern (changes requested):
The image crate has limits (https://docs.rs/image/0.25.10/image/struct.Limits.html), I'd try to rely on that instead.

sdk/src/utils/thumbnail.rs
ssanthosh: Performed via https://github.com/contentauth/c2pa-rs/pull/2062/commits/0b65a1cb0cbe2c1b348bd7f85f57b793756b24bf
But kept the total_bytes check also due to the following reason : 
```
Limits::default() sets max_alloc = 512 MiB and
that budget is stored on the decoder by set_limits(), but DynamicImage::from_decoder
calls decoder_to_image which allocates the pixel buffer directly via
vec![0u8; total_bytes as usize] — that allocation does not consult
Limits::max_alloc. A 16384×16384 RGBA PNG (1 GiB) or an 8192×8192 RGBA32F TIFF
(1 GiB) would still allocate 1 GiB before Limits had a chance to act. The
explicit total_bytes() check closes this gap on the only path that allocates.
```

ssanthosh:
@tmathern / @mauricefisher64 : Performed the review comment and still the total bytes check was also needed. Please check

tmathern (approved):

mauricefisher64 (approved):

PR #2063, opened 2026-04-16
fix: Harden BMFF hash merkel tree creation with max merkel leaves limits

ssanthosh:
# Security Fix: BmffHash Merkle Tree Unbounded Leaf Count (DoS)

## Issue

`create_merkle_tree_for_merkle_map()` and `validate_merkle_maps_mdat_boxes()` in
`sdk/src/assertions/bmff_hash.rs` build Merkle tree leaves by looping over a mdat box
in `fixed_block_size`-sized chunks. The only validation on the attacker-controlled
`fixed_block_size` field is a check for zero (or ≤ 1 in the second site).

The `variable_block_sizes` path has the same issue: a CBOR-embedded list of 500 001
one-byte entries produces an equally oversized leaf set.

## Fix

The cap is expressed in **memory** (bytes) rather than leaf count. Per-leaf
footprint depends on the hash algorithm, so a flat leaf-count cap means very
different memory budgets at SHA-256 (32 B) vs. SHA-512 (64 B). A memory cap is
easier to reason about for operators sizing containers.

## Tests Added

| Test | Validates |
|------|-----------|
| `test_merkle_tree_fixed_block_size_too_small_rejected` | Both call sites (`create_merkle_tree_for_merkle_map` and `validate_merkle_maps_mdat_boxes`) return `Err(InvalidAsset)` when `num_leaves × hash_size` exceeds 32 MB. Uses SHA-512 + 524_289 leaves (33_554_496 B > 33_554_432 B). |
| `test_merkle_tree_variable_block_sizes_too_many_rejected` | `create_merkle_tree_for_merkle_map` returns `Err(InvalidAsset)` for a `variable_block_sizes` list with 524_289 entries at SHA-512. |

## Checklist

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -2006,7 +2006,7 @@ impl BmffHash {
 
     /// Reject a Merkle tree whose leaf vector would exceed the
     /// `MAX_MERKLE_LEAVES_SIZE` memory budget for the resolved hash algorithm.
-    /// Centralises the budget check used by both Merkle-tree construction
+    /// Centralizes the budget check used by both Merkle-tree construction
     /// (`create_merkle_tree_for_merkle_map`) and validation
     /// (`validate_merkle_maps_mdat_boxes`).
     fn check_merkle_leaf_memory(
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -31,6 +31,13 @@ use sha2::{Digest, Sha256, Sha384, Sha512};
 /// Practical BMFF files rarely have more than 4 mdat boxes.
 const MAX_MDAT_BOXES: usize = 4;
 
+/// Maximum number of Merkle tree leaves allowed when building or validating a Merkle map.
+/// Prevents resource exhaustion when an attacker sets `fixed_block_size` to a tiny value
+/// or embeds a huge `variable_block_sizes` list in a BmffHash assertion.
+/// 500 000 leaves supports a 50 GB mdat with 100 KB blocks — sufficient for all practical
+/// streaming-video content authentication granularities.
+const MAX_MERKLE_LEAVES: u64 = 500_000;
+
 use crate::{
     assertion::{Assertion, AssertionBase, AssertionCbor},
     assertions::labels,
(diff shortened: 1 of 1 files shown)


sdk/src/assertions/bmff_hash.rs
```diff
@@ -31,6 +31,13 @@ use sha2::{Digest, Sha256, Sha384, Sha512};
 /// Practical BMFF files rarely have more than 4 mdat boxes.
 const MAX_MDAT_BOXES: usize = 4;
 
+/// Maximum number of Merkle tree leaves allowed when building or validating a Merkle map.
+/// Prevents resource exhaustion when an attacker sets `fixed_block_size` to a tiny value
+/// or embeds a huge `variable_block_sizes` list in a BmffHash assertion.
+/// 500 000 leaves supports a 50 GB mdat with 100 KB blocks — sufficient for all practical
+/// streaming-video content authentication granularities.
+const MAX_MERKLE_LEAVES: u64 = 500_000;
```
tmathern, line 39: Does this need to be configurable? @mauricefisher64 @gpeacock ?

mauricefisher64: If we have this variable it should probably be expressed in terms of memory usage.  500000 leaves at the worst case is 64 bytes (Sha512) only uses 32MB of memory, so not an outrageous about.  I think if you did this than someone working on an 8GB docker machines could say for example only use 10MB would of memory for the leaves.   Memory is also easier to think about rather that leaves which will not mean much to most people.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -2021,6 +2028,13 @@ impl BmffHash {
                 return Err(Error::BadParam("fixed block size is 0".to_string()));
             }
 
+            let num_leaves = bytes_left.div_ceil(fixed_block_size);
```
mauricefisher64, line 2031: So if you are doing memory check it world be somethings like this. 

if num_leaves * hash_alg_size_in_bytes > MAX_MERKLE_LEAVES_SIZE {
...
}

sdk/src/assertions/bmff_hash.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2063/commits/9d30d1987e1385697ec8dd53ba27eb2b06c9d2a9. Used 32MB as the limits for maximum memory that a single BmffHash Merkle leaf vector may consume

sdk/src/assertions/bmff_hash.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/9d30d1987e1385697ec8dd53ba27eb2b06c9d2a9

ssanthosh:
@mauricefisher64 / @tmathern : Performed the necessary changes requested. Please check

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1978,6 +2004,30 @@ impl BmffHash {
         Ok(())
     }
 
+    /// Reject a Merkle tree whose leaf vector would exceed the
+    /// `MAX_MERKLE_LEAVES_SIZE` memory budget for the resolved hash algorithm.
+    /// Centralises the budget check used by both Merkle-tree construction
```
tmathern:
```suggestion
    /// Centralizes the budget check used by both Merkle-tree construction
```

tmathern (approved):

mauricefisher64 (approved):

PR #2064, opened 2026-04-17
fix: Harden against deep linear ingredient chains with max limits

ssanthosh:
# Security Fix: Store Recursive Ingredient Traversal Stack Overflow (DoS)

## Vulnerability Issue

`get_claim_referenced_manifests_impl()`, `ingredient_checks()`, and
`ingredient_checks_async()` in `sdk/src/store.rs` walk the ingredient reference
graph recursively with no depth limit.

A crafted C2PA manifest store containing a linear chain of 3 000 ingredient
references (A → B → C → … → N) produces 3000 nested stack frames. With the
default 8 MB thread stack each frame consumes ~2.7 KB, exhausting the stack and
causing the process to terminate with exit code 134 (SIGABRT / stack overflow).

## Fix

Added module-level constant MAX_INGREDIENT_DEPTH of 200 to restrict the length of deep linear ingredient chain

## Tests Added

| Test | Validates |
|------|-----------|
| `test_ingredient_depth_limit_get_claim_referenced_manifests` | Returns `Err(InvalidAsset)` when `claim_label_path.len() >= MAX_INGREDIENT_DEPTH` |
| `test_ingredient_depth_limit_ingredient_checks` | Returns `Err(InvalidAsset)` when `depth >= MAX_INGREDIENT_DEPTH` |

## Checklist

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -91,6 +91,15 @@ const MANIFEST_STORE_EXT: &str = "c2pa"; // file extension for external manifest
 #[cfg(feature = "fetch_remote_manifests")]
 const DEFAULT_MANIFEST_RESPONSE_SIZE: usize = 10 * 1024 * 1024; // 10 MB
 
+/// Maximum depth of nested ingredient references allowed during validation.
+///
+/// Prevents stack overflow when processing a manifest store that contains a deep
+/// linear chain of ingredient references (A → B → C → … → N). With the default
+/// 8 MB thread stack, ~3 000 levels causes an unrecoverable stack overflow (exit 134).
+/// A limit of 200 uses only ~540 KB of stack, providing a large safety margin while
+/// accommodating any realistic provenance chain depth.
+const MAX_INGREDIENT_DEPTH: usize = 200;
+
 pub(crate) struct ManifestHashes {
     pub manifest_box_hash: Vec<u8>,
     pub signature_box_hash: Vec<u8>,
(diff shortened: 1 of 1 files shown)


sdk/src/store.rs
```diff
@@ -1528,7 +1537,14 @@ impl Store {
         asset_data: &mut ClaimAssetData<'_>,
         validation_log: &mut StatusTracker,
         context: &Context,
+        depth: usize,
```
tmathern, line 1540: I would try to keep the Context as last param? If we change the signature anyway?

sdk/src/store.rs
```diff
@@ -1954,6 +1978,7 @@ impl Store {
                         asset_data,
                         validation_log,
                         context,
+                        depth + 1,
```
tmathern, line 1981: MAX_INGREDIENT_DEPTH < MAC int, but I would make sure still that this can't overflow

tmathern (changes requested):

sdk/src/store.rs
mauricefisher64: Let me know when this is checked in.  There is an optimization that we can do here that I should apply

sdk/src/store.rs
```diff
@@ -3969,6 +4002,14 @@ impl Store {
         validation_log: &mut StatusTracker,
         claim_label_path: &mut Vec<&'a str>,
     ) -> Result<()> {
+        if claim_label_path.len() >= MAX_INGREDIENT_DEPTH {
```
mauricefisher64, line 4005: This makes no sense to me.  How does the number of ingredients relate to the ingredient depth?  I can have 1000000 ingredients and the max ingredient depth I hit could be 1.

ssanthosh: `claim_label_path` is the chain of claims currently being walked recursively: each entry pushes one entry in beginning of function and pops one at the end of function and hence its length is the current recursion depth and not the total number of ingredients seen.

sdk/src/store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2064/commits/22bd79274348d5ef1b3c8d7f252ab81d4bb99776

sdk/src/store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2064/commits/22bd79274348d5ef1b3c8d7f252ab81d4bb99776

ssanthosh:
@mauricefisher64 / @tmathern : Performed the necessary changes for the review comments. Please check

tmathern (approved):

mauricefisher64 (approved):

PR #2066, opened 2026-04-17
fix: Harden builder and resource store addition against path traversal zip slip attacks

ssanthosh:
 Security Fix: Zip Slip Path Traversal in Builder and ResourceStore

## Vulnerability Issue

`Builder::old_from_archive()`, `Builder::merge_resources_from_store()`,
`Builder::add_resource()` in `sdk/src/builder.rs`, and `ResourceStore::add()`
in `sdk/src/resource_store.rs` were vulnerable to a Zip Slip path traversal
attack.

### Root cause — two independent issues

**Issue 1: `sanitize_archive_path` return value discarded in `builder.rs`**

A prior partial fix (commit `3afed367`) introduced `sanitize_archive_path()` —
a helper that validates archive entry paths — but **discarded its return value
everywhere it was called**:

```rust
// Prior incomplete fix pattern — all five call sites:
let _sanitized_id = sanitize_archive_path(id)?;   // result discarded
...
builder.resources.add(id, data)?;                  // ORIGINAL unsanitized path stored
```

Although the function ran and would reject blatant traversal strings, the
sanitized / normalized path was never used. An attacker could craft archive
entries whose paths passed the validator yet whose normalized form differed from
the raw path — for example `./secret` normalized to `secret` — or embed crafted
path components that escape the intended extraction directory when used with
`ResourceStore`'s `base_path`.

Additionally, the old `sanitize_archive_path` implementation used a manual
string-split approach that could miss path components on Windows (backslash
separators, drive prefixes, UNC paths).

**Issue 2: `ResourceStore::add()` performs no path validation**

`ResourceStore::add()` is a public API. When a `base_path` is set (the
`file_io` feature), it writes resource data to disk by calling `base.join(id)`
directly with no sanitization:

```rust
// Vulnerable code in ResourceStore::add():
let path = base.join(id.into());   // id = "../evil.txt" escapes base
create_dir_all(path.parent()...)?;
write(path, value.into())?;        // file written outside base directory
```

This is exploitable independently of the Builder layer: anyone calling
`ResourceStore::add("../evil.txt", data)` on a store with `base_path` set
writes the file outside the intended directory. The old fix left this path
entirely unprotected.

### Affected locations

| File | Location | Issue |
|---|---|---|
| `builder.rs` | `add_resource` | Sanitized id discarded, original `id` stored |
| `builder.rs` | `old_from_archive` resources block | `_sanitized_id` discarded, original `id` stored |
| `builder.rs` | `old_from_archive` manifests block | `_sanitized_label` discarded |
| `builder.rs` | `old_from_archive` ingredients block | `_sanitized_id` discarded, original `id` stored |
| `builder.rs` | `merge_resources_from_store` | `_sanitized_id` discarded, original `id` stored and used for existence check |
| `resource_store.rs` | `ResourceStore::add()` | `base.join(id)` with no path validation — public API |

### `sanitize_archive_path` replaced (`builder.rs`)

Uses `std::path::Path::components()` for platform-correct parsing:
- `..` (`Component::ParentDir`) — rejected
- `/` prefix (`Component::RootDir`) — rejected
- Windows drive/UNC (`Component::Prefix`) — rejected
- `.` (`Component::CurDir`) — stripped silently
- `Normal` components — collected and re-joined with `/`

Returns the normalized forward-slash path, or `Err(Error::BadParam)` for any
dangerous input.

### All `builder.rs` call sites updated to use the sanitized key

```rust
// Fixed pattern — all call sites now use sanitized_id:
let sanitized_id = sanitize_archive_path(id)?;
...
builder.resources.add(sanitized_id, data)?;   // normalized path stored
```

`add_resource` also updated its `exists` check to use the sanitized key,
ensuring consistent normalization for both lookup and insertion.

### `ResourceStore::add()` validates path before `base.join()` (`resource_store.rs`)

When `base_path` is set, the id is now validated using `Path::components()`
before any file system operation:

```rust
use std::path::Component;
let id_str: String = id.into();
for component in Path::new(&id_str).components() {
    match component {
        Component::Normal(_) | Component::CurDir => {}
        _ => {
            return Err(crate::Error::BadParam(format!(
                "Path traversal not allowed: {id_str}"
            )));
        }
    }
}
let path = base.join(&id_str);
```

This closes the root-cause bypass: even if a caller reaches `ResourceStore::add()`
directly (bypassing all Builder-layer sanitization), traversal paths are
rejected before any file is written.

### `builder::tests` — 7 unit tests

| Test | Validates |
|---|---|
| `sanitize_archive_path_normal_path_accepted` | Clean path passes through unchanged |
| `sanitize_archive_path_dot_stripped` | Leading `./` is stripped |
| `sanitize_archive_path_parent_dir_rejected` | `../etc/passwd` rejected |
| `sanitize_archive_path_inner_parent_dir_rejected` | `resources/../../../etc/passwd` rejected |
| `sanitize_archive_path_absolute_rejected` | `/etc/passwd` rejected |
| `sanitize_archive_path_empty_rejected` | Empty string rejected |
| `sanitize_archive_path_dot_only_rejected` | `"."` (normalises to empty) rejected |

### `resource_store::tests::zip_slip_tests` — 4 unit tests (requires `file_io`)

These tests reproduce the root-cause exploit that the old fix left unprotected:
calling `ResourceStore::add()` directly with a traversal path when `base_path`
is set.

| Test | Validates |
|---|---|
| `add_with_base_path_rejects_parent_dir_traversal` | `"../evil.txt"` rejected; confirms no file written outside temp dir |
| `add_with_base_path_rejects_absolute_path` | `"/etc/passwd"` rejected |
| `add_with_base_path_accepts_normal_resource` | `"thumbnail.jpg"` accepted, file written correctly |
| `add_with_base_path_accepts_subdir_resource` | `"subdir/thumbnail.jpg"` accepted, file written correctly |

`add_with_base_path_rejects_parent_dir_traversal` would **fail** (file written
outside base) against the old fix (3afed367) and **passes** with this fix.

### `sdk/tests/test_builder.rs` — existing test updated

`test_archive_path_traversal_protection` (added in 3afed367) checked for
`"Absolute path not allowed"` as a separate message for `/etc/passwd`. The new
unified implementation uses `"Path traversal not allowed"` for both `..` and
absolute paths. The test's match arm was updated to reflect this.

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -93,58 +93,52 @@ impl ArchiveKind {
 /// Version of the Builder Archive file
 const ARCHIVE_VERSION: &str = "1";
 
-/// Sanitizes a path to prevent directory traversal attacks.
+/// Validate and canonicalize an archive entry path.
 ///
-/// This function validates that the path:
-/// - Does not contain '..' components
-/// - Does not contain absolute path markers
-/// - Does not escape the intended directory structure
+/// Uses [`std::path::Path::components`] for platform-correct parsing. This
+/// catches `..` (`ParentDir`), leading separators (`RootDir`), and Windows
+/// drive/UNC prefixes (`Prefix`) on all host platforms — cases that a
+/// hand-rolled string split can miss when they appear as inner components.
 ///
-/// # Arguments
-/// * `path` - The path string to sanitize
-///
-/// # Returns
-/// * The sanitized path if valid
-///
-/// # Errors
-/// * Returns an [`Error::BadParam`] if the path contains dangerous components
+/// Returns the normalized, forward-slash-separated path (`.` components are
+/// stripped), or an error if the path is empty, absolute, contains a `..`
+/// traversal, or includes a Windows drive/UNC prefix.
 fn sanitize_archive_path(path: &str) -> Result<String> {
-    // Reject empty paths
+    use std::path::{Component, Path};
+
     if path.is_empty() {
         return Err(Error::BadParam("Empty path not allowed".to_string()));
     }
 
-    // Reject paths that start with '/' (absolute paths)
-    if path.starts_with('/') || path.starts_with('\\') {
-        return Err(Error::BadParam(format!(
-            "Absolute path not allowed: {path}"
-        )));
-    }
+    let mut sanitized = String::new();
 
-    // Check for drive letters on Windows (e.g., "C:")
-    if path.len() >= 2 && path.chars().nth(1) == Some(':') {
-        return Err(Error::BadParam(format!("Drive letter not allowed: {path}")));
-    }
-
-    // Split the path and check each component
-    let components: Vec<&str> = path.split(&['/', '\\'][..]).collect();
-
-    for component in &components {
-        // Reject '..' components
-        if *component == ".." {
-            return Err(Error::BadParam(format!(
-                "Path traversal not allowed: {path}"
-            )));
-        }
-
-        // Reject empty components (which could come from '//')
-        if component.is_empty() {
-            continue; // Allow empty components from trailing slashes
+    for component in Path::new(path).components() {
+        match component {
+            Component::Normal(part) => {
+                let part = part.to_str().ok_or_else(|| {
+                    Error::BadParam(format!("Non-UTF-8 path component in: {path}"))
+                })?;
+                if !sanitized.is_empty() {
+                    sanitized.push('/');
+                }
+                sanitized.push_str(part);
+            }
+            // Silently drop current-directory markers (`.`).
+            Component::CurDir => {}
+            // Absolute paths (`/`), Windows drive/UNC prefixes, and `..` are all rejected.
+            Component::RootDir | Component::Prefix(_) | Component::ParentDir => {
+                return Err(Error::BadParam(format!(
+                    "Path traversal not allowed: {path}"
+                )));
+            }
         }
     }
 
-    // Normalize the path to use forward slashes
-    Ok(path.replace('\\', "/"))
+    if sanitized.is_empty() {
+        return Err(Error::BadParam("Empty path not allowed".to_string()));
+    }
+
+    Ok(sanitized)
 }
 
 /// Use a ManifestDefinition to define a manifest and to build a `ManifestStore`.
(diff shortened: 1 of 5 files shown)


sdk/src/builder.rs
```diff
@@ -93,58 +93,52 @@ impl ArchiveKind {
...
-/// * The sanitized path if valid
-///
-/// # Errors
-/// * Returns an [`Error::BadParam`] if the path contains dangerous components
+/// Returns the normalized, forward-slash-separated path (`.` components are
+/// stripped), or an error if the path is empty, absolute, contains a `..`
+/// traversal, or includes a Windows drive/UNC prefix.
 fn sanitize_archive_path(path: &str) -> Result<String> {
-    // Reject empty paths
+    use std::path::{Component, Path};
```
tmathern, line 107: Move in on top of the file?

sdk/src/resource_store.rs
```diff
@@ -289,7 +289,19 @@ impl ResourceStore {
     {
         #[cfg(feature = "file_io")]
         if let Some(base) = self.base_path.as_ref() {
-            let path = base.join(id.into());
+            use std::path::Component;
```
tmathern, line 292: This looks similar to the changes in builder.rs, can there be code shared/refactored?

sdk/src/builder.rs
```diff
@@ -93,58 +93,52 @@ impl ArchiveKind {
...
 const ARCHIVE_VERSION: &str = "1";
 
-/// Sanitizes a path to prevent directory traversal attacks.
+/// Validate and canonicalize an archive entry path.
 ///
-/// This function validates that the path:
-/// - Does not contain '..' components
-/// - Does not contain absolute path markers
-/// - Does not escape the intended directory structure
+/// Uses [`std::path::Path::components`] for platform-correct parsing. This
```
tmathern, line 98: I think there needs to be additional checks, because of the platform dependence.

On Linux, `\` can be used in filenames, and Path::components therefore will not flag it. I think we need to keep the previous logic that flagged this as path separator, or I think we can miss edge cases when switching between Linux and Windows (that could happen, since part of archives data). A zip is a portable artifact: someone can author it on any OS, using any path separator they want, and send it to anyone else...

tmathern (changes requested):

sdk/src/builder.rs
ssanthosh: Resolved with https://github.com/contentauth/c2pa-rs/pull/2066/commits/c80e68efbcba8b358356e08e4e5e973db36a48ac

sdk/src/resource_store.rs
ssanthosh: Resolved with https://github.com/contentauth/c2pa-rs/pull/2066/commits/c80e68efbcba8b358356e08e4e5e973db36a48ac. Added new utility file for reusing the code

sdk/src/builder.rs
ssanthosh: Resolved with https://github.com/contentauth/c2pa-rs/pull/2066/commits/c80e68efbcba8b358356e08e4e5e973db36a48ac

ssanthosh:
@tmathern : Performed the necessary changes to address the review comments. Please check.

tmathern (approved):

PR #2067, opened 2026-04-17
fix: Harden against empty certificates during OCSP certificate validation

ssanthosh:
# Security Fix: OCSP Empty-Certs DoS (index-out-of-bounds panic)

## Vulnerability

The c2pa-rs SDK panics with `index out of bounds` when validating a COSE
signature whose unprotected header contains an OCSP response with a
present-but-empty `certs` field (`certs = Some([])`).

Because the unprotected header is **not covered by the COSE signature**, an
attacker can inject a crafted OCSP DER blob without any signing keys. This
turns certificate-revocation checking into a remote DoS vector: any validator
processing attacker-supplied content panics.

### Root cause

`OcspResponse::from_der_checked` in `sdk/src/crypto/ocsp/mod.rs` guarded
against `certs = None` with `if let Some(ocsp_certs) = ...`, but did not
guard against an empty `Vec`. The first element was then accessed with a
direct index `ocsp_certs[0]`, which panics when the vec is present but empty:

```rust
// Vulnerable code — panics when certs = Some([])
if let Some(ocsp_certs) = basic_response.certs {
    ...
    let signing_key_der =
        rasn::der::encode(&ocsp_certs[0].tbs_certificate.subject_public_key_info)?;
    //                    ^^^^^^^^^^^^ index out of bounds if vec is empty
```

The same pattern appeared as defense-in-depth access sites in
`sdk/src/crypto/cose/ocsp.rs` (`check_stapled_ocsp_response` and
`fetch_and_check_ocsp_response`).

### Attack surface

The OCSP DER is read from the COSE `rVals`/`ocspVals` unprotected header
field, which is not authenticated by the COSE signature. Any attacker who can
present a crafted C2PA asset to a validator can trigger the panic.

### Affected locations

| File | Function | Issue |
|---|---|---|
| `sdk/src/crypto/ocsp/mod.rs` | `from_der_checked` | Primary panic: `ocsp_certs[0]` when `certs = Some([])` |
| `sdk/src/crypto/cose/ocsp.rs` | `check_stapled_ocsp_response` | Defense-in-depth: `&ocsp_certs[0]` |
| `sdk/src/crypto/cose/ocsp.rs` | `fetch_and_check_ocsp_response` | Defense-in-depth: `&ocsp_certs[0]` |

## Fix

All three `[0]` index accesses replaced with `.first()` + `let-else` guards
that return `Ok(OcspResponse::default())` (unknown/ignored OCSP status) rather
than panicking:

```rust
// Fixed pattern — all three sites:
let Some(first_cert) = ocsp_certs.first() else {
    return Ok(output);  // or Ok(OcspResponse::default())
};
```

### `sdk/src/crypto/ocsp/mod.rs` — `from_der_checked` (primary fix)

```rust
// grab the signing key from the first cert; reject if certs is empty
let Some(first_cert) = ocsp_certs.first() else {
    return Ok(output);
};
let Ok(signing_key_der) =
    rasn::der::encode(&first_cert.tbs_certificate.subject_public_key_info)
else {
    return Ok(output);
};
```

### `sdk/src/crypto/cose/ocsp.rs` — two defense-in-depth sites

`check_stapled_ocsp_response`:
```rust
if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
    let Some(first_cert) = ocsp_certs.first() else {
        return Ok(OcspResponse::default());
    };
    // ... rest of validation unchanged
}
```

`fetch_and_check_ocsp_response`:
```rust
if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
    let Some(first_cert) = ocsp_certs.first() else {
        return Ok(OcspResponse::default());
    };
    // ... rest of validation unchanged
}
```

### `sdk/src/crypto/ocsp/mod.rs` — 1 regression test

| Test | Validates |
|---|---|
| `from_der_checked_empty_certs_returns_default_not_panic` | Crafted DER with `certs = Some([])` returns `Ok(OcspResponse::default())` instead of panicking |

The test uses `OCSP_EMPTY_CERTS_DER` — 95 hand-crafted bytes representing a
structurally valid `OcspResponse` with `responseStatus = successful`,
`id-pkix-ocsp-basic`, a dummy SHA-256 signature, and `certs = Some([])`.
Running this test against the unfixed code panics at `ocsp_certs[0]`.

## Checklist

### DIFF
--- a/sdk/src/crypto/cose/ocsp.rs
+++ b/sdk/src/crypto/cose/ocsp.rs
@@ -361,12 +361,16 @@ fn check_stapled_ocsp_response(
 
     // If we get a valid response, validate the certs.
     if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
+        let Some(first_cert) = ocsp_certs.first() else {
+            return Ok(OcspResponse::default());
+        };
+
         // make sure this is an OCSP signing EKU
         let mut new_ctp = ctp.clone();
         new_ctp.clear_ekus();
         new_ctp.add_valid_ekus(OCSP_OID_STR.as_bytes()); // ocsp signing EKU
         if check_end_entity_certificate_profile(
-            &ocsp_certs[0],
+            first_cert,
             &new_ctp,
             validation_log,
             tst_info.as_ref(),
@@ -378,11 +382,7 @@ fn check_stapled_ocsp_response(
 
         // validate the trust
         if new_ctp
-            .check_certificate_trust(
-                ocsp_certs,
-                &ocsp_certs[0],
-                signing_time.map(|t| t.timestamp()),
-            )
+            .check_certificate_trust(ocsp_certs, first_cert, signing_time.map(|t| t.timestamp()))
             .is_err()
         {
             return Ok(OcspResponse::default());
@@ -457,13 +457,16 @@ pub(crate) fn fetch_and_check_ocsp_response(
 
     // If we get a valid response validate the certs.
     if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
+        let Some(first_cert) = ocsp_certs.first() else {
+            return Ok(OcspResponse::default());
+        };
+
         // make sure this is an OCSP signing EKU
         let mut new_ctp = ctp.clone();
         new_ctp.clear_ekus();
         new_ctp.add_valid_ekus(OCSP_OID_STR.as_bytes()); // ocsp signing EKU
 
-        if check_end_entity_certificate_profile(&ocsp_certs[0], &new_ctp, validation_log, None)
-            .is_err()
+        if check_end_entity_certificate_profile(first_cert, &new_ctp, validation_log, None).is_err()
         {
             return Ok(OcspResponse::default());
         }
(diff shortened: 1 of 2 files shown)


sdk/src/crypto/cose/ocsp.rs
```diff
@@ -361,12 +361,16 @@ fn check_stapled_ocsp_response(
 
     // If we get a valid response, validate the certs.
     if let Some(ocsp_certs) = &ocsp_data.ocsp_certs {
+        let Some(first_cert) = ocsp_certs.first() else {
```
tmathern, line 364: If present but empty is an attack scenario, shouldn't we report something? Isn't it suspicious enough to trigger an error?
I know we don't have good logging solutions, but it would be nice to know when this happens.

tmathern (approved):
Not sure that this shouldn't actually error out since it's suspicious payload (rather than silently continuing).

sdk/src/crypto/cose/ocsp.rs
mauricefisher64: They are not required to provide the cert list.  If it is not present we cannot validate the OCSP response and treat it as if it were not present.

mauricefisher64 (approved):

PR #2070, opened 2026-04-17
fix: Harden against unbounded HTTP response body read in DID web server

ssanthosh:
# Security Fix: did:web Unbounded Response Body Read (OOM DoS)

## Vulnerability

The `did:web` resolver in `sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs`
reads HTTP response bodies without any size limit. A malicious DID Web server
can crash any application using c2pa-rs by returning an infinitely large
response body, causing out-of-memory process termination.

### Root cause

`get_did_doc` calls `read_to_end` directly on the response body with no cap:

```rust
// Vulnerable code — reads unlimited bytes into memory
let mut document = Vec::new();
body.read_to_end(&mut document)
    .map_err(|e| DidWebError::Response(e.into()))?;
```

`Vec::read_to_end` will keep allocating until either the server closes the
connection or the process is killed by OOM.

### Attack surface

Any DID identity validation that resolves a `did:web` DID causes an outbound
HTTP request. An attacker controls the server at the domain encoded in the DID.
If they return a streaming/infinite body, the validator's memory grows without
bound until the OS kills the process.

### Affected location

| File | Function | Issue |
|---|---|---|
| `sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs` | `get_did_doc` | `body.read_to_end(&mut document)` — no size limit |

## Fix

The body size limit is enforced inside `AsyncGenericResolver::http_resolve_async`
in `sdk/src/http/mod.rs` via an opt-in config. No existing callers are affected.

### New config on `AsyncGenericResolver`

```rust
pub fn with_max_response_body_size(mut self, limit: u64) -> Self {
    self.max_response_body_size = Some(limit);
    self
}
```

When set, `AsyncGenericResolver::http_resolve_async` applies two guards before
returning the response to the caller:

**Layer 1 — Content-Length fast-fail** (before reading body bytes):

```rust
let reported_len = response
    .headers()
    .get(http::header::CONTENT_LENGTH)
    .and_then(|v| v.to_str().ok())
    .and_then(|s| s.parse::<u64>().ok());
if let Some(len) = reported_len {
    if len > limit {
        return Err(HttpResolverError::ResponseTooLarge);
    }
}
```

**Layer 2 — hard body cap** (defense-in-depth for servers that omit or lie about
`Content-Length`):

```rust
let (parts, mut body) = response.into_parts();
let mut buf = Vec::new();
body.take(limit + 1).read_to_end(&mut buf)?;
if buf.len() as u64 > limit {
    return Err(HttpResolverError::ResponseTooLarge);
}
return Ok(http::Response::from_parts(
    parts,
    Box::new(Cursor::new(buf)) as Box<dyn Read>,
));
```

The response returned to the caller always has a body bounded by `limit` bytes.
A new `HttpResolverError::ResponseTooLarge` variant is added.

### `get_did_doc` opts in and maps the error

```rust
let response = AsyncGenericResolver::with_redirects()
    .unwrap_or_default()
    .with_max_response_body_size(MAX_DID_DOC_SIZE)
    .http_resolve_async(request)
    .await
    .map_err(|e| match e {
        HttpResolverError::ResponseTooLarge => DidWebError::ResponseTooLarge,
        e => DidWebError::Request(url.to_owned(), e),
    })?;
```

The body read in `get_did_doc` is now a plain `read_to_end` — the resolver has
already bounded the bytes:

```rust
let mut document = Vec::new();
body.read_to_end(&mut document)
    .map_err(|e| DidWebError::Response(e.into()))?;
```

`MAX_DID_DOC_SIZE = 1 MiB` remains in `did_web.rs`.

### Why two layers?

`Content-Length` is advisory: a malicious server can omit it or advertise less
than it actually sends. Layer 1 fast-fails for honest large responses; Layer 2
catches dishonest ones. Together they bound allocation in all cases.

### `did_web::tests::resolve` — 2 async unit tests

| Test | Layer | Validates |
|---|---|---|
| `content_length_above_limit_rejected` | Layer 1 | Mock server sends explicit `Content-Length: MAX_DID_DOC_SIZE + 1` with matching body; `resolve()` returns `Err(DidWebError::ResponseTooLarge)` before body bytes reach the caller |
| `oversized_response_returns_error` | Layer 1 + 2 | Mock server returns `MAX_DID_DOC_SIZE + 1` bytes (httpmock auto-adds Content-Length); `resolve()` returns `Err(DidWebError::ResponseTooLarge)` |
## Checklist

### DIFF
--- a/sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
+++ b/sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
@@ -18,8 +18,13 @@
 // specific language governing permissions and limitations under
 // each license.
 
+use std::io::Read;
+
 use super::{did::Did, did_doc::DidDocument};
 use crate::http::{AsyncGenericResolver, AsyncHttpResolver, HttpResolverError};
+
+/// Maximum number of bytes accepted from a DID Web server response body.
+pub(crate) const MAX_DID_DOC_SIZE: u64 = 1024 * 1024; // 1 MiB
 
 const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
 
@@ -55,6 +60,9 @@ pub enum DidWebError {
 
     #[error("invalid web DID: {0}")]
     InvalidWebDid(String),
+
+    #[error("response body exceeded size limit of {MAX_DID_DOC_SIZE} bytes")]
+    ResponseTooLarge,
 }
 
 pub(crate) async fn resolve(did: &Did<'_>) -> Result<DidDocument, DidWebError> {
@@ -96,8 +104,13 @@ async fn get_did_doc(url: &str) -> Result<Vec<u8>, DidWebError> {
     };
 
     let mut document = Vec::new();
-    body.read_to_end(&mut document)
+    body.take(MAX_DID_DOC_SIZE + 1)
+        .read_to_end(&mut document)
         .map_err(|e| DidWebError::Response(e.into()))?;
+
+    if document.len() as u64 > MAX_DID_DOC_SIZE {
+        return Err(DidWebError::ResponseTooLarge);
+    }
 
     Ok(document)
 }
@@ -190,7 +203,7 @@ mod tests {
         use super::did;
         use crate::identity::claim_aggregation::w3c_vc::{
             did_doc::DidDocument,
-            did_web::{self, PROXY},
+            did_web::{self, DidWebError, MAX_DID_DOC_SIZE, PROXY},
         };
 
         #[tokio::test]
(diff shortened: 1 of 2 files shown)


tmathern (commented):
- Is body always a lazy reader pulling bytes from the socket (on demand) then? Because if `AsyncGenericResolver` in its resolution (the .await) takes the full response body into memory, the fix is too alte to prevent OOM. I would check what happens there, to make sure no case is missed where there can be OOM before we get to get_did_doc.
- If streaming, you may also be able to check content-length in the header, and return a responseTooLargeError. (This is more for the case where a too large reponse was generated without attack intent, since content-length is a honor system… But may help to catch things early).

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -18,9 +18,14 @@
 // specific language governing permissions and limitations under
 // each license.
 
+use std::io::Read;
+
 use super::{did::Did, did_doc::DidDocument};
 use crate::http::{AsyncGenericResolver, AsyncHttpResolver, HttpResolverError};
 
+/// Maximum number of bytes accepted from a DID Web server response body.
+pub(crate) const MAX_DID_DOC_SIZE: u64 = 1024 * 1024; // 1 MiB
```
mauricefisher64, line 27: Is this big enough?  I would check with @scouten-adobe

mauricefisher64 (approved):

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
ssanthosh: @scouten-adobe : Can you please check and confirm ?

tmathern (approved):

ssanthosh:
> * Is body always a lazy reader pulling bytes from the socket (on demand) then? Because if `AsyncGenericResolver` in its resolution (the .await) takes the full response body into memory, the fix is too alte to prevent OOM. I would check what happens there, to make sure no case is missed where there can be OOM before we get to get_did_doc.
> * If streaming, you may also be able to check content-length in the header, and return a responseTooLargeError. (This is more for the case where a too large reponse was generated without attack intent, since content-length is a honor system… But may help to catch things early).

@tmathern : Handled both the comments with https://github.com/contentauth/c2pa-rs/pull/2070/commits/c1519c6ec7be42953ed298f3c390573ae3b6a84c which creates a max response body config for `AsyncGenericResolver` and only `get_did_doc` sets and uses that config to limit the size of the response body. We also have a early guard for max response body when response header contains content-length.

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
ssanthosh: I feel 1MB should be sufficient and If needed we can increase it to a larger value later

PR #2075, opened 2026-04-20
fix: parse PEM cert chain to DER in RemoteSigner construction

byh-trufo:
The Remote variant of SignerSettings::c2pa_signer() was passing the raw PEM string bytes as a single cert blob, rather than parsing individual PEM certificates into DER. This caused two issues:

1. certs contained raw PEM text instead of DER-encoded certificates
2. reserve_size was based on PEM text length rather than DER cert sizes

Parse the PEM cert chain with pem::parse_many() to extract individual DER-encoded certificates, matching the behavior of the Local variant which uses create_signer::from_keys() for proper PEM handling.

## Checklist

### DIFF
--- a/sdk/src/settings/signer.rs
+++ b/sdk/src/settings/signer.rs
@@ -118,13 +118,21 @@ impl SignerSettings {
                 tsa_url,
                 referenced_assertions: _,
                 roles: _,
-            } => Ok(Box::new(RemoteSigner {
-                url,
-                alg,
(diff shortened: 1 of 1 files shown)


trufo-ai:
need to sign the CLA

mpgopala:
Created https://github.com/contentauth/c2pa-rs/pull/2414 for this. The PR needed few corrections and tests.

PR #2078, opened 2026-04-20
Byh/ingredient digital source type support

byh-trufo:

## Checklist

### DIFF
--- a/sdk/src/assertions/actions.rs
+++ b/sdk/src/assertions/actions.rs
@@ -183,6 +183,12 @@ pub enum DigitalSourceType {
     Other(String),
 }
 
+impl From<String> for DigitalSourceType {
+    fn from(s: String) -> Self {
+        serde_json::from_value::<DigitalSourceType>(serde_json::Value::String(s)).unwrap()
+    }
+}
+
 impl fmt::Display for DigitalSourceType {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.serialize(f)
--- a/sdk/src/assertions/actions.rs
+++ b/sdk/src/assertions/actions.rs
@@ -962,8 +962,7 @@ pub mod tests {
     fn digital_source_type_from_string() {
         assert_eq!(
             DigitalSourceType::from(
-                "http://cv.iptc.org/newscodes/digitalsourcetype/trainedAlgorithmicMedia"
-                    .to_owned(),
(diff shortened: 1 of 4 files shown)


gpeacock:
We know we need to add this, but we are figuring out conformance requirements first.

byh-trufo:
Because the use of this field breaks existing `c2pa-rs` validation (before this PR), should we add a test in the Conformance v0.2 rubrics to check that validators will be able to handle this before attaining conformant status if v2.4 is specified?

Happy to set it up ([here?](https://github.com/contentauth/c2pa-conformance-tool-cli/tree/main/crates/c2pa-validate/tests)) if the task force is not opposed to doing so at this point in time.

sdk/src/assertions/ingredient.rs
```diff
@@ -1216,6 +1233,42 @@ pub mod tests {
         assert!(v3_decoded.is_v3_compatible());
```
tmathern, line 1246: Will those changes affect `is_v3_compatible` checks? Are all the checks in there https://github.com/contentauth/c2pa-rs/blob/781c8dbdc47cf50d35f5382cc767ce7b96a7de6a/sdk/src/assertions/ingredient.rs#L179-L186 still needed?

sdk/src/assertions/actions.rs
```diff
@@ -183,6 +183,12 @@ pub enum DigitalSourceType {
     Other(String),
 }
 
+impl From<String> for DigitalSourceType {
+    fn from(s: String) -> Self {
+        serde_json::from_value::<DigitalSourceType>(serde_json::Value::String(s)).unwrap()
```
tmathern: Is the unwrap here safe? What about error propagation or error matching in callers?

sdk/src/assertions/ingredient.rs
byh-trufo: updated the checks to match v2.4 specs in a new commit

sdk/src/assertions/actions.rs
```diff
@@ -183,6 +183,13 @@ pub enum DigitalSourceType {
     Other(String),
 }
 
+impl From<String> for DigitalSourceType {
+    fn from(s: String) -> Self {
+        serde_json::from_value(serde_json::Value::String(s.clone()))
+            .unwrap_or(DigitalSourceType::Other(s))
```
tmathern, line 189: @lrosenthol Ok with that?

mauricefisher64:
Addressed in 2447

PR #2085, opened 2026-04-24
fix: Prevent duplicate timestamp assertion

timmurphydev:
When the `auto_timestamp_assertion` builder setting is used with `skip_existing` the SDK populates a list of claim URIs based on the `fetch_scope` setting. Then it loops through the ingredients removing any claim URIs that have a COSE timestamp. It also needs to remove any claim URIs present in a `c2pa.time-stamp assertion` on the ingredient. The timestamp assertion check gets skipped by the `continue;` statement when the ingredient has a COSE timestamp.

This PR removes the continue so both checks always run.

## Checklist


ok-nick (approved):
Nice catch!

PR #2089, opened 2026-04-27
fix: `c2pa-c-ffi` UB detected by miri

ok-nick:
This fixes UB found in `c2pa-c-ffi` detected by running [`miri`](https://github.com/rust-lang/miri/) on the test suite.

We should add this to our CI, although we may need to skip some tests that can take up to 30min to execute.

Command used for testing:
```bash
MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tree-borrows" cargo +nightly miri nextest run -p c2pa-c-ffi --no-default-features --features rust_native_crypto,file_io --no-fail-fast
```
Note, `nextest` **significantly** reduces execution time by parallelizing tests, however, it is not required.


c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -416,45 +415,35 @@ macro_rules! cstr_or_return {
             $crate::CimplError::null_parameter(stringify!($ptr)).set_last();
             return $err_val;
         } else {
-            // SAFETY: We create a bounded slice up to MAX_CSTRING_LEN.
-            // Caller must ensure ptr is valid for reading and points to a
-            // null-terminated string within MAX_CSTRING_LEN bytes.
-            let bytes = unsafe {
-                std::slice::from_raw_parts(ptr as *const u8, $crate::macros::MAX_CSTRING_LEN)
```
ok-nick, line 423: `slice::from_raw_parts` [safety requires](https://github.com/contentauth/c2pa-rs/pull/2089):
> data must be non-null, valid for reads for len * size_of::<T>() many bytes, and it must be properly aligned.

The existing behavior breaches that contract and is detected by `miri` as UB. The new implementation expects the caller to uphold that the string ends with a null-terminator, although it will still error if the string exceeds a particular length.

tmathern:
+1 on adding to CI!

tmathern (approved):

c2pa_c_ffi/src/c2pa_stream.rs
```diff
@@ -173,7 +173,7 @@ impl Seek for C2paStream {
             std::io::SeekFrom::End(pos) => (pos, C2paSeekMode::End),
         };
 
-        let new_pos = unsafe { (self.seeker)(&mut (*self.context), pos as isize, mode) };
+        let new_pos = unsafe { (self.seeker)(self.context, pos as isize, mode) };
```
tmathern, line 176: I wonder if those fixes will have side-effects. I'd suggest releasing switfly after merging, in case there are follow-up fixes to do in downstream SDKs (Hyrum's law in action).

cli/Cargo.toml
```diff
@@ -27,7 +27,7 @@ networking = [
 
 [dependencies]
 anyhow = "1.0"
-atree = "=0.5.2"
+atree = "0.5.4"
```
tmathern, line 30: Did atree fix the recent issues it had? We pinned it because it caused panics. Running `check-release` to verify that.

ok-nick: Yes, v0.5.2 also has UB, I upgraded it to the latest which fixes that. Long-term we should look at #1981.

c2pa_c_ffi/src/c2pa_stream.rs
ok-nick: The parameter was already a mutable pointer, so we were dereferencing it, then converting it to a mutable ref, then auto dereferencing it to a mutable pointer.

tmathern: Yes, but better safe than sorry. Trust, but verify. Updating the downstream SDKs is the ultimate test, especially if UB was involved.

cli/Cargo.toml
tmathern: (I meant it caused panics :) But seems resolved)

c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -416,45 +415,35 @@ macro_rules! cstr_or_return {
...
-            };
-            match std::ffi::CStr::from_bytes_until_nul(bytes) {
```
gpeacock, line 425: This seemed nice since it can return an error, but since it is based on a false pretense, it doesn't help. I get it. Thanks for catching this.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -4250,7 +4253,7 @@ verify_after_sign = true
     fn test_c2pa_reader_detailed_json() {
         use std::ffi::CStr;
 
-        let source_image = include_bytes!(fixture_path!("cloud.jpg"));
```
gpeacock, line 4253: why the change to C.jpg from cloud.jpg? I don't object, but I'm curious since it could change behavior

gpeacock (approved):
It took me a bit to make sure these changes are right. In general we make no assumptions about context, but in the specific case of the TestStream, it's ok.

c2pa_c_ffi/src/c_api.rs
ok-nick: I changed them for tests that don't actually need to test remote manifest behavior. Otherwise we are executing a network request when we don't have to. The networking libraries also don't play nice with `miri`.

PR #2093, opened 2026-04-28
feat: add `init trust`, trust sidecars, and atomic sidecar writes, plus fixes

gpeacock:
**PR Description:**

Adds c2patool init trust to bootstrap C2PA trust-list files locally, and auto-loads those files as sidecars on subsequent runs — eliminating the need to pass --trust_anchors etc. manually. Also fixes claim_generator_info resolution and hardens logging initialization.

**init trust subcommand**

init trust fetches the official C2PA trust list PEM and writes it atomically next to --settings
init trust --legacy additionally fetches interim Content Credentials anchors, store config, and allowed-SHA list
Sidecar files are auto-loaded on every run; finding any sidecar enables trust checks automatically
Writes use a temp-file + fsync + rename pattern to avoid partial writes on crash

**SDK fix: claim_generator_info resolution** 
[CAI-10486](https://jira.corp.adobe.com/browse/CAI-10486)
the claim_generator_info from settings was never getting used in practice. This fixes things to work in the correct order:
manifest defintion -> settings -> default.
ManifestDefinition::claim_generator_info now deserializes to an empty Vec when absent or [], resolved at claim-build time via settings → library default
Fixes a logic inversion that could suppress or duplicate the tool's own generator entry
c2patool will now report itself as the generator if no other claim_generator specified (was using c2pa-rs library)

**allow periods in pathnames**
[CAI-11758](https://jira.corp.adobe.com/browse/CAI-11758)

**Logging hardening**
[CAI-10487](https://jira.corp.adobe.com/browse/CAI-10487)
Replaces unsound set_var + env_logger::init() with env_logger::Builder::from_env
Tests: integration tests for log-level behavior; unit tests for claim_generator_info deserialization and settings-based resolution

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -162,8 +162,13 @@ pub struct ManifestDefinition {
     /// This is typically a reverse domain name.
     pub vendor: Option<String>,
 
-    /// Claim Generator Info is always required with an entry
-    #[serde(default = "default_claim_generator_info")]
+    /// Software that generated the claim, as a list of [`ClaimGeneratorInfo`].
+    ///
+    /// In JSON, when this key is **omitted** (or the array is empty), the value is
+    /// resolved at claim-building time: settings.builder.claim_generator_info if set, otherwise
+    /// on the active [`Context`] is used if set, otherwise [`ClaimGeneratorInfo::default`].
+    /// A non-empty list in the definition is used as given.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub claim_generator_info: Vec<ClaimGeneratorInfo>,
 
     /// Optional manifest metadata. This will be deprecated in the future; not recommended to use.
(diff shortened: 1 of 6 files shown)


sdk/src/builder.rs
```diff
@@ -1337,19 +1332,20 @@ impl Builder {
...
-                &self.context.settings().builder.claim_generator_info;
-            match claim_generator_info_settings {
-                Some(claim_generator_info_settings) => {
-                    claim_generator_info.push(claim_generator_info_settings.clone().try_into()?);
-                }
-                _ => {
-                    claim_generator_info.push(ClaimGeneratorInfo::default());
-                }
-            }
+            //Manifest omitted `claim_generator_info` (or `[]`): one entry from settings, else
```
tmathern, line 1336: We pick the first entry from settings if there are multiple? Do settings even let you define multiple currently?

tmathern (approved):

sdk/src/builder.rs
gpeacock: always the first setting, this is a leftover from  v1 claim support which we have not yet removed.

PR #2095, opened 2026-04-29
fix: Improve archive APIs to be able to link archives to actions

tmathern:
The recently added archive APIs simplify the handling of ingredient archives, but there is a bug where when those APIs are used, two ingredients can't be linked to actions when a Builder signs (see test scenario `test_two_ingredient_archives_link_distinctly_via_actions`).

This fixes that. 

## Checklist


sdk/src/builder.rs
```diff
@@ -3266,7 +3266,11 @@ impl Builder {
             }
         }
 
-        let ingredient = reader.to_ingredient()?;
+        let archive_ingredient_id = reader.active_archive_ingredient_id();
```
gpeacock: we don't know if the saved id came from the label or the instanceId here. We should just save and restore the  user assigned label here, and it should be done in reader.to_ingredient.()

sdk/src/builder.rs
```diff
@@ -3244,21 +3258,26 @@ impl Builder {
...
-                )));
-            }
-            None => {
-                return Err(Error::BadParam(format!(
-                    "expected a C2PA ingredient archive (org.contentauth.archive.metadata with archive:type {:?}); use add_ingredient_from_reader or add_ingredient_from_stream for other stores",
-                    labels::ARCHIVE_TYPE_INGREDIENT
-                )));
+        let ingredient_id = match reader.active_archive_kind() {
+            Some(ArchiveKind::Ingredient { ingredient_id }) => Some(ingredient_id),
+            // we should return an error here, but be tolerant in the transition if this has a parent.
```
scouten-adobe, line 3267: Consider adding commented-out code here that would return an error and a link to a ticket that would have us make that change. Otherwise I suspect we'll forget about this.

scouten-adobe (approved):
One minor suggestion; otherwise :+1:.

sdk/src/builder.rs
gpeacock: I'll write a ticket for it.

PR #2099, opened 2026-04-30
fix: Harden against integer underflow attack in GIF XMP validation

ssanthosh:
# Security Fix: GIF XMP Magic Trailer usize Underflow (Panic / Incorrect Output)

## Vulnerability

`CAIReader::read_xmp` in `sdk/src/asset_handlers/gif_io.rs` subtracts the
constant `257` from `bytes.len()` (a `usize`) without first confirming that the
decoded payload is at least 257 bytes long. When an attacker supplies a GIF with
a short XMP Application Extension the subtraction underflows:

- **Debug builds** — Rust panics with `"attempt to subtract with overflow"`,
  crashing the host process.
- **Release builds** — `usize` wraps to `usize::MAX - N`; the bounds check at
  `bytes.get(usize::MAX - N)` returns `None` (first validation is silently
  skipped), and `bytes.truncate(usize::MAX - N)` is a no-op; the function
  returns a garbage XMP string instead of `None`.

### Root cause

```rust
// Vulnerable — no minimum-length guard before subtraction
if let Some(byte) = bytes.get(bytes.len() - 257) { // panics when len < 257
    ...
}
...
bytes.truncate(bytes.len() - 257); // panics when len < 257
```

### Attack surface

Any caller of `Reader::from_stream("image/gif", ...)` on an attacker-supplied
GIF. A ~30-byte crafted GIF with a short `XMP Data` application extension is
sufficient. The panic is unconditional in debug builds; in release it silently
produces incorrect XMP data.

### Affected location

| File | Function | Issue |
|---|---|---|
| `sdk/src/asset_handlers/gif_io.rs` | `CAIReader::read_xmp` | `bytes.len() - 257` underflows when `bytes.len() < 257` (lines 58 and 69) |

## Fix

Add a minimum-length guard before the first subtraction. If the decoded payload
is shorter than 257 bytes it cannot carry a valid XMP magic trailer; `read_xmp`
returns `None` immediately:

```rust
// The XMP magic trailer is exactly 257 bytes. A block shorter than that
// cannot carry a valid trailer; reject it rather than underflowing the
// usize subtraction below (panic in debug, wrap-to-MAX in release).
if bytes.len() < 257 {
    return None;
}
```

The single guard protects both subtraction sites. No other changes were needed.

## Test Added

`test_xmp_short_trailer_does_not_panic` in `asset_handlers::gif_io::tests`:

Constructs a minimal valid GIF89a (header + logical screen descriptor + XMP
Application Extension with a 10-byte payload + trailer) and calls `read_xmp`.
Asserts the return value is `None` — not a panic, not garbage XMP.

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/gif_io.rs
+++ b/sdk/src/asset_handlers/gif_io.rs
@@ -53,6 +53,13 @@ impl CAIReader for GifIO {
             .ok()?
             .map(|marker| marker.block.data_sub_blocks.to_decoded_bytes())?;
 
+        // The XMP magic trailer is exactly 257 bytes. A block shorter than that
+        // cannot carry a valid trailer; reject it rather than underflowing the
+        // usize subtraction below (panic in debug, wrap-to-MAX in release).
+        if bytes.len() < 257 {
+            return None;
+        }
+
         // TODO: this should be validated on construction
         // Validate the 258-byte XMP magic trailer (excluding terminator).
         if let Some(byte) = bytes.get(bytes.len() - 257) {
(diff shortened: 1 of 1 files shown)


sdk/src/asset_handlers/gif_io.rs
```diff
@@ -53,6 +53,13 @@ impl CAIReader for GifIO {
             .ok()?
             .map(|marker| marker.block.data_sub_blocks.to_decoded_bytes())?;
 
+        // The XMP magic trailer is exactly 257 bytes. A block shorter than that
+        // cannot carry a valid trailer; reject it rather than underflowing the
+        // usize subtraction below (panic in debug, wrap-to-MAX in release).
+        if bytes.len() < 257 {
```
tmathern: Should `257` become a constant then? makes the code more readable

tmathern (approved):

sdk/src/asset_handlers/gif_io.rs
ssanthosh: Performed the changes with https://github.com/contentauth/c2pa-rs/pull/2099/commits/c100cccced8abe267a5f283c6e19fb943ddafc91

PR #2103, opened 2026-04-30
perf: lazy-load ingredient manifest stores from shared Arc<Store> 

gpeacock:
we wanted to make ingredient manifest available via reader if someone needed them so we generate them for every ingredient now on Reader. But most of the time no one needs these, and generating them for complex manifests with many ingredients can be costly. Videos can have thousands of ingredients. This PR switches to lazy generation of manifests, so we only generate ingredient manifest when someone asks for one. This also helps operations like Reader::into_builder. It is both a speed and memory improvement. It also helps prevent some DOS type attacks.


tmathern (approved):

sdk/src/ingredient.rs
```diff
@@ -322,9 +327,25 @@ impl Ingredient {
...
+            Ok(cow) => Some(cow),
+            Err(_) => self
+                .try_get_deferred_manifest_data(&r.identifier)
+                .ok()
+                .map(Cow::Owned),
+        }
+    }
+
+    fn try_get_deferred_manifest_data(&self, id: &str) -> Result<Vec<u8>> {
+        let mdr = self.manifest_data.as_ref().ok_or_else(|| Error::NotFound)?;
```
ok-nick, line 341:
```suggestion
        let mdr = self.manifest_data.as_ref().ok_or(Error::NotFound)?;
```

sdk/src/ingredient.rs
```diff
@@ -1255,11 +1293,14 @@ impl Ingredient {
...
-            })
+            self.resources
+                .get(id)
+                .or_else(|_| {
+                    resources
+                        .ok_or_else(|| Error::NotFound)
+                        .and_then(|r| r.get(id))
+                })
+                .or_else(|_| self.try_get_deferred_manifest_data(id).map(Cow::Owned))
         };
```
ok-nick, line 1304: It looks like we use this closure to get a lot of different resource types. Not a huge deal, but maybe we should move the try get deferred logic only to the place we get the manifest data.

sdk/src/ingredient.rs
```diff
@@ -489,6 +512,21 @@ impl Ingredient {
...
+    pub(crate) fn set_deferred_manifest_data(&mut self, source_store: Arc<Store>) -> Result<()> {
+        let active = self
+            .active_manifest()
+            .map(str::to_owned)
+            .ok_or_else(|| Error::NotFound)?;
+        if source_store.get_claim(&active).is_none() {
+            return Ok(());
+        }
+        self.source_store = Some(source_store);
+        self.manifest_data = Some(ResourceRef::new("application/c2pa", active));
```
ok-nick, line 526: Is it possible that some code path may try looking for `manifest_data` but fail to find it in the resource store? Unsure if we should set a resource ref here if we aren't positive it's in the resource store.

gpeacock: The point here is that we will generate the manifest when you request it, so we know it isn't there,but it will be.

PR #2117, opened 2026-05-05
Add text/plain asset handler via c2pa-text crate

erik-sv:
## Summary

Adds a `TextIO` asset handler for `text/plain` (C2PA 2.4 Appendix A.8, unstructured text) with manifest read, write, and remove support. The handler uses the [c2pa-text](https://github.com/encypherai/c2pa-text) reference implementation crate, which encodes JUMBF manifest bytes as invisible Unicode variation selectors. The encoding is reversible and preserves the visible text. NFC normalization is handled inside the crate.

## Feature gating

## Changes

- `sdk/src/asset_handlers/text_io.rs` (new): `TextIO` implementing `CAIReader`, `CAIWriter`, and `AssetIO`, gated behind `plain_text`.
- `sdk/src/asset_handlers/mod.rs`, `sdk/src/jumbf_io.rs`: cfg-gated module and handler registration, plus a feature-parity assertion in `test_get_supported_list` mirroring the `pdf` one.
- `sdk/src/utils/mime.rs`: `txt` and `text/plain` mappings.
- `sdk/Cargo.toml`: `c2pa-text = { version = "3.0.0", optional = true }` (caret requirement, so any 3.x release satisfies it) and the `plain_text` feature.

- Rebased onto current `main`.
- `c2pa-text` bumped from 2.0.0 to 3.0.0, the current crates.io release. The 3.x API is source-compatible for this handler; the major bump covered validation-code strings.
- `write_cai` now preserves the manifest position (@mauricefisher64's review question on #2190, which applies here too). An existing wrapper is replaced in place at its original byte offset via `extract_manifest` plus `encode_wrapper` splicing instead of strip-and-append. A new `text_io_write_preserves_wrapper_position` test covers this.
- Stream reads are bounded with `stream_len` plus `read_to_vec`, replacing the unbounded `read_to_end` (@mauricefisher64's feedback on #2190, applied across the stack).
- `store::tests::test_unsupported_type_without_external_manifest` is now `#[cfg(not(feature = "plain_text"))]`. Its "unsupported" fixture is a `.txt` file, which this feature makes supported. Default CI still runs it.

|---|---|---|---|

- `cargo check -p c2pa` (default features): clean. The feature gates compile out entirely.
- `cargo test -p c2pa --lib --features plain_text,file_io`: 863 passed, 0 failed.
- `cargo clippy --features plain_text,file_io`: no warnings. `cargo +nightly fmt --check`: clean.
- Round-trip, position preservation, object locations, remove, and no-manifest cases covered in the `text_io` tests.


dcondrey:
A few things I noticed:

1. The git dependency on `c2pa-text` won't work for downstream consumers pulling `c2pa-rs` from crates.io. This needs to be a published crate or vendored.
2. The spec requires NFC normalization before hashing (Section A.7.5). I don't see that in the handler or in `c2pa-text` — if the crate doesn't normalize, hashes will differ across platforms that store text in different normalization forms.
3. The PR description mentions AssetPatch support but I don't see an implementation.
4. `get_object_locations_from_stream` uses `text.len()` for byte length — are `result.offset` and `result.length` from `c2pa-text` guaranteed to be byte offsets rather than character offsets? The hash exclusion range needs to be in bytes.
5. Only `text/plain` and `txt` are registered. The spec's unstructured text method applies to any text file without a format-specific embedding section — worth considering whether this should handle additional `text/*` subtypes.

erik-sv:
@dcondrey Thanks for the close read. Addressed inline:

**1. Git dependency.** You're right - crates.io rejects git sources. `c2pa-text` is already published on crates.io at v1.1.0. Fixed in 40c165d; the dependency now points to the registry version.

**2. NFC normalization.** The `c2pa-text` crate handles this. `embed_manifest` normalizes input text to NFC before appending the wrapper, and `extract_manifest` normalizes clean text on output. See [lib.rs L162](https://github.com/encypherai/c2pa-text/blob/main/rust/src/lib.rs) - the crate depends on `unicode-normalization` for this. The handler doesn't need to normalize separately.

**3. AssetPatch.** Good catch - the original PR description was wrong. The handler implements `CAIReader`, `CAIWriter`, and `AssetIO`, not `AssetPatch`. Updated the description. The `c2pa-text` crate does expose `encode_wrapper_padded` for fixed-size wrappers that would enable in-place patching, but wiring that up is follow-up work.

**4. Byte offsets.** `ExtractionResult.offset` and `.length` are byte offsets. The extraction code uses Rust's `char_indices()`, which yields `(byte_index, char)` tuples, and the wrapper end is either `chars[j].0` (byte position) or `text.len()` (byte count). These are in the same unit as `text.len()` in the handler, so the hash exclusion ranges are correct.

**5. Additional text/ subtypes.** Agreed this is worth considering. Starting with `text/plain` keeps the initial surface small. Formats with structural syntax (markdown, HTML, CSV) may need their own handlers to avoid embedding in positions that break format parsers, so they're better handled incrementally.

erik-sv:
@sherifhanna-google Would you be able to review this PR? We discussed text/plain asset handler support in the May 22 Conformance TF meeting and it's a prerequisite for the text conformance rubrics (c2pa-org/conformance#368). Andy mentioned he'd push for a merge. Thanks!

erik-sv:
### Update summary

We've refreshed this PR against the latest `main` and pinned the published reference crate:

- **`c2pa-text` dependency bumped to `2.0.0`** (the released crates.io version). `TextIO` relies only on `embed_manifest` / `extract_manifest`, whose behavior is unchanged in 2.0.0, so this is a drop-in version bump.
- **Resolved a merge artifact** that a prior "Merge `main`" had left in `sdk/Cargo.toml`: kept the `c2pa-text` entry and dropped the `config` dependency, which `main` had already removed (it is unused in the SDK).
- **Verified locally:** `cargo check -p c2pa` is clean and the `text_io` round-trip / object-location / remove / no-manifest tests pass against `c2pa-text 2.0.0`.

### Scope

This handler intentionally implements a single method: **unstructured text per C2PA 2.4 Appendix A.8** — the Manifest Store (JUMBF) is encoded as invisible Unicode Variation Selectors. The only media type it registers is **`text/plain`** (extension `.txt`).

### Companion work (separate PRs, for clean review)

The remaining C2PA text methods are distinct mechanisms and are best reviewed independently. We're opening them as companion PRs rather than expanding this one:

- **HTML — Appendix A.7** (`<script type="application/c2pa">` / `<link rel="c2pa-manifest">`): a dedicated `HtmlIO` handler.
- **Structured text — Appendix A.9** (ASCII-armour comment/front-matter block carrying a URL or `data:` reference): covers Markdown, source code, YAML/TOML, etc. `image/svg+xml` stays with the existing `SvgIO` (A.3.3). We'll also file a tracking issue on the handling of generic XML media types (`text/xml`, `application/xml`, `application/xhtml+xml`) relative to A.9.

Each builds on the `c2pa-text 2.0.0` foundation this PR introduces.

mauricefisher64:
Is this PR superceded by 2190?

dcondrey (commented):
**On direction:** these three PRs do cover distinct spec sections — #2117 (A.8, variation-selector carrier), #2188 (A.7, `<script type="application/c2pa">`), #2190 (A.9, ASCII-armour comment) — but all three route the reference implementation's text support through the external `c2pa-text` crate, and #2190 covers only a subset of A.9's formats. I'd rather see text embedding implemented natively in `c2pa-rs` as one dependency-free surface spanning A.7–A.9 with the full A.9 format range. That's what I'm putting together, and I'm raising it here so maintainers can weigh that direction before committing to the external-crate path across the series.

The spec's own hierarchy reinforces this: the structured and format-specific methods are normatively preferred, and the variation-selector method is explicitly last-resort — "should only be used where no other embedding method is feasible" — and still marked under review. A native implementation can follow that hierarchy directly and cover the full A.7–A.9 range in one dependency-free surface.

**Dependency / NFC:** the git-source issue is resolved now that `c2pa-text` is pulled from crates.io (40c165d). The NFC-normalization question is still open, though — 40c165d only touched `Cargo.toml`, so whether producer and consumer normalize to NFC before `c2pa.hash.data` is computed still needs confirming. Without it, hashes diverge across platforms that store text in different normalization forms.

**`stream_to_string`:** the unbounded `read_to_end` in all three handlers should read into a bounded, `try_reserve`-backed buffer — a large input can otherwise OOM the process.

erik-sv:
No. The three PRs implement three different embedding methods from the 2.4 spec: this one is A.8 unstructured text, #2188 is the A.7 HTML method, #2190 is the A.9 structured-text method. #2188 and #2190 are stacked on this PR, which carries the shared c2pa-text dependency. That is why their diffs overlap with this one until it merges. The description now has the full layering table.

All three are rebased on current main, gated behind non-default features per Eric's request, and the write-position and bounded-read feedback is addressed in code with regression tests.

scouten-adobe:
Closed in favor of #2494.

PR #2120, opened 2026-05-07
chore: Enable sync CAWG validation

gpeacock:
We currently will auto validate CAWG identity when the verify apis are called async, but not when they are called synchronously (it silently logs a warning and skips validation).

This allows cawg verification in sync calls by default. Identity validation can be disabled with a setting.

c_ffi no longer needs a tokio runtime to block on async validation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .cargo/config.toml
 
 .vscode
+.claude
 
 /semver-checks/target/
 sdk/target/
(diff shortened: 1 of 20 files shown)


tmathern (approved):

tmathern:
For even more improvements, please consider: https://github.com/contentauth/c2pa-rs/pull/2152 (Other topic, other PR).

scouten-adobe (approved):
This looks really good. Thank you!

PR #2132, opened 2026-05-11
refactor!: remove `config` crate and use `serde_json` directly

ok-nick:
This PR removes the `config` crate in favor of using `serde_json` to handle merging and setting values directly. Running benchmarks show significant performance (~90-98%) and memory (up to ~1.5-77x depending on benchmark) improvements when dealing with settings. This change also removes about 9 dependencies (including duplicates due to versioning).

### Performance benchmarks via `criterion`:
image

### Memory benchmarks via `dhat`:
Before:
image
After:
image
Visualized with courtesy of claude:
image

### Semver changes
- The `T: Into<config::Value>` bound is now `T: Into<serde_json::Value>` (should be a near equivalent).
- Changed `T: Deserialize<'de>` to `T: DeserializeOwned`. It is unlikely a caller would be deserializing to a borrowed ref.
- The `with_`... were changed to take `&self`, that is backwards compatible.


scouten-adobe (approved):
Very nice find, @ok-nick!

sdk/src/settings/mod.rs
```diff
@@ -1014,29 +918,79 @@ impl SettingsValidate for Settings {
     }
 }
 
+/// Overlays `overlay` onto `target`. Objects are merged key-by-key, and any
+/// other value (e.g. `null` and arrays) replaces the target value.
+fn merge_json(target: &mut Value, overlay: Value) {
+    match (target, overlay) {
+        (Value::Object(target_map), Value::Object(overlay_map)) => {
+            for (key, overlay_value) in overlay_map {
+                merge_json(target_map.entry(key).or_insert(Value::Null), overlay_value);
```
gpeacock: We should either flatten this recursion or put some limit check on it. It somehow always seems to come up.

gpeacock (approved):
Thank you! That config crate was always awful for performance. This really helps. 
I added one comment about checkin recursion. Otherwise this looks great.

sdk/src/settings/mod.rs
ok-nick: Added.

tmathern (approved):

PR #2134, opened 2026-05-11
ci: run walltime read and sign benches with macro runners

ok-nick:
Experimenting with Codspeed macro runners for walltime benchmarks (accurate/stable I/O measurements) on read and sign benches.


gpeacock (approved):
I don't know if there are other CI config issues with this,but it looks good to me

PR #2135, opened 2026-05-11
fix: archive preserves duplicate label assertions

alextrnnn:
Restoring a builder from an archive used the instance-suffixed label (test__1) instead of the base label, so re-signing tacked on another suffix and produced the malformed label test__1__1.


gpeacock (approved):
Yes, I've seen this but had not tracked it down, thank you!

PR #2138, opened 2026-05-12
fix: case-insensitive enums in settings to preserve backwards compatibility with `config` crate

ok-nick:
Allows for case-insensitive enums to preserve backwards compatibility with the `config` crate.


tmathern (approved):

gpeacock (approved):
cleaner

PR #2140, opened 2026-05-12
fix: Harden against integer overflow attacks in hash_stream_by_alg_with_progress range validation

ssanthosh:
# Security Fix: HashRange u64 Addition Overflow → Panic (DoS)

## Vulnerability

`hash_stream_by_alg_with_progress` in `sdk/src/utils/hash_utils.rs` performs two
unchecked `u64` additions on `HashRange` fields (`start`, `length`) that are
deserialized directly from attacker-controlled CBOR in a C2PA manifest. When
`start + length` exceeds `u64::MAX` the process panics with
`"attempt to add with overflow"` and terminates (exit code 101).

This crashes **any** application that verifies a C2PA manifest. No valid
signature is required because manifest verification continues past COSE
signature failure.

### Affected asset types

- **DataHash** exclusion path — any file type (JPEG, PNG, GIF, TIFF, …)
- **BmffHash** inclusion path — BMFF-based assets (MP4, MOV, HEIC, AVIF)

### Root cause

```rust
// Line 260 — exclusion path: summary range_end check
let range_end = hr[num_blocks - 1].start() + hr[num_blocks - 1].length();
//             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//             Panics in debug; wraps silently in release (bypass bounds check)

// Line 358 — inclusion path: per-range end calculation
let end = inclusion.start() + inclusion.length() - 1;
//        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//        Panics when start + length > u64::MAX
```

Note: the per-range exclusion calculation (lines 286–291) was **already** using
`checked_add`/`checked_sub`. Only the two sites above were unguarded.

### Affected locations

| File | Line | Expression | Issue |
|---|---|---|---|
| `sdk/src/utils/hash_utils.rs` | 260 | `hr[last].start() + hr[last].length()` | Unchecked u64 addition — exclusion and inclusion paths |
| `sdk/src/utils/hash_utils.rs` | 358 | `inclusion.start() + inclusion.length() - 1` | Unchecked u64 addition — inclusion path only |

## Fix

Replace both bare `+` operations with `checked_add`, propagating an
`Error::BadParam` on overflow:

## Tests Added

Two regression tests in `utils::hash_utils::tests`:

| Test | Path exercised | HashRange used |
|---|---|---|
| `test_exclusion_range_overflow_returns_error` | Exclusion (`is_exclusion = true`) | `start = u64::MAX - 10, length = 20` |
| `test_inclusion_range_overflow_returns_error` | Inclusion (`is_exclusion = false`) | `start = u64::MAX, length = 1` |

Both assert `result.is_err()` — the function must return `Err`, not panic.

## Checklist

### DIFF
--- a/sdk/src/utils/hash_utils.rs
+++ b/sdk/src/utils/hash_utils.rs
@@ -257,7 +257,10 @@ where
 
             // verify structure of blocks
             let num_blocks = hr.len();
-            let range_end = hr[num_blocks - 1].start() + hr[num_blocks - 1].length();
+            let range_end = hr[num_blocks - 1]
+                .start()
+                .checked_add(hr[num_blocks - 1].length())
+                .ok_or(Error::BadParam("hash range overflow".to_string()))?;
             let data_end = data_len - 1;
 
             // range extends past end of file so fail
@@ -355,7 +358,11 @@ where
                         continue;
                     }
 
-                    let end = inclusion.start() + inclusion.length() - 1;
+                    let end = inclusion
+                        .start()
+                        .checked_add(inclusion.length())
+                        .ok_or(Error::BadParam("inclusion range overflow".to_string()))?
+                        - 1;
                     let inclusion_start = inclusion.start();
 
                     // add new BMFF V2 offset as a new range to be included so that we can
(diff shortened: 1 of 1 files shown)


sdk/src/utils/hash_utils.rs
```diff
@@ -257,7 +257,10 @@ where
 
             // verify structure of blocks
             let num_blocks = hr.len();
-            let range_end = hr[num_blocks - 1].start() + hr[num_blocks - 1].length();
+            let range_end = hr[num_blocks - 1]
```
tmathern, line 260: While at it, saturating substract?

tmathern (approved):

sdk/src/utils/hash_utils.rs
```diff
@@ -257,7 +257,10 @@ where
...
+                .start()
+                .checked_add(hr[num_blocks - 1].length())
+                .ok_or(Error::BadParam("hash range overflow".to_string()))?;
             let data_end = data_len - 1;
```
tmathern, line 264: Protect that one too with saturating_sub?

sdk/src/utils/hash_utils.rs
```diff
@@ -355,7 +358,11 @@ where
                         continue;
                     }
 
-                    let end = inclusion.start() + inclusion.length() - 1;
+                    let end = inclusion
+                        .start()
+                        .checked_add(inclusion.length())
+                        .ok_or(Error::BadParam("inclusion range overflow".to_string()))?
+                        - 1;
```
tmathern, line 365: Protect that one too with saturating_sub?

tmathern (commented):
I think saturating_sub should be used throughout here for the fix to be more solid.

sdk/src/utils/hash_utils.rs
ssanthosh: It is not needed as the match expression in line 253 checks `if !hr.is_empty()` and hence `num_blocks - 1` should not have any underflow issues

sdk/src/utils/hash_utils.rs
ssanthosh: It is not needed as we have a guard on line 248  which ensures data_len is >= 1 : 
```
if data_len < 1 {
        return Err(Error::OtherError("no data to hash".into()));
    }
```

sdk/src/utils/hash_utils.rs
ssanthosh: It is not needed as we have the guard for zero size inclusion in line 357 : 
```
if inclusion.length() == 0 {
                        continue;
}
```

tmathern (approved):

PR #2147, opened 2026-05-13
fix: error with validation results on invalid manifest after verify after sign

### LINKED ISSUE #1704: Signing an update manifest with format `application/c2pa` will attempt to validate hash bindings
When signing an update manifest with an empty vec as the input and output stream, with the intention being to get only C2PA bytes, the SDK will try verifying any ingredient's hash bindings. Since the original ingredient bytes are gone, this will always fail with "Hashes do not match."

A solution is to ignore the `Claim::verify_hash_bindings` call when the format is `application/c2pa`. However, the builder will write the c2pa bytes to the output stream and return it as a vec, which is redundant. This can be combatted by specifying `io::empty()` as the output stream, although this is not an elegant solution. Another option is to have a method to sign an update manifest with no input or output stream, and it would return only a C2PA manifest.  

Example of where the problem occurs:
```rust
let mut builder = Builder::new();
builder.set_intent(BuilderIntent::Update);

builder.add_ingredient_from_stream(
    json!({
        "relationship": "parentOf"
    })
    .to_string(),
    "image/jpeg",
    "some/file.jpg"
).unwrap();

builder.sign(
    &Settings::signer()?,
    "application/c2pa",
    &mut Cursor::new(Vec::new()),
    &mut Cursor::new(Vec::new()),
}.unwrap();
```
This is a "fix" to the problem since we don't verify stores where the stream length is 0 (since `io::empty()` never writes). However, we should still be verifying the store in this case, which is another problem.
```rust
builder.sign(
    &Settings::signer()?,
    "application/c2pa",
    &mut io::empty(),
    &mut io::empty(),
}.unwrap();
```

Context: https://github.com/contentauth/c2pa-rs/pull/1638/changes#r2620856635

### LINKED ISSUE #1875: Fix `verify_after_sign` setting
The `verify_after_sign` setting was not running prior to #1638, which "fixed" it to run. There was a problem discovered in the implementation in that it stops on the first validation error, even in the case that you attest to an ingredient's validation error. Thus, this setting was disabled by default in #1746.

There is also an additional edge case documented in #1704 when signing with the "application/c2pa" format.

ok-nick:
Fixes verify after sign so that it builds up the validation results and returns them in an err if deemed an invalid manifest. It also adds the verify after sign check to the other signing paths, including the embeddable APIs.

The PR introduces a new `ValidationFailureSummary` so that on error, it will display a concise summary of what failures, on what ingredients occurred. This can also be obtained via `ValidationResults::failure_summary` for the general case. `ValidationResults` also now impls `Display` with a more user-friendly interface.

Also adds a new setting `verify_after_sign_hash` which rehashes the asset and validates against the hard binding assertion stored. By default this setting is false due to the performance implications. I'm keeping `verify_after_sign` default to false for now, let's get the fix out first.

- Closes #1875
- Closes #1704


c2pa_c_ffi/src/c_api.rs
```diff
@@ -3319,6 +3319,10 @@ mod tests {
     #[test]
     #[cfg(feature = "file_io")]
     fn test_c2pa_sign_file_null_source_path() {
+        let settings = CString::new(r#"{"verify": {"verify_after_sign": false}}"#).unwrap();
+        let format = CString::new("json").unwrap();
+        let _ = unsafe { c2pa_load_settings(settings.as_ptr(), format.as_ptr()) };
```
tmathern: If updating the test, why not move to the new APIs? That is the thread-local one, isn't it?

sdk/src/asset_handlers/c2pa_io.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
     error::{Error, Result},
 };
 
-static SUPPORTED_TYPES: [&str; 3] = [
+pub static SUPPORTED_TYPES: [&str; 3] = [
```
tmathern: Should this move somewhere else to be more accessible?

sdk/src/error.rs
```diff
@@ -391,6 +392,9 @@ pub enum Error {
     #[error("C2PA Validation Error: {0}")]
     C2PAValidation(String),
 
+    #[error("manifest failed validation after signing")]
```
tmathern: This will be a breaking change.
Also is it really different to `C2PAValidation`?
Do we really need the enum, or should we all a validation error "source"?
(I can live with this, just asking).

sdk/src/asset_handlers/c2pa_io.rs
ok-nick: This is conventionally how they are stored for all of the asset handlers.

c2pa_c_ffi/src/c_api.rs
ok-nick: It is, the test case was built using the thread-local variant of the API. I will consider rewriting the test to use the new APIs.

sdk/src/error.rs
ok-nick: `Error::C2PAValidation` contains a string with only one validation error code as documented. The new variant contains a struct with all validation codes. I am working on a `ValidationErrors`-type of struct that can be obtained from `ValidationResults` so that the error outputs detailed/formatted info about all errors.

Previously when we hit the first error we would return it, which is incorrect since those errors may've been attested to in the validation results of an ingredient. The new implementation properly validates everything and then returns the error deltas.

ok-nick:
I kept `verify_after_sign` disabled for now (as it was already). Let's get the fix out before we change the default.

gpeacock (changes requested):
This isn't a fix, this is an entirely new way of reporting errors.
I can see how we may want to validate the manifest when we have no asset, but that should report an invalid hash if the manifest is associated with an asset  ( and it should always be, except for archives.

sdk/src/store.rs
```diff
@@ -1999,43 +1996,53 @@ impl Store {
...
+    }
+
+    /// Verifies the store and returns [`Error::InvalidManifest`] if the validation state is
+    /// invalid. Pass `asset_data` to also verify hash bindings against the asset.
+    #[async_generic(async_signature(
+        &mut self,
+        asset_data: Option<&mut ClaimAssetData<'_>>,
+        context: &Context,
+    ))]
+    pub(crate) fn validate_manifest(
```
mauricefisher64: How is this used?  Should callers never call verify_store directly? Could this have just been done in verify_store?

make_test_images/src/make_test_images.rs
```diff
@@ -217,33 +224,26 @@ impl MakeTestImages {
...
-        builder.add_ingredient(parent);
+        let ingredient = builder.add_ingredient_from_stream(
+            json!({
+                "title": name,
+                "relationship": relationship
+            })
+            .to_string(),
+            format,
+            &mut source,
+        )?;
```
gpeacock, line 243: this drops the thumbnail generation logic, but I don't recall why we need that vs letting it be generated.
But you can still do this
let mut ingredient = builder.add_ingredient_from_stream()...
source.rewind()?;
let (format, thumbnail) =
        make_thumbnail_from_stream(format, &mut source).context("making thumbnail")?;
 ingredient.parent.set_thumbnail(format, thumbnail)?;

sdk/benches/sign.rs
```diff
@@ -19,6 +21,18 @@ fn create_signer() -> CallbackSigner {
     CallbackSigner::new(ed_signer, SigningAlg::Ed25519, CERTS)
 }
 
+fn context() -> Arc<Context> {
```
gpeacock, line 24: you can also Context::new().into_shared() shared instead of Arc::new()

sdk/src/settings/mod.rs
```diff
@@ -348,6 +348,12 @@ pub struct Verify {
     ///
     /// [`Builder`]: crate::Builder
     pub verify_after_sign: bool,
+    /// Whether to include asset hash validation when verifying after signing.
+    ///
+    /// The default value is false.
+    ///
+    /// Has no effect when [`Verify::verify_after_sign`] is false.
+    pub verify_after_sign_hash: bool,
```
gpeacock: I'm not sure we should ever hash after signing. The only real reason we need the verify is validate things that we don't check for when we create the manifest.

sdk/src/store.rs
```diff
@@ -2931,27 +2958,19 @@ impl Store {
...
 
-                let verify_after_sign = settings.verify.verify_after_sign;
-                // Also catch the case where we may have written to io::empty() or similar
-                if verify_after_sign && output_stream.seek(SeekFrom::End(0))? > 0 {
-                    // verify the store
-                    let mut validation_log =
-                        StatusTracker::with_error_behavior(ErrorBehavior::StopOnFirstError);
+                if context.settings().verify.verify_after_sign {
+                    let output_len = output_stream.seek(SeekFrom::End(0))?;
+                    let validate_hash = context.settings().verify.verify_after_sign_hash;
```
gpeacock, line 2963: Whe would we actually need this? If we just made the hash, it should be correct.

sdk/src/validation_results.rs
```diff
@@ -359,6 +364,104 @@ impl ValidationResults {
...
+    /// Returns a summary of why validation failed.
+    ///
+    /// The `Display` impl outputs a human-readable view of the failures, suitable for error messages.
+    /// If there are no failures, an empty string is output.
+    pub fn failure_summary(&self) -> ValidationFailureSummary<'_> {
+        ValidationFailureSummary(self)
+    }
+}
+
+impl Display for ValidationResults {
```
gpeacock, line 392: So all of this is to provide a more condensed form of validationResults than we get with serde serialize?
the goal is to return an error string that is easier to read than validation results?

make_test_images/src/make_test_images.rs
```diff
@@ -151,15 +152,21 @@ fn file_name(path: &Path) -> Option<&str> {
...
+    context: Arc<c2pa::Context>,
 }
 
 impl MakeTestImages {
-    pub fn new(config: Config) -> Self {
+    pub fn new(config: Config) -> Result<Self> {
         let output = config.output_path.to_owned();
-        Self {
+        let context = c2pa::Context::new()
+            .with_settings(json!({"verify": {"verify_after_sign": false}}))
```
gpeacock: not sure if we should disable this here, but I can look at that later.

make_test_images/src/make_test_images.rs
```diff
@@ -151,15 +152,21 @@ fn file_name(path: &Path) -> Option<&str> {
...
+            .unwrap()
```
gpeacock: This should be returning an error instead of unwrap() since you also changed new() to return a result.

sdk/src/asset_handlers/c2pa_io.rs
gpeacock: pub (crate) ?

sdk/src/error.rs
```diff
@@ -391,6 +392,9 @@ pub enum Error {
     #[error("C2PA Validation Error: {0}")]
     C2PAValidation(String),
 
+    #[error("manifest failed validation with: {}", .0.failure_summary())]
+    InvalidManifest(ValidationResults),
```
gpeacock, line 396: need to map this to something in c_ffi

sdk/src/validation_results.rs
gpeacock: I'd make the string result be 
code: description (url)

gpeacock (changes requested):

sdk/src/settings/mod.rs
ok-nick: It can be used to ensure our validation logic works against our generation logic. Should we private it or do you suggest removing it entirely?

sdk/src/store.rs
ok-nick: https://github.com/contentauth/c2pa-rs/pull/2147#discussion_r3318730373

sdk/src/validation_results.rs
ok-nick: Exactly, there is lots of information we can include but I simplified it to make it easy to read. Let me see if I can incorporate some of your suggestions.

make_test_images/src/make_test_images.rs
ok-nick: It was already being done in the tests here: https://github.com/contentauth/c2pa-rs/blob/ef4b6186a3fe955e2b8d05c60c41eae2c86e5267/make_test_images/src/make_test_images.rs#L686-L707

Maybe we can remove it.

sdk/src/asset_handlers/c2pa_io.rs
ok-nick: It's not exposed in the public API, but no harm in making it `pub(crate)`.

make_test_images/src/make_test_images.rs
ok-nick: Good catch, I glossed over this when I changed the return signature.

make_test_images/src/make_test_images.rs
ok-nick: I figured c2pa-rs is already generating the thumbnail for us. Is there a reason we need to do it here again?

sdk/src/store.rs
ok-nick: The function name isn't ideal, I can try to come up with something better. The reason this logic isn't in `Store::verify_store` is because we don't want to hard error on an invalid manifest in the reader.

sdk/src/error.rs
ok-nick: Mapped it to the `Verify` c2pa-c-ffi error variant.

sdk/src/store.rs
ok-nick: Renamed it to `Store::verify_store_strict`.

sdk/src/validation_results.rs
ok-nick: Here's an example of the updated output:
```
state: Invalid
    success: claimSignature.validated, claimSignature.insideValidity
    failure:
      assertion.dataHash.mismatch: hash mismatch on data hash assertion (self#jumbf=...)
    ingredient [urn:uuid:1234]:
      failure:
        claimSignature.mismatch: ingredient signature did not validate
```

sdk/src/settings/mod.rs
gpeacock: I'll leave that up to Maurice. I think private would be fine if he wants to keep it.

make_test_images/src/make_test_images.rs
gpeacock: I updated the known good results. If running make images results in no changes, I think we are good.

gpeacock (approved):

sdk/src/settings/mod.rs
ok-nick: Ok, I made it private for now. @mauricefisher64 thoughts?

mauricefisher64: The user should never have control over whether or where we hash.  Definitely should be private.

PR #2148, opened 2026-05-13
docs: Add deprecation policy

scouten-adobe:


docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
+# Deprecation policy
+
+## Preamble
+
+The Content Authenticity Initiative SDKs are evolving projects. Prior to their 1.0 release, the APIs may change as we refine our design. That said, we are committed to making those changes in a transparent, predictable way so that developers can plan accordingly.
+
+This policy applies to the Rust SDK and all language bindings (JavaScript, Node.js, C, C++, Swift, Kotlin, and Python).
```
crandmck:
```suggestion
This policy applies to the Rust library and all language bindings (JavaScript, Node.js, C, C++, Swift, Kotlin, and Python).
```
Prefer to refer to each repo as a "library" vs. the entire set as the "SDK".  This is the terminology used in the docs.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Versioning and stability guarantees
+
+We follow [Semantic Versioning (SemVer)](https://semver.org/). Version 1.0.0 will define the public API, and after that release, the way version numbers are incremented will be dependent on how the public API changes: patch for backward-compatible bug fixes, minor for backward-compatible additions, and major for incompatible changes.
```
crandmck:
```suggestion
We follow [Semantic Versioning (SemVer)](https://semver.org/). Version 1.0.0 defines the public API, and subsequent version numbers are based on how the public API changes in the release: 
- **Patch** release for backward-compatible bug fixes.
- **Minor** release for backward-compatible additions.
- **Major** release for incompatible changes.
```
Small wording change so that the doc will not be time-specific: "1.0.0 will define..." implies the release is in future--which it is now, but won't be later.  And formatting change to make it easier to read.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
+# Deprecation policy
+
+## Preamble
```
crandmck:
```suggestion
```

No section heading is needed for this introductory section. "Preamble" is a bit grandiose for this document IMHO.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Versioning and stability guarantees
+
+We follow [Semantic Versioning (SemVer)](https://semver.org/). Version 1.0.0 will define the public API, and after that release, the way version numbers are incremented will be dependent on how the public API changes: patch for backward-compatible bug fixes, minor for backward-compatible additions, and major for incompatible changes.
+
+**Before 1.0:** Major version zero (`0.y.z`) is for initial development — anything may change at any time, and the public API should not be considered stable. In the Rust/Cargo ecosystem, this means that a change from `0.2.3` to `0.3.0` may include incompatible API changes. We will, however, make a good-faith effort to follow the deprecation process below even before 1.0, so that users have advance warning before breakage occurs.
+
+**After 1.0:** Breaking changes will only ship in major version bumps. Before completely removing functionality in a new major release, there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API. We will publish and retain historical documentation for at least each minor point release.
```
crandmck:
```suggestion
**After 1.0:** Breaking changes will only ship in major version increments. Before completely removing functionality in a new major release, there will be at least one minor release that contains the deprecation so that API consumers can smoothly transition to the new API. We will publish and retain historical documentation for at least each minor point release.
```

Small wording changes: 
- "should" -> "will" (assumig that this will always be true.  If not, then we should specify exceptions)
- "users" -> "API consumers" to clarify that we are not talking about end-users.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+
+## What counts as a breaking change
+
+Not all changes are equal. Breaking changes are assumed to be major changes, but not all breaking changes are major. The goal is that the same code should be able to run against different minor revisions, and minor changes should require at most a few local annotations. (This document is Rust-specific; we will treat other languages as closely to this list as is feasible.)
```
crandmck: This is confusing.  Do you mean "Breaking changes are assumed to be major changes, but not all major changes are breaking changes." ?  

If so, maybe reword as "All breaking changes are major changes, but not all major changes are breaking."

If not, then what exactly does this mean?  Most breaking changes are major, but in some cases they are not? When are breaking changes NOT major?? 

Please clarify.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+
+Changes we consider **breaking** (requiring a major version bump post-1.0):
```
crandmck:
```suggestion
Breaking changes that require a major version change  (post-1.0) include:
```

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+- Moving a public type, function, method, trait, or constant from one parent module to another
+- Removing or renaming a public type, function, method, trait, or constant
+- Changing the signature of a public function (parameter types, return types, or generics) except to the extent such changes are generally considered non-breaking (e.g. changing a `&mut Type` to `&Type`)
+- Changing the behavior of a public API in a way that violates previously documented contracts
+- Removing or renaming public enum variants or struct fields
+- Adding public enum variants or struct fields (unless `#[non_exhaustive]` was applied)
+- Breaking changes to upstream or third-party libraries to the extent that those APIs are re-published by our library and thus break our own API compatibility
+- Any other change flagged by `cargo-server-checks` (or an equivalent tool for any other language) as breaking compatibility
+
+Changes we consider **non-breaking** (minor or patch):
```
crandmck:
```suggestion
Non-breaking changes that require a minor or patch version change:
```

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+### Stage 1: Deprecation notice (minor release)
+
+- The item is marked deprecated in source code using the appropriate language mechanism. (See "Language-specific deprecation annotations" below.)
+- The deprecation message includes: what is deprecated, why, what to use instead, and the earliest planned removal timeline. (See stage 2.)
+- The change is documented in the CHANGELOG under a `### Deprecated` heading.
+- An announcement is posted in the project's Discord and, where applicable, linked from the relevant GitHub issue or PR.
+
+### Stage 2: Grace period
+
+The grace period is a timeframe during which the deprecated API remains operational before being retired. Our minimums are:
```
crandmck:
```suggestion
During the grace period, the deprecated API remains operational before being retired. The  _minimum_ grace periods are shown here.
```

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+- Deprecating a public item without removing it
+- Bug fixes that restore documented behavior
+
+## The deprecation process
+
+When we decide to remove or replace part of the public API, we follow a three-stage process:
+
+### Stage 1: Deprecation notice (minor release)
+
+- The item is marked deprecated in source code using the appropriate language mechanism. (See "Language-specific deprecation annotations" below.)
```
crandmck: This should be a numbered list, because the steps are performed in the specified order.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+
+- Adding new public items (types, functions, trait impls)
+- Deprecating a public item without removing it
+- Bug fixes that restore documented behavior
+
+## The deprecation process
+
+When we decide to remove or replace part of the public API, we follow a three-stage process:
+
+### Stage 1: Deprecation notice (minor release)
```
crandmck, line 53:
```suggestion
### Stage 1: Deprecation notice (minor release)

The initial stage provides advance notice of the deprecation:
```

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+The grace period is a timeframe during which the deprecated API remains operational before being retired. Our minimums are:
+
+| SDK maturity | Minimum grace period |
+| -- | -- |
+| Pre-1.0 | 60 days |
+| Post-1.0 | 90 days |
+
+During the grace period, the deprecated API will continue to work without functional regression. **Exception:** We may remove deprecated APIs before this window expires if needed to address serious security issues or vulnerabilities.
+
+### Stage 3: Removal
```
crandmck, line 77:
```suggestion
### Stage 3: Removal

In the final stage, the item is actually removed from the API: 
```

crandmck (changes requested):
Mostly minor suggestions. 

Please review punctuation of all list items to ensure they follow our style guide: Don’t end each bulleted item with a period unless it is a complete sentence, even if it completes a sentence begun in the introduction. (Claude can help with this.)

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+- The migration guide (see "Migration guides", below) is updated to reflect the removal as permanent.
+
+## Language-specific deprecation annotations
+
+Deprecation warnings will be expressed using each language's idiomatic mechanism so that developers are alerted by their compiler or toolchain.
+
+### Rust
+
+```rust
+#[deprecated(since = "0.5.0", note = "Use `Builder::new_v2()` instead. Will be removed in 0.7.0.")]
```
gpeacock: There's no way we are going to know what version will remove the code. Versions are not fixed to a timeline. It's hard enough to estimate the version we deprecate given the way our builds go.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+- Bug fixes that restore documented behavior
+
+## The deprecation process
+
+When we decide to remove or replace part of the public API, we follow a three-stage process:
+
+### Stage 1: Deprecation notice (minor release)
+
+- The item is marked deprecated in source code using the appropriate language mechanism. (See "Language-specific deprecation annotations" below.)
+- The deprecation message includes: what is deprecated, why, what to use instead, and the earliest planned removal timeline. (See stage 2.)
```
gpeacock: I don't think it is practical or necessary to add the earliest removal timeline. that's very subjective. Stage 2 already lets people know our general rules. I think this would just be misleading most of the time.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,139 @@
...
+
+Use `@available(*, deprecated, renamed: "newFunction", message: "Use newFunction() instead.")`.
+
+### Kotlin
+
+Use `@Deprecated(message = "...", replaceWith = ReplaceWith("newFunction()"))`.
+
+## Migration guides
+
+Every deprecation will be accompanied by a migration guide. We will provide alternatives or newer versions for deprecated features — if an item is scheduled for removal, users should know the recommended replacement.
```
gpeacock: I agree with the principal of letting people know that something is being removed and why. We should also let them know about alternatives. But the idea that every deprecation requires a before/after code comparison seems too much. Some should have this treatment, but not all. We have a LOT to deprecate. For some things no one ever knew about them and there were no previous code examples.

gpeacock (changes requested):
I left some comments in code about issues I found.

docs/deprecation-policy.md
scouten-adobe: I'll partially accept this. IMHO we should retain the future-tense wording and change the language when we actually reach 1.0.

docs/deprecation-policy.md
scouten-adobe: I disagree. It's pretty easy to do the date math on the day we declare something deprecated. Having that date in the deprecation notice (and also as a substitute for your suggestion at line 74 above) will remind _us_ to actually remove it when we can.

docs/deprecation-policy.md
crandmck: > change the language when we actually reach 1.0.

Just don't forget that part!

docs/deprecation-policy.md
crandmck: I softened the wording a bit to say that the deprecation notice includes the planned removal timeline IF POSSIBLE, while noting that it must always respect the grace period.

docs/deprecation-policy.md
crandmck: Added "ideally" to indicate that the before/after code comparison is good to have, but not strictly required.

crandmck (approved):

docs/deprecation-policy.md
crandmck: Changed this example to say `Will be removed in a future major release.`

crandmck:
@gpeacock I tried to address your concerns while maintaining the original intent as much as possible.

I also noticed that it didn't clearly say WHEN to provide the migration guide (just to update it to be permanent in stage 3), so I clarified that "Stage 1: Deprecation notice" should include the migration doc.

gpeacock (approved):
This looks good to me. 
One thing is that there may be some undocumented functionality that changes due to bug fixes or implementation issues. We do not promise support for undocumented features. (in some cases we have chosen not to document, in others we may not even know how the apis are being used, so I think we need this safeguard. I don't want to be maintaining undocumented bugs for compatibility like Flash did.

docs/deprecation-policy.md
```diff
@@ -0,0 +1,161 @@
...
+Not all changes are equal. 
+
+A breaking change always results in a major version increment, but a major version increment does not alwasy require a breaking change. A major version can also be used to release a rewritten API or significant new features: It primarily serves as a signal that the update requires careful review for compatibility. 
+
+The goal is that the same code should be able to run against different minor revisions, and minor changes require at most a few local annotations. (This document is Rust-specific; we will treat other languages as closely to this list as is feasible.)
+
+Changes considered **breaking** (requiring a major version increment post-1.0):
+
+- Moving a public type, function, method, trait, or constant from one parent module to another
+    - **EXCEPTION:** APIs that are not publicly _documents_ may be removed prior to 1.0 without following this policy.
```
crandmck:
```suggestion
    - **EXCEPTION:** APIs that are not _publicly  documented_ may be removed prior to 1.0 without following this policy.
```

crandmck (changes requested):

PR #2151, opened 2026-05-15
fix: Share tokio runtime in C FFI and pool reqwest HTTP clients

tmathern:
- Share a (Tokio) runtime in the C FFI. This should prevent thread (pool) recreation and the related memory consumption. Also expose an API to set a runtime in the C FFI (from a caller), so a host application could set a runtime to be shared.
- Pool reqwest clients (avoid client recreation and related memory consumption).

## Checklist


sdk/src/assertions/uuid_assertion.rs
```diff
@@ -27,6 +27,7 @@ pub struct Uuid {
 
 impl Uuid {
     /// Create new Identity instance
+    #[allow(dead_code)]
```
tmathern, line 30: Clippy started having opinions, not totally sure where this is used?

gpeacock (changes requested):
Better yet, use this PR instead and don't have a tokio runtime in the c_ffi a all!
https://github.com/contentauth/c2pa-rs/pull/2120

tmathern:
> Better yet, use this PR instead and don't have a tokio runtime in the c_ffi a all! #2120

Great! Let's try to get that one in soon! In the meantime, I think we still also need https://github.com/contentauth/c2pa-rs/pull/2152

PR #2152, opened 2026-05-15
fix: Share reqwest HTTP clients

tmathern:
- Share reqwest clients (avoid client recreation and related memory consumption).

## Checklist


sdk/src/assertions/uuid_assertion.rs
```diff
@@ -27,6 +27,7 @@ pub struct Uuid {
 
 impl Uuid {
     /// Create new Identity instance
+    #[allow(dead_code)]
```
tmathern, line 30: Clippy opinions. Why now? Small mystery.

sdk/src/http/reqwest.rs
```diff
@@ -17,26 +17,40 @@
...
-            .unwrap()
+        SYNC_CLIENT
+            .get_or_init(|| {
+                // By default `reqwest::blocking::Client::new()` unwraps if the TLS backend cannot be initialized.
+                // The behavior here is equivalent, except with a custom configuration.
+                #[allow(clippy::unwrap_used)]
+                reqwest::blocking::Client::builder()
+                    .redirect(reqwest::redirect::Policy::none())
+                    .build()
+                    .unwrap()
```
gpeacock: The unwrap feels like a bad thing here, though it is no worse than what we had before.

tmathern: `expect` at least can give us a message.
Removing the unwrap/expect might be a nice thing, but it will also be quite invasive since all caclelrs will need to deal with a signature change

sdk/src/http/reqwest.rs
```diff
@@ -17,26 +17,40 @@
...
+    };
 
     use http::{Request, Response};
 
     use crate::http::{HttpResolverError, SyncHttpResolver};
 
     pub type Impl = reqwest::blocking::Client;
 
+    static SYNC_CLIENT: OnceLock<reqwest::blocking::Client> = OnceLock::new();
+    static SYNC_CLIENT_REDIRECTS: OnceLock<reqwest::blocking::Client> = OnceLock::new();
```
ok-nick, line 32: IMO I prefer to not by default store global state like this. I'd rather the caller of the library chooses to store the HTTP client as global state if they'd like. I'm not sure if it has meaningful impact on memory/perf. I defer to @gpeacock on the final decision.

tmathern: Normally, the client itself is stateless (it is not state that changes behavior). You are reusing HTTP plumbing, so you do not recreate it, with all the underlying plumbing and needed memory, on the flight, on each call you need it. 

A new reqwest client goes through [reqwest::new](https://github.com/seanmonstar/reqwest/blob/master/src/blocking/client.rs#L1376), which spawns a [tokio runtime current thread on this thread](https://github.com/seanmonstar/reqwest/blob/master/src/blocking/client.rs#L1384), with its conenction pool.

Now imagine this in a server under load. N concurrent requests, each on its own worker thread, each going through this (and a few fetches), rebuilding on each call. Without sharing, every worker thread spawns its own runtime thread, and undergoes init again. (I don't think it impacts perf yet, but memory: threads do not come for free).

The caller could, choose, yes, but we also should have our default part reuse this for better defaults. The two statics only govern the default built-in resolver, where the library owns the lifecycle, not the caller. And the OnceLock is lazy, so if the network path is never touched, nothing is allocated.

gpeacock (approved):

sdk/src/http/reqwest.rs
gpeacock: I think this is ok, but I'm nervous about it after the problems we had with settings that were held in a thread local

tmathern:
(Also fixes an unused dependency warning flagged by the CI/CD jobs)

PR #2154, opened 2026-05-18
perf: avoid buffering stream when computing placeholder PNG C2PA chunk

ok-nick:
Reduces memory usage when reading/signing a PNG that doesn't contain a C2PA manifest by optimizing `get_object_locations_from_stream` to quickly find where the C2PA chunk goes rather than computing it via `write_cai`. When signing a 100MB PNG, the total memory usage decreases by 29.4%.

Before:
Screenshot 2026-05-18 at 5 25 14 PM
After:
Screenshot 2026-05-18 at 5 25 30 PM

### DIFF
--- a/sdk/src/asset_handlers/png_io.rs
+++ b/sdk/src/asset_handlers/png_io.rs
@@ -355,13 +355,24 @@ impl CAIWriter for PngIO {
             let file_end = input_stream.seek(SeekFrom::End(0))? as usize;
             (ps, file_end)
         } else {
-            let mut output_stream = Cursor::new(Vec::new());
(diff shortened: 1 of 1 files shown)


gpeacock (commented):
I'm wondering if there are side effects caused by modifying the input stream. We probably don't expect that. It is mut so that we can seek, not modify.

ok-nick:
@gpeacock I'm not sure what you mean. The input stream is read-only and this change only seeks?

mauricefisher64 (approved):

ok-nick:
Discussed offline with Gavin.

PR #2156, opened 2026-05-20
fix: make GIF box map C2PA placeholder len 0 rather than 1

### LINKED ISSUE #1668: GIF boxes hash placeholder is of length 1 when it should be 0
GIF returns a placeholder box for boxes hash with length 1 when it should be 0. We should also document this behavior (as well as placeholders in general).

ok-nick:
It can affect the box hashing code when generating hashes with `minimal=true`. This change also allows the code to be greatly simplified, I switched from declarative iterator to imperative-style for clarity.

This line here is removed:
```rust
/// ...
if should_insert_placeholder {
    offset += 1;
    /// ...
```

- Closes #1668


tmathern:
(Merging in latest main fixes an unused dependency issue flagged by the CI/CD jobs)

tmathern:
@ok-nick `It can affect the box hashing code when generating hashes with minimal=true.`. Can you elaborate what the symptom/observed bug is in this case?

gpeacock (approved):

tmathern (approved):

ok-nick:
@tmathern When doing a minimal form box hash (everything before c2pa, the c2pa, and after c2pa), the old GIF code gave the empty placeholder C2PA box a length of 1. That would offset all of the subsequent boxes by 1, which is incorrect and would hash the wrong data.

See here: https://github.com/contentauth/c2pa-rs/blob/2c0695e47f0886a89efc755064d87b718432ecef/sdk/src/assertions/box_hash.rs?plain=1#L260-L358

PR #2157, opened 2026-05-20
test(c2pa_c_ffi): move allow(deprecated) next to deprecated functions

ok-nick:
Narrows the `#[allow(deprecated)]` next to the use of the deprecated function rather than on the module-level, so that we aren't writing new tests with deprecated functions.

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -2933,7 +2933,6 @@ unsafe fn c2pa_mime_types_to_c_array(strs: Vec<String>, count: *mut usize) -> *c
 }
 
 #[cfg(test)]
-#[allow(deprecated)]
 mod tests {
     use std::{ffi::CString, io::Seek, panic::catch_unwind};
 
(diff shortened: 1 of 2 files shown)


c2pa_c_ffi/src/c_api.rs
```diff
@@ -3312,10 +3333,14 @@ mod tests {
     #[test]
     fn test_c2pa_builder_no_embed() {
         let manifest_def = CString::new("{}").unwrap();
+        #[allow(deprecated)]
```
tmathern: What about flagging the tests which use the deprecated APIs with #deprecated instead of in the tests? Bonus is we easily see which tests would verify deprecated APIs, and once we clean up deprecated APIs, they are easier to remove (and to justify removing).

gpeacock (approved):
this change is ok,but I agree with Tania it would be cleaner just to allow deprecated on the tests that test deprecated function than on every call.

c2pa_c_ffi/src/c_api.rs
ok-nick: Sure, let me change that.

PR #2161, opened 2026-05-21
perf: skip intermediate stream copy for common signing path

gpeacock:
When no XMP remote-ref needs embedding and no manifest stripping is required, input_stream is now passed directly to the hash and write passes instead of being copied into a temporary intermediate_stream first. For large video assets this saves one full-file read+write pass per signing call.

The fast_path flag captures the condition; source_is_intermediate threads the choice through all downstream call sites. Operations that genuinely need a writable buffer (box-hash compression, Merkle UUID insertion) lazily populate intermediate_stream on demand and flip the flag. Also adds output_stream.flush() before the non-BMFF hash-readback rewind to drain buffering/FFI-backed streams before seeking.

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -2988,6 +2988,11 @@ impl Store {
 
         let io_handler = get_assetio_handler(format);
 
+        let is_bmff = is_bmff_format(format);
+        // fast_path applies to all formats: when there is no XMP embed and no manifest removal,
+        // we can pass input_stream directly to the write/hash steps, skipping one full-file copy.
+        let fast_path = url.is_none() && !remove_manifests;
+
         context.check_progress(ProgressPhase::Writing, 1, 2)?;
 
         if let Some(io_handler) = &io_handler {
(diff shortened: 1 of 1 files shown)


mauricefisher64 (approved):

sdk/src/store.rs
```diff
@@ -3225,6 +3283,10 @@ impl Store {
             if needs_hashing {
                 let pc = self.provenance_claim_mut().ok_or(Error::ClaimEncoding)?;
 
+                // Flush before seeking: FFI-backed streams (and any stream that
+                // buffers writes) must be flushed before seeking back to read.
+                output_stream.flush()?;
```
ok-nick, line 3288: Hm, how come we only flush here? Can FFI handle this?

ok-nick (approved):

sdk/src/store.rs
gpeacock: ffi has supports flush on streams.

PR #2164, opened 2026-05-23
fix: Load ingredient manifest store esp. to load resources on search/load miss

tmathern:
Since https://github.com/contentauth/c2pa-rs/pull/2103, ingredients manifest stores are lazy loaded. But if a caller is trying to get to resources of a lazy loaded ingredient, there may be misses as the resources may not have yet been loaded.

Especially, this breaks the legacy approach to extract ingredients from a Builder to reattach them to a new Builder.

This was detected by 3 c2pa-cpp failing when running with a build from c2pa-rs main (since those changes are currently unreleased), that exercise the legacy way of doing things (all fail with a resource not found error, due to resources not being loaded at time of checking them/wanting to use them):
[  FAILED  ] BuilderTest.ExtractIngredientsFromArchive
[  FAILED  ] BuilderTest.ExtractIngredientsFromArchiveToBuilder
[  FAILED  ] BuilderTest.ExtractIngredientsFromArchives

I *believe* we don't want to break that yet, to give time for users who may rely on the legacy approach to switch to the new API. This PR offers a fallback approach on this (now edge?) case when the resource is not found, to still try to load it. 
(I am not sure if this side-effect was intended from the perf PR, but here we are. We either need a fix of the perf PR, or the fallback that eager loads on miss).

## Checklist

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -1748,10 +1748,8 @@ pub mod tests {
     /// that is lazily deferred in `source_store`.
     #[test]
     fn test_resource_to_stream_retrieve_deferred_manifest_data() -> Result<()> {
-        let reader = Reader::default().with_stream(
-            "image/jpeg",
-            Cursor::new(IMAGE_WITH_INGREDIENT_MANIFEST),
-        )?;
+        let reader = Reader::default()
+            .with_stream("image/jpeg", Cursor::new(IMAGE_WITH_INGREDIENT_MANIFEST))?;
 
         let active = reader.active_manifest().unwrap();
         let ingredient = active
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -882,7 +882,7 @@ impl Reader {
     pub fn resource_to_stream(
         &self,
         uri: &str,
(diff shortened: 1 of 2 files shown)


sdk/src/reader.rs
```diff
@@ -917,7 +917,30 @@ impl Reader {
                 Ok(resource) => resource.write_stream(&relative_uri, stream),
                 Err(_) => match find_resource(&absolute_uri) {
                     Ok(resource) => resource.write_stream(&absolute_uri, stream),
-                    Err(e) => Err(e),
+                    Err(_) => {
+                        // Fallback: ingredient manifest_data may be lazily loaded.
```
gpeacock, line 921: write_stream() should handle this. I'll take a look. lazy loading should be entirely transparent.

gpeacock: I have another idea that would defer all resources, but, until then. this is about as good as anything else I can come up with.

gpeacock (approved):
This actually works fine for now. I have another bigger idea about how to fix this in the long term.

sdk/src/reader.rs
tmathern: ack'ed!

PR #2168, opened 2026-05-26
fix: CAI-10364 (Server-Side Request Forgery (SSRF) via CAWG did:web Resolution during Post-Validation )

mpgopala:
1. Gate localhost→HTTP to #[cfg(test)]
2. Block IP literals in to_url()
3. Pass a resolver parameter get_did_doc so that above two fixes are applied for did:web processing

## Checklist


sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
```diff
@@ -154,7 +157,8 @@ impl SignatureVerifier for IcaSignatureVerifier {
         // TO DO (CAI-7970): Add support for VC version 1.
         let mut ica_credential = self.parse_ica_vc_v2(payload_bytes, status_tracker)?;
 
-        self.check_issuer_signature_async(&sign1, &ica_credential)
+        let _resolver = Context::new().resolver_async();
+        self.check_issuer_signature_async(&sign1, &ica_credential, _resolver.as_ref())
```
ok-nick: This is where the problem becomes tricky. The resolver here isn't actually the one passed in by the user with restrictions and such. In the reader we call a higher-level function that eventually boils down to here.

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
```diff
@@ -54,6 +56,21 @@ use crate::{
...
 pub struct IcaSignatureVerifier {
     // TO DO (CAI-7980): Add option to configure trusted ICA issuers.
+    resolver: Arc<dyn AsyncHttpResolver>,
+}
+
+// WASM is single-threaded; Arc<dyn AsyncHttpResolver> satisfies MaybeSync but
+// not the unconditional Sync required by SignatureVerifier. Safe because WASM
+// has no threads.
+#[cfg(target_arch = "wasm32")]
+unsafe impl Sync for IcaSignatureVerifier {}
```
ok-nick: If you change `trait SignatureVerifier: Sync` to `trait SignatureVerifier: MaybeSync`, we can get rid of this.

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -112,22 +158,29 @@ async fn get_did_doc(url: &str, resolver: &dyn AsyncHttpResolver) -> Result<Vec<
...
+        .unwrap_or_default()
+        .http_resolve(request)
+        .map_err(|e| match e {
+            HttpResolverError::ResponseTooLarge => DidWebError::ResponseTooLarge,
+            e => DidWebError::Request(url.to_owned(), e),
+        })?;
+    let (parts, body) = response.into_parts();
+    check_response_status(parts.status, url)?;
+    read_body_with_limit(body, url)
 }
```
ok-nick: The HTTP resolver should be passed down in the sync case too. Can we also use `#[async_generic(...)]` to avoid the duplication between sync/async functions.

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
mpgopala: Fixed.

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
mpgopala: Done

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
mpgopala: Done

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -156,46 +175,43 @@ async fn get_did_doc(url: &str) -> Result<Vec<u8>, DidWebError> {
 pub(crate) fn resolve(did: &Did<'_>) -> Result<DidDocument, DidWebError> {
     let url = prepare_url(did)?;
     // TODO: https://w3c-ccg.github.io/did-method-web/#in-transit-security
-    let bytes = get_did_doc_sync(&url)?;
+    let bytes = get_did_doc(&url, &SyncGenericResolver::new())?;
```
ok-nick: If we are constructing a new resolver anywhere in the CAWG path, it will not respect the settings set by the user when creating a reader.

The problem gets complicated when you start to deal with the traits and structs storing them. Solving it by passing it through the stack will result in a lot of duplication, though solving it without doing that will require quite a bit of refactoring.

mpgopala:
@ok-nick @scouten-adobe : Please review. I have passed the Context through the stack as discussed.

sdk/src/signer.rs
```diff
@@ -91,13 +91,10 @@ pub trait Signer {
         if let Some(url) = self.time_authority_url() {
             if let Ok(body) = self.timestamp_request_body(message) {
                 let headers: Option<Vec<(String, String)>> = self.timestamp_request_headers();
+                let context = Context::new();
```
scouten-adobe, line 96: Please flag this with a TODO note that creating a bare `Context` is not recommended practice in production code and we should look for a way to avoid this. Not a blocker for this PR IMHO.

sdk/src/signer.rs
```diff
@@ -219,16 +216,11 @@ pub trait AsyncSigner: MaybeSend + MaybeSync {
     async fn send_timestamp_request(&self, message: &[u8]) -> Option<Result<Vec<u8>>> {
         if let Some(url) = self.time_authority_url() {
             if let Ok(body) = self.timestamp_request_body(message) {
-                use crate::http::AsyncGenericResolver;
-
                 let headers: Option<Vec<(String, String)>> = self.timestamp_request_headers();
+                let context = Context::new();
```
scouten-adobe, line 224: Same comment as above.

scouten-adobe (approved):

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
mpgopala: Fixed the issue in below commits.

sdk/src/signer.rs
mpgopala: Done

sdk/src/signer.rs
mpgopala: Done

PR #2172, opened 2026-05-27
fix: Sync resolver for CAWG web DID

tmathern:
There are code paths which are purely sync, and which still need to be able to resolve web did.
Example: Reader from files (especially if called through the C FFI in C++).
This PR implements a sync resolver, with the obvious tradeoff that... the resolver is sync. I tried to share as much code as possible between sync and async implementations, hence some small refactors (new functions that get reused).

## Checklist
- [ ] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment. _(Existing TODOs were left in place, not sure they were linked to issues already...)._

### DIFF
--- a/sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
+++ b/sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
@@ -459,17 +459,9 @@ impl IcaSignatureVerifier {
 
             "web" => {
                 let did_doc = if _sync {
-                    did_web::resolve_sync(&primary_did).map_err(|e| {
-                        ValidationError::SignatureError(IcaValidationError::InvalidDidDocument(
-                            e.to_string(),
-                        ))
-                    })?
+                    did_web::resolve_sync(&primary_did)?
                 } else {
-                    did_web::resolve(&primary_did).await.map_err(|e| {
-                        ValidationError::SignatureError(IcaValidationError::InvalidDidDocument(
-                            e.to_string(),
-                        ))
-                    })?
+                    did_web::resolve(&primary_did).await?
                 };
 
                 let Some(vm1) = did_doc.verification_relationships.assertion_method.first() else {
--- a/sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
+++ b/sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
@@ -459,9 +459,9 @@ impl IcaSignatureVerifier {
 
             "web" => {
                 let did_doc = if _sync {
-                    did_web::resolve_sync(&primary_did)?
+                    did_web::resolve(&primary_did)?
                 } else {
-                    did_web::resolve(&primary_did).await?
+                    did_web::resolve_async(&primary_did).await?
                 };
 
                 let Some(vm1) = did_doc.verification_relationships.assertion_method.first() else {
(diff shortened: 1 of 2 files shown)


sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -114,6 +119,47 @@ async fn get_did_doc(url: &str) -> Result<Vec<u8>, DidWebError> {
     Ok(document)
 }
 
+pub(crate) fn resolve_sync(did: &Did<'_>) -> Result<DidDocument, DidWebError> {
+    let url = prepare_url(did)?;
+    // TODO: https://w3c-ccg.github.io/did-method-web/#in-transit-security
```
tmathern, line 137: Same as L87

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
```diff
@@ -458,56 +458,51 @@ impl IcaSignatureVerifier {
             }
 
             "web" => {
-                if _sync {
+                let did_doc = if _sync {
```
tmathern, line 461: Only the resolution changes. 
The rest came out of the else statement (unconditional now), and lost and indentation level.

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
```diff
@@ -458,56 +458,51 @@ impl IcaSignatureVerifier {
...
+                let Some(jwk_prop) = vm1.properties.get("publicKeyJwk") else {
+                    return Err(ValidationError::SignatureError(
+                        IcaValidationError::InvalidDidDocument(
+                            "DID document's assertionMethod doesn't contain a publicKeyJwk entry"
+                                .to_string(),
+                        ),
+                    ));
+                };
+
+                // OMG SO HACKY!
```
gpeacock, line 492: glad you kept this in :)

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -99,19 +127,30 @@ async fn get_did_doc(url: &str) -> Result<Vec<u8>, DidWebError> {
...
-    match parts.status {
-        http::StatusCode::OK => (),
-        http::StatusCode::NOT_FOUND => return Err(DidWebError::NotFound(url.to_string())),
-        _ => return Err(DidWebError::Server(parts.status.to_string())),
-    };
-
-    let mut document = Vec::new();
-    body.read_to_end(&mut document)
-        .map_err(|e| DidWebError::Response(e.into()))?;
+pub(crate) fn resolve_sync(did: &Did<'_>) -> Result<DidDocument, DidWebError> {
```
gpeacock: General practice in the code is to add async to the async methods and leave the sync off on the sync ones, it would be nice to align that here since it won't make sense in retrospect.

gpeacock (approved):
Nice that you did the refactoring work for common code.
One nit is that we should follow the _async pattern for async methods and not use _sync. I guess there could be an exception if the async method is already public

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
tmathern: It makes me smile :)

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
tmathern: Updating...

PR #2173, opened 2026-05-27
perf: lazy load all ingredient resources & remove old to_archive zip format.

gpeacock:
This is a draft PR to test this optimization


sdk/src/utils/mime.rs
```diff
@@ -58,8 +58,8 @@ pub fn format_to_mime(format: &str) -> String {
     .to_string()
 }
 
-/// Converts a format to a file extension
-#[cfg(feature = "file_io")]
+/// Converts a format to a file extension (not used anymore but maybe we want it later?)
```
tmathern, line 62: Maybe we want it later: In the public API surface?

sdk/src/ingredient.rs
```diff
@@ -1078,24 +1080,21 @@ impl Ingredient {
...
                         .map(|assertion| {
-                            let embedded_data = EmbeddedData::from_assertion(assertion)?;
-                            ingredient.resources.add_uri(
-                                &data_uri.url(),
-                                &embedded_data.content_type,
-                                embedded_data.data,
-                            )
+                            Ok::<ResourceRef, Error>(ResourceRef::new(
+                                assertion.content_type(),
+                                &uri,
```
tmathern: Above, the thumbnail branch stores `&absolute_uri`. Here we have `&uri`. Are the variables really different (uri vs absolute uri), or do they contain the same (aka: it's just a wording/naming choice here for the var name)? Do they need to contain the same? I think so, because ti affects resolution of the actual resource.

sdk/src/ingredient.rs
```diff
@@ -1078,24 +1080,21 @@ impl Ingredient {
...
+                            ))
                         })
                 }
                 uri if uri.contains(jumbf::labels::DATABOXES) => store
                     .get_data_box_from_uri_and_claim(data_uri, claim_label)
                     .map(|data_box| {
-                        ingredient
-                            .resources
-                            .add_uri(&uri, &data_box.format, data_box.data.clone())
+                        Ok::<ResourceRef, Error>(ResourceRef::new(&data_box.format, &uri))
```
tmathern: Same question regarding uri vs absolute URI as in https://github.com/contentauth/c2pa-rs/pull/2173/changes#r3416945248 (since they relate).

sdk/src/builder.rs
```diff
@@ -1188,23 +1128,20 @@ impl Builder {
...
     /// # Errors
     /// * Returns an [`Error`] if the archive cannot be written.
     pub fn to_archive(&self, mut stream: impl Write + Seek) -> Result<()> {
-        if let Some(true) = self.context.settings().builder.generate_c2pa_archive {
-            let c2pa_data = self.working_store_sign(ArchiveKind::Builder)?;
-            stream.write_all(&c2pa_data)?;
-            return Ok(());
+        if self.context.settings().builder.generate_c2pa_archive == Some(false) {
+            return Err(Error::NotImplemented(
+                "ZIP archive format is no longer supported; remove the generate_c2pa_archive=false setting".to_string(),
```
tmathern, line 1139: Thanks for the useful error message. Does it also need a deprecation warning though (not sure where to put it...)?

sdk/src/builder.rs
```diff
@@ -9706,39 +9627,15 @@ mod tests {
         let builder = Builder::from_context(Context::new().with_settings(settings).unwrap());
 
         let mut archive = Cursor::new(Vec::new());
-        builder.to_archive(&mut archive).unwrap();
-        archive.rewind().unwrap();
+        // we now get an error when trying to write the archive because the old format is not supported for writing.
+        builder
+            .to_archive(&mut archive)
+            .expect_err("Expected error for old format");
```
tmathern, line 9629: I wonder - if people have old format archives laying around... Should we offer a script or tool to "upconvert"? Is it even possible without the source asset?

sdk/src/ingredient.rs
```diff
@@ -495,19 +471,61 @@ impl Ingredient {
...
+                    self.manifest_data = Some(ResourceRef::new("application/c2pa", active));
+                }
+            }
         }
-        self.source_store = Some(source_store);
-        self.manifest_data = Some(ResourceRef::new("application/c2pa", active));
-        Ok(())
+        self.resources
+            .set_resolver(std::sync::Arc::new(move |uri: &str| {
+                use crate::jumbf::labels::{manifest_label_from_uri, ASSERTIONS, DATABOXES};
```
tmathern: I have a tendency to put imports at the top of the file... Or the top of the mod, for tests. If possible.

sdk/src/manifest.rs
```diff
@@ -388,6 +388,23 @@ impl Manifest {
         Ok(self)
     }
 
+    /// Installs a store-backed resolver on this manifest's resources so that
```
tmathern:
```suggestion
    /// Adds a store-backed resolver on this manifest's resources so that
```
?

sdk/src/ingredient.rs
```diff
@@ -495,19 +471,61 @@ impl Ingredient {
...
-    /// Lazily resolve [`Ingredient::manifest_data`] from `source_store` using
-    /// [`Ingredient::active_manifest`] as the provenance claim label.
-    pub(crate) fn set_deferred_manifest_data(&mut self, source_store: Arc<Store>) -> Result<()> {
-        let active = self
-            .active_manifest()
-            .map(str::to_owned)
-            .ok_or_else(|| Error::NotFound)?;
-        if source_store.get_claim(&active).is_none() {
-            return Ok(());
+    /// Installs a resolver on this ingredient's [`ResourceStore`] that looks up
```
tmathern: Why "install"? Why not "add"?
```suggestion
    /// Adds a resolver on this ingredient's [`ResourceStore`] that looks up
```

sdk/src/manifest.rs
```diff
@@ -388,6 +388,23 @@ impl Manifest {
...
+    /// JUMBF-URI identifiers (e.g. claim thumbnails) can be resolved lazily.
+    pub(crate) fn set_store_resolver(&mut self, store: std::sync::Arc<Store>) {
+        use crate::jumbf::labels::ASSERTIONS;
```
tmathern: Imports on top? I think this call will be hit more than once every two blue moons (correct me if the assumption is wrong), so the import cost would be justified.

sdk/src/resource_store.rs
```diff
@@ -82,25 +83,27 @@ impl UriOrResource {
         match self {
             UriOrResource::ResourceRef(r) => Ok(UriOrResource::ResourceRef(r.clone())),
             UriOrResource::HashedUri(h) => {
-                let (format, data) = if h.url().contains(DATABOXES) {
+                let url = to_absolute_uri(claim.label(), &h.url());
+                if h.url().contains(DATABOXES) {
+                    // v1 databoxes: eagerly materialize; no store-level resolver for these yet.
```
tmathern:
```suggestion
                    // v1 databoxes: eagerly materialize; no store-level resolver for these.
```
(Except if you plan adding those?)

sdk/src/ingredient.rs
ok-nick: I agree.

sdk/src/ingredient.rs
```diff
@@ -495,19 +471,61 @@ impl Ingredient {
...
+                        .map(|db| db.data.clone());
+                }
+                if uri.contains(ASSERTIONS) {
+                    let assertion = store.get_assertion_from_uri(uri)?;
+                    // Try EmbeddedData first (claim thumbnails, data assertions).
+                    // Fall back to raw assertion bytes (ingredient thumbnails stored as binary).
+                    if let Ok(embedded) = EmbeddedData::from_assertion(assertion) {
+                        return Some(embedded.data);
+                    }
+                    return Some(assertion.data().to_vec());
```
ok-nick: Can we check for the `c2pa.embedded-data` label instead?

sdk/src/resource_store.rs
```diff
@@ -170,6 +173,46 @@ pub struct ResourceStore {
...
+}
+
+#[cfg(target_arch = "wasm32")]
+impl std::ops::Deref for StoreResolver {
+    type Target = dyn Fn(&str) -> Option<Vec<u8>>;
+
+    fn deref(&self) -> &Self::Target {
+        &*self.0
+    }
 }
```
ok-nick: Hm, I'd prefer if we had an explicit method to get the closure from a `StoreResolver`. I find it ambiguous to implicitly convert the struct to a closure.

sdk/src/resource_store.rs
```diff
@@ -170,6 +173,46 @@ pub struct ResourceStore {
...
+    #[cfg_attr(feature = "json_schema", schemars(skip))]
+    resolver: Option<StoreResolver>,
+}
+
+/// Type alias for the resolver closure stored in [`StoreResolver`].
+/// On wasm32 the store is not `Sync`, so we omit the `Send + Sync` bounds.
+#[cfg(not(target_arch = "wasm32"))]
+type ResolverFn = Arc<dyn Fn(&str) -> Option<Vec<u8>> + Send + Sync>;
+#[cfg(target_arch = "wasm32")]
+type ResolverFn = Arc<dyn Fn(&str) -> Option<Vec<u8>>>;
```
ok-nick: Can you use `MaybeSend + MaybeSync` here instead?

sdk/src/resource_store.rs
```diff
@@ -170,6 +173,46 @@ pub struct ResourceStore {
     base_path: Option<PathBuf>,
     #[serde(skip_serializing_if = "Option::is_none")]
     label: Option<String>,
+    /// Optional resolver that can look up resource bytes by URI from an external source
+    /// (e.g. a `Store`). Used to defer materialization of bytes during reading.
+    #[serde(skip)]
+    #[cfg_attr(feature = "json_schema", schemars(skip))]
```
ok-nick: `schemars` automatically skips if it's marked as `serde` skip iirc.

sdk/src/builder.rs
gpeacock: The method isn't deprecated, only the ability to use it to write the old archive format. That already requires overriding the default generate_c2pa_archve setting. So I think this error message should cover it. There are all kinds of problems with the old format and the sooner we get rid of it, the better. We can still read the old format, but we won't write it anymore.

sdk/src/builder.rs
gpeacock: we sill support reading the old format, so you can basically just load the old archive into a builder and then archive the builder again to convert it to the new format.

sdk/src/resource_store.rs
gpeacock: I was debating about if it is worth doing or not

sdk/src/builder.rs
tmathern: Great! Thanks. (I really do like the error message, makes it very, very clear).

sdk/src/resource_store.rs
tmathern: How old is this code path? How likely is it to be hit?

sdk/src/ingredient.rs
```diff
@@ -495,19 +471,61 @@ impl Ingredient {
...
+                    let hashed_uri = crate::hashed_uri::HashedUri::new(uri.to_owned(), None, &[]);
+                    return store
+                        .get_data_box_from_uri_and_claim(&hashed_uri, &label)
+                        .map(|db| db.data.clone());
+                }
+                if uri.contains(ASSERTIONS) {
+                    let assertion = store.get_assertion_from_uri(uri)?;
+                    // Try EmbeddedData first (claim thumbnails, data assertions).
+                    // Fall back to raw assertion bytes (ingredient thumbnails stored as binary).
+                    if let Ok(embedded) = EmbeddedData::from_assertion(assertion) {
```
ok-nick: How come `Store::get_assertion_from_uri` doesn't include embedded data? Why do we need something custom here?

sdk/src/ingredient.rs
```diff
@@ -1134,14 +1133,11 @@ impl Ingredient {
         let mut thumbnail = None;
         // for Builder model, ingredient resources may be in the manifest
         let get_resource = |id: &str| {
-            self.resources
-                .get(id)
-                .or_else(|_| {
-                    resources
-                        .ok_or_else(|| Error::NotFound)
-                        .and_then(|r| r.get(id))
```
ok-nick, line 1142: How come you don't chain them?

sdk/src/manifest.rs
```diff
@@ -388,6 +388,23 @@ impl Manifest {
         Ok(self)
     }
 
+    /// Installs a store-backed resolver on this manifest's resources so that
+    /// JUMBF-URI identifiers (e.g. claim thumbnails) can be resolved lazily.
+    pub(crate) fn set_store_resolver(&mut self, store: std::sync::Arc<Store>) {
```
ok-nick: Import `Arc` at the top?

sdk/src/manifest.rs
```diff
@@ -388,6 +388,23 @@ impl Manifest {
...
+        use crate::jumbf::labels::ASSERTIONS;
+        self.resources
+            .set_resolver(std::sync::Arc::new(move |uri: &str| {
+                if uri.contains(ASSERTIONS) {
+                    let assertion = store.get_assertion_from_uri(uri)?;
+                    if let Ok(embedded) = EmbeddedData::from_assertion(assertion) {
+                        return Some(embedded.data);
+                    }
+                    return Some(assertion.data().to_vec());
```
ok-nick: Same here, why don't we check the embedded assertion label?

sdk/src/ingredient.rs
gpeacock: what do you mean? I do chain them, I just don't need the extra deferred call since that is now handled in resources.

sdk/src/ingredient.rs
gpeacock: we don't really use the data field, and certainly not from other manifests. But I changed this to align with the other code just in case

sdk/src/ingredient.rs
gpeacock: you are right, we get the same result. I updated this

sdk/src/ingredient.rs
gpeacock: I changed this code

sdk/src/manifest.rs
gpeacock: done

sdk/src/resource_store.rs
gpeacock: removed

sdk/src/resource_store.rs
gpeacock: good idea

sdk/src/resource_store.rs
gpeacock: I made the requested change..

sdk/src/resource_store.rs
gpeacock: Can't do that, Rust limitation on this kind of type definition.

sdk/src/ingredient.rs
gpeacock: I reworked this to also use absolute uris

sdk/src/ingredient.rs
```diff
@@ -1076,25 +1071,26 @@ impl Ingredient {
...
+                                &absolute_uri,
                             ))
                         })
                 }
-                uri if uri.contains(jumbf::labels::DATABOXES) => store
+                absolute_uri if absolute_uri.contains(jumbf::labels::DATABOXES) => store
                     .get_data_box_from_uri_and_claim(data_uri, claim_label)
                     .map(|data_box| {
-                        Ok::<ResourceRef, Error>(ResourceRef::new(&data_box.format, &uri))
+                        Ok::<ResourceRef, Error>(ResourceRef::new(&data_box.format, &absolute_uri))
```
ok-nick: Why do these need to return `Ok`? I think we can remove the result there.

sdk/src/resource_store.rs
```diff
@@ -364,7 +407,15 @@ impl ResourceStore {
...
         if let Some(base) = self.base_path.as_ref() {
             let path = base.join(id);
-            return path.exists();
+            if path.exists() {
+                return true;
+            }
+        }
+
+        if let Some(resolver) = &self.resolver {
+            if resolver.resolve(id).is_some() {
```
ok-nick: Wouldn't resolve load it into memory? Does that mean every time we check if something exists the bytes needs to be loaded into memory?

sdk/src/ingredient.rs
```diff
@@ -495,19 +472,55 @@ impl Ingredient {
...
+                }
+                if uri.contains(ASSERTIONS) {
+                    let assertion = store.get_assertion_from_uri(uri)?;
+                    return Some(assertion.data().to_vec());
+                }
+                // Manifest: uri is a bare claim label (set by manifest_data ref above).
+                if let Some(claim) = store.get_claim(uri) {
+                    return Store::build_flat_ingredient_store(&store, claim)
+                        .and_then(|s| s.to_jumbf_internal(0))
+                        .ok();
```
ok-nick: Maybe it would make sense if the resolver closure returned a Result so that we can propagate errors back to the caller.

sdk/src/ingredient.rs
ok-nick: I figured the resolver would be used here and since the resolver is chained, we don't need to check `self.resources` and the parameter `resources`.

tmathern (approved):

sdk/src/ingredient.rs
gpeacock: Ingredient is $self here, it is not mutable. this is called by builder::to_claim which is also not mutable. I'd need to change both of those in order to use chaning here. It isn't worth the effort and would cause other undesired side effects.

sdk/src/ingredient.rs
gpeacock: good catch, much cleaner

sdk/src/resource_store.rs
gpeacock: I reworked the resolver logic so that now there is a get/ keys and has method. Exists calls has.

sdk/src/ingredient.rs
gpeacock: New get returns Option Result

sdk/src/reader.rs
```diff
@@ -978,20 +950,36 @@ impl Reader {
...
+
+        for claim in self.store.claims() {
+            let claim_label = claim.label();
+            // Write binary (media) assertions — thumbnails, icons, embedded data
+            for ca in claim.claim_assertion_store() {
+                let assertion = ca.assertion();
+                if !assertion.content_type().starts_with("application/") {
+                    let uri = to_assertion_uri(claim_label, &ca.label());
+                    if let Some(a) = self.store.get_assertion_from_uri(&uri) {
+                        let _ = write_bytes(uri_to_path(&uri, Some(claim_label)), a.data());
```
tmathern: No need to check the result here?

sdk/src/resource_store.rs
```diff
@@ -180,6 +226,53 @@ impl ResourceStore {
...
+        });
+    }
+
+    /// Chains the resolver from `other` as an additional fallback on this store.
+    ///
+    /// When [`get`](Self::get) finds no local match it tries the existing resolver
+    /// chain first, then falls through to `other`'s resolver.  This allows a
+    /// builder's resource store to transparently serve bytes that live in an
+    /// ingredient's manifest store without eagerly copying them.
+    pub(crate) fn chain_resolver_from(&mut self, other: &ResourceStore) {
```
tmathern, line 233: Should the depth of the resolution chain be limited? There were other issues where depth was a concern.

tmathern (approved):

sdk/src/manifest.rs
```diff
@@ -353,39 +349,68 @@ impl Manifest {
...
+
+    /// Adds a store-backed resolver on this manifest's resources so that
+    /// JUMBF-URI identifiers (e.g. claim thumbnails) can be resolved lazily.
+    pub(crate) fn set_store_resolver(&mut self, store: Arc<Store>) {
+        let label = self.label().unwrap_or_default().to_owned();
+        let store_get = store.clone();
+        let store_has = store.clone();
+        let store_keys = store;
+        let label_keys = label;
+        self.resources.set_resolver(
```
ok-nick: Should this also handle databoxes like the ingredient one?

sdk/src/reader.rs
```diff
@@ -978,20 +950,36 @@ impl Reader {
...
+            }
+            std::fs::write(file_path, data).map_err(Error::IoError)
+        };
+
+        for claim in self.store.claims() {
+            let claim_label = claim.label();
+            // Write binary (media) assertions — thumbnails, icons, embedded data
+            for ca in claim.claim_assertion_store() {
+                let assertion = ca.assertion();
+                if !assertion.content_type().starts_with("application/") {
```
ok-nick: Are we trying to ignore writing c2pa manifests here? How come we don't check explicitly for that mime type?

sdk/src/resource_store.rs
```diff
@@ -170,6 +163,59 @@ pub struct ResourceStore {
...
+    }
+
+    fn has(&self, id: &str) -> bool {
+        (self.has_fn)(id)
+    }
+
+    fn keys(&self) -> Vec<String> {
+        (self.keys_fn)()
+    }
 }
```
ok-nick: I think we should consider making this a trait that you implement on a struct. It feels like that's what's trying to be done but using closures instead. The outer struct can be wrapped in an Arc.

sdk/src/reader.rs
```diff
@@ -894,69 +895,40 @@ impl Reader {
...
-                    }
-                },
-            }
+        let bytes: &[u8] = if absolute_uri.contains(ASSERTIONS) {
+            self.store
+                .get_assertion_from_uri(&absolute_uri)
+                .map(|a| a.data())
+                .ok_or_else(|| Error::ResourceNotFound(uri.to_owned()))?
+        } else if absolute_uri.contains(DATABOXES) {
+            let hashed_uri = HashedUri::new(absolute_uri.clone(), None, &[]);
```
ok-nick: Are we sure it's okay to leave an empty hash here? Looking at the store we compare it:
https://github.com/contentauth/c2pa-rs/blob/8486c327039a9b42037793ffe758bf5cdf1ce861/sdk/src/claim.rs?plain=1#L1596

sdk/src/reader.rs
gpeacock: fixed

sdk/src/reader.rs
gpeacock: I changed the check

sdk/src/reader.rs
gpeacock: reworked

sdk/src/resource_store.rs
gpeacock: I reworked this code, but I'm not very concerned about that since this is an internal method only used to chain an ingredient to a manifest.

sdk/src/resource_store.rs
gpeacock: ok

sdk/src/manifest.rs
gpeacock: I doubt that would ever come up,but I added it just to be safe.

ok-nick (approved):

PR #2176, opened 2026-05-28
fix: propagate hard errors when parsing ingredient streams

ok-nick:
When calling a function such as `Builder::add_ingredient_from_stream` and we fail to parse the stream, errors are ignored. This may happen, for example, if the wrong format string is passed in, or if there is a problem parsing the stream.

This change propagates those hard errors to the caller. Note for validation errors there is no hard error and it's added to the ingredient's validation results.


sdk/src/ingredient.rs
```diff
@@ -721,25 +721,7 @@ impl Ingredient {
...
-                    .collect();
-
-                for status in statuses {
-                    results.add_status(status.clone());
-                }
-                self.validation_status = results.validation_errors();
-                self.validation_results = Some(results);
-                Ok(())
-            }
+            Err(e) => Err(e),
```
gpeacock: This seems like the opposite of what you want to do. We were adding these errors to the validation log, now were aren't. is this because we are adding them now at some other layer?

gpeacock (commented):
We didn't want to fail in this code because we want to add ingredients even if the validation fails. I'm not sure what the full impact of this change would be. Is this now captured at some different level? If so, please comment on how that works and why we no longer need to log the errors.

sdk/src/ingredient.rs
ok-nick: If there was a validation error, the store wouldn't return a hard error. We handle that case in the other match paths:
https://github.com/contentauth/c2pa-rs/blob/6fee6c0de010d65308938edb1e49ff8a9a466800/sdk/src/ingredient.rs#L640-L723

If there is a hard error, such as failing to parse the stream, it would be caught in the path I made the change and I believe it should be propagated back to the caller.

sdk/src/builder.rs
```diff
@@ -3720,6 +3720,25 @@ mod tests {
...
+        })
+        .to_string();
+
+        let mut builder = Builder::default()
+            .with_definition(simple_manifest_json())
+            .unwrap();
+
+        // pass incorrect format
+        let result = builder.add_ingredient_from_stream(ingredient_json, "image/png", &mut source);
+        assert!(result.is_err());
```
gpeacock, line 4262: This would be a breaking change. Is it worth it? We intentionally give a lot of slack when adding ingredients. The idea is to catch issues in the validation status but otherwise add things if we can. In this case we can't parse the format but we could still add an ingredient and report the format mismatch in the validation_status.
Note that we specifically allow for unsupported type errors and return OK()
We should be adding an informational status that there was no provenance in those cases, but it is not an error and the ingredient will still be added. We could do something similar for an unparsable vs unknown.
```
            Err(Error::JumbfNotFound)
            | Err(Error::ProvenanceMissing)
            | Err(Error::UnsupportedType) => Ok(()), // no claims but valid file
            Err(Error::BadParam(desc)) if desc == *"unrecognized file type" => Ok(()),
```

ok-nick: This case always failed because ingredient v3s must have an `activeManifest` specified if there are `validationResults`:
https://github.com/contentauth/c2pa-rs/blob/22444234e41c2030d061169e2bb9fb674d3418f1/sdk/src/assertions/ingredient.rs?plain=1#L426-L433

The spec is interesting because [in validation it says](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_performing_recursive_validation):
> If the ingredient assertion does not have an activeManifest (or c2pa_manifest) field, record an ingredient.unknownProvenance informational code...

However, [for ingredient assertions it says](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_v3_ingredient_assertions):
> In a v3 ingredient assertion with no activeManifest field, the validationResults field shall not be present.

So the spec sort of contradicts itself here.

In any case, I think it would be ideal if adding an ingredient from a stream failed explicitly, so the user can choose to add the ingredient without the stream. I understand we already skip this in other cases, so maybe it makes sense to apply that behavior more generally to any type of hard error.

gpeacock: I think the spec is wrong in saying that there can be no validationResults without and activeManifest, for exactly this reason. It leaves us with no way to report errors that that cause us to no be able to include the manifest, including invalid format issues. I should file a ticket on the spec for that.

ok-nick: I've been thinking about this a little more. I think allowing validation results in an ingredient without an active manifest can make things complicated. Ingredient validation results are really there so you can validate against the ingredient's manifest and compare the deltas. I'm not sure if saying that we failed to parse the asset in the manifest is useful.

Ultimately the SDK should allow the user to explicitly handle this scenario. IMO I think hard error is the right choice so the user can choose to add the ingredient without a stream. This is non-breaking (besides the different error) because the code already failed due to missing both `activeManifest` and `validationResults` in the ingredient

gpeacock: These are requirements in the spec:

> If the ingredient assertion does not have an activeManifest (or c2pa_manifest) field, record an ingredient.unknownProvenance informational code unless the value of the relationship field is inputTo, and then skip to the next ingredient assertion until they are all exhausted. At that point, return from the current recursion level.

also:

> If a manifest was documented to exist in a remote location, but is not present there, or the location is not currently available (such as in an offline scenario), the manifest.inaccessible error code shall be used to report the situation.

and

> If an ingredient’s manifest is [remote](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_external_manifests), and the claim generator is unable to retrieve the manifest, it should use an error code of manifest.inaccessible to reflect that.

But the spec also says we cannot have validation_results without manifest data

> In a v3 ingredient assertion with no activeManifest field, the validationResults field shall not be present.

There is no information on how to return information or error codes for an ingredient without validation_status so we are in a stalemate.

Your proposal is that we should return a hard error if there is no manifest, but there's nothing that says an ingredient must have a manifest. In fact, the only way we know if it has a manifest or not is to read the stream. So we must read the stream to add an ingredient if there is a stream to read. There are also things like the xmp instanceID and documentID that become part of an ingredient even if there is no manifest.

ok-nick: > Your proposal is that we should return a hard error if there is no manifest

I'm proposing that it be a hard error if we fail to parse the stream. If there is no manifest, the ingredient can be added without `validationResults` and without an `activeManifest`. If there is a hard error due to an issue with the stream or our parsing logic, the user can choose to add the ingredient without the stream (`Builder::add_ingredient`) or handle it according to their needs.

Doing the former will result in an `ingredient.unknownProvenance` for that ingredient when validating the top-level manifest.

gpeacock: So if they pass in a stream and we have a hard error they would follow that up by making another call to add a kind of generic ingredient reference? I suppose that could work but it will definitely be a breaking change. PS would be very surprised by the change if we don't take special care to warn them about it.

I presume you would also do a hard error for a remote manifest we cannot retrieve (due to server errors or being offline). We need to work with the C2PA spec to come up with some new way to report this.

ok-nick: It isn't a breaking chance since the code already failed due to missing both `activeManifest` and `validationResults` in the ingredient. I would actually consider this change to be more of a clarification of the returned error. The behavior I'm proposing is what we already do today.

For the remote manifest, it's already being handled separately in the same match block, although looking at it now it seems it suffers from the same problem. For that case it makes sense to me if the ingredient reported `manifest.inaccessible`. There would need a spec change.

ok-nick: See #2327

gpeacock (changes requested):
If we fix https://github.com/contentauth/c2pa-rs/pull/2367 to return errors or report validation status, then we don't need this change. There are essentially two different behaviors we need to test and validate. If ignore_ingredient_errors is false then we get the functionality defined here. if it is true, then we really need to remove the validation check that doesn't allow validation_results without a manifest. Leonard says that is the correct behavior and the spec isn't right on that. If we don't want to wait for that to be updated in the spec, then don't add the validation results, but then we need to file another issue to fix that.

sdk/src/ingredient.rs
```diff
@@ -744,24 +745,12 @@ impl Ingredient {
...
-                    .logged_items()
-                    .iter()
-                    .filter_map(ValidationStatus::from_log_item)
-                    .collect();
-
-                for status in statuses {
-                    results.add_status(status.clone());
+            Err(err) => {
+                if context.settings().builder.ignore_ingredient_errors {
+                    debug!("ignoring ingredient error: {err:?}");
```
gpeacock, line 750: If we are ignoring the ingredient errors here, I think we should also add them to the validation_status/results like we did before. That's the safest thing to do for the compatibility case.

gpeacock (changes requested):
When the ignore errors flag is set, we should still add any new status to the validation_status/results like we used to do. I'm not sure what those errors would be, but it seems like the safest path, and it won't be the default anymore.

sdk/src/ingredient.rs
ok-nick: Added 👍

gpeacock (approved):

PR #2177, opened 2026-05-28
perf: avoid buffering PNG when writing XMP and removing manifest

ok-nick:
Avoids buffering the PNG into memory when writing XMP and removing manifests. Reduces total memory usage for writing XMP w/o embedded manifest in a 100MB PNG by 16% and peak memory usage by 26.7%.

Before:
Screenshot 2026-05-28 at 1 50 23 PM

After:
Screenshot 2026-05-28 at 1 50 13 PM

PR #2178, opened 2026-05-28
perf: avoid buffering intermediate streams if stream len is greater than threshold

ok-nick:
Avoids buffering assets into memory when the input stream len exceeds the threshold defined in settings. When signing a simple 100MB PNG with the threshold set 50MB, total memory usage decreases by 79.7%.

Before:
Screenshot 2026-05-28 at 2 46 09 PM

After:
Screenshot 2026-05-28 at 2 47 20 PM

These benchmarks are given the `backing_store_memory_threshold_in_mb` is set to 50 on a 100MB PNG. I also included the changes in #2177 to these benchmarks.

### DIFF
--- a/sdk/src/utils/io_utils.rs
+++ b/sdk/src/utils/io_utils.rs
@@ -121,8 +121,8 @@ pub(crate) fn stream_len<R: Read + Seek + ?Sized>(reader: &mut R) -> Result<u64>
 
 #[cfg(target_arch = "wasm32")]
 pub(crate) fn stream_with_fs_fallback(
-    _threshold: usize,
-    _expected_size: u64,
+    _threshold_mb: usize,
+    _expected_size_bytes: u64,
 ) -> Result<impl Read + Write + Seek> {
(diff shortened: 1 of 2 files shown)


sdk/src/utils/io_utils.rs
```diff
@@ -119,29 +119,39 @@ pub(crate) fn stream_len<R: Read + Seek + ?Sized>(reader: &mut R) -> Result<u64>
     Ok(len)
 }
 
+#[cfg(target_arch = "wasm32")]
+pub(crate) fn stream_with_fs_fallback(
```
tmathern, line 123: Ok to do the signature change for wasm here? I assume so.

tmathern (approved):

sdk/src/utils/io_utils.rs
ok-nick: Yep, tested it and CI tests it.

gpeacock (approved):

PR #2180, opened 2026-05-29
fix: Warn instead of error when redaction cannot be applied

ale-adobe:
- Update logic in `builder.rs` and `claim.rs` to only emit a warning when a redaction cannot be applied, rather than throwing an `AssertionRedactionNotFound` error.
- Remove `AssertionRedactionNotFound` entirely; use a boolean to communicate whether applying a redaction was successful or not
- Update `redaction.md` to remove the requirement that URIs in the redactions list must match an assertion

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -1488,7 +1488,9 @@ impl Builder {
             let applied = claim.redactions().map(|r| r.as_slice()).unwrap_or_default();
             for redaction in redactions {
                 if !applied.contains(redaction) {
-                    return Err(Error::AssertionRedactionNotFound);
+                    log::warn!(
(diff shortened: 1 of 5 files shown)


sdk/src/builder.rs
ok-nick, line 1: How come this is being changed to a warning? I actually think it should hard error and the caller should make sure they don't supply an invalid URI. I'd imagine it can be a surprise if redaction is implicitly skipped.

mauricefisher64:
I think we have come up with a better solution here that would avoid the issue of false positive error.  I think we should leave the code as is and I will submit a fix to handle the underlying concern.  That is we should not report errors if the redaction is already accounted for elsewhere.

ale-adobe:
Closing PR since @mauricefisher64 has a better solution in mind.

PR #2181, opened 2026-05-29
feat: Auto format detect for Reader

gpeacock:
Check magic bytes to detect format and fall back to provided format/extension on Reader

If the supplied extension is supported and we have detected another format from the same asset handler, we will use the extension provided.
If the detected handler does not match the handler for the supplied extension we use the detected handler.
if we don't detect a handler and we have a handler for the supplied extension we use it.
otherwise we return that it is not supported.

- Add `format_from_stream()` to `utils/mime.rs`: reads ≤16 bytes of
  magic signatures (JPEG, PNG, GIF, TIFF/BigTIFF, JXL, RIFF, BMFF,
  FLAC, ID3, MPEG, PDF) and rewinds before returning
- Wire detection into `Reader::with_stream()` and `Reader::with_file()`;
  caller-supplied format is used as fallback when detection returns None
- Add `"jxl" => "image/jxl"` to `extension_to_mime` (was missing)
- Swap `tests/fixtures/sample1.heic` / `sample1.heif` so file names
  match their actual BMFF `ftyp` brands (`heic` / `mif1`)
- Add `create_test_stream()` helper to `utils/test.rs`
- Rewrite mime tests to use real fixture files via `detect_fixture()`;
  retain synthetic tests only for BigTIFF (no fixture) and ID3-tagged
  FLAC (no real fixture with ID3 prepended)


sdk/src/utils/mime.rs
```diff
@@ -103,3 +104,268 @@ pub fn format_from_path<P: AsRef<std::path::Path>>(path: P) -> Option<String> {
...
+///
+/// For streams carrying ID3-tagged audio the function decodes the ID3v2
+/// sync-safe tag size and peeks past the tag header to determine whether
+/// the payload is FLAC (`fLaC` marker) or MP3 (anything else).
+pub fn format_from_stream<R: std::io::Read + std::io::Seek>(stream: &mut R) -> Option<String> {
+    use std::io::SeekFrom;
+
+    stream.rewind().ok()?;
+    let mut buf = [0u8; 16];
+    // `read` may return fewer bytes than requested for short streams.
```
tmathern: Should it then return `ErrorKind::UnexpectedEof`, and let the caller handle it? I imagine this may happen with (future) streaming cases?
(In tests, I think things always fill buffer eagerly, since we use Cursor, so this won't be hit easily).

sdk/src/utils/mime.rs
```diff
@@ -103,3 +104,268 @@ pub fn format_from_path<P: AsRef<std::path::Path>>(path: P) -> Option<String> {
...
+        return Some("image/png".to_string());
+    }
+
+    // GIF87a or GIF89a
+    if n >= 6 && &buf[0..3] == b"GIF" && (&buf[3..6] == b"87a" || &buf[3..6] == b"89a") {
+        return Some("image/gif".to_string());
+    }
+
+    // TIFF (standard and BigTIFF), both byte orders
+    if n >= 4
```
tmathern: Can this confound with DNG? Asking because I thought DNG is (loosely?) based on TIFF, but with enough differences that a mismatch here may be relevant. If we get a DNG (I think we have some tests about this in downstream SDKs), will this still work in an acceptable fashion so DNGs are handled properly?

sdk/src/utils/mime.rs
```diff
@@ -103,3 +104,268 @@ pub fn format_from_path<P: AsRef<std::path::Path>>(path: P) -> Option<String> {
...
+        return Some("image/jxl".to_string());
+    }
+
+    // RIFF family: "RIFF" at bytes 0-3, four-CC format tag at bytes 8-11
+    if n >= 12 && &buf[0..4] == b"RIFF" {
+        return Some(
+            match &buf[8..12] {
+                b"WEBP" => "image/webp",
+                b"AVI " => "video/avi",
+                _ => "audio/wav", // WAVE and other unknown RIFF variants
```
tmathern: All the unknowns can be handled like wav?

sdk/src/utils/mime.rs
gpeacock: This is intended as a helper method, if it fails there should always be a fallback. So I think None is ok for a general meaning of failed to detect anything. Returning an error on a file in an unexpected format might be misleading. What if it is a text file with 2 bytes? The right answer is we didn't detect anything, not that there was an unexpected Eof.

sdk/src/utils/mime.rs
gpeacock: I rewrote the logic on this. Now 
if the supplied extension is supported and we have detected another format from that asset handler, we will use the extension provided.
If the detected handler does not match the handler for the supplied extension we use the detected handler.
if we don't detect a handler and we have a handler for the supplied extension we use it.
otherwise we return that it is not supported.

sdk/src/utils/mime.rs
gpeacock: I rewrote how this was done, see above.

tmathern (approved):

mauricefisher64 (approved):

Issue #2182, opened 2026-05-29
Remove `#[ignore]` on `test_multiple_identities_valid` by mocking networking


ok-nick:
The test intermittently fails due to networking problems. Ideally these tests wouldn't use networking and we mock the backend HTTP calls.

ok-nick:
Close in favor of #2184

PR #2183, opened 2026-05-29
test: ignore `test_multiple_identities_valid` due to intermittent networking failures

ok-nick:
The `test_multiple_identities_valid` intermittently fails due to networking issues. This is presumably due to the new changes that allow network requests to be called for CAWG in sync contexts. See #2172 and #2120.

Ideally we mock the networking, but for the time being let's ignore it so CI isn't failing.

See #2182 for more information.

### DIFF
--- a/sdk/src/identity/validator.rs
+++ b/sdk/src/identity/validator.rs
@@ -99,6 +99,8 @@ mod tests {
         );
     }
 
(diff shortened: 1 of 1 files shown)


ok-nick:
Close in favor of #2184

PR #2184, opened 2026-05-29
chore: Replace live HTTP GET with a local mock server for Connected Identities service

scouten-adobe:
## Root cause (confirmed)

`test_multiple_identities_valid` (and `test_connected_identities_valid`) validate JPEGs whose CAWG identity assertions are issued by `did:web:connected-identities.identity-stage.adobe.com`. Validation resolved that DID by making a **live HTTPS GET** to `https://connected-identities.identity-stage.adobe.com/.well-known/did.json`. Any network/server hiccup on that Adobe stage host produced an intermittent failure.

## The fix

Generalized the test PROXY hook in `did_web.rs`: replaced the single-value `PROXY: Option<String>` (which only rewrote `localhost`) with a `PROXIES: HashMap<domain → url-prefix>`, plus `set_proxy()` / `clear_proxies()` helpers. `to_url()` now redirects resolution for any registered domain. Updated the three existing `did_web` resolve tests to the new API.

Captured the issuer DID document as a fixture: `connected_identities_did.json`.

Made both validator tests hermetic in `validator.rs`: a `mock_connected_identities_did()` helper starts an `httpmock` server serving the fixture and registers a proxy entry for the Adobe stage domain. The mock setup is gated `#[cfg(not(target_arch = "wasm32"))]`, matching the existing `did_web` network tests (`httpmock` is a native-only dev-dep); Wasm behavior is unchanged.

## Checklist

### DIFF
--- a/sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
+++ b/sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
@@ -32,11 +32,32 @@ pub(crate) const MAX_DID_DOC_SIZE: u64 = 1024 * 1024; // 1 MiB
 const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
 
 #[cfg(test)]
-use std::cell::RefCell;
+use std::{cell::RefCell, collections::HashMap};
 
 #[cfg(test)]
 thread_local! {
-    pub(crate) static PROXY: RefCell<Option<String>> = const { RefCell::new(None) };
+    /// Maps a `did:web` domain name to a URL prefix (ending in `/`) that should
+    /// be used in place of the real `https://{domain}/` origin. Tests register
+    /// entries here to redirect DID document resolution to a local mock server,
+    /// keeping the suite hermetic.
+    pub(crate) static PROXIES: RefCell<HashMap<String, String>> = RefCell::new(HashMap::new());
+}
+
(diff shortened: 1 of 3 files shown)


sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -32,11 +32,32 @@ pub(crate) const MAX_DID_DOC_SIZE: u64 = 1024 * 1024; // 1 MiB
...
 
 #[cfg(test)]
 thread_local! {
-    pub(crate) static PROXY: RefCell<Option<String>> = const { RefCell::new(None) };
+    /// Maps a `did:web` domain name to a URL prefix (ending in `/`) that should
+    /// be used in place of the real `https://{domain}/` origin. Tests register
+    /// entries here to redirect DID document resolution to a local mock server,
+    /// keeping the suite hermetic.
+    pub(crate) static PROXIES: RefCell<HashMap<String, String>> = RefCell::new(HashMap::new());
+}
```
ok-nick, line 44: I don't think this would be an issue in the general case but if two tests are on the same thread they can run into conflicts here.

ok-nick (approved):
Cool solution, it's like a mini DNS.

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
scouten-adobe: They shouldn't be running on the same thread at the same time.

PR #2186, opened 2026-05-29
chore: no-op CI trigger for Rust 1.96.0 rust-lld bus-error repro

scouten-adobe:
Adds an HTML comment to README.md so this PR is a true no-op and can be labeled "check-release" to reproduce the rust-lld SIGBUS observed on ubuntu-latest + stable doctest link in PR #2184 CI.

Do not merge.


scouten-adobe:
Withdrawing. Looks like there was a transient failure.

PR #2187, opened 2026-05-29
chore: probe — force system ld on ubuntu-latest+stable tier-1b job

scouten-adobe:
PROBE BRANCH (do not merge). Targets only the matrix entry that crashes in rust-lld with SIGBUS on Rust 1.96.0 stable (ubuntu-latest, OpenSSL features). Sets RUSTFLAGS/RUSTDOCFLAGS="-C link-arg=-fuse-ld=ld" via matrix include so other entries are untouched.

If this run passes where the no-op repro PR fails, rust-lld is confirmed as the cause and the file-a-bug path opens.

Related: PR #2184 CI run 26646459500.


scouten-adobe:
Closing. Looks like we were investigating a transient failure.

PR #2188, opened 2026-05-29
Add HtmlIO asset handler for text/html (C2PA 2.4 Appendix A.7)

erik-sv:
## Summary

Adds an `HtmlIO` asset handler implementing the C2PA 2.4 Appendix A.7 HTML embedding method, built on the [`c2pa-text`](https://crates.io/crates/c2pa-text) crate. The Manifest Store is Base64-encoded into a `<script type="application/c2pa">` element in the document `<head>`. External `<link rel="c2pa-manifest">` references are recognized on read and treated as an external, non-embedded store. The `c2pa.hash.data` exclusion range covers the script element per A.7.1.3.

## Feature gating

This handler is gated behind the non-default `html` feature.

- Rebased onto current `main` and re-stacked on the updated #2117 (`c2pa-text` 3.0.0).
- `write_cai` now replaces an existing manifest element in place at its original byte offset, preserving the manifest location for update manifests (@mauricefisher64's review comment). Insertion into `<head>` happens only when no C2PA element exists yet. A new `html_io_write_preserves_element_position` test covers this.
- Stream reads are bounded with `stream_len` plus `read_to_vec` (review feedback).

## Scope

Only the HTML method (A.7). Unstructured text (A.8) is #2117. Structured text (A.9) is #2190. SVG stays with the existing `SvgIO` (A.3.3).

|---|---|---|---|

- `cargo test -p c2pa --lib --features plain_text,html,file_io`: 868 passed, 0 failed.
- `cargo test -p c2pa --lib --features html,file_io` (feature standing alone): 864 passed, 0 failed.
- `cargo check -p c2pa` (default features): clean. Clippy and nightly fmt: clean.
- Combined build with all three text features (this PR merged with #2190): 874 passed, 0 failed.


sdk/src/asset_handlers/html_io.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn write_cai(
+        &self,
+        input_stream: &mut dyn CAIRead,
+        output_stream: &mut dyn CAIReadWrite,
+        store_bytes: &[u8],
+    ) -> Result<()> {
+        let text = stream_to_string(input_stream)?;
+
+        // Remove any existing inline manifest so writes are idempotent.
+        let clean = strip_manifest_script(&text);
```
mauricefisher64: See my comment in write_cai here:  https://github.com/contentauth/c2pa-rs/pull/2190

We expect our writers to preserve the location of original manifest if it exists. This enables support for update manifests.

sdk/src/asset_handlers/text_io.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn write_cai(
+        &self,
+        input_stream: &mut dyn CAIRead,
+        output_stream: &mut dyn CAIReadWrite,
+        store_bytes: &[u8],
+    ) -> Result<()> {
+        let text = stream_to_string(input_stream)?;
+
+        // Strip any existing wrapper to get clean text.
+        let clean = match extract_manifest(&text) {
```
mauricefisher64: Same comment as html_io:write_cai

dcondrey (commented):
The architectural concern I raised on #2117 applies here too: this makes the reference implementation's HTML embedding (spec A.7) depend on the external `c2pa-text` crate rather than owning it in-tree.

Independent of direction, the one code item worth fixing is the unbounded `read_to_end` in `stream_to_string` — a maliciously large input can OOM the process, so it should read into a bounded, `try_reserve`-backed buffer.

I think HTML embedding belongs in the native text support I'm putting together for `c2pa-rs` alongside A.8 and A.9, so the whole text surface is one dependency-free implementation.

erik-sv:
Rebased on current main, gated behind a non-default feature, and the review feedback is addressed in code with a regression test. Details in the updated description under "Updated 2026-07-27, response to review".

Issue #2189, opened 2026-05-29
SvgIO applies the SVG (A.3.3) embedding method to generic XML media types, diverging from A.9


erik-sv:
## Observation

`SvgIO` registers the SVG metadata-element embedding method for a set of generic XML media types in addition to SVG itself:

```
["svg", "application/svg+xml", "xhtml", "xml",
 "application/xhtml+xml", "application/xml", "image/svg+xml", "text/xml"]
```

For `image/svg+xml`, this is correct and spec-compliant: **C2PA 2.4 Appendix A.3.3** prescribes embedding the Base64-encoded Manifest Store as a `c2pa:manifest` element inside the `<metadata>` element, with `xmlns:c2pa="http://c2pa.org/manifest"` on the `<svg>` root (the recent namespace-placement fix matches this exactly).

However, for the **non-SVG** XML types — `text/xml`, `application/xml`, `application/xhtml+xml`, `xhtml` — the same SVG metadata-element method is applied, which diverges from the specification:

- **Appendix A.9 (Embedding Manifests into Structured Text)** states that XML-based formats that do not have a format-specific embedding section *may use the structured text method with XML comment syntax* (`<!-- -----BEGIN C2PA MANIFEST----- … -----END C2PA MANIFEST----- -->`).
- A.9 explicitly **excludes only `image/svg+xml`** from the structured-text method (because it has its own A.3.3 section). Generic XML has no format-specific section, so A.9 is the prescribed method for it.

In other words, `text/xml` / `application/xml` / `application/xhtml+xml` are currently embedded with the SVG-specific mechanism rather than the A.9 comment-block mechanism the spec assigns to them.

## Why it matters now

Structured-text (A.9) support is being added to `c2pa-rs` (companion to #2117 and the HTML A.7 handler). An A.9 handler would naturally claim `text/xml` / `application/xml` / `application/xhtml+xml`, which `SvgIO` currently also claims — so the two handlers would contend for the same media types, and assets produced by one would not validate under the other.

## Suggested direction

- Narrow `SvgIO` to **`image/svg+xml`** (and the `svg` extension), keeping the A.3.3 metadata-element method for SVG only.
- Route generic XML media types to the forthcoming **A.9 structured-text** handler (XML comment syntax).

Happy to align the structured-text PR with whatever the maintainers prefer here — flagging it up front so the handler boundaries are settled before that PR lands.

mauricefisher64:
[...]

PR #2190, opened 2026-05-29
Add StructuredTextIO asset handler for structured text (C2PA 2.4 Appendix A.9)

erik-sv:
## Summary

Adds a `StructuredTextIO` asset handler implementing the C2PA 2.4 Appendix A.9 structured-text embedding method, built on the [`c2pa-text`](https://crates.io/crates/c2pa-text) crate. The Manifest Store is carried as a `data:application/c2pa;base64,...` reference inside an ASCII-armour block (`-----BEGIN C2PA MANIFEST----- ... -----END C2PA MANIFEST-----`) placed in a host-format comment line. Markdown uses `<!-- -->`; YAML and TOML use `#`. The `c2pa.hash.data` exclusion covers the block line per A.9.4.

Registers `md`, `markdown`, `text/markdown`, `yaml`, `yml`, `application/yaml`, `text/yaml`, and `toml`. Per A.9, Markdown is structured text because it has comment syntax, so it lives here rather than in the `text/plain` handler.

## Feature gating

This handler is gated behind the non-default `structured_text` feature.

- Rebased onto current `main` and re-stacked on the updated #2117 (`c2pa-text` 3.0.0).
- `write_cai` now replaces an existing manifest block line in place, preserving its original position and line terminator for update manifests (@mauricefisher64's review comment). Insertion at document start happens only when no block exists. A new `structured_write_preserves_block_position` test covers this.
- Stream reads are bounded with `stream_len(reader)` plus `reader.read_to_vec(len)`, replacing the unbounded `read_to_end`, as suggested in review. An oversized asset now fails safely with `BadParam` instead of aborting.

## Scope and the XML boundary

This handler does not claim the XML media types (`text/xml`, `application/xml`, `application/xhtml+xml`). Per A.9 those should use the structured-text comment method, but they are currently claimed by `SvgIO`, which applies the A.3.3 metadata-element method. The overlap is tracked in #2189. `image/svg+xml` remains with `SvgIO` in all cases.

|---|---|---|---|

- `cargo test -p c2pa --lib --features plain_text,structured_text,file_io`: 869 passed, 0 failed.
- `cargo test -p c2pa --lib --features structured_text,file_io` (feature standing alone): 865 passed, 0 failed.
- `cargo check -p c2pa` (default features): clean. Clippy and nightly fmt: clean.
- Combined build with all three text features (this PR merged with #2188): 874 passed, 0 failed.


sdk/src/asset_handlers/text_io.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn write_cai(
+        &self,
+        input_stream: &mut dyn CAIRead,
+        output_stream: &mut dyn CAIReadWrite,
+        store_bytes: &[u8],
+    ) -> Result<()> {
+        let text = stream_to_string(input_stream)?;
+
+        // Strip any existing wrapper to get clean text.
+        let clean = match extract_manifest(&text) {
```
mauricefisher64: Does this combination of extract_manifest and embed_manifest ensure that the new manifest is written at the exact same position?  This is required for update manifests to work correctly.

sdk/src/asset_handlers/structured_text_io.rs
```diff
@@ -0,0 +1,342 @@
...
+        "md" | "markdown" | "text/markdown" => ("<!--", "-->"),
+        // YAML, TOML, INI and similar use a hash line comment.
+        _ => ("#", ""),
+    }
+}
+
+/// Read the entire stream as a UTF-8 string.
+fn stream_to_string(reader: &mut dyn CAIRead) -> Result<String> {
+    reader.rewind()?;
+    let mut buf = Vec::new();
```
mauricefisher64: This will panic if the string is too big for memory.  If you use 

let stream_len = stream_len(reader)?;
let mut buf = reader.read_to_vec(stream_len)?;

This will fail safely if some tries to pass you a bad buffer.

dcondrey (commented):
Same architectural concern as on #2117: this routes structured-text embedding (spec A.9) through the external `c2pa-text` crate instead of owning it in the reference implementation, and it covers only a subset of the formats A.9 defines (Markdown / YAML / TOML) rather than the full range.

A.9 is also the spec's preferred general text method — the variation-selector method is marked last-resort and under review — so getting the full structured-text range right in the reference implementation matters more here than anywhere in the series.

Independent of direction, the unbounded `read_to_end` in `stream_to_string` is worth fixing — large inputs can OOM the process, so it needs a bounded buffer.

I'm building native A.7–A.9 text support for `c2pa-rs` with no added dependency and the full A.9 format range, so I'd rather fold structured text into that.

erik-sv:
Rebased on current main, gated behind a non-default feature, and the review feedback is addressed in code with a regression test. Details in the updated description under "Updated 2026-07-27, response to review".

gpeacock:
We have recently updated the asset_Io api to be more consistent and avoid code duplication. It is mostly just renaming and cutting code that is no longer needed. Please update

scouten-adobe:
Thank you for this submission, @erik-sv. We've decided to take the smaller dependency in #2283 instead.

PR #2199, opened 2026-06-02
fix: Dup redaction

mauricefisher64:
If we already have recorded a redaction for the assertion don't do it again.  This avoids false error when an item is already redacted.  

## Checklist


sdk/src/store.rs
```diff
@@ -6466,35 +6529,90 @@ pub mod tests {
...
+        )
+        .unwrap();
+
+        // make sure the redaction stuck
+        let redacted_claim = resolved_store.get_claim(pc.label()).unwrap();
         assert!(redacted_claim
             .get_assertion(labels::SCHEMA_ORG, 0)
             .is_none());
+
+        // make sure if we redact again it does not cause issues
```
ale-adobe: Leftover comment?

ale-adobe (approved):

sdk/src/store.rs
```diff
@@ -4131,15 +4132,18 @@ impl Store {
...
-            to_both.retain(|f| !svi.redactions.contains(f));
-            final_redactions.append(&mut to_both);
+            final_redactions.retain(|f| !svi.redactions.contains(f));
         }
 
         let claims_to_add: Vec<Claim> = i_store_mut.claims().into_iter().cloned().collect();
         claim.add_ingredient_data(
             claims_to_add,
-            Some(final_redactions),
+            Some(final_redactions.into_iter().collect()),
```
tmathern, line 4146: This means order if non-deterministic. I think it is fine, but worth checking in downstream SDKs before merging. (If ordering changes break downstream, which I do not expect, we can switch to `BTreeSet` instead of a `Hashet`).

tmathern (approved):

PR #2202, opened 2026-06-03
feat!: Split c2pa-raw-crypto into its own crate

scouten-adobe:
## Summary

The raw cryptographic signing/validation primitives have been extracted out of the `c2pa` SDK into a new crate, **`c2pa_raw_crypto`**. This is phase 1 of a larger `crypto` extraction. COSE, OCSP, ASN.1, and RFC 3161 time stamping deliberately **stay** in the `c2pa` crate; only the raw-signature layer (plus `ec_utils`) moves.

## What moved

- `sdk/src/crypto/raw_signature/**` → `c2pa_raw_crypto/src/**`
- `sdk/src/crypto/ec_utils.rs` → `c2pa_raw_crypto/src/ec_utils.rs`
- The two `check_certificate_trust` backends were **not** moved to the new crate (they depend on COSE trust types); they were relocated within the SDK to `sdk/src/crypto/cose/certificate_trust/{rust_native,openssl}.rs`.

## Key design decisions

- **`RawSigner` is private-key-only.** It exposes `sign`, `alg`, and `max_signature_size` (renamed from `reserve_size`, since it reports only the raw signature length) — and deliberately exposes neither the private key nor the certificate chain. Time stamping, OCSP, the TSA-URL constructor argument, and all `async` variants were removed from the new crate. The constructor is `signer_from_private_key(private_key, alg)`.
- **The signing certificate chain lives in the SDK.** It is no longer parsed, stored, or returned by the raw signer. The SDK parses the PEM chain (`crypto::cert_chain_pem_to_der`) and threads it through its own signer holders (`RawSignerWrapper`, the CAWG/identity holders, etc.). This removed X.509/PEM certificate parsing from the raw signers entirely.
- **Time stamp + OCSP moved "up" into the SDK.** A new SDK-internal `CoseSigner` / `AsyncCoseSigner` trait (in `crypto/cose/cose_signer.rs`) carries those concerns plus the cert chain; `cose::sign` consumes it. The overall COSE box reserve is computed SDK-side from the raw `max_signature_size` plus the cert chain, framing, time stamp, and OCSP.
- **Feature model:** `rust_native_crypto` (default) and `openssl`. Both may be enabled (e.g. via workspace feature unification); `rust_native_crypto` takes precedence at runtime and the OpenSSL backend goes unused — not a compile error. Enabling **neither** is allowed in the `c2pa_raw_crypto` crate, but not in the `c2pa` crate.
- **`SigningAlg` is `#[non_exhaustive]`** (forward-compat for future spec algorithms) and is re-exported as `c2pa::SigningAlg`. `JsonSchema` stays in the SDK (the new crate has no `schemars` dependency).
- **New, narrowly scoped error enums** (`RawSignerError`, `RawSignatureValidationError`) including `UnsupportedAlgorithm(SigningAlg)` and `NoCryptoBackend`.
- **`Oid` wrapper** hides `bcder` from the public API; `validator_for_sig_and_hash_algs` is public and takes `&Oid`.
- Removed the now-unused `Signer::raw_signer()` / `AsyncSigner::async_raw_signer()` hooks (no remaining consumers once COSE wraps every signer uniformly).
- Deleted the orphaned `sdk/src/openssl/openssl_trust_handler.rs` and the dead wasm-only async-validator path.

## Dependencies

The new crate keeps a deliberately small footprint. Backend crypto crates are gated behind `rust_native_crypto` / `openssl`; `der` and `nom` were dropped as direct deps (routed through `pkcs8::der` / unused), and `asn1-rs` is now `rust_native_crypto`-only (needed solely by a `derive(DerSequence)`).

Moving the cert chain out removed X.509/PEM parsing from the signers, but `x509-parser` remains the one always-on heavyweight (still used by `oids`, `ec_utils`, and the validators). Fully removing it — migrating those to the RustCrypto stack (`const_oid` / `spki` / `der`) — is left as a follow-up.

## Public API changes (pre-1.0)

- `c2pa_raw_crypto::signer_from_private_key(private_key, alg)` replaces the former `signer_from_cert_chain_and_private_key`.
- Identity constructors now take the cert chain explicitly, e.g. `X509CredentialHolder::from_raw_signer(signer, cert_chain)` and `IdentityAssertionSigner::new(signer, cert_chain)`.

(Neither of these are formally documented in the public API.)

## Test plan

- [x] `cargo test -p c2pa-raw-crypto` (default / rust-native): 54 pass
- [x] `--no-default-features --features openssl`: 31 pass
- [x] `--no-default-features` (neither backend) + `wasm32` build
- [x] both backends enabled: compiles (rust-native selected at runtime)
- [x] full SDK lib tests, default (openssl) and `rust_native_crypto`: 730 pass each
- [x] workspace builds; no warnings (incl. `missing_docs`) across the matrix
- [x] full SDK integration test suite (`sdk/tests/**`)


crandmck:
Will this change potentially trickle down to other language bindings (when they are updated/released)?

The only (not auto-generated) doc that I think might be affected is for Node.js: https://github.com/contentauth/c2pa-node-v2#identityassertionsigner but I'm not sure if/when the Node signature will change or if there will be other more impacts (e.g. new import required).  Please advise.

scouten-adobe:
> Will this change potentially trickle down to other language bindings (when they are updated/released)?
> 
> The only (not auto-generated) doc that I think might be affected is for Node.js: https://github.com/contentauth/c2pa-node-v2#identityassertionsigner but I'm not sure if/when the Node signature will change or if there will be other more impacts (e.g. new import required). Please advise.

Minimally. The one thing that I'm currently aware of is the `SigningAlg` type which becomes `#[non_exhaustive]`, but that would have had to happen anyway. (C2PA is likely to add new algorithms in future versions of the spec.)

tmathern (changes requested):
Question on feature model to resolve first:

I have a question on this claim: `Feature model: rust_native_crypto (default) and openssl. Both may be enabled (e.g. via workspace feature unification); rust_native_crypto takes precedence at runtime and the OpenSSL backend goes unused — not a compile error. Enabling neither is allowed (remote/custom signer, no validation); the built-in constructors then return RawSignerError::NoCryptoBackend / None at runtime.`

But there is also still this:
https://github.com/contentauth/c2pa-rs/blob/78ec04d00791b33c988522280bf59ae8eac0b1b5/sdk/src/lib.rs#L325 

Note in the bottom:
```
#[cfg(not(any(feature = "openssl", feature = "rust_native_crypto")))]
compile_error!("Either 'openssl' or 'rust_native_crypto' feature must be enabled.");
```

Should it go away or stay then? Doesn't that force at least one to be enabled?

scouten-adobe:
> Question on feature model to resolve first:
> 
> I have a question on this claim: `Feature model: rust_native_crypto (default) and openssl. Both may be enabled (e.g. via workspace feature unification); rust_native_crypto takes precedence at runtime and the OpenSSL backend goes unused — not a compile error. Enabling neither is allowed (remote/custom signer, no validation); the built-in constructors then return RawSignerError::NoCryptoBackend / None at runtime.`
> 
> But there is also still this:
> 
> https://github.com/contentauth/c2pa-rs/blob/78ec04d00791b33c988522280bf59ae8eac0b1b5/sdk/src/lib.rs#L325
> 
> Note in the bottom:
> 
> ```
> #[cfg(not(any(feature = "openssl", feature = "rust_native_crypto")))]
> compile_error!("Either 'openssl' or 'rust_native_crypto' feature must be enabled.");
> ```
> 
> Should it go away or stay then? Doesn't that force at least one to be enabled?

Good call-out, @tmathern. I think, for now, I'd like to leave the code as-is and update the description of this PR to match what actually happens. If we need a no-backend configuration, there are roughly 300 unit tests that will have to be updated to either not be executed or to verify the no-backend / error-out behavior. I'd prefer to not to include that in the scope of this PR.

sdk/src/crypto/raw_signature/tests/mod.rs
```diff
@@ -11,14 +11,17 @@
...
-//! Tests are grouped here because the same test should pass using the same test data regardless of which implementation is used.
+//! Backends for checking X.509 certificate trust against a
+//! [`CertificateTrustPolicy`](super::CertificateTrustPolicy).
+//!
+//! The active backend mirrors the cryptography backend selected for
+//! `c2pa-raw-crypto`.
 
-#![allow(clippy::expect_used)]
-#![allow(clippy::panic)]
-#![allow(clippy::unwrap_used)]
```
ok-nick, line 18: 👍

sdk/src/crypto/cose/certificate_trust/openssl.rs
```diff
@@ -11,16 +11,17 @@
...
 
-use crate::crypto::{
-    cose::{CertificateTrustError, CertificateTrustPolicy, TrustAnchorType},
-    raw_signature::openssl::OpenSslMutex,
-};
+use crate::crypto::cose::{CertificateTrustError, CertificateTrustPolicy, TrustAnchorType};
 
+// Dead when both `openssl` and `rust_native_crypto` are enabled (the rust-native
+// trust backend is selected at runtime in that case).
+#[allow(dead_code)]
```
ok-nick: Can we feature flag this rather than dead code?

sdk/src/crypto/cose/certificate_trust_policy.rs
```diff
@@ -147,22 +147,19 @@ impl CertificateTrustPolicy {
...
                 end_entity_cert_der,
                 signing_time_epoch,
             );
         }
 
-        #[cfg(all(
-            feature = "openssl",
-            not(all(feature = "rust_native_crypto", target_arch = "wasm32"))
-        ))]
+        #[cfg(all(feature = "openssl", not(feature = "rust_native_crypto")))]
```
ok-nick: It's not pretty, but the original reason I added the WASM check here is because otherwise compiling with `--all-features` on WASM will break.

sdk/src/crypto/cose/certificate_trust/rust_native.rs
```diff
@@ -170,79 +170,46 @@ fn check_chain_order(certs: &[Vec<u8>]) -> Result<(), CertificateTrustError> {
...
+        "rsa512" => pair(&RSA_OID, &SHA512_OID),
+        "es256" => pair(&EC_PUBLICKEY_OID, &SHA256_OID),
+        "es384" => pair(&EC_PUBLICKEY_OID, &SHA384_OID),
+        "es512" => pair(&EC_PUBLICKEY_OID, &SHA512_OID),
+        "ps256" => pair(&RSA_PSS_OID, &SHA256_OID),
+        "ps384" => pair(&RSA_PSS_OID, &SHA384_OID),
+        "ps512" => pair(&RSA_PSS_OID, &SHA512_OID),
+        "ed25519" => pair(&ED25519_OID, &SHA512_OID),
+        _ => None,
     }
```
ok-nick, line 192: I would love to see this as an enum, but perhaps out of scope for this PR.

sdk/src/crypto/cose/sign.rs
```diff
@@ -468,11 +475,43 @@ const PAD: &str = "pad";
...
+    raw_signature_size: usize,
+    certs: &[Vec<u8>],
+    has_time_stamp: bool,
+    ocsp_response: Option<&[u8]>,
+) -> usize {
+    /// COSE/JUMBF framing overhead beyond the items accounted for explicitly.
+    const COSE_OVERHEAD: usize = 1024;
+
+    /// Reserve for an RFC 3161 time stamp token in the COSE unprotected header.
+    const TIME_STAMP_RESERVE: usize = 10000;
```
ok-nick, line 495: Hm, can we make these configurable?

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -875,20 +881,5 @@ async fn validate_timestamp_sig_async(
...
-        crate::crypto::raw_signature::validator_for_sig_and_hash_algs(sig_alg, hash_alg)
-    {
-        validator
-            .validate(&sig_val.to_bytes(), tbs, signing_key_der)
-            .map_err(|_| TimeStampError::InvalidData)
-    } else {
-        Err(TimeStampError::UnsupportedAlgorithm)
-    }
+    validate_timestamp_sig(sig_alg, hash_alg, sig_val, tbs, signing_key_der)
 }
```
ok-nick: Can we remove this function since it's not WASM nor async specific anymore?

sdk/src/identity/builder/async_identity_assertion_signer.rs
```diff
@@ -28,31 +27,37 @@ use crate::{
 /// [`Manifest`]: crate::Manifest
 pub struct AsyncIdentityAssertionSigner {
     #[cfg(not(target_arch = "wasm32"))]
-    signer: Box<dyn AsyncRawSigner + Sync + Send>,
+    signer: Box<dyn RawSigner + Sync + Send>,
 
     #[cfg(target_arch = "wasm32")]
-    signer: Box<dyn AsyncRawSigner>,
+    signer: Box<dyn RawSigner>,
```
ok-nick, line 33: Can simplify this into one field with `MaybeSend + MaybeSync`.

sdk/src/identity/x509/async_x509_credential_holder.rs
```diff
@@ -40,31 +41,36 @@ pub struct AsyncX509CredentialHolder(Box<dyn AsyncRawSigner + Send + Sync + 'sta
 /// [`AsyncCredentialHolder`]: crate::identity::builder::AsyncCredentialHolder
 /// [§8.2, X.509 certificates and COSE signatures]: https://cawg.io/identity/1.1-draft/#_x_509_certificates_and_cose_signatures
 #[cfg(target_arch = "wasm32")]
-pub struct AsyncX509CredentialHolder(Box<dyn AsyncRawSigner + 'static>);
+pub struct AsyncX509CredentialHolder {
+    signer: Box<dyn RawSigner + 'static>,
+    cert_chain: Vec<Vec<u8>>,
+}
```
ok-nick, line 47: Same here for `MaybeSend + MaybeSync` (and the other places).

sdk/Cargo.toml
```diff
@@ -35,19 +35,10 @@ fetch_remote_manifests = ["dep:wasi"]
...
-    "dep:ecdsa",
-    "dep:num-bigint-dig",
-    "dep:p256",
-    "dep:p384",
-    "dep:p521",
-    "dep:pkcs1",
-    "dep:rsa",
-    "dep:spki",
-]
+rust_native_crypto = ["c2pa_raw_crypto/rust_native_crypto"]
```
ok-nick, line 41: 👍

sdk/src/crypto/cose/cose_signer.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::{
+    crypto::time_stamp::{AsyncTimeStampProvider, TimeStampProvider},
+    maybe_send_sync::{MaybeSend, MaybeSync},
+};
+
+/// A `CoseSigner` produces a raw signature and supplies the time stamp and OCSP
+/// information needed to assemble a COSE `Sign1` structure.
+///
+/// If an implementation _can_ be asynchronous, prefer [`AsyncCoseSigner`].
+pub(crate) trait CoseSigner: TimeStampProvider {
```
ok-nick, line 34: On a related note, do we need the `RawSigner` trait? Can we remove it and simplify it to just the `Signer` trait?

sdk/src/crypto/cose/sign.rs
scouten-adobe: I'm going to leave this as-is for now. Planning a refactoring of the COSE logic next and that will provide an opportunity to add such things.

sdk/src/identity/builder/async_identity_assertion_signer.rs
scouten-adobe: I'd like to consider that in a separate PR.

sdk/src/crypto/cose/cose_signer.rs
scouten-adobe: `RawSigner` is an abstraction now published by the `c2pa_raw_ crypto` crate. Yes, we need it.

scouten-adobe:
Merging after a Slack :+1: from @ok-nick.

tmathern:
https://github.com/contentauth/c2pa-raw-crypto#cryptography-backend--crate-features states:
```
Enabling neither is allowed (e.g. when only the type definitions are needed, or when signing is delegated to a remote service and no validation is performed); in that case the built-in signer/validator constructors report an error / return None at runtime.
```

Is this correct too for that crate?
That sounds like contradicting contracts?

sdk/src/identity/x509/x509_credential_holder.rs
```diff
@@ -28,18 +27,22 @@ use crate::{
...
     ///
-    /// The [`RawSigner`] implementation actually holds (or has access to)
-    /// the relevant certificates and private key material.
-    ///
-    /// [`RawSigner`]: crate::crypto::raw_signature::RawSigner
-    pub fn from_raw_signer(signer: Box<dyn RawSigner + Sync + Send + 'static>) -> Self {
-        Self(signer)
+    /// [`RawSigner`]: c2pa_raw_crypto::RawSigner
+    pub fn from_raw_signer(
+        signer: Box<dyn RawSigner + Sync + Send + 'static>,
```
tmathern, line 42: RawSigner comes from c2pa_raw_crypto::RawSigner. This means anyone doing custom signers now needs to depend on the new crate too, since type aren't reexported. Is this intended?
This may lead to added dependency + forced lockstep for others to match exactly the dependency here, otherwise there will be errors that 2 different version of the same crate was used.

tmathern: Affects those types: RawSigner, RawSignerError, RawSignatureValidationError, UnknownAlgorithmError, signer_from_private_key.

The RawSigner* ones may be debatable, UnknownAlgorithmError and signer_from_private_key may be more problematic.

sdk/src/crypto/cose/certificate_trust_policy.rs
tmathern: https://github.com/contentauth/c2pa-raw-crypto likely needs similar guards too.

tmathern: Also I expect this may need to have downstream WASM builds updated.

sdk/src/crypto/mod.rs
```diff
@@ -16,9 +16,25 @@
...
-pub mod raw_signature;
 pub mod time_stamp;
+
+/// Parse a PEM certificate chain into DER-encoded certificates, end-entity
+/// first.
+///
+/// Accepts JSON-encoded PEM (with literal `\n` escapes) as well as standard
+/// PEM. This lives here because the raw signer no longer tracks the signing
+/// certificate chain; that is now the SDK's responsibility.
+pub(crate) fn cert_chain_pem_to_der(cert_chain_pem: &[u8]) -> crate::Result<Vec<Vec<u8>>> {
```
tmathern, line 30: Maybe this one should be pub? It would be useful for callers to create the chains. Looks like some constructors ask for that format, so that helper was helpful.

sdk/src/lib.rs
```diff
@@ -243,11 +243,30 @@ pub use assertions::DigitalSourceType;
 #[doc(inline)]
 pub use assertions::Relationship;
 pub use builder::{Builder, BuilderIntent, HashType, ManifestDefinition};
+pub use c2pa_raw_crypto::SigningAlg;
```
tmathern, line 246: Re-exports SigningAlg here, but not the associated error type UnknownAlgorithmError, which I beleive is here only: https://github.com/contentauth/c2pa-raw-crypto/blob/main/src/signing_alg.rs#L52-L53.

Things like "es256".parse::<c2pa::SigningAlg>() will compile, but any downstream code that names the error type, storing it in a struct, writing it in a return type like Result<SigningAlg, UnknownAlgorithmError>, or matching on it now has no path to it through c2pa-rs only. 

Those have to add the dependency to the new crate for that, and keep the lockstep with what c2pa-rs uses here. I'd suggest fixing that to not force down a new dependency for users who are somewhat likely to run into this.

tmathern:
Note on https://github.com/contentauth/c2pa-raw-crypto:
Since openssl/rust_native_crypto are meant to be interchangeable, they should test same behavior on exposed surfaces. For instance, [this seems a quite relaxed check](https://github.com/contentauth/c2pa-raw-crypto/blob/0976b295e5b6d931d96133d09c45c47a60d620bd/src/rust_native/validators/ecdsa_validator.rs#L71), that it may be worth testing openssl behaves the same. So that the backends are truly interchangeable. Or at least the public API used by c2pa-rs should be cross-tested that way.

tmathern:
On https://github.com/contentauth/c2pa-node-v2:
This will break the node SDK as-is. For instance, it had imports of now-gone things: https://github.com/contentauth/c2pa-node-v2/blob/main/src/neon_identity_assertion_signer.rs#L19-L22. There is another example here: https://github.com/contentauth/c2pa-node-v2/blob/main/src/neon_signer.rs#L18-L20.

The node SDK also implemented async related signers: https://github.com/contentauth/c2pa-node-v2/blob/main/src/neon_signer.rs#L332. It also used the RawSigner directly it seems: https://github.com/contentauth/c2pa-node-v2/blob/main/src/neon_signer.rs#L236-L238. I am not sure of the impact/if they can switch to sync. [This error here](https://github.com/contentauth/c2pa-node-v2/blob/main/src/neon_signer.rs#L424-L426) looks mildly worrisome regarding that switch, or at least like a need for a redesign of the neon side. Not totally sure of the implications here, but it's not just a simple switch.

PR #2203, opened 2026-06-03
fix: Exact ingredient redaction URI matching

tmathern:
Follow-up of https://github.com/contentauth/c2pa-rs/pull/2199. 
Exact ingredient redaction URI matching when looking for redactions.

## Checklist

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -1821,10 +1821,13 @@ impl Claim {
 
         // delete assertion or databox
         if assertion_uri.contains(ASSERTION_STORE) {
+            // Extract the last path segment from the URI to match exactly,
+            // avoids substring false-positives.
+            let uri_label = assertion_uri.rsplit('/').next().unwrap_or(assertion_uri);
             if let Some(index) = self
                 .assertion_store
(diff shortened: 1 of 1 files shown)


sdk/src/claim.rs
```diff
@@ -1821,10 +1821,13 @@ impl Claim {
 
         // delete assertion or databox
         if assertion_uri.contains(ASSERTION_STORE) {
+            // Extract the last path segment from the URI to match exactly,
+            // avoids substring false-positives.
```
mauricefisher64: Good catch.  We do have a function that will return the label with this instance identifier if you want to use that

mauricefisher64: let exact_label = Claim::label_with_instance(x.label(), x.instance);

mauricefisher64 (approved):

ok-nick (approved):

ale-adobe (approved):
LGTM, good catch!

sdk/src/claim.rs
tmathern: Updated!

Issue #1764, opened 2026-01-22
Trust model defined in CAWG identity assertion 1.2 is not implemented


scouten-adobe:
### Component

c2pa-rs

### Version

0.76.0

### Platform

Linux

### What happened?

The trust model that was added in the CAWG identity assertion specification as of version 1.2 is not enforced. This is a tracking issue while the team works on this issue.

### What did you expect to happen?

Trust model should be enforced by default.

dcondrey:
Is there a planned path to eventually include process-based evidence? (IMU/EDL data) When the user has an incentive to forge their own provenance this could be evolved from 'who signed this' to 'is the evidence falsifiable'.

mikeargento:
The question from @dcondrey touches on an interesting boundary worth being precise about. As I understand the CAWG 1.2 trust model, it primarily answers "who signed this," rather than the stronger question of whether an artifact could exist without traversing an enforced creation path.

Adding IMU/EDL data as additional signed claims certainly improves provenance richness, but if the signing step remains separable from the sensor capture step, a motivated identity holder could still construct provenance that appears structurally valid.

Curious how the team is thinking about that threat model: is the planned enforcement intended to address cases where the identity holder themselves is adversarial, or is the scope primarily third-party credential validation?

dcondrey:
@mikeargento The CAWG 1.2 trust model is doing what it should, which is credential validation. It's not trying to solve the adversarial identity holder problem, and it shouldn't be. That's a different layer.

That's the problem space I've been working in within the IETF RATS working group ([draft-condrey-rats-pop-protocol](https://datatracker.ietf.org/doc/draft-condrey-rats-pop-protocol/)). The approach collects timestamped, cryptographically chained process evidence (keystroke dynamics, edit deltas, IMU where applicable) during creation and binds it to the output, making the authorship claim verifiable rather than just asserted.

This is intentionally separate from C2PA, but the two are naturally complementary. Identity answers who signed it. Process evidence answers whether the claimed creation path is physically plausible. The manifest architecture is already extensible enough to reference external evidence standards without requiring changes to C2PA itself.

PR #2209, opened 2026-06-04
fix!: Enforce trusted-issuer allow-list for CAWG ICA credentials (CAI-11347)

scouten-adobe:
### Problem

The CAWG identity claims aggregation (ICA) verifier accepted an attacker-controlled self-issued `did:jwk` (or any DID) as its own trust root. Because the issuer's public key was taken directly from the credential, a self-consistent signature was surfaced as a *trusted* creator identity — letting an attacker forge `verifiedIdentities` (e.g. claim any `username`) on a normally-signed asset and have the SDK/CLI accept it as legitimate provenance. (CAI-11347; the underlying gap was tracked as the unimplemented CAI-7980.)

### Fix

ICA issuers must now appear on an explicit, secure-by-default allow-list:

- **New setting** `cawg_trust.trusted_ica_issuers` — an exact-match list of issuer DIDs (any DID method). The default is empty, meaning **no ICA issuer is trusted** unless explicitly configured.
- **Enforcement** — `IcaSignatureVerifier` carries `trusted_issuers` and, after a valid signature, checks the issuer DID (with any fragment stripped) against the list. An off-list issuer logs the failure code `cawg.ica.untrusted_issuer` and withholds `cawg.ica.credential_valid`, per the CAWG spec ("MUST issue the failure code … but MAY continue validation").
- **Scoped to the assertion** — `ValidationResults::validation_state()` explicitly ignores `cawg.ica.untrusted_issuer` in both its `is_valid` and `is_trusted` checks, so an untrusted ICA issuer never changes the enclosing manifest's state. The manifest's trust derives from its own C2PA signer, not from any identity claimed within it.

### Settings threading (Context by value)

`validate_partial_claim` previously read `Context::new()`, which returns pure defaults and ignored all caller configuration. Settings are now threaded by value:

- `validate_partial_claim` takes `&Settings`; `Manifest::from_store` passes its real settings.
- `CawgValidator` carries `Settings` — construct it with `CawgValidator::new(&context)` (or `CawgValidator::default()`).
- Incidental correctness improvement: the X.509 CAWG trust path now also honors the threaded context settings.

A duplicate `cawg.ica.credential_valid` emission was removed so the verifier is the single source of truth for that status.

### ⚠️ Breaking change

`CawgValidator` is no longer a unit struct. External callers using `&CawgValidator {}` must switch to `CawgValidator::new(&context)` or `CawgValidator::default()`.

### Tests

- `did_is_untrusted` / `issuer_trusted_on_allow_list` — direct-verifier unit tests (mirroring the X.509 `untrusted_cert` precedent).
- `ica_issuer_untrusted_does_not_affect_manifest_state` — end-to-end through `Context` / `CawgValidator::new`, asserting the failure is recorded, `credential_valid` is withheld, and `validation_state` stays `Valid`.
- Full `c2pa` lib suite passes (786 passed, 6 pre-existing ignored).

### Docs

`docs/cawg-identity.md` documents the new setting, the secure default, and the assertion-scoped semantics.

## Checklist

Out of scope (tracked separately): ICA issuer revocation (CAI-7993) and `verifiedIdentities` inspection (CAI-7994).


c2pa_c_ffi/src/c_api.rs
```diff
@@ -3537,22 +3537,57 @@ mod tests {
...
         assert!(json_str
             .to_str()
             .unwrap()
             .contains("cawg.ica.credential_valid"));
-        unsafe { c2pa_reader_free(reader) };
+
+        unsafe {
+            c2pa_free(settings as *mut c_void);
+            c2pa_free(context as *mut c_void);
+            c2pa_reader_free(configured_reader);
```
gpeacock, line 3356: c2pa_free should be ok here, unless I'm missing something.

sdk/src/validation_results.rs
```diff
@@ -1105,6 +1136,7 @@ pub mod validation_codes {
             | TIME_OF_SIGNING_INSIDE_VALIDITY
             | INGREDIENT_PROVENANCE_UNKNOWN
             | ASSERTION_DATAHASH_ADDITIONAL_EXCLUSIONS => LogKind::Informational,
+            CAWG_ICA_UNTRUSTED_ISSUER => LogKind::Failure,
```
gpeacock: Not necessary to add this here since all other status codes are considered LogKind::Failure. I think we would need add all the possible failure status codes otherwise.

sdk/tests/test_builder.rs
```diff
@@ -166,7 +166,8 @@ fn test_builder_cyclic_ingredient() -> Result<()> {
         let mut reader =
             Reader::from_context(no_verify_context).with_stream(format, cyclic_ingredient)?;
         // Ideally we'd use a sync path for this. There are limitations for tokio on WASM.
-        tokio::runtime::Runtime::new()?.block_on(reader.post_validate_async(&CawgValidator {}))?;
+        tokio::runtime::Runtime::new()?
```
gpeacock, line 171: This should not need async anymore. We can just call reader.post_validate()

sdk/src/validation_results.rs
```diff
@@ -220,6 +220,18 @@ impl ValidationResults {
...
             //
+            // A failure to trust the issuer of a CAWG identity claims aggregation
+            // (ICA) credential is scoped to that individual identity assertion. It
+            // is recorded as a failure for that assertion but, like an untrusted
+            // signing certificate, does not on its own invalidate the enclosing
+            // manifest. Unlike an untrusted signing certificate, it also does not
+            // prevent the manifest from being considered trusted, since the
+            // manifest's trust derives from its own C2PA signer, not from any
+            // identity claimed within it.
+            let manifest_trust_unaffected = |status: &ValidationStatus| {
```
gpeacock: This is really unfortunate, why isn't this just informational? It seems very odd to force an error code that then we have to manually ignore. We will need to document that all errors except for this one are errors!

sdk/src/settings/mod.rs
```diff
@@ -85,6 +85,19 @@ pub struct Trust {
...
+    /// Each entry is a full DID string (any DID method) that is compared, after
+    /// stripping any fragment, against the `issuer` of an ICA verifiable
+    /// credential. An issuer that is not present on this list is reported with
+    /// the failure code `cawg.ica.untrusted_issuer` for that identity assertion.
+    ///
+    /// The default value is empty, meaning that NO ICA issuer is trusted. This
+    /// is a deliberate secure default: a self-issued `did:jwk` (or any other
+    /// issuer) is not trustworthy simply because its signature is
+    /// self-consistent. Populate this list with the DIDs of issuers you trust.
+    pub trusted_ica_issuers: Option<Vec<String>>,
```
ok-nick, line 104: Maybe it's time to create a separate `CawgTrust` struct so it's not mixed in with C2PA.

sdk/src/validation_results.rs
ok-nick: I agree with Gavin, I think this would make more sense as an informational code at least until we are able to separate validation results for identity assertions.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -3537,22 +3537,57 @@ mod tests {
...
+        let builder = unsafe { c2pa_context_builder_new() };
+        assert!(!builder.is_null());
+
+        let settings = unsafe { c2pa_settings_new() };
+        assert!(!settings.is_null());
+
+        let json = CString::new(
+            r#"{"cawg_trust": {"trusted_ica_issuers": ["did:jwk:eyJhbGciOiJFZERTQSIsImt0eSI6Ik9LUCIsImNydiI6IkVkMjU1MTkiLCJ4IjoiTXA1LTBlODNuTmdRaGRoQlc4UnNoa2p5OTBzYTFBOUpJemtJdGNEcUN1SSJ9"]}}"#,
+        )
+        .unwrap();
```
ok-nick: Why do we need to set this explicitly? Why is it not inherited from default settings?

sdk/src/identity/validator.rs
```diff
@@ -20,12 +20,33 @@ use serde_json::Value;
...
+impl CawgValidator {
+    /// Create a `CawgValidator` whose trust configuration is drawn from the
+    /// provided [`Context`].
+    pub fn new(context: &Context) -> Self {
+        Self {
+            settings: context.settings().clone(),
+        }
+    }
+}
+
```
ok-nick: Settings can be expensive to clone, especially if trust lists are stored in them. Maybe consider storing the context instead since we will need the HTTP resolvers in the future here anyways.

c2pa_c_ffi/src/c_api.rs
scouten-adobe: The \`#[cfg(test)]\` default in `Trust::default()` that pre-trusts the bundled fixture issuers only applies to the **c2pa crate's own** unit tests. When c2pa is compiled as a dependency (as it is here for c2pa-c-ffi), it builds *without* `cfg(test)`, so the secure production default applies (empty allow-list → no ICA issuer trusted). The test therefore has to configure the issuer explicitly to obtain `cawg.ica.credential_valid`.

sdk/tests/test_builder.rs
scouten-adobe: `CawgValidator` only implements `AsyncPostValidator` (this is true on `main` as well — ICA validation can require network I/O for `did:web` resolution), so there is no sync `post_validate()` path to call here. Switching to sync would mean adding a new `PostValidator` impl, which is out of scope for this PR, so I left it as the async `block_on` form that `main` uses.

sdk/src/validation_results.rs
scouten-adobe: Done in 2d3ac0be. `cawg.ica.untrusted_issuer` is now recorded via `.informational()`, and `validation_state()` no longer special-cases it — an informational code never appears among the failures, so the manual `manifest_trust_unaffected` carve-out is gone. `cawg.ica.credential_valid` is still withheld for an untrusted issuer, so the security behavior is unchanged.

sdk/src/validation_results.rs
scouten-adobe: Addressed in 2d3ac0be — `cawg.ica.untrusted_issuer` is now classified as `LogKind::Informational` (folded into the informational arm), so the explicit `=> LogKind::Failure` line is gone.

sdk/src/identity/validator.rs
scouten-adobe: Resolved during the merge with `main`: `main` had already refactored `CawgValidator`/`IcaSignatureVerifier` to carry `&Context` (the CAI-7980 plumbing from the SSRF fix #2168). I adopted that shape, so the trusted-issuer list is now read from `context.settings().cawg_trust.trusted_ica_issuers` rather than from a cloned `Settings`.

sdk/src/settings/mod.rs
scouten-adobe: Agreed. To keep this security fix focused I've deferred it to a follow-up — tracked as CAI-12709, and referenced in a `TO DO (CAI-12709)` comment on the `trusted_ica_issuers` field. That ticket covers introducing a dedicated `CawgTrust` struct and removing `trusted_ica_issuers` from the shared C2PA `Trust`.

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
```diff
@@ -545,6 +575,60 @@ impl<'a> IcaSignatureVerifier<'a> {
...
+    /// the enclosing C2PA manifest invalid or untrusted (see
+    /// [`ValidationResults::validation_state`]).
+    ///
+    /// [`Context`]: crate::Context
+    /// [`ValidationResults::validation_state`]: crate::validation_results::ValidationResults::validation_state
+    fn check_issuer_trust(
+        &self,
+        ica_credential: &IcaCredential,
+        status_tracker: &mut StatusTracker,
+        ok: &mut bool,
```
ok-nick, line 603: How come we don't return the boolean? Why pass it as a mutable ref?

ok-nick (approved):

c2pa_c_ffi/src/c_api.rs
```diff
@@ -3308,22 +3308,57 @@ mod tests {
     #[test]
     #[allow(deprecated)]
     fn test_c2pa_reader_from_stream_cawg() {
+        // This fixture's identity claims aggregation (ICA) credential is signed
+        // by a `did:jwk` issuer. ICA issuers are untrusted by default, so we must
+        // add that issuer to `cawg_trust.trusted_ica_issuers` for the credential
+        // to be reported as valid.
+        let builder = unsafe { c2pa_context_builder_new() };
+        assert!(!builder.is_null());
```
gpeacock: I know our tests are always checking but these c2pa_.._new() methods cannot return null. It does no harm, but it also does no good.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -3308,22 +3308,57 @@ mod tests {
...
+        let settings = unsafe { c2pa_settings_new() };
+        assert!(!settings.is_null());
+
+        let json = CString::new(
+            r#"{"cawg_trust": {"trusted_ica_issuers": ["did:jwk:eyJhbGciOiJFZERTQSIsImt0eSI6Ik9LUCIsImNydiI6IkVkMjU1MTkiLCJ4IjoiTXA1LTBlODNuTmdRaGRoQlc4UnNoa2p5OTBzYTFBOUpJemtJdGNEcUN1SSJ9"]}}"#,
+        )
+        .unwrap();
+        let format = CString::new("json").unwrap();
+        let result =
+            unsafe { c2pa_settings_update_from_string(settings, json.as_ptr(), format.as_ptr()) };
```
gpeacock: When setting a single value like this, the simpler way is to use a path:
c2pa_settings_set(settings, path, value)

sdk/src/identity/tests/claim_aggregation/validation.rs
```diff
@@ -900,16 +902,125 @@ async fn did_doc_without_assertion_method() {
...
 
-// #[test]
-// #[ignore]
-// fn did_is_untrusted() {
-//     // The validator SHALL verify that the issuer’s DID is present or can be
-//     // traced to its preconfigured list of trustable entities. If the issuer
-// is     // not verifiably trusted, the validator MUST issue the failure code
-//     // `cawg.ica.untrusted_issuer` but MAY continue validation.
+#[c2pa_test_async]
+async fn did_is_untrusted() {
```
gpeacock, line 906: Everything should work sync or async now. Is there reason these tests need to be async?

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
scouten-adobe: This follows the existing pattern of the sibling validation helpers in this file — `check_content_type` and `cross_check_signer_payload` both take `ok: &mut bool` and return `Result<()>`. The `Result` is used for `?`-propagating the hard errors (e.g. a malformed issuer DID from `Did::new`), while `ok` accumulates the "valid but not fatal" state that `check_signature` folds together before deciding whether to emit `cawg.ica.credential_valid`. Returning a bool here would make this one helper diverge from the others, so I kept it consistent. I'm happy to revisit the return-value style across all of these helpers, but I'd prefer to do that in a separate PR — let me know if you'd like me to file an issue to track it.

c2pa_c_ffi/src/c_api.rs
scouten-adobe: Good point — removed the `is_null()` asserts on the infallible `c2pa_context_builder_new()` / `c2pa_settings_new()` constructors in 70473836.

c2pa_c_ffi/src/c_api.rs
scouten-adobe: Done in 70473836 — switched to `c2pa_settings_set_value(settings, path, value)` with `path = "cawg_trust.trusted_ica_issuers"`, which accepts the string array directly, so the JSON blob + `c2pa_settings_update_from_string` is gone.

sdk/src/identity/tests/claim_aggregation/validation.rs
scouten-adobe: These tests exercise the lower-level identity path rather than `Reader::post_validate`: they call `IdentityAssertion::validate(..).await` and `read_manifest(..).await`, both of which are async-only (`validate` ultimately drives ICA validation, which can require network I/O for `did:web` resolution). There's no sync equivalent at this level, and every sibling test in this file is `#[c2pa_test_async]` for the same reason, so I kept these consistent with them.

sdk/src/identity/claim_aggregation/ica_signature_verifier.rs
scouten-adobe: Filed #2272 to track migrating these helpers to the return-value style in a follow-up PR.

sdk/src/validation_results.rs
```diff
@@ -233,6 +245,7 @@ impl ValidationResults {
                 && (active_manifest.failure().is_empty()
                     || active_manifest.failure().iter().all(|status| {
                         status.code() == validation_status::SIGNING_CREDENTIAL_UNTRUSTED
+                            || manifest_trust_unaffected(status)
```
gpeacock: It probably makes more sense to just have a list of status codes that can only impact trust here. But I can rework this later. The logic is just kind of confusing this way, and I suspect we will end up with more exceptions over time.

sdk/src/validation_results.rs
gpeacock: except that it looks like it is still here

sdk/tests/test_builder.rs
gpeacock: There's actually no need to call post_validate at all anymore, it is already done in reader. the problem is that thsi is the only remaining call to post_validate, so if we get rid of it we have no other tests.

ok-nick (approved):

gpeacock (approved):

caiopensrc:
- #2310

lucatrevisanii:
Hi! First off, thanks for this fix. We spent a good while confused about why our `cawg_trust` trust anchors never reached the CAWG X.509 check, and the change here (reading settings from the caller's `context` instead of a fresh `Context::new()`) turned out to be exactly what we needed.

Is there a rough timeline for a release that includes it? As of 0.90.15 the published crate still has the old behavior, so the identity assertion reports `signingCredential.untrusted` even when its certificate chains to a configured anchor. For now we're carrying a vendored 0.90.4 with just this change cherry-picked, and we'd be happy to get back to a plain crates.io pin.

Thanks again for the work on this. In case it helps triage, this also relates to #1764 and #2247.

PR #2217, opened 2026-06-09
ci: Use latest `release-plz` and add a dry run job on release PRs

### LINKED ISSUE #2299: New option request: Always trigger release of downstream crates when a dependent crate in same repo will be published
## Motivations

In the [content authenticity project](https://github.com/contentauth/c2pa-rs/), we have a core library (`c2pa`) which is reflected in two downstream crates (`c2pa-c-ffi` and `c2patool`). All three of these are hosted in the same repo.

We would like to always release the downstream crates whenever the core library is updated.

Currently release-plz has a hard policy of not releasing if there are no changes to the source code of those crates, which is at odds with the use case stated above.

- Would you like to implement this feature? _yes, if agreed to_

## Solution

I'd like to see a new option which would override the check for direct changes to the source code of a crate if a downstream dependency in the same repo is being updated/re-published. In other words, if `Cargo.toml`'s `dependencies` section is being updated, the `version` field should also be updated and a corresponding changelog entry should be generated as part of the `release-pr` workflow.

## Alternatives

See https://github.com/contentauth/c2pa-rs/pull/1222 for a hacky workaround to the problem we're trying to solve.

ok-nick:
Updates `release-plz` and cleans up the patches used to fix https://github.com/release-plz/release-plz/issues/2299 which is now resolved. This should solve the issue that forced us to sometimes create noop PRs.

There is also now a dry run of `release-plz` on release PRs so we can catch issues before merging the release PR. `c2pa` dependency versions are also now specified in the workspace so there is less churn in the diff.

I've tested updating and releasing locally with a dry run although when this is merged I'll create a noop release just to be sure.

### DIFF
--- a/c2pa_c_ffi/Cargo.toml
+++ b/c2pa_c_ffi/Cargo.toml
@@ -28,7 +28,7 @@ http = ["c2pa/http_reqwest", "c2pa/http_reqwest_blocking"]
 add_thumbnails = ["c2pa/add_thumbnails"]
 
 [dependencies]
-c2pa = { path = "../sdk", version = "0.86.0", default-features = false, features = [
+c2pa = { workspace = true, features = [
     "fetch_remote_manifests",
     "file_io",
     "pdf",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -24,7 +24,7 @@ networking = [
 [dependencies]
 anyhow = "1.0"
 atree = "0.5.4"
-c2pa = { path = "../sdk", version = "0.86.0", features = [
+c2pa = { workspace = true, features = [
     "fetch_remote_manifests",
     "file_io",
     "add_thumbnails",
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,43 +33,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_PLZ_TOKEN }}
           CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_SECRET }}
-        run: |
-          echo "-- Running release-plz release-pr --"
-          release_pr_output=$(release-plz release-pr \
-            --git-token "${GITHUB_TOKEN}" \
-            --repo-url "https://github.com/${GITHUB_REPOSITORY}" \
-            -o json)
-          echo "release_pr_output: $release_pr_output"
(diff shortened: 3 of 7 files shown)


c2pa_c_ffi/Cargo.toml
```diff
@@ -28,7 +28,7 @@ http = ["c2pa/http_reqwest", "c2pa/http_reqwest_blocking"]
 add_thumbnails = ["c2pa/add_thumbnails"]
 
 [dependencies]
-c2pa = { path = "../sdk", version = "0.86.1", default-features = false, features = [
+c2pa = { workspace = true, features = [
```
scouten-adobe, line 31: Please vet this approach in a side project before landing this. My memory is that crates.io has a stricter policy on version references than cargo and I've been burned by trying something like this before.

Reason I'm concerned is that I believe the content that `cargo publish` sends to crates.io is the _subset_ of the repo that contains only the source code for _this_ crate. If I remember correctly, the workspace shell gets lost in that upload but the source code doesn't get rewritten to remove the `workspace = true` reference, so the crate becomes unbuildable.

cli/Cargo.toml
```diff
@@ -24,7 +24,7 @@ networking = [
 [dependencies]
 anyhow = "1.0"
 atree = "0.5.4"
-c2pa = { path = "../sdk", version = "0.86.1", features = [
+c2pa = { workspace = true, features = [
```
scouten-adobe, line 27: Same concern as in `c2pa_c_ffi`.

scouten-adobe (changes requested):

c2pa_c_ffi/Cargo.toml
ok-nick: Ah that's quite obscure. Let me make a test project then and do a real upload.

ok-nick: Crate A: https://crates.io/crates/test-crate-90543123487
Crate B: https://crates.io/crates/test-crate-dep-90543123487
Workspace: [test-publish.zip](https://github.com/user-attachments/files/28763356/test-publish.zip)

Crate A depends on crate B and uses the same structure with `workspace = true`.

I also tested by adding an independent crate outside of the workspace that depends on crate A via `crates.io` and had no issues.

scouten-adobe (approved):
@ok-nick has demonstrated that this works in a side project. Approving.

PR #2225, opened 2026-06-10
fix: Improve ingredientMismatch error messages

mpgopala:
When `verify_actions` fires `assertion.action.ingredientMismatch`, the explanation now includes the 0-based action index, the action type, each referenced ingredient URL, and the actual relationship found on that ingredient (vs. the expected one).

Before:
  "action must have valid ingredient with ComponentOf relationship"

After:
  "action[3] ('c2pa.placed') must have valid ingredient with componentOf
   relationship; ingredient(s): ['self#jumbf=c2pa.assertions/c2pa.ingredient.v3'
   (title='A.jpg', relationship='parentOf')]"

Changes:
- Add `enumerate()` to the inner action loop in `verify_actions` to track the 0-based action index
- Add `Claim::ingredient_mismatch_msg` helper that resolves each ingredient URL back to its assertion, reads the actual `Relationship`, and formats a diagnostic string
- Apply the richer message at all three mismatch sites: `c2pa.opened` (parentOf), `c2pa.placed`/`c2pa.removed` (componentOf), and `c2pa.transcoded`/`c2pa.repackaged` (parentOf)

Validation status codes and public API are unchanged; only the `explanation` string improves.

Fixes: CAI-12311

## Checklist


docs/cai-12311-plan.md
```diff
@@ -0,0 +1,193 @@
+# CAI-12311: Better error messages for assertion.action.ingredientMismatch
```
tmathern: If you keep that, rename the file to something that explains what the doc is for

docs/cai-12311-plan.md
```diff
@@ -0,0 +1,193 @@
...
+| Actual relationship | Do a secondary lookup into `claim.ingredient_assertions()` after `found_good != 1` |
+| Ingredient title | `Ingredient::from_assertion(...).title` |
+
+The `Ingredient` struct (defined in `sdk/src/assertions/ingredient.rs:57`) exposes:
+- `pub title: Option<String>`
+- `pub relationship: Relationship`  — enum variants `ParentOf`, `ComponentOf`, `InputTo`
+
+The `Relationship` enum derives `Debug` but has no `Display`. Its serde rename strings (`"parentOf"`, `"componentOf"`, `"inputTo"`) are the canonical user-facing names.
+
+## Proposed changes
```
tmathern: If you keep file + code changes, they're not proposed changes anymore

sdk/src/claim.rs
```diff
@@ -2005,6 +2005,66 @@ impl Claim {
...
+    // action's index, type, each referenced ingredient URL, and the actual relationship
+    // found (vs. the expected one).
+    fn ingredient_mismatch_msg(
+        claim: &Claim,
+        action_index: usize,
+        action_type: &str,
+        params: &assertions::ActionParameters,
+        expected_relationship: &str,
+    ) -> String {
+        use crate::jumbf::labels::assertion_label_from_uri;
```
tmathern: Imports on top of the file please?

sdk/src/claim.rs
```diff
@@ -2005,6 +2005,66 @@ impl Claim {
...
+        };
+
+        let details: Vec<String> = urls
+            .iter()
+            .map(|url| {
+                let found = assertion_label_from_uri(url).and_then(|target| {
+                    claim.ingredient_assertions().iter().find_map(|i| {
+                        if i.label() == target {
+                            Ingredient::from_assertion(i.assertion()).ok().map(|ing| {
+                                let rel = match &ing.relationship {
```
tmathern: You could write a string conversion for that? Might comme in handy in the future, if it something like that does not already exist in the codebase

tmathern (changes requested):
I like the improved logs!
But some minor changes to do still.

docs/cai-12311-plan.md
mpgopala: Sorry. Pushed by mistake. Will delete that.

sdk/src/claim.rs
mpgopala: The conversion is very specific to the error string. I could add a utility. But that would have to be generic without taking the filter like this one. We should probably handle it when needed?

sdk/src/claim.rs
mpgopala: Done

docs/cai-12311-plan.md
mpgopala: The file is deleted.

sdk/src/claim.rs
tmathern: I was asking because usually those enums are serializable, so they flowlessly can be logged/printed, and then you don't need to do matching like that.

mpgopala: I misunderstood your request. Added as_str to Relationship now. I thought you wanted such a utility for the entire `ingredient_assertions`.

tmathern (approved):

tmathern (approved):

sdk/src/assertions/ingredient.rs
```diff
@@ -52,6 +52,16 @@ pub enum Relationship {
     InputTo,
 }
 
+impl Relationship {
```
tmathern, line 55: I fear this can drift from what serde serialization does, especially since hand baked. At least maybe add a test verifying they don't drift.

mpgopala: Done.

PR #2232, opened 2026-06-11
chore: Add breaking-changes label automation to verify dependent crates

scouten-adobe:
## What

Adds a `breaking-changes` label that, when applied to a c2pa-rs PR, fans out to the org's downstream binding repos and verifies each still builds against the PR's branch — gating the PR on the results as a required status check.

## How it works

- **`.github/workflows/breaking-changes.yml`** (orchestrator + aggregator):
  - On every PR it posts a single always-present required context, **`breaking-changes-gate`** (success when the label is absent, so unlabeled PRs never hang).
  - When a same-repo PR carries the label, the gate goes pending and a `repository_dispatch` (`c2pa-rs-breaking-change`) fans out to each dependent in `DEPENDENTS`. If `BINARY_DEPENDENTS` is non-empty, it also triggers one branch build of `library-release.yml` (via `CROSS_ORG_PR_TOKEN`, since `GITHUB_TOKEN` can't start a run).
  - A `status`-triggered job aggregates the per-dependent `breaking-changes / <repo>` results back into the gate.
  - Uses `pull_request_target` (never checks out PR code) so the gate can be posted on fork PRs too; fork PRs pass-through with a note.

- **Receivers** under `docs/breaking-change-automation/receivers/` — one per dependent, tailored to how each consumes c2pa-rs:
  - `c2pa-node-v2`, `c2pa-js` — rewrite the Cargo `c2pa` dep to a git+branch source, build, open a draft PR (Variant A).
  - `c2pa-python`, `c2pa-cpp` — build from a local c2pa-rs checkout, status-only.
  - `c2pa-ios`, `c2pa-android` — download the branch's `library-release.yml` artifacts, stage them, build, status-only. **Unvalidated** (need a real run + runners).

- **`library-release.yml`**: `release` job gated to `c2pa-v*` tags (so branch dispatch produces artifacts without a release), plus a `verify_sha` input + `run-name` so receivers can locate the branch build.

## Required follow-up before this is live

1. ✅ **Done** — org secret **`CROSS_ORG_PR_TOKEN`** is live (Contents + Pull requests write on the dependents; Commit statuses + Actions write on c2pa-rs).
2. Add **`breaking-changes-gate`** as a required status check on `main` / `v1_api`. Do **not** require the per-dependent contexts.
3. Merge the receiver PRs, then add each live repo to `DEPENDENTS` (and `BINARY_DEPENDENTS` for ios/android). See the README's "To activate iOS/Android".

## Receiver PRs

- Source-buildable (ready): contentauth/c2pa-node-v2#70, contentauth/c2pa-js#131, contentauth/c2pa-python#284, contentauth/c2pa-cpp#234
- Prebuilt-binary (unvalidated drafts): contentauth/c2pa-ios#116, contentauth/c2pa-android#92

### DIFF
--- a/docs/breaking-change-automation/README.md
+++ b/docs/breaking-change-automation/README.md
@@ -80,7 +80,8 @@ ones where they exist:
 | `c2pa-js` | [`receivers/c2pa-js.yml`](./receivers/c2pa-js.yml) | Variant A (git-branch dep, builds wasm), opens a draft PR |
 | `c2pa-python` | [`receivers/c2pa-python.yml`](./receivers/c2pa-python.yml) | `build-from-source` via `C2PA_RS_PATH`, status-only |
 | `c2pa-cpp` | [`receivers/c2pa-cpp.yml`](./receivers/c2pa-cpp.yml) | `build-from-source` via `C2PA_BUILD_FROM_SOURCE`, status-only |
-| `c2pa-ios`, `c2pa-android` | start from [`test-c2pa-rs-branch.yml`](./test-c2pa-rs-branch.yml) (Variant B) | need the artifact bridge — see Phase 2 |
+| `c2pa-ios` | [`receivers/c2pa-ios.yml`](./receivers/c2pa-ios.yml) | downloads branch artifacts → `C2PA_LOCAL_ARTIFACTS`, status-only (unvalidated) |
+| `c2pa-android` | [`receivers/c2pa-android.yml`](./receivers/c2pa-android.yml) | stages branch `.so` into `jniLibs`, `assembleRelease`, status-only (unvalidated) |
 
 When adapting the generic template, edit the two marked sections:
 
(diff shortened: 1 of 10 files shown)


docs/breaking-change-automation/README.md
```diff
@@ -0,0 +1,180 @@
+# Breaking-change verification across dependent crates
+
+When a c2pa-rs change breaks API/ABI compatibility, the downstream binding repos
+may need code changes before the c2pa-rs PR can land. This automation surfaces
+that *before* merge: apply the **`breaking-changes`** label to a c2pa-rs PR and
+every dependent repo is built against the PR's branch, with the results gating
+the PR as a required status check.
```
ok-nick, line 7: We should also consider running this suite on every release. That may be planned already, though it would be good to call out. Sometimes things don't break API compatibility but they do break test cases/functionality in downstream SDKs that we can't predict. Same idea when the build system changes.

It may also be worth considering requiring this as as status check before finally merging. If we know it doesn't affect anything (such as not a feat: , fix: , build: PR, or just in general) it can be explicitly overridden.

docs/breaking-change-automation/README.md
```diff
@@ -0,0 +1,180 @@
...
+  against the new commit. Removing the label passes the gate again.
+
+The orchestrator runs on **`pull_request_target`** rather than `pull_request`.
+That's required so the gate status can be posted on fork PRs too — under
+`pull_request` the `GITHUB_TOKEN` is read-only for forks, which would leave the
+required gate context unset and block every external-contributor PR. The
+fan-out job is safe under `pull_request_target` because it never checks out or
+executes PR code; it only reads trusted event metadata and posts statuses /
+dispatches. **Do not add a checkout of the PR head to that workflow.**
+
```
ok-nick, line 45: I think we just need to be careful not to run CI on PRs that edit this workflow. I've seen recommendations to use `workflow_run` to ensure any functionality accessing secrets is gated on the main branch rather than the fork.

docs/breaking-change-automation/README.md
```diff
@@ -0,0 +1,180 @@
...
+c2pa-rs and all six dependent repos. (Rename throughout if you prefer a different
+name — it appears in `breaking-changes.yml` and the receiver template.)
+
+### 3. Install the receiver in each dependent repo
+
+Each dependent gets `.github/workflows/test-c2pa-rs-branch.yml`. Because the
+repos consume c2pa-rs differently, the receivers differ — use the ready-made
+ones where they exist:
+
+| Dependent | Receiver to install | Notes |
```
ok-nick, line 77: I wonder if we can move the build logic to the other repo, maybe as a reusable workflow we call or a makefile that way we don't need custom logic here for each SDK.

docs/breaking-change-automation/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+Only `c2pa-ios` and `c2pa-android` truly need prebuilt native libraries (the
+other four build from source). They consume the binaries that c2pa-rs publishes
+from [`library-release.yml`](../../.github/workflows/library-release.yml).
+
+**The artifact bridge is in place.** That workflow already uploads per-target
+workflow artifacts (`release-artifacts-<os>-<target>`, containing the same
+`c2pa-v*-<target>.zip` files as the release assets) on every run, and its
+`release` job is gated to `c2pa-v*` tag pushes — so dispatching it against a
+branch produces downloadable artifacts without cutting a release:
```
ok-nick, line 125: This sounds cool, so we will have all the artifacts available directly in c2pa-rs?

ok-nick (commented):
Excited to see this, much needed!

scouten-adobe:
Withdrawing for now pending some conversations about our release process.

Issue #2247, opened 2026-06-24
Distinguishing C2PA Untrusted and CAWG Untrusted


byh-trufo:
### Component

c2pa-rs

### Version

v0.89.0

### Platform

Linux

### What happened?

Currently, when the CAWG identity signature or the C2PA claim signature do not trace back to a configured trust list, a `signingCredential.untrusted` is emitted. Many downstream validators may inadvertently fail the entire C2PA manifest with "Validation Failed" if the CAWG identity is not trusted.

This has been a source of confusion for a number of companies so far.

For example, `c2pa-conformance-tool` de-duplicates failures by the `c2pa-rs` error code, and untrusted CAWG is treated in the same way as untrusted C2PA.

### What did you expect to happen?

Ideally, C2PA and CAWG validation should be independent -- especially since CAWG is not designed to have a single "correct" trust list config, so CAWG validation will oftentimes fail even though both the C2PA content and CAWG identity components are both fine.

I had originally prepared a PR in `c2pa-conformance-tool` to distinguish C2PA and CAWG failures. However, before diving into implementation, I wanted to check if the more proper solution is to distinguish the failures at the `c2pa-rs` error code level, before it is too late to do so.

Perhaps, we can use `identityCredential.untrusted` or `cawg.identity.untrusted` when CAWG fails the trust list check?

scouten:
:white_check_mark: Jira issue https://jira.corp.adobe.com/browse/CAI-12718 is successfully created for this GitHub issue.

n8fr8:
I can't seem to get the c2patool to trust my Trufo CAWG OV or SSL.COM CAWG IV certificate at all. It is always just "valid", even if I ensure I am adding the Trufo and SSL Root CA Certs into my "--trust_anchors" file.

Would this fix address that as well?

byh-trufo:
The most recent commit in the fork should address the CAWG trust list setup issue!

mauricefisher64:
This issue is address in PR: https://github.com/contentauth/c2pa-rs/pull/2545

PR #2248, opened 2026-06-24
feat: report CAWG identity trust with cawg.identity.* status codes

byh-trufo:
Addresses Issue: https://github.com/contentauth/c2pa-rs/issues/2247.

Distinguish an untrusted (or trusted) CAWG creator-identity credential from the C2PA claim signer by emitting cawg.identity.untrusted / cawg.identity.trusted instead of the signingCredential.* codes.

- Add CredentialKind and a non-breaking Verifier::with_credential_kind builder so the shared COSE trust check selects the right status namespace; CAWG paths bind CredentialKind::CawgIdentity.
- Add CAWG_IDENTITY_TRUSTED/UNTRUSTED constants and classify them in log_kind; treat cawg.identity.untrusted like signingCredential.untrusted for validity (Valid, never Trusted) via is_untrusted_failure.

Implements cawg.identity.trusted (ref CAI-7980).

## Checklist

### DIFF
--- a/sdk/src/identity/identity_assertion/assertion.rs
+++ b/sdk/src/identity/identity_assertion/assertion.rs
@@ -306,7 +306,7 @@ impl IdentityAssertion {
         status_tracker: &mut StatusTracker,
         context: &Context,
     ) -> Result<serde_json::Value, ValidationError<String>> {
-        let settings = Context::new().settings().clone();
+        let settings = context.settings().clone();
         self.check_padding(status_tracker)?;
 
         self.signer_payload
--- a/sdk/src/identity/identity_assertion/assertion.rs
+++ b/sdk/src/identity/identity_assertion/assertion.rs
@@ -23,7 +23,7 @@ use serde_bytes::ByteBuf;
 
 use crate::{
     crypto::{
-        cose::{parse_cose_sign1, CertificateTrustPolicy, CoseError, Verifier},
+        cose::{parse_cose_sign1, CertificateTrustPolicy, CoseError, CredentialKind, Verifier},
         raw_signature::RawSignatureValidationError,
     },
     dynamic_assertion::PartialClaim,
@@ -351,8 +351,12 @@ impl IdentityAssertion {
                 parse_cose_sign1(&self.signature, &signer_payload_cbor, status_tracker)
                     .map_err(|e| ValidationError::SignatureError(e.to_string()))?;
 
+            // Verify as a CAWG creator-identity credential so trust results are
+            // reported with the `cawg.identity.*` status codes rather than the
+            // C2PA claim-signer `signingCredential.*` codes.
+            let cawg_verifier = cose_verifier.with_credential_kind(CredentialKind::CawgIdentity);
             let cert_info = if _sync {
-                cose_verifier.verify_signature(
+                cawg_verifier.verify_signature(
                     &self.signature,
                     &signer_payload_cbor,
                     &[],
@@ -360,7 +364,7 @@ impl IdentityAssertion {
                     status_tracker,
                 )
             } else {
-                cose_verifier
+                cawg_verifier
                     .verify_signature_async(
                         &self.signature,
                         &signer_payload_cbor,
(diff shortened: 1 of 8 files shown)


scouten-adobe:
FYI we probably won't be able to merge this PR until CAWG ratifies the 1.3 spec that contains these changes (hoping that occurs in July or early August).

sdk/src/validation_results.rs
```diff
@@ -27,6 +27,23 @@ use crate::{
     validation_status::{self, log_kind, ValidationStatus},
 };
 
+/// Returns `true` if `code` is an untrusted-credential failure that does not, by
+/// itself, invalidate a manifest.
+///
+/// This tolerates both the C2PA claim signer ([`SIGNING_CREDENTIAL_UNTRUSTED`])
```
scouten-adobe, line 33: I don't that this is the desired outcome: If a C2PA manifest is entirely valid according to C2PA rules but the identity assertion is invalid according to CAWG rules, I _think_ we will want to report the C2PA status as trusted, but that does mean a CAWG-interested consumer would need to dig deeper to learn that the CAWG identity assertion is untrusted or invalid. (FWIW it's not common, but entirely possible that there are _multiple_ CAWG identity assertions representing multiple named actors; each of those could be valid or invalid independently from each other.)

scouten-adobe (changes requested):
Two things blocking us from merging:

* We need to discuss amongst ourselves the question I raised about whether a CAWG assertion can impact the C2PA validation state of the overall manifest. (Note that Adobe is closed next week for a company-wide summer holiday, so this likely won't happen until we return.)
* CAWG needs to ratify the 1.3 draft spec that introduces these new validation status codes. I hope to submit that through the CAWG/DIF process in July or early August.

Once those issues are cleared, I'll be eager to merge this. Until then I'm using the "Request changes" status to prevent merging prematurely.

sdk/src/validation_results.rs
byh-trufo: In this case, would we want the `ValidationState.is_valid/is_trusted` result to ignore any error codes prefixed with `cawg.*`? This would be a slightly larger change, but it makes sense to me as well.

Increasing the scope a bit, would we then want to introduce a `CawgValidationState.is_valid/is_trusted` result (potentially per CAWG identity assertion)? This would probably require a crjson addendum from CAWG as well.

scouten-adobe: Yes, those sound like good changes IMHO (and, ideally, per identity assertion).

sdk/src/identity/identity_assertion/assertion.rs
Ignasio-Qualabs: I was having the same issue so tested with this branch. This line is throwing the context away and creating a new one from scratch.
I think this should be let settings = context.settings().clone()

n8fr8: Regarding " caller-configured CAWG trust settings (trust_anchors, user_anchors, trust_config, allowed_list, verify_trust_list)" are these only available in the yaml/json settings or can they be set via cli arguments? Do you have an example of how to specify CAWG specific trust settings that work with this PR?

Ignasio-Qualabs: No they are not exposed as CLI flags. they live in the Settings struct and are loaded via JSON, TOML, or programmatically through the SDK.

The cawg_trust block takes the same shape as the existing trust block:

{
  "cawg_trust": {
    "verify_trust_list": true,
    "trust_anchors": "<PEM bundle of your trusted CAs>",
    "trust_config": "1.3.6.1.5.5.7.3.4\n1.3.6.1.5.5.7.3.36\n"
  }
}

trust_anchors is a PEM bundle of the CA certificates you want to accept as trust roots for CAWG identity assertions.

n8fr8: Thanks. Missed that separate cawg_trust{} block. Appreciate it!

n8fr8: Everything working here with your custom branch and these new settings!

 {
          "code": "cawg.identity.trusted",
          "url": "self#jumbf=/c2pa/urn:c2pa:765db996-108d-492e-9851-b3e5fadba754/c2pa.assertions/cawg.identity",
          "explanation": "signing certificate trusted, found in System trust anchors"
        },
        {
          "code": "cawg.identity.well-formed",
          "url": "self#jumbf=/c2pa/urn:c2pa:765db996-108d-492e-9851-b3e5fadba754/c2pa.assertions/cawg.identity",
          "explanation": "CAWG X.509 identity signature valid"
        }

mauricefisher64:
adressed in https://github.com/contentauth/c2pa-rs/pull/2516

PR #2250, opened 2026-06-25
ci: Revise open-source release process

### LINKED ISSUE #2229: Track release-plz `release_commits` breaks `version_groups` setting
`release-plz` fails to respect `version_groups` (`c2pa` and `c2pa-c-ffi`) when `release_commits` is specified. For now it is commented out, but should be reintroduced when the upstream bug is fixed: https://github.com/release-plz/release-plz/issues/2891

scouten-adobe:
Implements the approved **Open source release process – DRAFT 2** (approved 23 June 2026) in CI and documentation. Applies to the **0.x (pre-1.0) phase**.

## Core idea

Govern changes by **breaking vs. non-breaking**, not by size:

- **Track 1 — additive (`0.x.y`)**: lands on `main`, is cherry-picked onto the active release line (`stable`) via a backport-on-label bot, and ships quickly on the current release line. Gated by `cargo-semver-checks`.
- **Track 2 — breaking (`0.x.0`)**: batched onto a scheduled **train** cut on the **second Monday of odd-numbered months at 16:00 UTC**, baked **≥ 3 business days** on an unpublished `0.(x+1).0-rc.N` candidate branch, then **manually promoted** to `stable`. Skip-if-empty; don't-hold-the-train.

## Branch model

| Branch | Role | Published? |
|---|---|---|
| `main` | nightly-like; always green, API-unstable; upstream-first | **No** |
| `stable` | active release line; additive + promoted breaking releases publish here | Yes |
| `v0.x` (e.g. `v0.89`) | retired line, snapshotted at promotion; rare backports | Yes |
| `0.(x+1).0-rc.N` | transient breaking candidate; bakes under CI | **No (by construction)** |

The `-rc.`-named candidate matches the Tier CI triggers (`*-rc.*`) but **not** the publish triggers (`stable`/`v0.*`), so a candidate can never be published.

## CI changes

**New workflows:** `release-pr.yml` (version/changelog PR) + `release.yml` (publish) split off `main`; `release-train-cut.yml` (scheduled RC cut w/ date gate + skip-if-empty); `backport.yml` (`korthout/backport-action`); `upstream-first-check.yml` (proactive: blocks a release-branch PR whose commits aren't already on `main`); `reconciliation.yml` (reactive upstream-first safety net via `git cherry`); `semver-checks.yml`; `check-no-patch-deps.yml` (blocks release if an in-flight extracted-crate `[patch]`/git dep is present); `canary-extracted-crates.yml` (cross-repo drift vs `c2pa_cbor`); `labels.yml` + `.github/labels.yml` (label sync).

**Modified:** `tier-1a.yml` runs on release/RC branches and **drops the defunct `v1_api` branch**; `tier-1b.yml`/`tier-2.yml` run on release/RC-targeting PRs and during an RC bake; `release-plz.toml` enables `release_commits` (closes #2229); `library-release.yml` comment clarified.

**Validation gating:** Tier 1A gates every merge to `main`. **Any PR targeting a release-line (`stable`, `v0.x`) or release-candidate (`*-rc.*`) branch — including backport PRs and the release-plz release PR — must pass the full Tier 1A + 1B + 2 suite** plus `cargo-semver-checks`. All three tiers also run against `main` on a daily (nightly) schedule. The `check-release` label opts a `main` PR into the full suite on demand.

**Supply chain:** every GitHub Action introduced by this PR is pinned to a full commit SHA.

## Docs

Rewrote [`docs/release-process.md`](docs/release-process.md) as the canonical process doc; aligned `deprecation-policy.md` (deprecate additively → removal rides the train; 60-day window unchanged), `support-tiers.md` (adds a tier-gating section cross-referencing the release process), `CONTRIBUTING.md` (upstream-first), and `README.md` to the two-track model.

## One-time setup (before the first train)

```sh
git fetch --tags
git branch stable c2pa-v0.89.0   # most-recent c2pa release tag at branch creation
git push origin stable
```

## Decisions baked in (per discussion)

- Backport engine: **`korthout/backport-action`** (self-contained; Mergify noted as alternative in docs).
- RC promotion: **manual**.
- Validation gate: **all three tiers on any release-line/RC-targeting PR (including backports)**, plus a nightly full-suite run against `main`.
- Extracted-crate example: **`c2pa_cbor`** (already extracted to `contentauth/c2pa-cbor`).

## Notes for reviewers

- `actionlint` is clean on all new/modified workflows.
- The cross-repo canary patches `c2pa_cbor` (`contentauth/c2pa-cbor`); extend the list as more crates are extracted.


.github/labels.yml
```diff
@@ -0,0 +1,31 @@
...
+  color: "fbca04"
+  description: "Tracking issue for an active breaking-change release train bake"
+
+- name: reconciliation
+  color: "d93f0b"
+  description: "A release branch has commits missing from main (upstream-first violation)"
+
+- name: canary
+  color: "b60205"
+  description: "Cross-repo canary detected drift with an extracted crate"
```
tmathern: Is this going to be different for errors reported, for isntance, by the downstream SDKs? (asking because it mentions crates explicitly).

docs/deprecation-policy.md
```diff
@@ -14,6 +14,8 @@ We follow [Semantic Versioning (SemVer)](https://semver.org/). Version 1.0.0 wil
 
 **Before 1.0:** Major version zero (`0.y.z`) is for initial development: Anything may change at any time, and the public API should not be considered stable. In the Rust/Cargo ecosystem, this means that a change from `0.2.3` to `0.3.0` may include incompatible API changes. We will, however, make a good-faith effort to follow the deprecation process below even before 1.0, so that users have advance warning before breakage occurs.
 
+Pre-1.0 this dovetails with our two-track [release process](release-process.md): a **deprecation is additive**, so it ships continuously as a patch release (the `y` in `0.x.y`) and starts the grace-period clock immediately. The eventual **removal is breaking**, so it rides the next scheduled breaking "train" (a bump of the middle number, `0.x.0`) once the grace period has elapsed. Users therefore get the replacement API and the deprecation warning right away, with a known schedule for when the old API disappears.
```
tmathern: Dovetails?

docs/support-tiers.md
```diff
@@ -55,6 +55,7 @@ A Tier 1B configuration will:
 
 * Have continuous integration tests that build and pass for every versioned release.
 A release will be blocked if the tests do not pass.
+In practice these run on the release-plz release PR (which carries the `release` label) immediately before a crates.io publish, and continuously on a release-candidate branch during a train's bake. You can also invoke them on any PR by adding the `check-release` label.
```
tmathern:
```suggestion
In practice these run on the `release-plz` release PR (which carries the `release` label) immediately before a crates.io publish, and continuously on a release-candidate branch during a train's bake. You can also invoke them on any PR by adding the `check-release` label.
```
(I usually place tool names in backticks too)

release-plz.toml
```diff
@@ -37,8 +37,11 @@ dependencies_update = true
 features_always_increment_minor = false
 pr_labels = ["release"]
 release_always = false
-# TODO: https://github.com/contentauth/c2pa-rs/issues/2229
-# release_commits = "^(feat|fix|docs|perf|refactor|revert|test|update)[!]?[(:]"
+# Only these commit types trigger a release PR. This keeps chore-only commits
+# (including chore cherry-picks brought over by the backport bot) from
+# generating spurious release PRs on a release-line branch.
+# See https://github.com/contentauth/c2pa-rs/issues/2229
+release_commits = "^(feat|fix|docs|perf|refactor|revert|test|update)[!]?[(:]"
```
tmathern, line 44: docs triggers a release?
`docs` commits may need case-by-case review, no? Difference between new docs and fixing a typo. Same note for `test`?

CONTRIBUTING.md
```diff
@@ -61,6 +61,16 @@ These include such things as Rust code format, Clippy/lint checks, and unit test
 We encourage you to raise an issue in GitHub before starting work on a major addition to the crate.
 This will give us an opportunity to discuss API design and avoid duplicate efforts.
 
+### Which branch to target
+
+Open your pull request against **`main`**. We follow an _upstream-first_ rule: every change lands on `main` first, and release-line branches (`stable`, `v0.x`) only ever receive changes by cherry-pick — nothing originates on them. This keeps `main` the superset of all releases so nothing is lost across a version bump.
+
+* If your change is **additive** (a new API, a bug fix), it can ship continuously. After it merges, a maintainer adds a `backport-stable` label and a bot opens a cherry-pick PR onto the current release line.
```
tmathern: `ship continuously` is a bit unclear. What I think we mean is that in can merge to main, and release in the next version (no need for a major).

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+This document describes how we release `c2pa-rs` and its related crates, and the branching model that supports it. It applies to the **0.x (pre-1.0) phase**; we will revisit it as we approach a 1.0 stability commitment.
 
-Release-plz leverages information contained in the various `Cargo.toml` files and its own [`release-plz.toml`](../release-plz.toml) configuration file. In this document, we will talk only about how it release-plz is configured for this project. For more detail about the available settings in release-plz, we recommend reading the [release-plz documentation site](https://release-plz.dev/docs), which is quite extensive.
+## Goals
 
-## How it works
+We have two goals that pull against each other:
```
tmathern:
```suggestion
We have two main goals:
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-We use the pre-packaged [release-plz GitHub Actions wrapper](https://github.com/release-plz/action) to trigger builds. This is configured in our [`release.yml`](../.github/workflows/release.yml) task.
+* **Move fast on pre-1.0 refactors.** While we're below 1.0, we want freedom to make breaking API changes and larger refactors.
```
tmathern:
```suggestion
* **Move fast on pre-1.0 refactors.** While we're below 1.0, we may make larger refactors and breaking API changes.
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+* **Be stable and predictable for users** (including ourselves and the other language bindings that build on this crate): a steady stream of features, plus breaking changes that arrive on a *known schedule* rather than at random.
```
tmathern:
```suggestion
* **Be stable and predictable as much as possible for users** (including ourselves and the other language bindings that build on this crate): a steady stream of features, plus breaking changes that arrive on a pre-determined *known schedule*
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-For additional information, please read the [release-plz documentation for GitHub Action](https://release-plz.dev/docs/github/quickstart).
+We reconcile them with one core principle.
```
tmathern: (I'd remove that)

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-In this repo, the build automation is triggered for each commit to the `main` branch.
+## Core principle: split by breaking vs. non-breaking
 
-This task runs using a GitHub access token tied to Eric's `@scouten-adobe` account, which is why all PRs and releases are under Eric's name. (See ticket CAI-7621 for moving this to a team-owned bot account.)
+What matters to someone depending on this crate isn't whether a change is "big" — it's whether it breaks their build. An additive feature costs them nothing; a changed or removed API forces a migration. So we govern those two kinds of change on two different tracks, which map directly onto pre-1.0 Cargo semantics:
```
tmathern:
```suggestion
What matters to someone depending on this crate isn't whether a change is "big" — it's whether it breaks their build. An additive feature costs them nothing; a changed or removed API forces a migration with potential fixes and updates to their build system. So we govern those two kinds of change on two different tracks, which map directly onto pre-1.0 Cargo semantics:
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+| -- | -- | -- | -- |
+| Additive / non-breaking | `0.x.y` (bump `y`) | compatible | **Track 1** — fast, continuous |
+| Breaking | `0.x.0` (bump `x`) | incompatible | **Track 2** — scheduled "train" |
 
-This runs [`release-plz release-pr`](https://release-plz.dev/docs/usage/release-pr). For each published crate in the repo, it looks for certain types of commits since the last tagged release _within_ the source tree for the crate. The exact list of commit types that trigger a release PR is configured by the `release-commits` section of [`release-plz.toml`](../release-plz.toml), but notably is set to ignore `chore` commits and include `feat`, `fix`, and `docs` commits. If it finds any matching, it will open a new release pull request or update the existing release. (NOTE: When Colin's [PR #2196 to release-plz](https://github.com/release-plz/release-plz/pull/2196) lands there will be a new trigger here which will reference intra-project dependency updates even if the downstream project has no commits of its own.)
+Most changes never wait for the train: anything additive ships continuously, and only breaking changes are batched and scheduled.
```
tmathern:
```suggestion
Most changes never wait for the train: anything additive merges to main continuously, and only breaking changes are batched and scheduled.
```
(I think? I don't think we continuously bake minor releases?)

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-* `cargo.toml`: **Update version number.**
-  * If any API breaking changes are detected, bump the minor version number. (This will change to a major version number once we do 1.0 release.)
-  * If any feature changes (API additions) are detected, bump the minor version number. (This will _not_ change when we do 1.0 release.)
-  * If only bug-fix changes are detected, bump the patch version number.
-  * For all crates, `release-plz` will evaluate the commit history since the previous release tag using [Conventional Commit syntax](https://www.conventionalcommits.org/en/v1.0.0/#summary) to determine API breaking changes, feature changes, or only bug-fix changes.
-  * For library crates _only,_ `release-plz` will also download the most recent release from [crates.io](https://crates.io) to verify any API surface changes. The larger scope (patch -> minor -> major) will be used to select the new version number.
+| Branch | Role | Published to crates.io? |
+| -- | -- | -- |
+| `main` | Always green but API-unstable ("nightly-like"). It must always compile and pass tests, but its public API is **not** guaranteed stable — it may contain unstabilized or feature-gated work. | **No** |
```
tmathern, line 27: Do you want to cross-reference the support tiers here?

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+| `stable` | Tracks the most-recent crates.io release and is the currently-active release line. Additive (`0.x.y`) releases, and the promoted breaking (`0.x.0`) release, are published from here. | **Yes** |
+| `v0.x` (e.g. `v0.89`) | A long-lived branch for a **retired** release line, snapshotted from `stable` when that line is retired. The target for the rare backport to a retired line. | Yes (rare backports) |
```
tmathern:
```suggestion
| `v0.x` (e.g. `v0.89`) | A long-lived branch for a **retired** release line, snapshotted from `stable` when that line is retired. The potential target for the rare backport to a retired line. | Yes (rare backports) |
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-* `CHANGELOG.md`: **Add description of new release.**
-  * This adds a new section to the top of the changelog file with a link to the diffs between this version and the previous release.
-  * The changelog section also contains a list of commits that affected this crate's source, grouped by the commit type. (See [Conventional Commit syntax](https://www.conventionalcommits.org/en/v1.0.0/#summary).)
+Two rules keep this coherent:
 
-> [!IMPORTANT]
-> It _is_ permissible to manually edit the proposed changelog in the PR, but be aware those changes will be overwritten if another update is triggered. release-plz uses force-push to update any existing release PR.
+* **Upstream-first.** Every change lands on `main` first. Release-line branches only ever *receive* changes (via cherry-pick); nothing originates on them. This guarantees `main` is always the superset and that nothing is lost across a major bump. A scheduled [reconciliation check](#reconciliation-check) enforces this.
+* **`main` stays releasable.** Destabilizing work happens on feature branches off `main`, merged only once coherent.
```
tmathern, line 37: I think then this needs a mention/cross-link with the support tiers indeed.

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-> [!IMPORTANT]
-> If the version number of the crate in `main` is _different_ from the version number on [crates.io](https://crates.io), this job will NOT run. See the next section, "Publish new crates."
+> **A note on cherry-picking.** Cherry-picking transfers individual fixes between branches without merging everything. Because actively-supported lines can diverge over time, a fix that applies cleanly on one branch may not cherry-pick directly onto another. In that case the change may need to be adapted to compile, integrate, and pass tests on the target branch — or, if the branches have diverged enough, implemented separately for each supported branch.
 
-### Publish new crates
+## Track 1 — Additive releases (`0.x.y`)
 
-For each crate in the repo, if the version number of the crate in `main` is _different_ from the version number on [crates.io](https://crates.io), release-plz will attempt to publish the crate to [crates.io](https://crates.io). This typically occurs after a maintainer has merged a release PR as generated in the previous step.
+Low-risk, non-breaking features and bug fixes ship continuously:
```
tmathern: `ship continuously` is really confusing everywhere. I would differentiate merge to main versus shipping as in "shipping a release".

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
-* **Creates a [GitHub release](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository)** for the uploaded crate. This creates an annotated git tag of the form `(crate-name)-v(version-number)`, which – among other things – is used to drive the previous-version comparison for future release PRs. This also creates a ZIP archive of the source at this crate version which is stored on GitHub with the release.
+Key points:
 
-### Publish c2patool binary builds
+* **Short (~1-day) bake.** Additive releases (`0.x.y`, y ≥ 1) don't need a full release-candidate stage, but we do hold a brief bake — approximately one business day — before the crates.io publish, to re-verify compatibility against the downstream projects we maintain (the other language bindings). This is far lighter than the breaking train's bake; it exists to catch integration surprises, not to soak the change. (An optional `0.x.y-rc.N` soak can additionally be used for a change that warrants extra caution.)
```
tmathern: That's the important difference between merge and ship/release

tmathern: "soak the change"?

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
 
-If a new version of `c2patool` is generated by the previous step, an additional job runs after the publish job to build and publish binary versions of `c2patool`.
+## Track 2 — The breaking train (`0.x.0`)
 
-This is specified by the `publish-c2patool-binaries` job in our [`release.yml`](../.github/workflows/release.yml) task.
+Breaking changes and larger refactors are batched onto a scheduled train:
 
-For each supported command-line platform (MacOS, Windows, and Ubuntu), it performs the following steps:
+1. On the scheduled date, a release-candidate branch `0.(x+1).0-rc.1` is cut from `main`. **RC branches are not published to crates.io** (their name keeps them off every publish trigger).
+2. **Bake period: minimum 3 days.** Only bug fixes are accepted during the bake, and they follow upstream-first (fix on `main`, cherry-pick to the candidate). Hold longer than 3 days if needed to validate across the downstream projects we maintain.
```
tmathern:
```suggestion
2. **Bake period: minimum 3 business days.** Only bug fixes are accepted during the bake, and they follow upstream-first (fix on `main`, cherry-pick to the candidate). Hold longer than 3 days if needed to validate across the downstream projects we maintain.
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+* **Extract-with-re-export:** moving code into a separate crate is additive as long as the public paths are preserved by re-exporting (`pub use`). See [extracted crates](#extracted-crates-multi-repo).
 
-A quick, non-authoritative, summary of the PR title rules follows. As a quick reminder, the summary line (and thus the PR title) must have this exact format, including punctuation:
+## Extracted crates (multi-repo)
 
-```
-type(scope): description
-```
+We are extracting stable, low-churn code out of `c2pa-rs` into independent crates, each in its own repository, so that this code is not rebuilt on every PR to `c2pa-rs` `main`. The first example is **`c2pa-raw-crypto`** — the raw cryptographic signature primitives, now in their own [`contentauth/c2pa-raw-crypto`](https://github.com/contentauth/c2pa-raw-crypto) repo. These crates are versioned **independently** (lockstep versioning would re-couple the build times we're trying to separate).
```
tmathern: The first example technical was `c2pa-cbor` :) . Might want to mention that crate anyway!

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
-* **`chore`**: Maintenance work. (Will not trigger a release PR and will not be included in changelog.)
-* **`docs`**: Revising documentation.
-* `build` or `ci`: Adjusting the build system.
-* `perf`: Performance optimization.
-* `refactor`
-* `revert`
-* `style`
-* `test`
-* `update`: Updating a dependency. (Used by Dependabot.)
+* **Extracted crates leave the train.** Each gets a simple, as-needed release (cut one when a change lands, `cargo-semver-checks`-gated). Because they rarely change, they have no meaningful wait period of their own.
```
tmathern:
```suggestion
* **Extracted crates leave this repo's release train.** Each gets a simple, as-needed release (cut one when a change lands, `cargo-semver-checks`-gated). Because they rarely change, they have no meaningful wait period of their own.
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+
+### Patch-dependency guard
+
+[`check-no-patch-deps.yml`](../.github/workflows/check-no-patch-deps.yml) fails if a `[patch]` section or a git dependency is present. It runs on release-branch PRs and as a required prerequisite of `release.yml`, so a `[patch]` left over from co-developing an extracted crate can never leak into a publish.
+
+### Release-train cut
+
+[`release-train-cut.yml`](../.github/workflows/release-train-cut.yml) runs every Monday and gates on a date check so it only acts on the second Monday of an odd-numbered month (or when dispatched manually with `force: true`). When it fires it computes the next breaking version, applies skip-if-empty, and — if there's breaking work to ship — pushes a new `0.(x+1).0-rc.1` candidate branch from `main` and opens a `release-train` tracking issue describing the bake and the manual-promotion step.
```
tmathern, line 169: When on Monday? (Hour + timezone?)

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+* **The release-plz release PR** carries the `release` label and therefore runs the full **Tier 1A + 1B + 2** suite. This is the comprehensive gate immediately before a crates.io publish, and the natural home for the short (~1-day) downstream re-verification.
+* To force the full suite on an individual risky backport before it reaches the release branch, add the `check-release` label to that backport PR.
 
-When performing the "Create a release pull request" step described above, release-plz sometimes updates the existing release branch and sometimes creates a new release branch with a new pull request. We don't understand how it makes that choice. When it creates a new pull request, it closes the old pull request, but does not delete the old release branch.
+See [docs/support-tiers.md](support-tiers.md) for what each tier covers.
 
-In order to reduce the signal noise in the git repo, we've added a step in the release process to search for and delete old release branches. Search for "Clean up stale release-plz branches" in the [`release.yml`](../.github/workflows/release.yml) task.
+## What we are deliberately NOT doing
```
tmathern: I would remove the section, we don't want to list "bad" precedents?

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+
+`scope` is optional; if present it must be one of `c2patool`, `export_schema`, `make_test_images`, `sdk`, or `c2pa_c_ffi`. If omitted, drop the parentheses too. `description` is a short sentence, capitalized, no trailing period, preferably under 70 characters.
+
+> MAINTENANCE NOTE: if these rules change, keep [`.github/workflows/pr_title.yml`](../.github/workflows/pr_title.yml) and [`.commitlintrc.yml`](../.commitlintrc.yml) in sync.
+
+## Known issues
+
+### Failure to update downstream crates
+
+In repos that host multiple crates, an earlier crate in the dependency chain can warrant a release while a crate that depends on it has no commits of its own. release-plz updates the downstream version reference but doesn't always cut a new release of the downstream crate. (See [issue #2164](https://github.com/release-plz/release-plz/issues/2164) and [PR #2196](https://github.com/release-plz/release-plz/pull/2196).) Workaround: post a no-op change (a whitespace tweak or comment) to the downstream crate to trigger a release PR.
```
tmathern:
```suggestion
In repos that host multiple crates, an earlier crate in the dependency chain can warrant a release while a crate that depends on it has no commits of its own. `release-plz` updates the downstream version reference but doesn't always cut a new release of the downstream crate. (See [issue #2164](https://github.com/release-plz/release-plz/issues/2164) and [PR #2196](https://github.com/release-plz/release-plz/pull/2196).) Workaround: post a no-op change (a whitespace tweak or comment) to the downstream crate to trigger a release PR.
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+
+### Left-behind release branches
+
+release-plz sometimes creates a new release branch + PR instead of updating the existing one, leaving the old branch behind. To reduce noise, [`release-pr.yml`](../.github/workflows/release-pr.yml) deletes stale `release-plz-*` branches.
```
tmathern:
```suggestion
`release-plz` sometimes creates a new release branch + PR instead of updating the existing one, leaving the old branch behind. To reduce noise, [`release-pr.yml`](../.github/workflows/release-pr.yml) deletes stale `release-plz-*` branches.
```

.github/workflows/release.yml
```diff
@@ -56,31 +49,31 @@ jobs:
...
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Run release-plz
+      # Make cargo-semver-checks available so release-plz uses it as an extra
+      # guard against an accidental breaking change shipping as an additive
+      # (0.x.y) release. (The hard gate is semver-checks.yml on the release PR.)
+      - name: Install cargo-semver-checks
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-semver-checks
```
ale-adobe, line 60: Should we pin a version on this tool call to keep behavior consistent and avoid unexpected slight changes?

docs/deprecation-policy.md
ale-adobe: This implies that when we deprecate an API that a replacement will also be released in the same patch. Are we committing to doing so? I agree it makes sense generally, but I think there are also some real world cases where we might want to deprecate an API as soon as possible to prevent more users from picking it up and using it before we can provide a new API, especially if a temporary workaround is already available.

CONTRIBUTING.md
ale-adobe: More specifically, maybe we can say that it will be released in the next patch version of the release line.

release-plz.toml
ale-adobe: +1

.github/workflows/tier-1a.yml
```diff
@@ -14,7 +14,9 @@ on:
   pull_request:
     branches:
       - main
-      - v1_api
+      - stable
+      - 'v0.*'
```
ale-adobe, line 22: After the initial 1.0 launch, I assume this will be updated to target anything matching `v*.*`?

.github/workflows/canary-extracted-crates.yml
```diff
@@ -0,0 +1,84 @@
...
+          c2pa-raw-crypto = { git = "https://github.com/contentauth/c2pa-raw-crypto", branch = "main" }
+          EOF
+
+      - name: Build and test
+        run: |
+          cargo build --workspace --all-features
+          cargo test --workspace --all-features
+
+      - name: Open or update canary issue on failure
+        if: failure()
```
ale-adobe, line 56: Is it worth having a success case as well, so that we can confirm more explicitly that nothing broke, rather than relying on "no failure reported" as the signal?

.github/workflows/labels.yml
```diff
@@ -0,0 +1,34 @@
+name: Sync labels
```
ale-adobe, line 1: Neat!

.github/workflows/reconciliation.yml
```diff
@@ -0,0 +1,96 @@
+name: Release-branch reconciliation
+
+# Safety net for the upstream-first rule. Everything is supposed to land on
```
ale-adobe: Some thoughts here...

This approach seems reactive: if something goes wrong, this job will detect that the commit exists on release-line branch but doesn't exist on main and then flags it *after the fact*.

Could we consider a different approach (perhaps in addition to this) where we proactively prevent this from happening to begin with? We can enforce branch protection on release-line branches, requiring a PR to be opened before merging, and run a required CI check on any such PRs that verifies all commits in the PR already exist on `main`.

It might be good to have both a proactive guard and a reactive check (this reconciliation job).

.github/workflows/release.yml
ale-adobe: I see that we use `obi1kenobi/cargo-semver-checks-action@v2` below...

.github/labels.yml
```diff
@@ -0,0 +1,31 @@
+# Version-controlled GitHub labels, synced by labels.yml workflow via
+# crazy-max/ghaction-github-labeler. Add labels here rather than creating them
+# by hand in the GitHub UI.
+#
+# NOTE: This file intentionally lists only the labels the release process
+# depends on. Per-line backport labels (backport-v0.<N>) are created on demand
+# at train-cut time and are not enumerated here.
```
ale-adobe: I'm a bit confused by how this would work on-demand. If those specific version labels are created on demand and not listed here, how would the other Git actions be aware of them and agree on what labels exist?

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+The model only works if we stay disciplined about keeping the fast lane non-breaking:
 
-We [enforce commit squashing](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) for pull requests in this repo to simplify our git history. When squashing commits for pull request merges, GitHub defaults to using the PR title as the first line of the commit message. For that reason, our [`pr-title.yml`](../.github/workflows/pr-title.yml) task is configured to read the PR title and ensure that it confirms to [Conventional Commit syntax](https://www.conventionalcommits.org/en/v1.0.0/#summary). This enforcement is configured by [`.commitlintrc.yml`](../.commitlintrc.yml), which is the definitive specification of what constitutes an acceptable PR title.
+* **Review norm:** "Can this ship additively? If yes, it goes out now. If it requires a break, it waits for the next train."
+* **Forward-compat tools:** prefer `#[non_exhaustive]` (enums/structs), sealed traits, and default trait methods so future additions stay compatible.
```
ale-adobe:
```suggestion
* **Forward-compatible tools:** prefer `#[non_exhaustive]` (enums/structs), sealed traits, and default trait methods so future additions stay compatible.
```

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+* **Deprecate-then-remove:** when we must break, add the replacement API additively now and mark the old one `#[deprecated]`. Removal of the old API rides a later train, after the deprecation window elapses. See the [deprecation policy](deprecation-policy.md). Users get the improvement immediately and a window to migrate.
```
ale-adobe, line 97: Relates to my other comment about having these two strictly paired.

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
-A quick, non-authoritative, summary of the PR title rules follows. As a quick reminder, the summary line (and thus the PR title) must have this exact format, including punctuation:
+## Extracted crates (multi-repo)
 
-```
-type(scope): description
-```
+We are extracting stable, low-churn code out of `c2pa-rs` into independent crates, each in its own repository, so that this code is not rebuilt on every PR to `c2pa-rs` `main`. The first example is **`c2pa-raw-crypto`** — the raw cryptographic signature primitives, now in their own [`contentauth/c2pa-raw-crypto`](https://github.com/contentauth/c2pa-raw-crypto) repo. These crates are versioned **independently** (lockstep versioning would re-couple the build times we're trying to separate).
 
-The `type` field describes the nature of changes you are making. This project requires the type to be one of these exact names (bold names are preferred in most cases):
+Once a subproject lives in its own repo and is published, `c2pa-rs` depends on it by version (`c2pa-raw-crypto = "0.2"`) like any third-party crate. That reframes the "wait period" from a scheduling problem into a dependency-edge problem:
```
ale-adobe: Any thoughts on pinning specific versions of subproject crates, or would we use `^` or something else?

docs/release-process.md
```diff
@@ -1,157 +1,231 @@
...
+* [`release.yml`](../.github/workflows/release.yml) runs `release-plz release`: when a release PR merges (a push to the release-line branch), it publishes the changed crates to crates.io, creates GitHub releases, and tags them `(crate-name)-v(version)`. It then builds and uploads the `c2patool` binaries and SBOMs. A push whose ref contains `-rc.` never publishes.
 
-MAINTENANCE NOTE: If this list of rules is changed, please keep in sync with [`.github/workflows/pr_title.yml`](../.github/workflows/pr_title.yml) and [`.commitlintrc.yml`](../.commitlintrc.yml).
+How release-plz chooses a version, per crate:
 
-## Known issues
+* If only bug-fix commits are detected, bump the patch number (`y`).
+* If API additions or breaking changes are detected, bump the middle number (`x`). (Pre-1.0, Cargo treats a middle-number bump as incompatible; this becomes the major-number bump after 1.0.)
```
ale-adobe, line 135: After 1.0, how and when would we use the middle-number (minor version)?

ale-adobe (commented):
Many of the jobs create issues on Jira for tracking purposes. Do we have plans to properly organize and make these issues easily viewable and trackable, perhaps during a recurring meeting? I'm sure it wouldn't be hard to set up, but just wondering if you have something in mind already.

release-plz.toml
ok-nick, line 44: This has not been fixed upstream yet, see #2229.

.github/labels.yml
scouten-adobe: Let's say `v0.95.x` train is retired (because there's `v0.96` train that is the current `stable` train), but we decide to back port a security fix to 0.95. We don't need an automated process to create the `backport-v0.95` label – we'd just do that manually for that specific case. (It's relatively easy to do that.)

What is probably worth saying in this file (and I'll make that change momentarily) should be used to create the common cases and we can create special-case labels by hand.

.github/labels.yml
scouten-adobe: Yes, this is specifically for the crates that we own that are _upstream_ of `c2pa-rs`. We currently have `c2pa-cbor` as one example of an extracted crate. If something changes in _that_ crate that breaks existing `c2pa-rs` behavior we'd get a `canary` PR urging us to investigate and fix that breakage.

The downstream SDKs should have their own canaries to track `c2pa-rs` `main` branch and we'd coordinate across projects to resolve those prior to the next release train.

.github/workflows/tier-1a.yml
scouten-adobe: Yes

docs/deprecation-policy.md
scouten-adobe: Odd English expression. Replaced it with "fits."

release-plz.toml
scouten-adobe: I agree that `docs` and `test` commits warrant case-by-case evaluation. This is the trigger for a new _release PR_ against the `stable` release train, so there are two inflection points where we can do that manual review:

1. We could decide not to back port a `docs` or `test` change from `main` to `stable`.
2. We could see the release PR and decide that it isn't worthy of a new release until something _else_ more significant lands.

So I think this is fine as is.

CONTRIBUTING.md
scouten-adobe: Good call-out. I've updated the wording.

docs/release-process.md
scouten-adobe: Done.

docs/release-process.md
scouten-adobe: Other way around: _Everything_ merges to main ASAP after passing CI gates. Those changes that are non-breaking are also ported to the stable branch and can then be released fairly quickly. I've updated the wording to that effect.

docs/release-process.md
scouten-adobe: Agreed. Adjusted the language slightly.

.github/workflows/release.yml
scouten-adobe: I think that's a good practice that we should adopt generally. I'll do a separate PR targeting main and then update this one once that's adopted.

docs/deprecation-policy.md
scouten-adobe: Yes, this consistent with our previously-discussed and agreed-to deprecation policy. See https://github.com/contentauth/c2pa-rs/blob/main/docs/deprecation-policy.md#stage-1-deprecation-notice-minor-release.

.github/workflows/tier-1a.yml
scouten-adobe: This policy is effective for the 0.x days. We'll need to revisit as we prepare for 1.0.

.github/workflows/canary-extracted-crates.yml
scouten-adobe: I'd prefer not to do that. This runs daily and that could be a lot of signal noise that we'd have to process. This raises a GitHub issue on failure and that (hopefully rare) signal is something we'd want to focus on.

docs/release-process.md
scouten-adobe: I _think_ the answer is that we'd use the middle number for each new release train as we are here. If there's nothing significantly new to release (i.e. semver would say don't bump the middle number), then I'd say we should skip that release train.

But that's a question we can resolve as we approach 1.0.

docs/release-process.md
scouten-adobe: Agree; that's confusing. Removed.

docs/release-process.md
scouten-adobe: Good call. Changed it. (Also, we haven't yet landed the `c2pa-raw-crypto` change.)

docs/release-process.md
scouten-adobe: Tight pins can create usability problems for library consumers, so I'm wary of doing that. We already have the nightly crate-drift canary and the semver pins should prevent us from introducing breaking changes.

.github/workflows/reconciliation.yml
scouten-adobe: Agreed. I've added a new verification run for that case.

.github/workflows/canary-extracted-crates.yml
tmathern: +1, this also avoids "alert fatigue".

docs/release-process.md
tmathern: (Also I think we already had bug reports sometimes asking to loosen that because it created issues downstream for users too).

docs/release-process.md
tmathern: I would have thought we'd follow semver. Maybe some release trains would only have bugfixes, who knows!

tmathern (approved):

docs/release-process.md
```diff
@@ -1,157 +1,242 @@
...
+* **Extract-with-re-export:** moving code into a separate crate is additive as long as the public paths are preserved by re-exporting (`pub use`). See [extracted crates](#extracted-crates-multi-repo).
+
+## Extracted crates (multi-repo)
+
+We are extracting stable, low-churn code out of `c2pa-rs` into independent crates, each in its own repository, so that this code is not rebuilt on every PR to `c2pa-rs` `main`. The first example is **`c2pa_cbor`** — the CBOR (de)serialization primitives used throughout C2PA manifests, now in their own [`contentauth/c2pa-cbor`](https://github.com/contentauth/c2pa-cbor) repo. These crates are versioned **independently** (lockstep versioning would re-couple the build times we're trying to separate).
+
+Once a subproject lives in its own repo and is published, `c2pa-rs` depends on it by version (`c2pa_cbor = "0.77"`) like any third-party crate. That reframes the "wait period" from a scheduling problem into a dependency-edge problem:
```
ok-nick: Prefer full x.y.z versions?

docs/release-process.md
```diff
@@ -1,157 +1,242 @@
...
+## Automation
+
+Cutting a release is mostly a CI action rather than manual toil. The pieces:
+
+### `release-plz`
+
+We use [`release-plz`](https://release-plz.dev) (via the [GitHub Action wrapper](https://github.com/release-plz/action)), configured by [`release-plz.toml`](../release-plz.toml). Its two responsibilities are split across two workflows, both of which run on the **release-line and release-candidate branches** — never on `main`:
+
+* [`release-pr.yml`](../.github/workflows/release-pr.yml) runs `release-plz release-pr`: for each published crate it inspects commits since the last tag and opens/updates a **release PR** that bumps the version and updates the changelog. Because that PR targets a release-line branch, it runs the full Tier 1A + 1B + 2 suite (see [validation gating](#validation-gating)).
+* [`release.yml`](../.github/workflows/release.yml) runs `release-plz release`: when a release PR merges (a push to the release-line branch), it publishes the changed crates to crates.io, creates GitHub releases, and tags them `(crate-name)-v(version)`. It then builds and uploads the `c2patool` binaries and SBOMs. A push whose ref contains `-rc.` never publishes.
```
ok-nick, line 130: Not a blocker, but we should consider using [GitHub immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) in this workflow.
> The use of this type of release increases security by blocking supply chain attacks. Attackers cannot:
> - Inject vulnerabilities or malware into current project releases.
> - Make changes to assets and tags that may break developer workflows.

docs/release-process.md
```diff
@@ -1,157 +1,242 @@
...
-* `make_test_images`
-* `sdk`
+The [support tiers](support-tiers.md) map directly onto the branching model:
 
-If `scope` is omitted, the parenthesis must also be omitted.
+* **Merging to `main`** requires **Tier 1A** — the merge gate for everyday development.
+* **Any PR targeting a release-line (`stable`, `v0.x`) or release-candidate (`*-rc.*`) branch** must pass the **full Tier 1A + 1B + 2 suite** before it can merge. This includes **backport PRs**, RC bake bugfix PRs, and the `release-plz` release PR — anything headed for a published (or soon-to-be-published) artifact gets the most thorough validation we have. PRs targeting release lines additionally run [`cargo-semver-checks`](#semver-checks).
+* During a train's bake, Tier 1A + 1B + 2 also run on every push to the `*-rc.*` branch.
+* **All three tiers also run against `main` on a daily schedule** (a nightly run), catching regressions that only appear under the heavier Tier 1B/2 configurations even when no release-targeting PR is open.
+* On a `main` PR you can opt into the full suite on demand by adding the `check-release` label (useful to assess release-readiness before a change is backported).
```
ok-nick, line 187: Can we make these part of the branch protection rules? We can require these workflows must succeed before merging and also that the branch being tested must be updated to the latest commit.

docs/release-process.md
```diff
@@ -1,157 +1,242 @@
...
+## Known issues
 
-In the meantime, the workaround that seems easiest is to post a no-op change to any downstream crates that need to be re-released. A whitespace change or a comment that could benefit from some proofreading is typically enough to cause release-plz to call for a new release.
+### Failure to update downstream crates
 
-### Left-behind release branches
+In repos that host multiple crates, an earlier crate in the dependency chain can warrant a release while a crate that depends on it has no commits of its own. `release-plz` updates the downstream version reference but doesn't always cut a new release of the downstream crate. (See [issue #2164](https://github.com/release-plz/release-plz/issues/2164) and [PR #2196](https://github.com/release-plz/release-plz/pull/2196).) Workaround: post a no-op change (a whitespace tweak or comment) to the downstream crate to trigger a release PR.
```
ok-nick: Maybe leave a comment on the issues to remove this in the release-process.md when fixed?

docs/release-process.md
```diff
@@ -1,157 +1,242 @@
...
 
-When performing the "Create a release pull request" step described above, release-plz sometimes updates the existing release branch and sometimes creates a new release branch with a new pull request. We don't understand how it makes that choice. When it creates a new pull request, it closes the old pull request, but does not delete the old release branch.
+### Left-behind release branches
 
-In order to reduce the signal noise in the git repo, we've added a step in the release process to search for and delete old release branches. Search for "Clean up stale release-plz branches" in the [`release.yml`](../.github/workflows/release.yml) task.
+`release-plz` sometimes creates a new release branch + PR instead of updating the existing one, leaving the old branch behind. To reduce noise, [`release-pr.yml`](../.github/workflows/release-pr.yml) deletes stale `release-plz-*` branches.
```
ok-nick: Is this a bug in release-plz? Maybe it's fixed or needs to be reported?

.github/workflows/semver-checks.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  semver-checks:
+    name: cargo-semver-checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          fetch-depth: 0
```
ok-nick: Can we remove fetch-depth 0 here?

.github/workflows/backport.yml
ok-nick, line 1: Maybe organize all of the release-related workflows in a sub-folder?

docs/release-process.md
scouten-adobe: Agree that it sounds like a good idea and I tried that a few months ago. It's incompatible with the workflow of uploading the c2patool binaries and SBOMs after the release is created.

If there's a way to stall creating the release until those artifacts are created, I'm open to it, but IMHO it's out of scope for this PR.

docs/deprecation-policy.md
scouten-adobe: Changed "dovetails" to "fits".

docs/support-tiers.md
scouten-adobe: Applied in the latest revision — thanks!

CONTRIBUTING.md
scouten-adobe: Reworded to make clear it means merging to `main` and shipping in the next patch — no major bump needed.

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Removed in the core-principle reword — thanks!

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Right — reworded to "ships fairly quickly by being backported to the stable release train" to avoid implying we continuously bake minor releases.

docs/release-process.md
scouten-adobe: Done — added a "How the tiers gate merges and releases" section to `docs/support-tiers.md` and a "Validation gating" section here, cross-linked both ways.

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Done — added a "How the tiers gate merges and releases" section to `docs/support-tiers.md` and a "Validation gating" section here, cross-linked both ways.

docs/release-process.md
scouten-adobe: Agreed this was the most confusing part. Replaced "ships continuously" throughout with "ships quickly on the current release train," dropped the "soak the change" phrasing, and reworked the text to distinguish landing on `main` from shipping a release.

docs/release-process.md
scouten-adobe: Agreed this was the most confusing part. Replaced "ships continuously" throughout with "ships quickly on the current release train," dropped the "soak the change" phrasing, and reworked the text to distinguish landing on `main` from shipping a release.

scouten-adobe: Removed that phrasing.

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Good catch — switched the example (and the canary) to `c2pa_cbor` / `contentauth/c2pa-cbor`, since `c2pa-raw-crypto` isn't extracted yet.

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Added: second Monday of odd months at 16:00 UTC, matching `release-train-cut.yml`.

docs/release-process.md
scouten-adobe: Removed the "What we are deliberately NOT doing" section.

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

.github/workflows/release.yml
scouten-adobe: Pinned — every action this PR adds is now pinned to a full commit SHA (and #2265 pinned the pre-existing ones).

.github/workflows/tier-1a.yml
scouten-adobe: Yes — this is 0.x-scoped; the branch patterns generalize post-1.0, which we'll revisit at the 1.0 commitment.

.github/workflows/labels.yml
scouten-adobe: 🙌

.github/workflows/reconciliation.yml
scouten-adobe: Added a proactive `upstream-first-check.yml` (blocks a release-branch PR whose commits aren't already on `main`, patch-id compare via `git cherry`) plus a "Branch protection" doc section; `reconciliation.yml` stays as the reactive backstop.

.github/labels.yml
scouten-adobe: The backport action derives the target branch from the label name (`backport-<branch>`), so labels don't need pre-registering — a `backport-v0.89` label works the moment it exists. The list here is just the always-present labels; the rare retired-line labels are created by hand only when needed. I'll clarify the file comment.

docs/release-process.md
scouten-adobe: Applied in the latest revision — thanks!

docs/release-process.md
scouten-adobe: Post-1.0 we move to standard SemVer (minor = additive, major = breaking); pre-1.0 the middle number carries "breaking" because Cargo treats `0.x` that way. This doc is explicitly 0.x-scoped and we'll revisit at the 1.0 commitment.

docs/release-process.md
scouten-adobe: Good idea — noted as a follow-up for `release.yml`; not blocking this PR.

docs/release-process.md
scouten-adobe: Added a "Branch protection" section recommending exactly that (require PRs, required checks, and require-branch-up-to-date). The setting itself is repo config, so it's documented rather than committed.

.github/workflows/backport.yml
scouten-adobe: GitHub only discovers workflows at the top level of `.github/workflows/` — files in subdirectories aren't executed — so we can't nest them. The filename prefixes and `name:` fields are the best grouping available.

release-plz.toml
scouten-adobe: Keeping both `docs` and `test`. `release-plz` only opens a release PR when a crate's version would actually change, and both of these only ever imply a **patch** bump (never breaking) — so a docs or test improvement can ship as a patch without extra ceremony. We'd rather keep that simple than gate it case-by-case.

docs/release-process.md
scouten-adobe: Addressed together with the deprecation-policy thread — deprecation and replacement stay paired; see the updated `deprecation-policy.md`.

docs/release-process.md
scouten-adobe: We'll use a normal caret requirement (e.g. `c2pa_cbor = "0.77"`) and let `Cargo.lock` pin the exact version for reproducible builds — same as any other dependency. Now documented in the extracted-crates section.

docs/release-process.md
scouten-adobe: We'll use a caret requirement with `Cargo.lock` pinning the exact version (same as any other dep) rather than `=x.y.z` in `Cargo.toml`. Now documented in the extracted-crates section.

.github/workflows/canary-extracted-crates.yml
scouten-adobe: We're going to skip an explicit success signal to avoid alert fatigue (per Tania's related note). The canary stays quiet on green and only opens/updates an issue on failure.

.github/workflows/semver-checks.yml
scouten-adobe: Done — removed. `cargo-semver-checks` baselines against the published crates.io release, so it doesn't need full git history.

docs/release-process.md
scouten-adobe: Filed our own tracking issue #2298 for this (rather than commenting on the upstream release-plz issue) and linked it from the note here, so we remember to trim it once it's resolved upstream.

docs/release-process.md
scouten-adobe: It's a `release-plz` quirk. Filed #2299 to track it (and to confirm/report upstream if it isn't already), linked from the note here.

.github/labels.yml
scouten-adobe: Reworded the label to scope it to upstream crates we maintain (e.g. `c2pa_cbor`) — it's specifically the cross-repo canary for our extracted crates, distinct from downstream language-binding error reporting.

.github/workflows/backport.yml
tmathern: ☝️ +1. I think we can handle the file's organization as it currently is.

tmathern (approved):

tmathern:
(Pulling in main so the docs error blocking PR checks goes away)

ok-nick (approved):

release-plz.toml
ok-nick: @scouten-adobe Please double check this before merging.

ale-adobe (approved):

release-plz.toml
scouten-adobe: @ok-nick sadly, the underlying release-plz issue is still open.

PR #2253, opened 2026-06-25
feat: Builder pre claim

gpeacock:
This is an experiment in using builder to construct a claim on the fly rather than only when we sign. It allows us to get usable HashedUris that can be inserted into other assertions, which is an increasing pattern in the spec.

This adds a pre-claim to builder, instead of creating the claim only at signing time, we create a claim up front and add assertions directly to it. This works as long as the user doesn't update the definition between setting it and signing. Since the existing api allows that, we need to handle this by tracking and merging additions.a

A new method, add_assertion_with_ref, returns a HashedURI that can be inserted directly into other assertions now.


docs/builder-pre-claim.md
```diff
@@ -0,0 +1,82 @@
+# Builder pre-claim: what changed and why
+
+## Summary
+
+Previously, `Builder` accumulated everything into `ManifestDefinition` (a JSON-serializable struct) and only built the actual `Claim` once, inside `to_claim()`, at signing time. This branch adds a `Claim` up front (`Builder::pre_claim: Option<Claim>`) and lets several methods write directly into it as they're called, instead of staging data in `definition` for later translation.
+
+`to_claim()` still runs the old translation logic for anything left in `definition` (JSON-set title/format, `definition.assertions`, `definition.ingredients`, `definition.thumbnail`, actions, etc.) — it clones `pre_claim` (if present) as its starting point, then does its usual pass over `definition`, skipping anything that's already been added directly. Both models coexist in the same `Builder`; nothing is removed.
```
tmathern:
```suggestion
`to_claim()` still runs the old translation logic for anything left in `definition` (JSON-set title/format, `definition.assertions`, `definition.ingredients`, `definition.thumbnail`, actions, etc.). It clones `pre_claim` (if present) as its starting point, then does its usual pass over `definition`, skipping anything that's already been added directly. Both models coexist in the same `Builder`.
```
(Shorter sentences are easier to read :) )

docs/builder-pre-claim.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Why
+
+Some assertions need to reference other assertions by hashed URI before the manifest is signed (this is an increasingly common pattern in the C2PA spec — e.g. an action referencing the ingredient it operated on). Under the old model, an assertion added via `add_assertion` doesn't have a knowable JUMBF URI/hash until `to_claim()` runs, so there was no way to embed such a reference directly; call sites either duplicated logic or referenced things positionally/by id, resolved later.
```
tmathern:
```suggestion
Some assertions need to reference other assertions by hashed URI before the manifest is signed (e.g. an action referencing the ingredient it operated on). Under the previous model, an assertion added via `add_assertion` doesn't have a knowable JUMBF URI/hash until `to_claim()` runs, so there was no way to embed such a reference directly; call sites either duplicated logic or referenced things positionally/by id, resolved later.
```

docs/builder-pre-claim.md
```diff
@@ -0,0 +1,82 @@
...
+
+Adds an ingredient assertion built from an already-read/validated `Reader`, without constructing a full `Ingredient`. It merges the reader's own manifest data into the pre-claim's ingredient store, fills in `activeManifest`/`claimSignature`/`validationResults` on the assertion, and returns the assertion's `HashedUri`.
+
+`T` accepts, via `TryInto<IngredientAssertion>`:
+- an `IngredientAssertion` (owned) or `&IngredientAssertion`
+- JSON text (`&str`/`String`) or `serde_json::Value`, matching the assertion's own v3 JSON schema (`dc:title`, `dc:format`, `relationship`, `thumbnail`, etc.)
+
+`redactions` is a list of JUMBF URIs of assertions to strip from the ingredient's manifest chain as it's merged in — find the URI via the `Reader` (`reader.active_label()` + `jumbf::labels::to_assertion_uri(label, assertion_label)`), then pass it here. The caller is responsible for separately recording a `c2pa.redacted` action documenting why, same as the existing redaction pattern elsewhere in `Builder` — this method only does the mechanical redaction.
```
tmathern: With this change, actually we could consider an add_reaction API which takes a URI and adds the action, no?

sdk/src/builder.rs
```diff
@@ -1561,19 +1798,26 @@ impl Builder {
...
-                &claim_generator,
-                definition.vendor.as_deref(),
-                self.claim_version().into(),
-            ),
-        }
-        .with_context(self.context.clone());
+        // If prepare_resources() pre-populated a claim, clone it so pre-registered
+        // EmbeddedData assertions (thumbnail, ingredient data) are already in the claim
+        // with stable JUMBF URIs. Otherwise build a fresh claim the normal way.
+        let mut claim = if let Some(pre) = &self.pre_claim {
```
tmathern, line 1914: pre_claim is v2 claim only (there is a check about this somewhere).

Are there potential side-effects here if we deal with a v1 claim? Especially around the claim generator then since it has special handling in the else?

I think it is fine as-is, but wanted to double-check that we don't change/loose things we shouldn't be losing.

sdk/src/builder.rs
```diff
@@ -827,15 +839,33 @@ impl Builder {
...
+        let instance_id = self.definition.instance_id.clone();
+        if self.claim_version() >= 2 {
+            let mut data = Vec::new();
+            stream.read_to_end(&mut data)?;
+            let hu = self.add_assertion_with_ref(&EmbeddedData::new(
+                labels::CLAIM_THUMBNAIL,
+                format_to_mime(&format),
+                data,
+            ))?;
+            let alg = self.pre_claim.as_ref().map(|p| p.alg().to_string());
```
tmathern, line 850: Since we have "claim" vs "pre_claim" branches, are there any risks we get algorithms mismatch depending on the order of calls on a Builder? 

Everything pre_claim should be handled with a grain of salt when the caller checks or changes those, right (since it is pre-claim/pre-signing, and only signing writes things in stone)? Then we should emphasize that in the docs.

sdk/src/builder.rs
```diff
@@ -449,6 +449,18 @@ pub struct Builder {
     // Accumulator to assist with hashing mdat bytes for BmffHash assertions.
     #[serde(skip)]
     bmff_hasher: MerkleAccumulator,
+
+    // Pre-populated claim built by add_resource(). When Some, to_claim() clones this as its
+    // starting claim so that assertions added by add_resource() are already present.
+    #[serde(skip)]
```
tmathern, line 455: What happens if you'd serialize a Builder that has data added through the pre_claim APIs?

I imagine this can happen server-side if Builder get passed around in-between server machines.

sdk/src/hashed_uri.rs
```diff
@@ -84,6 +84,19 @@ impl HashedUri {
     }
 }
 
+// `salt` is deliberately excluded: it's `skip_serializing`/`skip_deserializing` (a
+// signing-time artifact, not part of a HashedUri's wire identity), so a `HashedUri`
+// obtained in-memory before signing (which carries the real salt) and the same
+// reference read back after a round trip (which never has one) should still compare
+// equal. Derived `PartialEq` would include `salt` and break that.
+impl PartialEq for HashedUri {
```
tmathern, line 92: Can we have the case were URIs would legitimately only differ by the salt, and now they collide because the salt is excluded?
(Wrongly considered equal when they should be different)

sdk/src/builder.rs
```diff
@@ -1253,6 +1423,73 @@ impl Builder {
...
+                .ingredients
+                .iter()
+                .find(|ing| ing.data_ref().is_some_and(|r| r.identifier == sanitized_id))
+            {
+                let fmt = ing.data_ref().map(|r| r.format.clone()).unwrap_or_default();
+                (labels::EMBEDDED_DATA, fmt)
+            } else {
+                // Resource not yet referenced in the definition — fall back to ResourceStore
+                // so callers that set definition refs after add_resource still work.
+                if self.resources.exists(&sanitized_id) {
```
tmathern, line 1578: Is this same or less or more strict than `self.resource_map` check? I suspect the intent of the check is the same.
(This is important to make sure we find the right resources with the right ids no matter how we check here).

sdk/src/ingredient.rs
```diff
@@ -1283,20 +1292,27 @@ impl Ingredient {
...
-                1 => claim.add_databox(
-                    &data_ref.format,
-                    box_data.into_owned(),
-                    data_ref.data_types.clone(),
-                )?,
-                _ => {
-                    let embedded_data = EmbeddedData::new(
-                        labels::EMBEDDED_DATA,
-                        format_to_mime(&data_ref.format),
+            let hash_uri = if let Some(h) = data_ref.hash.as_ref() {
```
tmathern, line 1295: Why not gate on the resource map that was built and is used in pre-claims paths?

sdk/src/builder.rs
```diff
@@ -449,6 +449,18 @@ pub struct Builder {
     // Accumulator to assist with hashing mdat bytes for BmffHash assertions.
     #[serde(skip)]
     bmff_hasher: MerkleAccumulator,
+
+    // Pre-populated claim built by add_resource(). When Some, to_claim() clones this as its
```
tmathern, line 453: Can this mutate somehow after a to_claim clone? Can we miss data that was added later here?

sdk/src/builder.rs
```diff
@@ -449,6 +449,18 @@ pub struct Builder {
...
+    // starting claim so that assertions added by add_resource() are already present.
+    #[serde(skip)]
+    pre_claim: Option<Claim>,
+
+    // Maps each user-provided resource id (e.g. "thumbnail.jpg") → ResourceRef with the
+    // JUMBF URI and hash of the corresponding assertion in pre_claim.
+    // Used by to_claim() to skip re-adding assertions that add_resource() already placed
+    // in pre_claim, and to pass updated ResourceRefs to ingredient.add_to_claim().
+    #[serde(skip)]
+    pub(crate) resource_map: HashMap<String, ResourceRef>,
```
tmathern, line 463: Are there cases during the lifetime of a Builder (including serialization/deserialization) where we need to go through the map to update it and make sure all string -> ResourceRef are current?

sdk/src/builder.rs
```diff
@@ -827,15 +839,33 @@ impl Builder {
...
-        self.definition.thumbnail = Some(ResourceRef::new(
-            format,
-            self.definition.instance_id.clone(),
-        ));
+        let format: String = format.into();
+        let instance_id = self.definition.instance_id.clone();
+        if self.claim_version() >= 2 {
+            let mut data = Vec::new();
+            stream.read_to_end(&mut data)?;
+            let hu = self.add_assertion_with_ref(&EmbeddedData::new(
```
```suggestion
            let hashed_uri = self.add_assertion_with_ref(&EmbeddedData::new(
```
(Opinion, makes code easier to read)

sdk/src/builder.rs
```diff
@@ -929,6 +959,24 @@ impl Builder {
...
+        // Actions assertion onto `definition.assertions` — `to_claim()` would sign both,
+        // producing a duplicate `c2pa.actions` assertion.
+        if let Some(pre) = self.pre_claim.as_mut() {
+            let existing = pre.claim_assertion_store().iter().find(|ca| {
+                let label = ca.label_raw();
+                let (match_label, _version, _instance) = parse_label(&label);
+                match_label.starts_with(Actions::LABEL)
+            });
+            if let Some(ca) = existing {
+                let mut actions = Actions::from_assertion(ca.assertion())?;
```
tmathern, line 971: If from_assertion hits a serialization path, can't these make the hashed uris handed out before obsolete? Would someone still be holding on possibly stale URIs here? E.g. uris `add_assertion_with_ref` handed out?

sdk/src/builder.rs
```diff
@@ -1609,8 +1853,9 @@ impl Builder {
 
         if let Some(thumb_ref) = definition.thumbnail.as_ref() {
             // Setting the format to "none" will ensure that no claim thumbnail is added
-            if thumb_ref.format != "none" {
-                //let data = self.resources.get(&thumb_ref.identifier)?;
+            if thumb_ref.format != "none" && !self.resource_map.contains_key(&thumb_ref.identifier)
+            {
+                // Not pre-registered via add_resource — add assertion normally.
```
tmathern, line 1968: I would remove the comment, the if statement is self-explanatory

tmathern (commented):
I just have a few questions on the code!

docs/builder-pre-claim.md
scouten-adobe: Why not automatically add the `c2pa.redacted` action? Leaving it to the caller seems error-prone.

scouten-adobe: Also, I'd argue for removing the `Option<...>`. An empty `Vec<>` does no heap allocation, so it's very cheap.

docs/builder-pre-claim.md
```diff
@@ -0,0 +1,82 @@
...
+
+For an ingredient with no manifest data of its own, skip this method: build the `IngredientAssertion` directly and add it with `add_assertion_with_ref`.
+
+### `Action::add_ingredient_ref(self, ingredient: HashedUri) -> Self`
+
+Adds an ingredient reference directly from a `HashedUri` (e.g. the one returned by `add_ingredient_with_reader`), instead of by id resolved later in `to_claim()`. Matches the existing `add_ingredient_id` incremental pattern — call it once per ingredient.
+
+### `IngredientAssertion` is now a public type
+
+`assertions::Ingredient` (re-exported as `IngredientAssertion`) was previously `pub(crate)` only. It's now public, and gained `Clone`, plus `TryFrom<&str>` / `TryFrom<String>` / `TryFrom<&String>` / `TryFrom<serde_json::Value>` / a `Deserialize` impl for its v3 JSON schema (this re-encodes the input as CBOR and delegates to the existing `from_assertion` decoder, so there's one source of truth for the v3 field set).
```
scouten-adobe:
```suggestion
`assertions::Ingredient` (re-exported as `IngredientAssertion`) was previously `pub(crate)` only. It's now public, and gained `Clone`, plus `TryFrom<&str>` / `TryFrom<String>` / `TryFrom<&String>` / `TryFrom<serde_json::Value>` / a `Deserialize` impl for its v3 JSON schema. (This re-encodes the input as CBOR and delegates to the existing `from_assertion` decoder, so there's one source of truth for the v3 field set.)
```

The parenthetical comment is a complete sentence, so should be punctuated accordingly.

sdk/src/assertions/actions.rs
```diff
@@ -756,6 +756,20 @@ impl Action {
         self
     }
 
+    /// Adds an ingredient [`HashedUri`] reference to this action directly.
+    ///
+    /// Use this when building actions via `Builder::add_assertion_with_ref` — the
```
scouten-adobe, line 761:
```suggestion
    /// Use this when building actions via [`Builder::add_assertion_with_ref`] — the
```

sdk/src/assertions/ingredient.rs
```diff
@@ -104,6 +106,65 @@ impl Serialize for Ingredient {
...
+
+/// Clone by value, so `&Ingredient` can be used anywhere an owned value is expected (e.g.
+/// [`crate::Builder::add_ingredient_with_reader`]).
+impl From<&Ingredient> for Ingredient {
+    fn from(value: &Ingredient) -> Self {
+        value.clone()
+    }
+}
+
+/// Implement TryFrom for &str (JSON string).
```
scouten-adobe, line 132: FWIW these doc comments are not required even when strict doc-comment mode is on. Not a strong preference, but I generally avoid doc comments that restate what the `impl ... for ...` syntax will say.

sdk/src/assertions/mod.rs
```diff
@@ -50,10 +50,10 @@ mod exif;
 #[allow(deprecated)]
 pub use exif::Exif;
 
-#[allow(dead_code)] // will become public later
+#[allow(dead_code)]
```
scouten-adobe, line 53: Is this `dead_code` bypass still needed? Can it be more narrowly-scoped (i.e. to a single function)?

sdk/src/builder.rs
scouten-adobe: I'd be surprised if hashed URIs become obsolete with the exception of the hard-binding assertions.

sdk/src/builder.rs
```diff
@@ -3406,6 +3709,7 @@ impl Builder {
     /// * `reader` - The Reader to get the ingredient from.
     /// # Returns
     /// * A reference to the added ingredient.
+    #[deprecated(note = "Use `Builder::add_ingredient_from_archive` instead.")]
```
scouten-adobe, line 3822:
```suggestion
    #[deprecated(note = "Use `Builder::add_ingredient_from_archive` instead. Will be deprecated on or after 2026-09-27.")]
```

(See https://github.com/contentauth/c2pa-rs/blob/main/docs/deprecation-policy.md#rust.)

sdk/src/builder.rs
```diff
@@ -9521,6 +9821,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(deprecated)]
```
scouten-adobe, line 9934: Can we scope this to the specific tests that call deprecated functions? Otherwise, I think this `#[allow(...)]` will outlive the eventual deletions.

scouten-adobe (changes requested):
Thanks @gpeacock. I traced the core mechanism and it holds up: `add_assertion_impl` folds the salt into the box hash at insertion time and persists it on the `ClaimAssertion`, so the `HashedUri` you hand back really is stable through signing. The direction is good; most of what follows is about the non-default configurations and a couple of loose ends. Several of these corroborate questions @tmathern already raised inline, and I've left some narrower notes in my own review.
## Correctness
1. **v1 claims lose `claim_generator`.** @tmathern flagged this on `to_claim()` (builder.rs:1804) and wondered whether it's fine as-is — I think it isn't. `add_assertion_with_ref` / `add_embedded_data` / `add_ingredient_with_reader` aren't gated to v2+ the way `set_thumbnail`/`add_resource` are, and `ensure_pre_claim` builds the pre-claim with an empty generator. When `to_claim()` clones the pre-claim it skips the branch that sets the v1 `claim_generator` string, and the later loop only appends v2-style `claim_generator_info` — so a v1 claim built through these APIs ends up with an empty generator. Let's either gate these methods to v2+ or set the generator on the cloned claim.
2. **Algorithm mismatch by call order.** This is exactly @tmathern's question on builder.rs:852. `ensure_pre_claim` uses the default alg, so assertions added early get hashed under it; a later `definition.hash_alg` change overwrites `claim.alg` *after* the clone in `to_claim()`, leaving those assertions hashed under a different alg than the claim advertises. `ensure_pre_claim` should honor `hash_alg` (and probably reject a later change). Agreed with her broader point that anything touched pre-signing needs a "not written in stone until signing" caveat in the docs.
3. **Serializing mid-build silently drops pre-claim work.** @tmathern raised this on the field decl (builder.rs:455) re: passing a `Builder` between server machines, and again on `resource_map` currency (453/463). `pre_claim` and `resource_map` are `#[serde(skip)]`, so a serialize-then-resume after `add_assertion_with_ref` yields a silently incomplete manifest. Given the stakes I'd want `Serialize` to error (or at minimum loudly warn) when `pre_claim.is_some()`, rather than relying on the "interactive only" doc note.
4. **`add_action` merge vs. "URIs don't go obsolete".** In my inline note I said I'd be surprised if handed-out URIs became obsolete outside the hard-binding assertions — but this is the one path that contradicts that, and it's what @tmathern was getting at on builder.rs:973. Merging into an existing pre-claim Actions assertion calls `replace_assertion`, which rewrites that assertion's bytes and hash, invalidating any `HashedUri` previously returned for it. Worth either documenting the exception clearly or reconsidering the merge.
5. **Salt-excluded equality.** @tmathern's question on hashed_uri.rs:92 is the right one: can two URIs legitimately differ *only* by salt and now compare equal? The rationale for excluding salt is sound and there's no `Hash` derive to break, but this changes equality crate-wide, not just on the new path — I'd like us to convince ourselves nothing in validation/dedup relies on salt-sensitive comparison before we lock it in.
## Cleanup (new — not yet covered inline)
- **Doc drift:** the `pre_claim` field comment says "built by `add_resource()`" (it's several methods), and both it and `set_data_direct` reference a `prepare_resources()` that no longer exists in the branch. Leftovers from an earlier iteration.
## Smaller items (see my inline comments)
- Redaction ergonomics: I'd auto-add the `c2pa.redacted` action rather than leaving it to the caller, and drop the `Option<Vec<...>>` in favor of a plain `Vec` (empty vec doesn't allocate).
- The `deprecated` note on `add_ingredient_from_reader` should carry a date per our [deprecation policy](https://github.com/contentauth/c2pa-rs/blob/main/docs/deprecation-policy.md#rust), and the blanket `#[allow(deprecated)]` should be scoped to the specific tests so it doesn't outlive the deletions.
- The `dead_code` bypass on the `ingredient` module and the restate-the-signature doc comments on the new `TryFrom`/`Deserialize` impls can both be tightened.
- +1 to @tmathern's readability nits (`hu` → `hashed_uri`, dropping the self-explanatory `else` comment) and worth considering her `add_reaction`-style convenience API as a follow-up.
## Tests
Good coverage on the happy paths and the duplicate-actions / auto-parent cases. The gaps line up with the above: v1 via `add_assertion_with_ref` (#1), custom `hash_alg` + pre-claim (#2), and serialize-then-resume (#3).

PR #2259, opened 2026-06-30
fix(riff): support image/x-webp in WebP signing path

ZakryDat:
_This PR extends RIFF/WebP signing support to handle WebP MIME aliases, specifically image/x-webp. It adds image/x-webp to the RIFF handler’s supported types and updates the WebP detection condition in the RIFF writer so WebP-specific VP8X/XMP handling runs for both extension-style and MIME-style WebP identifiers._

## Checklist


sdk/src/asset_handlers/riff_io.rs
```diff
@@ -34,11 +34,13 @@ use crate::{
     },
 };
 
-static SUPPORTED_TYPES: [&str; 12] = [
+static SUPPORTED_TYPES: [&str; 14] = [
     "avi",
     "wav",
     "webp",
+    "x-webp",
```
gpeacock: did anyone actually use an x-webp extension? I don't think we need to add this here.

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -183,7 +185,7 @@ where
         if let Some(xmp) = xmp_data {
             if is_riff_chunk && !xmp.is_empty() {
                 // if this is a webp doc we must also update VP8X
-                if format == "webp" {
+                if format == "webp" || format == "x-webp" || format == "image/webp" || format == "image/x-webp" {
```
gpeacock: let's just add the image/x-webp mime type. I don't think anyone used x-web as an extension

gpeacock (changes requested):

ZakryDat:
good catch - I removed the unnecessary extras, limit the change to just adding the image/x-webp mimetype

ZakryDat:
change is now minimal - just adding the image/x-webp mimetype.

ZakryDat:
@gpeacock sorry to raise this again, but hoping it can make it into the next release

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33196167913](https://github.com/contentauth/c2pa-rs/actions/runs/33196167913).

PR #2264, opened 2026-07-06
fix: ensure created has a dst and don't add parent when actions already exist

gpeacock:
Do not automatically add a parent ingredient and opened action if a created or opened action already exists.
Ensures that c2pa_created has an associated digitalSourceType

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -1823,7 +1823,33 @@ impl Builder {
             self.intent(),
             Some(BuilderIntent::Edit | BuilderIntent::Update)
         );
-        if auto_parent && !self.definition.ingredients.iter().any(|i| i.is_parent()) {
+        // Don't auto-add a parentOf ingredient if the user's manifest already declares a
+        // c2pa.created or c2pa.opened action — those are mutually exclusive with auto-parent.
+        let has_created_or_opened = self.definition.assertions.iter().any(|a| {
+            if !a.label.starts_with(crate::assertions::Actions::LABEL) {
+                return false;
+            }
+            let AssertionData::Json(value) = &a.data else {
+                return false;
+            };
+            value
+                .get("actions")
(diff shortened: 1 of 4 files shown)


tmathern (approved):

sdk/src/claim.rs
```diff
@@ -2274,10 +2274,10 @@ impl Claim {
                             "verify_actions"
                         )
                         .validation_status(validation_status::ASSERTION_ACTION_INGREDIENT_MISMATCH)
-                        .failure_no_throw(
```
ok-nick, line 2277: When do you decide to use `failure_no_throw` vs. `failure`?

sdk/src/builder.rs
```diff
@@ -1823,7 +1823,27 @@ impl Builder {
             self.intent(),
             Some(BuilderIntent::Edit | BuilderIntent::Update)
         );
-        if auto_parent && !self.definition.ingredients.iter().any(|i| i.is_parent()) {
+        // Don't auto-add a parentOf ingredient if the user's manifest already declares a
+        // c2pa.created or c2pa.opened action — those are mutually exclusive with auto-parent.
```
ok-nick, line 1827: Should we hard error if the user explicitly sets an intent but we ignore it because we know they are doing something different?

ok-nick (approved):

sdk/src/claim.rs
gpeacock: The problem is that we use the same code for signing post validate and for reading validation. For signing we'd want to hard fail, for reading we want to capture and report all errors, so we need to rework this code to handle both cases.

sdk/src/builder.rs
gpeacock: This probably needs more thought. Should an intent require that the user NOT try do to anything on their own, or should it just verify that the intent is fulfilled?

PR #2265, opened 2026-07-07
ci: pin GitHub Actions to full commit SHAs

scouten-adobe:
## Summary

Pins every third-party GitHub Action referenced in the workflows to a full 40-character commit SHA instead of a mutable tag or branch ref, following [GitHub's supply-chain hardening guidance](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). Each pin carries a trailing `# version` comment so it stays auditable and Dependabot can keep updating it.

All 131 active `uses:` lines across 9 workflow files are now SHA-pinned; no unpinned refs remain.

## Behavior-encoding actions

A few actions encode behavior in the ref name, so pinning to a SHA requires moving that behavior into an explicit input:

| Action | Was | Now |
|---|---|---|
| `dtolnay/rust-toolchain` | `@stable` / `@beta` / `@nightly` / `@master` | pinned to the **master** SHA; channel moved to an explicit `toolchain:` input |
| `taiki-e/install-action` | `@cargo-llvm-cov` | pinned to the **v2** SHA + `tool: cargo-llvm-cov` |
| `dtolnay/install` | `@cargo-docs-rs` | pinned to that ref's SHA (crate is hardcoded in the ref, no input needed) |

For `rust-toolchain` steps that already passed a `toolchain:` (e.g. matrix versions, pinned nightlies), only the ref was swapped — the existing input is untouched.

## Verification

- All workflow files parse as valid YAML.
- `actionlint` reports no new issues (only pre-existing `SC2086` shellcheck info warnings on unrelated `run:` scripts).
- Every SHA was resolved from the upstream repo at the tag/branch previously referenced.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -37,11 +37,11 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
       - name: Check codspeed build cache
         id: cache-codspeed-build
-        uses: actions/cache/restore@v5
+        uses: actions/cache/restore@caa296126883cff596d87d8935842f9db880ef25 # v5
         with:
           path: target/codspeed/
           key: codspeed-output-${{ runner.os }}-${{ hashFiles('**/Cargo.lock', 'sdk/benches/**/*.rs', 'sdk/src/**/*.rs') }}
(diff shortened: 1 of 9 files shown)


tmathern (approved):

tmathern:
Approving with the caveat that those are amtching recent commit SHA-1s...

ale-adobe (approved):
LGTM. Didn't check the actual SHA1 strings, of course, but trusting that they are correct. 😄

ok-nick (approved):

PR #2271, opened 2026-07-08
fix: Harden against arbitrary Local file read and exfiltration via manifest relative path traversal attacks

ssanthosh:
# Security Fix: Path Traversal via Manifest Ingredient `data.identifier`

## Vulnerability

`c2patool` can be tricked into reading attacker-chosen local files outside the
manifest directory when signing an asset with an untrusted manifest definition.
A crafted `manifest.json` using a path-traversal payload such as `../secret.txt`
or `..\password.txt` in an ingredient `data.identifier` (or `thumbnail`,
`manifest_data`, `ocsp_responses`) causes the tool to resolve that path
relative to the manifest directory, read the target file from disk, and embed
its contents silently into the generated signed output.

Because `c2patool` runs with the invoking user's permissions, any file readable
by that user is in scope — SSH keys, cloud credentials, `.env` files, CI
secrets, etc. The extracted bytes are embedded as an `EmbeddedData` /
databox assertion inside the signed asset, enabling exfiltration through
normal signing and file-sharing workflows.

### Root cause

`sdk/src/resource_store.rs::ResourceStore::add()` already sanitized identifiers
with `sanitize_archive_path` before writing files, but the **read** side of the same store
was left unprotected. Four methods used `base.join(id)` verbatim with the
caller-supplied identifier:

| Method | Sink |
|---|---|
| `ResourceStore::get()` | `read(base.join(id))` — reads file into a `Cow<Vec<u8>>` |
| `ResourceStore::write_stream()` | `File::open(base.join(id))` streamed to caller |
| `ResourceStore::exists()` | `base.join(id).exists()` — leaks file existence |
| `ResourceStore::path_for_id()` | returns `base.join(id)` unfiltered |

When `c2patool` invokes `builder.set_base_path(manifest_dir)`
([cli/src/main.rs:1112](cli/src/main.rs#L1112)) and then signs, the ingredient
resolution path in [sdk/src/ingredient.rs:1271](sdk/src/ingredient.rs#L1271)
calls `get_resource(&data_ref.identifier)`, which falls through to
`ResourceStore::get()` when the id is not in-memory. With `id = "../secret.txt"`,
`base.join(id)` escapes `manifest_dir` and the target file is read into the
databox that gets embedded and signed.

### Attack

```json
{
  "claim_generator": "attacker",
  "ingredients": [{
    "title": "innocent.jpg",
    "format": "image/jpeg",
    "relationship": "componentOf",
    "data": {
      "format": "text/plain",
      "identifier": "../../../home/victim/.ssh/id_rsa"
    }
  }]
}
```

Running `c2patool -m manifest.json -o out.jpg photo.jpg` reads
`~/.ssh/id_rsa`, embeds its contents into `out.jpg` as a signed databox, and
publishes it under the victim's signature.

## Fix

Route every disk-fallback read through the existing
`sanitize_archive_path` sanitizer (already imported by
`resource_store.rs`). The sanitizer rejects `..`, absolute prefixes, Windows
drive/UNC prefixes, and backslash separators on all platforms — identical to
the guard applied to `ResourceStore::add()`.

- `get()` and `write_stream()` — sanitize first; on rejection, return
  `Error::ResourceNotFound(id)` so callers see a normal missing-resource error
  instead of an information-disclosure crash.
- `exists()` — sanitize first; on rejection, skip the disk probe and continue
  checking the resolver / in-memory map (returns `false` for pure traversal
  ids). Prevents existence-oracle leaks about files outside `base_path`.
- `path_for_id()` — sanitize first; on rejection, return `None`.

The fix is scoped to the disk-fallback branch. In-memory resource lookups
(which use the identifier as a `HashMap` key) are unaffected, so legitimate
JUMBF URIs, `self#jumbf=...` references, and `HashMap`-registered thumbnails
continue to work unchanged.

All four read methods now route through a single `resolve_within_base()` helper that applies **two layers of defense**:

1. **String sanitization** (existing `sanitize_archive_path`) — rejects `..` traversal, absolute paths, backslash separators, and Windows drive/UNC prefixes before touching the filesystem.
2. **Symlink containment** (new) — if the resolved target exists, it is `canonicalize()`d (following symlinks) and verified to still live within the canonicalized `base_path`. This blocks an innocuously-named symlink (e.g. a bundle shipping `thumb.jpg → /etc/passwd`) that string checks alone would miss. Both sides are canonicalized, so a legitimately symlinked base (`/tmp → /private/tmp` on macOS) and symlinks that stay *inside* the base are **not** falsely rejected — only escapes are blocked.

## Tests Added

Added to `sdk/src/resource_store.rs::tests::zip_slip_tests` (gated
`#[cfg(all(feature = "file_io", not(target_arch = "wasm32")))]`):

| Test | Validates |
|---|---|
| `get_with_base_path_rejects_parent_dir_traversal` | `get("../secret.txt")` returns `ResourceNotFound`; secret file outside base is never read |
| `get_with_base_path_rejects_absolute_path` | `get("/etc/passwd")` returns `ResourceNotFound` |
| `get_with_base_path_rejects_backslash_traversal` | `get("..\\secrets\\password.txt")` rejected end-to-end on all hosts |
| `get_with_base_path_accepts_normal_resource` | Legitimate `thumb.jpg` still resolves via disk-fallback |
| `write_stream_with_base_path_rejects_parent_dir_traversal` | Streaming API is closed too; output stream remains empty on rejection |
| `exists_with_base_path_rejects_parent_dir_traversal` | `exists` returns `false` for traversal / backslash / absolute ids |
| `path_for_id_rejects_traversal` | Path-derivation helper returns `None` for hostile ids, `Some` for legitimate ids |
| `get_with_base_path_accepts_nested_resource` | legitimate nested `resources/thumbs/thumb.jpg` still reads
| `get_with_base_path_rejects_nested_traversal` | buried `..` (`resources/../../secret.txt`) rejected
| `get_with_base_path_rejects_escaping_symlink` / `| ..._allows_internal_symlink` (`#[cfg(unix)]`)
| `exists_with_base_path_rejects_escaping_symlink` | (`#[cfg(unix)]`)
| `write_stream_with_base_path_rejects_escaping_symlink` | (`#[cfg(unix)]`)

## Checklist

### DIFF
--- a/sdk/src/resource_store.rs
+++ b/sdk/src/resource_store.rs
@@ -336,12 +336,12 @@ impl ResourceStore {
             match self.base_path.as_ref() {
                 Some(base) => {
                     // Reject identifiers that would escape base_path (traversal, absolute
-                    // paths, backslash separators). Prevents attacker-supplied manifest
-                    // definitions from exfiltrating files outside the manifest directory.
-                    let sanitized_id = sanitize_archive_path(id)
+                    // paths, backslash separators, or symlinks pointing outside base_path).
+                    // Prevents attacker-supplied manifest definitions from exfiltrating
+                    // files outside the manifest directory.
+                    let path = resolve_within_base(base, id)
                         .map_err(|_| Error::ResourceNotFound(id.to_string()))?;
                     // read the file, save in Map and then return a reference
-                    let path = base.join(&sanitized_id);
                     let value = read(&path).map_err(|_| {
                         Error::ResourceNotFound(path.to_string_lossy().into_owned())
                     })?;
@@ -378,10 +378,9 @@ impl ResourceStore {
             match self.base_path.as_ref() {
                 Some(base) => {
                     // Reject identifiers that would escape base_path (see get()).
-                    let sanitized_id = sanitize_archive_path(id)
+                    let path = resolve_within_base(base, id)
                         .map_err(|_| Error::ResourceNotFound(id.to_string()))?;
                     // read from, the file to stream
-                    let path = base.join(&sanitized_id);
                     let mut file = std::fs::File::open(path)?;
                     return std::io::copy(&mut file, &mut stream).map_err(Error::IoError);
                 }
@@ -412,25 +411,25 @@ impl ResourceStore {
     }
 
     /// Returns `true` if the resource has been added or exists as a file or in the resolver.
-    pub fn exists(&self, id: &str) -> bool {
-        if self.resources.contains_key(id) {
+    pub fn exists(&self, path: &str) -> bool {
+        if self.resources.contains_key(path) {
             return true;
         }
 
         #[cfg(feature = "file_io")]
         if let Some(base) = self.base_path.as_ref() {
             // Skip disk probes for identifiers that would escape base_path — a
-            // hostile id like `../etc/passwd` must not leak existence via this API.
-            if let Ok(sanitized_id) = sanitize_archive_path(id) {
-                let path = base.join(&sanitized_id);
-                if path.exists() {
+            // hostile id like `../etc/passwd`, or a symlink pointing outside
+            // base_path, must not leak existence via this API.
+            if let Ok(resolved) = resolve_within_base(base, path) {
+                if resolved.exists() {
                     return true;
                 }
             }
         }
 
         if let Some(resolver) = &self.resolver {
-            if resolver.has(id) {
+            if resolver.has(path) {
                 return true;
             }
         }
@@ -453,8 +452,8 @@ impl ResourceStore {
     #[cfg(feature = "file_io")]
     // Returns the full path for an ID.
     pub fn path_for_id(&self, id: &str) -> Option<PathBuf> {
-        let sanitized_id = sanitize_archive_path(id).ok()?;
-        self.base_path.as_ref().map(|base| base.join(&sanitized_id))
+        let base = self.base_path.as_ref()?;
+        resolve_within_base(base, id).ok()
     }
 }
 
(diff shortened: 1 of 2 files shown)


sdk/src/resource_store.rs
```diff
@@ -335,11 +335,15 @@ impl ResourceStore {
...
+                    // definitions from exfiltrating files outside the manifest directory.
+                    let sanitized_id = sanitize_archive_path(id)
+                        .map_err(|_| Error::ResourceNotFound(id.to_string()))?;
                     // read the file, save in Map and then return a reference
-                    let path = base.join(id);
-                    let value = read(path).map_err(|_| {
-                        let path = base.join(id).to_string_lossy().into_owned();
-                        Error::ResourceNotFound(path)
+                    let path = base.join(&sanitized_id);
+                    let value = read(&path).map_err(|_| {
```
tmathern, line 373: Is the goal here also to catch symlinks pointing to other files? (You know, name a link differently to the actual file it points to?).

sdk/src/resource_store.rs
```diff
@@ -618,5 +630,106 @@ mod tests {
             let result = store.add("..\\..\\etc\\passwd", b"attacker data".to_vec());
             assert!(result.is_err());
         }
+
```
tmathern, line 757: For tests: I would add some more tests with nested paths, symlinks if doable in tests.

sdk/src/resource_store.rs
mpgopala: Minor: Can you rename the `id` parameter to `path` or something similar? Passing id and getting back id is a bit confusing!
> if let Ok(sanitized_id) = sanitize_archive_path(id) {

mpgopala (commented):
If we want to catch symlinks, Also add tests for symlinked directories.
If we do not want to allow symlinks, modify sanitize_archive_paths to reject on a symlink.

sdk/src/resource_store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2271/commits/510305b68c02a562cb51786ff2b9f064cd2c4080

sdk/src/resource_store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2271/commits/510305b68c02a562cb51786ff2b9f064cd2c4080

sdk/src/resource_store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2271/commits/510305b68c02a562cb51786ff2b9f064cd2c4080

sdk/src/resource_store.rs
```diff
@@ -731,5 +765,127 @@ mod tests {
...
+            assert!(
+                matches!(result, Err(Error::ResourceNotFound(_))),
+                "expected ResourceNotFound, got {result:?}"
+            );
+        }
+
+        // Symlink defense-in-depth: a symlink planted inside base_path (e.g. by
+        // an extracted manifest bundle) that points to a file *outside* base_path
+        // must not be read, even though its name contains no traversal sequence.
+        #[cfg(unix)]
```
mpgopala, line 902: Why cfg(unix)? Windows supports symlinks.

mpgopala (approved):

sdk/src/resource_store.rs
ssanthosh: Tests uses `std::os::unix::fs::symlink()` fn which works only for unix platform. Fix is cross platform and we can add separate tests for windows but Creating a symlink on Windows needs privilege. Unlike Unix, CreateSymbolicLink requires either Administrator rights or Developer Mode enabled. The tests could fail on CI/CD runner. Added new symbolic link tests for windows platform with https://github.com/contentauth/c2pa-rs/pull/2271/commits/6a1cfeebbbe595db27c4c8ad500b62e5fab67df6

tmathern (approved):

caiopensrc:
- #2326

PR #2278, opened 2026-07-08
fix: Split CAWG trust config into a dedicated CawgTrust struct (CAI-12709)

scouten-adobe:
## Summary

`Settings.cawg_trust` reused the same `Trust` struct as the C2PA `Settings.trust`, so CAWG-only settings appeared on both. In particular, `verify_trust_list` is documented as *"ONLY applicable to CAWG"* yet lived on the C2PA `Trust`, where no code path reads it — setting `trust.verify_trust_list` silently had no effect.

This introduces a dedicated `CawgTrust` struct that models only the settings that apply to CAWG validation:

- **New `CawgTrust` struct** with `verify_trust_list`, `user_anchors`, `trust_anchors`, `trust_config`, `allowed_list` (its own `Default`, including the `#[cfg(test)]` fixture loading, and `SettingsValidate`).
- **`Settings.cawg_trust`** changed from `Trust` to `CawgTrust`.
- **C2PA `Trust`** slimmed to the four fields C2PA actually reads; the CAWG-only `verify_trust_list` removed.
- The shared PEM-parsing helpers (`load_trust_from_data`, `test_load_trust`) are lowered to module-private free functions — they never used `self` — so both structs reuse them without duplication.

This is the CAI-12709 follow-up to CAI-11347. On `main` today there is no `trusted_ica_issuers` field yet, so this PR is purely the struct split. When CAI-11347 lands, its `trusted_ica_issuers` field and `IcaSignatureVerifier` read site attach to `CawgTrust` (not `Trust`), and its `TO DO (CAI-12709)` marker resolves.

## Backward compatibility

Not a breaking change for CAWG configuration. Every `cawg_trust.*` setting path is preserved. A stray `trust.verify_trust_list` is now silently ignored (serde has no `deny_unknown_fields`) rather than accepted-but-ineffective — behavior is identical.

## Testing

- `cargo check` clean on `--lib`, `--workspace`, and `--features json_schema`; `cargo clippy` clean.
- `settings::` tests pass (28); `identity::` tests pass (73), including the CAWG x509 path that relies on the moved `cawg_trust` fixtures.
- Added `test_cawg_trust_is_distinct_from_c2pa_trust` locking in the split and the backward-compat path.

## Docs

No doc changes needed: `docs/context-settings.md` already documented C2PA `trust` *without* `verify_trust_list`, and CAWG trust content in `docs/cawg-identity.md` arrives with CAI-11347.


sdk/src/settings/mod.rs
```diff
@@ -87,61 +81,152 @@ pub struct Trust {
...
+    ///
+    /// <div class="warning">
+    /// Verifying trust is REQUIRED by the CAWG spec. This option should only be used for development or testing.
+    /// </div>
+    pub(crate) verify_trust_list: bool,
+    /// List of additional user-provided trust anchor root certificates as a PEM bundle.
+    pub user_anchors: Option<String>,
+    /// List of default trust anchor root certificates as a PEM bundle.
+    ///
+    /// Normally this option contains the official C2PA-recognized trust anchors found here:
```
ok-nick: CAWG trust anchors here.

sdk/src/settings/mod.rs
ok-nick: CAWG uses different config/anchors?

sdk/src/settings/mod.rs
```diff
@@ -87,61 +81,152 @@ pub struct Trust {
...
+        Ok(())
+    } else {
+        Err(Error::CoseInvalidCert)
+    }
+}
 
-            // sanity check that that is is base64 encoded and outside of certificate block
-            if !inside_cert_block && base64::decode(&l).is_ok() && !l.is_empty() {
-                found_der_hash = true;
+#[allow(clippy::derivable_impls)]
```
ok-nick, line 138: How come we need this?

sdk/src/settings/mod.rs
```diff
@@ -56,23 +56,17 @@ pub(crate) trait SettingsValidate {
...
 pub struct Trust {
-    /// Whether to verify certificates against the trust lists specified in [`Trust`]. This
-    /// option is ONLY applicable to CAWG.
-    ///
-    /// The default value is true.
-    ///
-    /// <div class="warning">
-    /// Verifying trust is REQUIRED by the CAWG spec. This option should only be used for development or testing.
-    /// </div>
-    pub(crate) verify_trust_list: bool,
```
ok-nick, line 75: 👍

sdk/src/settings/mod.rs
```diff
@@ -87,61 +81,152 @@ pub struct Trust {
...
+/// modeled separately from the C2PA [`Trust`] because several of these settings
+/// (such as `verify_trust_list`) apply only to CAWG validation and have no
+/// effect on C2PA manifest verification.
+#[cfg_attr(
+    feature = "json_schema",
+    derive(schemars::JsonSchema),
+    schemars(default)
+)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct CawgTrust {
```
ok-nick, line 208: Should #2209 be merged before this? or maybe base this branch off `cai-11347-ica-trust-infrastructure` so that the other new settings fields are put in the `CawgTrust` struct.

scouten-adobe: Good call — #2209 has now merged to main. I've merged main into this branch and routed its `trusted_ica_issuers` field into `CawgTrust` (not the C2PA `Trust`), which resolves the `TO DO (CAI-12709)` marker that PR left behind. The `IcaSignatureVerifier` read site already used `cawg_trust.trusted_ica_issuers`, so it needed no change.

Note #2209 was a breaking release (`fix!`), so this change rides in the same train.

sdk/src/settings/mod.rs
scouten-adobe: These docs were copied verbatim from the C2PA `Trust` and still pointed at the C2PA conformance trust-list, which is misleading here. `cawg_trust.*` is independent from `trust.*`: it configures the trust lists used only when validating CAWG X.509 identity signatures. Reworded the field docs accordingly and dropped the C2PA-specific link.

sdk/src/settings/mod.rs
scouten-adobe: Yes — `cawg_trust.*` is fully independent from the C2PA `trust.*` settings; it is consulted only for CAWG identity (`cawg.x509.cose`) validation. Clarified this in the field docs.

sdk/src/settings/mod.rs
scouten-adobe: It's needed on `Trust` because the `#[cfg(not(test))]` default is all-`None`, which clippy flags as derivable (the `#[cfg(test)]` variant loads bundled fixtures, so we can't just `#[derive(Default)]`). Added a comment explaining it. On `CawgTrust` the attribute was not actually needed — its `verify_trust_list: true` default is not derivable — so I removed it there.

sdk/src/settings/mod.rs
scouten-adobe: Thanks!

sdk/src/settings/mod.rs
```diff
@@ -85,80 +79,172 @@ pub struct Trust {
...
+    /// Whether to verify certificates against the trust lists specified in [`CawgTrust`].
+    ///
+    /// The default value is true.
+    ///
+    /// <div class="warning">
+    /// Verifying trust is REQUIRED by the CAWG spec. This option should only be used for development or testing.
+    /// </div>
+    pub(crate) verify_trust_list: bool,
```
ok-nick, line 216: We have `verify.verify_trust` setting specifically for C2PA. Would you prefer that we move that setting to the trust-related struct? We can do that in this PR since the setting is private (not necessarily a breaking change).

sdk/src/settings/mod.rs
```diff
@@ -85,80 +79,172 @@ pub struct Trust {
...
+    /// stripping any fragment, against the `issuer` of an ICA verifiable
+    /// credential. An issuer that is not present on this list is reported with
+    /// the informational code `cawg.ica.untrusted_issuer` for that identity
+    /// assertion and its `cawg.ica.credential_valid` success code is withheld.
+    ///
+    /// The default value is empty, meaning that NO ICA issuer is trusted. This
+    /// is a deliberate secure default: a self-issued `did:jwk` (or any other
+    /// issuer) is not trustworthy simply because its signature is
+    /// self-consistent. Populate this list with the DIDs of issuers you trust.
+    pub trusted_ica_issuers: Option<Vec<String>>,
```
ok-nick, line 265: Does the CAWG spec point to a place where we can find a list of DIDs to trust?

sdk/src/settings/mod.rs
```diff
@@ -85,80 +79,172 @@ pub struct Trust {
...
+    /// consulted only for CAWG identity validation.
+    pub user_anchors: Option<String>,
+    /// List of default trust anchor root certificates as a PEM bundle, used when
+    /// validating CAWG X.509 identity signatures.
+    pub trust_anchors: Option<String>,
+    /// List of allowed extended key usage (EKU) object identifiers (OID) that
+    /// CAWG identity certificates must have.
+    pub trust_config: Option<String>,
+    /// List of explicitly allowed CAWG identity certificates as a PEM bundle.
+    pub allowed_list: Option<String>,
```
ok-nick, line 251: Not a blocker, but it would be nice if we pointed to the URL in the CAWG spec of where to find the official values.

sdk/src/settings/mod.rs
ok-nick: These test values are for C2PA not CAWG?

ok-nick (approved):

sdk/src/settings/mod.rs
scouten-adobe: No, and it probably won't for a while yet. Known issue, but it will take some time to answer that question.

sdk/src/settings/mod.rs
scouten-adobe: Good catch — those were the shared C2PA test trust fixtures (`store.cfg` + `test_cert_root_bundle.pem`), and they turned out to be redundant on the CAWG default: the CAWG X.509 unit tests validate against `CertificateTrustPolicy::default()`, which already bundles the test credential suite (and `#[cfg(test)]` only affects in-crate unit tests — integration tests and the FFI already ran against the all-`None` production default). Removed in `d14cf00d`; the CAWG test default now mirrors production, keeping only the CAWG-specific `trusted_ica_issuers` DIDs. Full identity + settings suites still pass.

sdk/src/settings/mod.rs
scouten-adobe: Good point on the asymmetry. I'd like to keep this PR focused on the CAWG/C2PA trust split — moving `verify.verify_trust` → `trust.verify_trust` changes the settings key and would silently drop existing configs that set it (serde ignores unknown keys), so it deserves its own change (possibly with a settings version bump). Filed as CAI-13121; also flagged `verify.verify_timestamp_trust` there for consistency.

sdk/src/settings/mod.rs
scouten-adobe: Done in `f2a5733f` — the `trust_anchors` / `trust_config` / `allowed_list` docs now cite the CAWG interim trust model (§8.2.4.1): the Mozilla S/MIME root store and the IPTC Origin Verified News Publishers lists, plus the required `id-kp-emailProtection` EKU and S/MIME certificate-policy OIDs.

ok-nick (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31038256764](https://github.com/contentauth/c2pa-rs/actions/runs/31038256764).

PR #2279, opened 2026-07-09
fix: Fix new Clippy warnings for Rust 1.97.0

scouten-adobe:


ok-nick (approved):

sdk/src/claim.rs
```diff
@@ -1585,10 +1585,9 @@ impl Claim {
             hr.url()
         } else {
             // make a full path
-            if let Some(box_name) = box_name_from_uri(&hr.url()) {
+            {
```
gpeacock: It looks like we have an extra set of parenthesis  here now

tmathern (approved):

gpeacock (approved):
one issue with an extra block that I don't think is needed anymore, but other than that it looks good

sdk/src/claim.rs
scouten-adobe: Good catch. Removed.

PR #2280, opened 2026-07-09
fix: Harden against integer overflow attacks in BMFF parser

ssanthosh:
# Security Fix: BMFF Integer Overflow in `_skip_bytes` and `offset_adjust`

## Vulnerabilities

Two related integer-overflow patterns in
[sdk/src/asset_handlers/bmff_io.rs](sdk/src/asset_handlers/bmff_io.rs):

### 1. `_skip_bytes` — `u64` size cast to `i64` can seek backward

```rust
fn _skip_bytes<R: Read + Seek + ?Sized>(reader: &mut R, size: u64) -> Result<()> {
    reader.seek(SeekFrom::Current(size as i64))?;   // ← wraps for size > i64::MAX
    Ok(())
}
```

A crafted BMFF file whose box size exceeds `i64::MAX` (0x7FFFFFFFFFFFFFFF)
would produce a *backward* seek via the wrapped negative `i64`. Any parser
routine using this helper would then re-parse earlier bytes as a new box,
enabling either a crash on malformed re-parse, an infinite loop, or bypass
of manifest validation via re-interpretation. The function is currently
unused inside the crate, so this is a latent trap rather than an active
sink — but hardening it removes the footgun before a future caller re-arms
it.

### 2. `offset_adjust` truncation — `usize as i32` corrupts stco/co64/iloc

Four sites computed the file-offset shift used by
`adjust_known_offsets`:

- `save_cai_store` (C2PA embed) — [bmff_io.rs:~2111](sdk/src/asset_handlers/bmff_io.rs#L2111)
- `remove_cai_store_from_stream` — [bmff_io.rs:~2260](sdk/src/asset_handlers/bmff_io.rs#L2260)
- `embed_reference_to_stream` (XMP write) — [bmff_io.rs:~2531](sdk/src/asset_handlers/bmff_io.rs#L2531)
- `inject_placeholder` — [bmff_io.rs:~2675](sdk/src/asset_handlers/bmff_io.rs#L2675)

All four cast `usize` box sizes to `i32`:

```rust
let offset_adjust: i32 = free_box_bytes.len() as i32;        // silently truncates > 2 GiB
let pad_size: i32 = new_c2pa_box_size as i32
                  - existing_c2pa_box_size as i32;           // both operands truncated
```

`adjust_known_offsets` then applied that (potentially wrong-sign, wrong-
magnitude) adjustment to every 32-bit `stco`, 64-bit `co64`, and `iloc`
offset in the file. For a C2PA box larger than 2 GiB the truncated
`offset_adjust` produces garbage offsets across the whole output MP4 —
sample chunks point outside the media data, the file becomes structurally
invalid, and, worse, a validator can be steered into interpreting the
adjusted offsets as if they pointed at legitimate assertion content.

### `_skip_bytes` — route through `SeekFrom::Start(u64)` with `checked_add`

```rust
fn _skip_bytes<R: Read + Seek + ?Sized>(reader: &mut R, size: u64) -> Result<()> {
    let current = reader.stream_position()?;
    let target = current
        .checked_add(size)
        .ok_or_else(|| Error::InvalidAsset("BMFF skip past u64::MAX".to_string()))?;
    reader.seek(SeekFrom::Start(target))?;
    Ok(())
}
```

`SeekFrom::Start` takes a `u64` natively — no `as i64` cast anywhere on the
path. Any skip that would push the target past `u64::MAX` surfaces as
`InvalidAsset` instead of wrapping.

### `offset_adjust` — widen to `i64`, gate every producer with `try_from`

`adjust_known_offsets` now takes `adjust: i64`. Every producer replaces
`usize as i32` with `i64::try_from(usize)` and propagates `InvalidAsset` on
overflow:

```rust
let new_c2pa_box_size_i64 = i64::try_from(new_c2pa_box_size)
    .map_err(|_| Error::InvalidAsset("C2PA box too large".to_string()))?;
let offset_adjust: i64 = if end == 0 {
    new_c2pa_box_size_i64
} else {
    let existing_i64 = i64::try_from(end - start)
        .map_err(|_| Error::InvalidAsset("existing C2PA box too large".to_string()))?;
    new_c2pa_box_size_i64 - existing_i64
};
```

Inside `adjust_known_offsets`, the existing `u32::try_from(adjust.abs())` /
`u64::try_from(adjust.abs())` chains already return `InvalidAsset` for values
that cannot fit the target offset table. Widening `adjust` to `i64` means
those `try_from` guards now correctly gate the full range — previously they
only ran inside the already-truncated `i32` domain, so an oversized box
that truncated to a small positive `i32` bypassed them entirely.

### Sites touched

- [bmff_io.rs:403](sdk/src/asset_handlers/bmff_io.rs#L403) — `_skip_bytes` hardened
- [bmff_io.rs:740](sdk/src/asset_handlers/bmff_io.rs#L740) — `adjust_known_offsets` signature widened to `i64`
- [bmff_io.rs:2111](sdk/src/asset_handlers/bmff_io.rs#L2111) — C2PA embed offset_adjust
- [bmff_io.rs:2260](sdk/src/asset_handlers/bmff_io.rs#L2260) — C2PA remove offset_adjust
- [bmff_io.rs:2531](sdk/src/asset_handlers/bmff_io.rs#L2531) — XMP write offset_adjust
- [bmff_io.rs:2675](sdk/src/asset_handlers/bmff_io.rs#L2675) — `inject_placeholder` offset_adjust

## Tests Added

Four regression tests in `bmff_io::tests`:

| Test | Validates |
|---|---|
| `test_skip_bytes_large_u64_does_not_seek_backward` | Calling `_skip_bytes` with `size = i64::MAX + 1` never produces a *backward* seek (invariant: `stream_position after >= before`). Would fail pre-fix. |
| `test_skip_bytes_u64_max_returns_error_not_backward_seek` | Calling `_skip_bytes` with `size = u64::MAX` surfaces as `Err(Error::InvalidAsset(_))` — no wrap. |
| `test_adjust_known_offsets_accepts_over_i32_max_positive` | `adjust_known_offsets` accepts `adjust = i32::MAX + 1` (i64) on an empty tree/map. Pre-fix signature `i32` could not represent this value. |
| `test_adjust_known_offsets_accepts_below_neg_i32_min_negative` | Same on the negative side — a shrink adjustment magnitude beyond 2 GiB is now representable. |

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/bmff_io.rs
+++ b/sdk/src/asset_handlers/bmff_io.rs
@@ -401,7 +401,15 @@ fn box_start<R: Read + Seek + ?Sized>(reader: &mut R, is_large: bool) -> Result<
 }
 
 fn _skip_bytes<R: Read + Seek + ?Sized>(reader: &mut R, size: u64) -> Result<()> {
-    reader.seek(SeekFrom::Current(size as i64))?;
+    // `size as i64` on a value greater than i64::MAX wraps to a negative
+    // number, which would seek *backward* rather than skip forward. Compute
+    // the absolute target with checked_add on u64 and seek to it via
+    // `SeekFrom::Start`, which takes a u64 directly.
+    let current = reader.stream_position()?;
+    let target = current
+        .checked_add(size)
+        .ok_or_else(|| Error::InvalidAsset("BMFF skip past u64::MAX".to_string()))?;
+    reader.seek(SeekFrom::Start(target))?;
     Ok(())
 }
 
@@ -741,7 +749,7 @@ fn adjust_known_offsets<W: Write + CAIRead + ?Sized>(
     mut output: &mut W,
     bmff_tree: &Arena<BoxInfo>,
     bmff_path_map: &HashMap<String, Vec<Token>>,
-    adjust: i32,
+    adjust: i64,
 ) -> Result<()> {
     let start_pos = output.stream_position()?; // save starting point
 
@@ -2107,14 +2115,18 @@ impl CAIWriter for BmffIO {
         // write ContentProvenanceBox
         output_stream.write_all(&new_c2pa_box)?;
 
-        // calc offset adjustments
-        let offset_adjust: i32 = if end == 0 {
-            new_c2pa_box_size as i32
+        // calc offset adjustments. Use i64 so a box larger than i32::MAX
+        // (2 GiB) does not silently truncate and corrupt embedded offsets.
+        let new_c2pa_box_size_i64 = i64::try_from(new_c2pa_box_size)
+            .map_err(|_| Error::InvalidAsset("C2PA box too large".to_string()))?;
+        let offset_adjust: i64 = if end == 0 {
+            new_c2pa_box_size_i64
         } else {
             // value could be negative if box is truncated
             let existing_c2pa_box_size = end - start;
-            let pad_size: i32 = new_c2pa_box_size as i32 - existing_c2pa_box_size as i32;
-            pad_size
+            let existing_i64 = i64::try_from(existing_c2pa_box_size)
+                .map_err(|_| Error::InvalidAsset("existing C2PA box too large".to_string()))?;
+            new_c2pa_box_size_i64 - existing_i64
         };
 
         // write content after ContentProvenanceBox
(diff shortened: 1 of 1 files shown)


tmathern (approved):

mpgopala (approved):

tmathern:
(I pulled in main, so Clippy gets appeased)

PR #2281, opened 2026-07-09
feat(sdk): add Builder action/ingredient removal API

cdmurph32:
The retain_actions/remove_actions methods filter actions by a caller predicate, always keeping the inception action (c2pa.created/opened) at index 0, setting allActionsIncluded=false on removal, rejecting an empty result, and preserving the original .v1/.v2 label.

The retain_ingredients method prunes ingredients while keeping any referenced by a surviving action or marked parentOf (the keep predicate can only rescue an orphan, never drop a referenced one), rewriting positional c2pa.ingredient.v3__N references afterward so they don't dangle.

The API carries no domain logic; referenced-ingredient matching uses label-or-instance_id to mirror how to_claim resolves ingredientIds. Adds tests covering inception guarding, re-index holes, edited+inputTo, parentOf at a non-zero index, diamonds, nested-chain preservation, predicate rescue, and an end-to-end sign+read asserting no action malformed/ingredientMismatch errors.

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -902,6 +902,199 @@ impl Builder {
         Ok(self)
     }
 
+    /// Retains only the actions for which `keep` returns true.
+    ///
+    /// The inception action (`c2pa.created`/`c2pa.opened`) is always kept regardless of `keep`,
+    /// and is moved to index 0 if needed, so the manifest stays valid per the C2PA spec. Sets
+    /// `allActionsIncluded = false` when anything is removed. This does not touch ingredients —
+    /// call [`Builder::retain_ingredients`] (`retain_ingredients(|_| false)` to drop all orphans)
+    /// afterwards if you also want to drop ingredients now orphaned by the removed actions.
+    ///
+    /// # Arguments
+    /// * `keep` - A predicate; the action is retained when it returns true.
+    /// # Returns
+    /// * A mutable reference to the [`Builder`]. A no-op if there is no actions assertion.
+    /// # Errors
+    /// * Returns an [`Error::BadParam`] if retention would leave zero actions (a `c2pa.actions`
+    ///   assertion must have a non-empty `actions` array).
+    pub fn retain_actions<F>(&mut self, mut keep: F) -> Result<&mut Self>
+    where
+        F: FnMut(&Action) -> bool,
+    {
+        // Find the actions assertion; no-op when there isn't one.
+        let Some(pos) = self
+            .definition
+            .assertions
+            .iter()
+            .position(|a| a.label().starts_with(Actions::LABEL))
+        else {
+            return Ok(self);
+        };
+
+        // Isolate: remove, mutate a local copy, re-add. This mirrors `add_action` and keeps the
+        // `FnMut` `&Action` borrow off `self.definition`.
+        let assertion_def = self.definition.assertions.remove(pos);
+        let original_label = assertion_def.label.clone();
+        let mut actions: Actions = assertion_def.to_assertion()?;
+
+        let original_len = actions.actions.len();
+        // The C2PA spec requires exactly one c2pa.created/c2pa.opened action, so keep it
+        // regardless of the caller predicate.
+        actions.actions.retain(|a| {
+            a.action() == c2pa_action::CREATED
+                || a.action() == c2pa_action::OPENED
+                || keep(a)
+        });
+
+        if actions.actions.is_empty() {
+            return Err(Error::BadParam(
+                "retain_actions would remove every action; at least one action is required"
+                    .to_string(),
+            ));
+        }
+
+        // The inception action must be at index 0. This is usually a no-op since
+        // Actions::add_action already inserts it there.
+        if let Some(idx) = actions.actions.iter().position(|a| {
+            a.action() == c2pa_action::CREATED || a.action() == c2pa_action::OPENED
+        }) {
+            if idx != 0 {
+                let inception = actions.actions.remove(idx);
+                actions.actions.insert(0, inception);
+            }
+        }
+
+        if actions.actions.len() < original_len {
+            actions.all_actions_included = Some(false);
+        }
+
+        // Preserve the original `.v1`/`.v2` label on rebuild.
+        self.add_assertion(original_label, &actions)?;
+        Ok(self)
+    }
+
+    /// Removes the actions for which `remove` returns true; the convenience inverse of
+    /// [`Builder::retain_actions`].
+    ///
+    /// The inception action (`c2pa.created`/`c2pa.opened`) is always kept, as with
+    /// [`Builder::retain_actions`].
+    ///
+    /// # Arguments
+    /// * `remove` - A predicate; the action is removed when it returns true.
+    /// # Returns
+    /// * A mutable reference to the [`Builder`]. A no-op if there is no actions assertion.
+    /// # Errors
+    /// * Returns an [`Error::BadParam`] if removal would leave zero actions.
+    pub fn remove_actions<F>(&mut self, mut remove: F) -> Result<&mut Self>
+    where
+        F: FnMut(&Action) -> bool,
+    {
+        self.retain_actions(|a| !remove(a))
+    }
+
+    /// Retains ingredients, then rewrites positional ingredient references so surviving actions
+    /// stay valid.
+    ///
+    /// An ingredient is kept if any of the following hold:
+    /// - it is referenced by a current action (via `parameters.ingredients` URLs or the
+    ///   `ingredientIds`/`instanceId` parameters),
+    /// - it is a [`Relationship::ParentOf`] ingredient (lineage for `c2pa.opened`), or
+    /// - `keep` returns true for it.
+    ///
+    /// `keep` therefore only ever rescues an otherwise-orphaned ingredient — it can never drop a
+    /// referenced or lineage ingredient (that would break spec validation). Use `keep` for
+    /// provenance-aware policy, e.g. "this ingredient's embedded manifest chain contains a signal
+    /// of interest, so retain the whole ingredient (and its nested chain)"; the closure may inspect
+    /// the ingredient's [`manifest_data`](Ingredient::manifest_data) to decide. Prune all orphans
+    /// with `retain_ingredients(|_| false)`. Call [`Builder::retain_actions`] first if you are also
+    /// removing actions: the keep-set is computed from whatever actions currently remain.
+    ///
+    /// # Arguments
+    /// * `keep` - A predicate that can rescue an otherwise-orphaned ingredient by returning true.
+    /// # Returns
+    /// * A mutable reference to the [`Builder`].
+    /// # Errors
+    /// * Returns an [`Error`] if an assertion cannot be decoded or rebuilt.
+    pub fn retain_ingredients<F>(&mut self, mut keep: F) -> Result<&mut Self>
+    where
+        F: FnMut(&Ingredient) -> bool,
+    {
+        // Snapshot ingredient ids in positional order BEFORE pruning so the positional `__N`
+        // references can be remapped afterwards. The id is label-or-instance_id, matching how
+        // `to_claim` keys ingredients (builder.rs `ingredient_map`) and how `ingredientIds`
+        // parameters resolve.
+        let pre_filter_ids: Vec<String> = self
+            .definition
+            .ingredients
+            .iter()
+            .map(ingredient_effective_id)
+            .collect();
+
+        // Pull out the actions assertion (if any) so we can compute the referenced-ingredient
+        // keep-set and later rewrite its positional URLs. Isolating it also keeps the `keep`
+        // `&Ingredient` borrow off `self.definition`.
+        let actions_pos = self
+            .definition
+            .assertions
+            .iter()
+            .position(|a| a.label().starts_with(Actions::LABEL));
+
+        let mut keep_set: HashSet<String> = HashSet::new();
+        let mut actions_opt: Option<(String, Actions)> = None;
+        if let Some(pos) = actions_pos {
+            let assertion_def = self.definition.assertions.remove(pos);
+            let original_label = assertion_def.label.clone();
+            let actions: Actions = assertion_def.to_assertion()?;
+            for action in &actions.actions {
+                // Symbolic references: ingredientIds / org.cai.ingredientIds / instanceId /
+                // deprecated instance_id. Relationship-agnostic — any referenced ingredient is
+                // kept whatever its relationship (e.g. `c2pa.edited` -> `inputTo`).
+                for id in action_ingredient_ids(action) {
+                    keep_set.insert(id);
+                }
+                // Positional `parameters.ingredients` HashedUri references -> ingredients[N].
+                if let Some(uris) = action.parameters().and_then(|p| p.ingredients.as_ref()) {
+                    for uri in uris {
+                        let url = uri.url();
+                        let label_start = url.rfind('/').map(|p| p + 1).unwrap_or(0);
+                        let (_, idx) = parse_positional_label(&url[label_start..]);
+                        if let Some(id) = pre_filter_ids.get(idx) {
+                            keep_set.insert(id.clone());
+                        }
+                    }
+                }
+            }
+            actions_opt = Some((original_label, actions));
+        }
+
+        // Prune: keep referenced, `parentOf`, or caller-rescued ingredients; drop the rest.
+        self.definition.ingredients.retain(|ing| {
+            keep_set.contains(&ingredient_effective_id(ing))
+                || matches!(ing.relationship(), &Relationship::ParentOf)
+                || keep(ing)
+        });
+
+        // Rewrite surviving positional `__N` references to the ingredients' new positions, then
+        // re-add the actions assertion (preserving its original label).
+        if let Some((original_label, mut actions)) = actions_opt {
+            let new_ids: Vec<String> = self
+                .definition
+                .ingredients
+                .iter()
+                .map(ingredient_effective_id)
+                .collect();
+            let id_to_new_idx: HashMap<&str, usize> = new_ids
+                .iter()
+                .enumerate()
+                .map(|(i, id)| (id.as_str(), i))
+                .collect();
+            rewrite_action_ingredient_urls(&mut actions, &pre_filter_ids, &id_to_new_idx);
+            self.add_assertion(original_label, &actions)?;
+        }
+
+        Ok(self)
+    }
+
     /// Request a trusted timestamp for manifests with the given label.
     ///
     /// This only records the label on the builder. During signing, any matching manifest(s) will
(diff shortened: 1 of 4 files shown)


sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+            .assertions
+            .iter()
+            .position(|a| a.label().starts_with(Actions::LABEL))
+        else {
+            return Ok(self);
+        };
+
+        // Isolate: remove, mutate a local copy, re-add. This mirrors `add_action` and keeps the
+        // `FnMut` `&Action` borrow off `self.definition`.
+        let assertion_def = self.definition.assertions.remove(pos);
```
tmathern: If we return in error state before re-adding, the Builder is in a broken state. This might be fine if we consider this errored and broken, but it might not be fine if we want to "restore" a Builder to try to recover from the error.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+        let original_label = assertion_def.label.clone();
+        let mut actions: Actions = assertion_def.to_assertion()?;
```
tmathern: For instance, we can error out here, and the assertions would be lost due to L936 removal.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+        // The C2PA spec requires exactly one c2pa.created/c2pa.opened action, so keep it
+        // regardless of the caller predicate.
+        actions.actions.retain(|a| {
+            a.action() == c2pa_action::CREATED
+                || a.action() == c2pa_action::OPENED
+                || keep(a)
+        });
+
+        if actions.actions.is_empty() {
+            return Err(Error::BadParam(
```
tmathern: For instance, we can error out here, and the assertions would be lost due to L936 removal.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+                actions.actions.insert(0, inception);
+            }
+        }
+
+        if actions.actions.len() < original_len {
+            actions.all_actions_included = Some(false);
+        }
+
+        // Preserve the original `.v1`/`.v2` label on rebuild.
+        self.add_assertion(original_label, &actions)?;
```
tmathern: For instance, we can error out here too due to the `?`, and the assertions would be lost due to L936 removal.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+        // Rewrite surviving positional `__N` references to the ingredients' new positions, then
+        // re-add the actions assertion (preserving its original label).
+        if let Some((original_label, mut actions)) = actions_opt {
+            let new_ids: Vec<String> = self
+                .definition
+                .ingredients
+                .iter()
+                .map(ingredient_effective_id)
+                .collect();
+            let id_to_new_idx: HashMap<&str, usize> = new_ids
```
tmathern: I think we need to make sure this works when an ingredient has either only a label, only an id, or both (label prevails usually) by adding test coverage covering all those options.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+                // deprecated instance_id. Relationship-agnostic — any referenced ingredient is
+                // kept whatever its relationship (e.g. `c2pa.edited` -> `inputTo`).
+                for id in action_ingredient_ids(action) {
+                    keep_set.insert(id);
+                }
+                // Positional `parameters.ingredients` HashedUri references -> ingredients[N].
+                if let Some(uris) = action.parameters().and_then(|p| p.ingredients.as_ref()) {
+                    for uri in uris {
+                        let url = uri.url();
+                        let label_start = url.rfind('/').map(|p| p + 1).unwrap_or(0);
```
tmathern: `unwrap_or(0)` feels dangerous if URIs are malformed or have any other issue.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+    /// with `retain_ingredients(|_| false)`. Call [`Builder::retain_actions`] first if you are also
+    /// removing actions: the keep-set is computed from whatever actions currently remain.
+    ///
+    /// # Arguments
+    /// * `keep` - A predicate that can rescue an otherwise-orphaned ingredient by returning true.
+    /// # Returns
+    /// * A mutable reference to the [`Builder`].
+    /// # Errors
+    /// * Returns an [`Error`] if an assertion cannot be decoded or rebuilt.
+    pub fn retain_ingredients<F>(&mut self, mut keep: F) -> Result<&mut Self>
```
tmathern: `retain` may be somewhat confusing here, maybe call it prune instead? Because of this: ``keep` therefore only ever rescues an otherwise-orphaned ingredient` (the keep param may be confusing).

sdk/src/builder.rs
```diff
@@ -3330,6 +3523,158 @@ impl std::fmt::Display for Builder {
...
+
+            if !rewrote {
+                return action;
+            }
+
+            let label = action.action().to_owned();
+            match action.clone().set_parameter("ingredients", remapped_uris) {
+                Ok(rewritten) => rewritten,
+                Err(e) => {
+                    log::error!("failed to remap ingredient URLs on action '{label}': {e:?}");
```
tmathern: Is it OK to just log and still return the action as-is?

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+            a.action() == c2pa_action::CREATED || a.action() == c2pa_action::OPENED
+        }) {
+            if idx != 0 {
+                let inception = actions.actions.remove(idx);
+                actions.actions.insert(0, inception);
+            }
+        }
+
+        if actions.actions.len() < original_len {
+            actions.all_actions_included = Some(false);
```
tmathern: `all_actions_included` is a more recent addition I think. Does this need gating on claim_version for instance? The SDK still supports writing older claims... We could hit this case here?

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+        // regardless of the caller predicate.
+        actions.actions.retain(|a| {
+            a.action() == c2pa_action::CREATED
+                || a.action() == c2pa_action::OPENED
+                || keep(a)
+        });
+
+        if actions.actions.is_empty() {
+            return Err(Error::BadParam(
+                "retain_actions would remove every action; at least one action is required"
```
```suggestion
                "at least one action is required"
```
Makes it simpler?

sdk/src/builder.rs
```diff
@@ -3330,6 +3523,158 @@ impl std::fmt::Display for Builder {
...
+                .iter()
+                .map(|uri| match rewrite_one_ingredient_uri(uri, pre_filter_ids, id_to_new_idx) {
+                    Some(new) => {
+                        rewrote = true;
+                        new
+                    }
+                    None => {
+                        // Either no rewrite needed or the lookup failed (stale ref).
+                        let url = uri.url();
+                        let label_start = url.rfind('/').map(|p| p + 1).unwrap_or(0);
```
tmathern: Isn't this doing the same parsing as on top of the function?

sdk/src/builder.rs
```diff
@@ -9674,4 +10019,568 @@ mod tests {
...
+            .collect();
+        assert_eq!(labels, vec!["ing_keep"]);
+    }
+
+    // a parentOf ingredient is kept even when no surviving action references it.
+    #[test]
+    fn retain_ingredients_keeps_parent_of() {
+        let mut b = removal_builder(json!({
+            "ingredients": [
+                { "title": "parent", "format": "image/jpeg", "relationship": "parentOf", "label": "ing_parent", "instance_id": "id_parent" },
```
tmathern, line 10276: I would add variations of tests with an ingredient having a label only, the other instance_id only (since that can happen).

tmathern (commented):
The overall design looks promising!
I left some comments/questions inline.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+        };
+
+        // Isolate: remove, mutate a local copy, re-add. This mirrors `add_action` and keeps the
+        // `FnMut` `&Action` borrow off `self.definition`.
+        let assertion_def = self.definition.assertions.remove(pos);
+        let original_label = assertion_def.label.clone();
+        let mut actions: Actions = assertion_def.to_assertion()?;
+
+        let original_len = actions.actions.len();
+        // The C2PA spec requires exactly one c2pa.created/c2pa.opened action, so keep it
```
tmathern: (The v2.* spec)

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+        });
+
+        if actions.actions.is_empty() {
+            return Err(Error::BadParam(
+                "retain_actions would remove every action; at least one action is required"
+                    .to_string(),
+            ));
+        }
+
+        // The inception action must be at index 0. This is usually a no-op since
```
tmathern: I would just keep the comment `The inception action must be at index 0.`. This here is to be robust and so we don't have to keep track of what add_action does and how it is implemented.

sdk/src/builder.rs
```diff
@@ -902,6 +902,199 @@ impl Builder {
...
+    /// # Errors
+    /// * Returns an [`Error::BadParam`] if removal would leave zero actions.
+    pub fn remove_actions<F>(&mut self, mut remove: F) -> Result<&mut Self>
+    where
+        F: FnMut(&Action) -> bool,
+    {
+        self.retain_actions(|a| !remove(a))
+    }
+
+    /// Retains ingredients, then rewrites positional ingredient references so surviving actions
```
```suggestion
    /// Retains ingredients, then rewrites positional ingredient references so linked actions
```

sdk/src/builder.rs
```diff
@@ -3330,6 +3523,158 @@ impl std::fmt::Display for Builder {
     }
 }
 
+/// Collects all ingredient instance IDs an action references, checking every linking mechanism:
```
tmathern: 👍

sdk/src/builder.rs
```diff
@@ -3330,6 +3523,158 @@ impl std::fmt::Display for Builder {
...
+    }
+    (label, 0)
+}
+
+/// Builds a positional label from a base and 0-based index.
+fn make_positional_label(base: &str, idx: usize) -> String {
+    if idx == 0 {
+        base.to_owned()
+    } else {
+        format!("{base}__{idx}")
```
tmathern: Don't we already have that logic somewhere else? COuld we reuse it here, so we're sure those labelled indices/URIs always have the same format?

sdk/src/builder.rs
```diff
@@ -9674,4 +10019,568 @@ mod tests {
         let future = builder.sign_async(&signer, "image/jpeg", &mut src, &mut dst);
         assert_send(future);
     }
+
+    /// Build a `Builder` (with test context) from a JSON manifest definition.
```
```suggestion
    /// Create a `Builder` (with test context) from a JSON manifest definition.
```

sdk/src/builder.rs
cdmurph32: Yes. I'll change it so that it isn't mutating.

sdk/src/builder.rs
cdmurph32: Good catch. I'm not familiar enough with v1 vs v2.

sdk/src/builder.rs
cdmurph32: Changed `keep` to `rescue`. Perhaps wording can be changed some more.

sdk/src/builder.rs
cdmurph32: Kind of. I've removed some other redundancy

sdk/src/builder.rs
cdmurph32: No. Rewriting it.

sdk/src/builder.rs
```diff
@@ -902,6 +902,212 @@ impl Builder {
...
+    {
+        // Find the actions assertion; no-op when there isn't one.
+        let Some(pos) = self
+            .definition
+            .assertions
+            .iter()
+            .position(|a| a.label().starts_with(Actions::LABEL))
+        else {
+            return Ok(self);
+        };
```
ok-nick: There can be multiple actions assertions. For instance, the spec requires that `c2pa.created`/`c2pa.opened` are the first actions in the actions assertion in the created assertions list. User-performed actions would go in a new actions assertion in the gathered assertions list.

sdk/src/builder.rs
```diff
@@ -902,6 +902,212 @@ impl Builder {
...
+        // intact — nothing is removed until all the fallible work is done.
+        let original_label = self.definition.assertions[pos].label.clone();
+        let original_created = self.definition.assertions[pos].created;
+        let mut actions: Actions = self.definition.assertions[pos].to_assertion()?;
+
+        let original_len = actions.actions.len();
+        // The v2.x C2PA spec requires exactly one c2pa.created/c2pa.opened action, so keep it
+        // regardless of the caller predicate.
+        actions.actions.retain(|a| {
+            a.action() == c2pa_action::CREATED || a.action() == c2pa_action::OPENED || keep(a)
```
ok-nick: Requires exactly one in one created assertions actions assertion, other actions assertions do not need one.

sdk/src/builder.rs
```diff
@@ -902,6 +902,212 @@ impl Builder {
...
+        });
+
+        if actions.actions.is_empty() {
+            return Err(Error::BadParam(
+                "at least one action is required".to_string(),
+            ));
+        }
```
ok-nick: Some settings, as well as intents, add the created/opened when signing, which happens at a later time.

sdk/src/builder.rs
ok-nick: It's an assertion version condition, not claim version. We always write the latest assertion version.

sdk/src/builder.rs
```diff
@@ -902,6 +902,212 @@ impl Builder {
...
+
+        // Encode first, then swap in place: `self` is never left without its actions assertion on
+        // error. Preserves the original `.v1`/`.v2` label and created/gathered flag.
+        let data = AssertionData::Cbor(c2pa_cbor::value::to_value(&actions)?);
+        self.definition.assertions[pos] = AssertionDefinition {
+            label: original_label,
+            data,
+            kind: None,
+            created: original_created,
+        };
```
ok-nick: Why not just set `self.definition.assertions[pos].data = new_data`?

sdk/src/builder.rs
```diff
@@ -902,6 +902,212 @@ impl Builder {
...
+    /// The inception action (`c2pa.created`/`c2pa.opened`) is always kept, as with
+    /// [`Builder::retain_actions`].
+    ///
+    /// # Arguments
+    /// * `remove` - A predicate; the action is removed when it returns true.
+    /// # Returns
+    /// * A mutable reference to the [`Builder`]. A no-op if there is no actions assertion.
+    /// # Errors
+    /// * Returns an [`Error::BadParam`] if removal would leave zero actions.
+    pub fn remove_actions<F>(&mut self, mut remove: F) -> Result<&mut Self>
```
ok-nick: IMO we should stick with one of retain or remove, the user can flip their conditional. I lean towards `filter_actions` to be consistent with `Iterator::filter`.

sdk/src/ingredient.rs
```diff
@@ -252,6 +252,19 @@ impl Ingredient {
...
+    ///
+    /// This is how the [`Builder`](crate::Builder) keys ingredients when resolving action
+    /// `ingredientIds` and emitting positional assertion labels, so any code matching actions to
+    /// ingredients must use the same rule.
+    pub fn effective_id(&self) -> String {
+        self.label()
+            .filter(|label| !label.is_empty())
+            .map(str::to_string)
+            .unwrap_or_else(|| self.instance_id().to_string())
+    }
```
ok-nick, line 268: I think we should keep this private.

sdk/src/builder.rs
```diff
@@ -3330,6 +3532,157 @@ impl std::fmt::Display for Builder {
...
+/// `self#jumbf=c2pa.assertions/c2pa.ingredient.v3__2` -> `c2pa.ingredient.v3__2`.
+/// When the URL has no path separator, the whole URL is treated as the label.
+///
+/// This borrows a slice of `url` (rather than reusing [`assertion_label_from_uri`], which returns
+/// an owned, normalized `String`) so callers can compute the label's byte offset within the
+/// original URL and rebuild the prefix. Where only the label value is needed, prefer
+/// [`assertion_label_from_uri`].
+fn ingredient_ref_label(url: &str) -> &str {
+    url.rsplit('/').next().unwrap_or(url)
+}
```
ok-nick: We have methods that do this in the `labels` module.

sdk/src/builder.rs
```diff
@@ -3330,6 +3532,157 @@ impl std::fmt::Display for Builder {
...
+        action.get_parameter::<c2pa_cbor::Value>("instanceId"),
+    ));
+    if ids.is_empty() {
+        #[allow(deprecated)]
+        if let Some(id) = action.instance_id() {
+            ids.push(id.to_owned());
+        }
+    }
+    ids
+}
```
ok-nick: We have a method `Actions::extract_ingredient_ids` which I believe solves this already.

sdk/src/builder.rs
```diff
@@ -902,6 +902,212 @@ impl Builder {
...
+    }
+
+    /// Retains ingredients, then rewrites positional ingredient references so linked actions
+    /// stay valid.
+    ///
+    /// An ingredient is kept if any of the following hold:
+    /// - it is referenced by a current action (via `parameters.ingredients` URLs or the
+    ///   `ingredientIds`/`instanceId` parameters),
+    /// - it is a [`Relationship::ParentOf`] ingredient (lineage for `c2pa.opened`), or
+    /// - `rescue` returns true for it.
```
ok-nick, line 1003: The use case feels very narrow here, it's not just filtering ingredients, it's for a specific workflow removing ingredients. 

For the `parentOf` ingredient condition, I think we should let people do what they want here and filter out by relationship if they choose. You may want to replace that ingredient with a new parent after all. Validation will catch any problems when the manifest is ready.

For the "referenced by existing action" case, maybe we tell them that in the callback if it's referenced and they decide if they still want to filter the ingredient. I think there is more to this answer though, what if you want more context about the action? What if you want to keep the action but remove the link? What if you want info about something else in the manifest?

I don't have the answers to these questions but perhaps there's a larger picture here.

tmathern (approved):

sdk/src/builder.rs
cdmurph32: I made some changes but I haven't really addressed your comment. 

It still keeps parentOf ingredients and action referenced ingredients rather that surfacing an `is_referenced` callback and letting the caller decide. That's all I really need for what I'm doing, but perhaps @tmathern or @gpeacock will have an opinion as to where it should be more flexible.

sdk/src/builder.rs
cdmurph32: The existing functions didn't do what I needed.
I've moved these into sdk/src/jumbf/labels.rs, so they will be next to the functions they resemble.

sdk/src/builder.rs
cdmurph32: I need a non-mutating version. I'll organize this better.

sdk/src/ingredient.rs
cdmurph32: I think we really need this to be public since it is used by to_claim() https://github.com/contentauth/c2pa-rs/blob/building_reduction/sdk/src/builder.rs#L1594C33-L1594C45
The user provided filter function needs to be able to use the same logic as to_claim to be able to resolve action ingredient IDs.

sdk/src/builder.rs
tmathern: I'm ok with the level of flexibility here for now.

I can see the point of switching out a parent ingredient, but maybe that is the point where instead of pruning you are better off restarting a new Builder and moving things over "manually"? (With resources & json wrangling etc).

tmathern (approved):

caiopensrc:
- #2337

PR #2282, opened 2026-07-10
fix: Harden against memory amplification attacks in copy of strip, tile and big table of TIFF file

ssanthosh:
# Security Fix: TIFF Strip / Tile / BigTable Amplification OOM

## Vulnerability

`c2patool` crashed with an OOM abort (exit code 134) when signing a crafted
293 KB TIFF file. The strip / tile / BigTable clone logic in
[sdk/src/asset_handlers/tiff_io.rs](sdk/src/asset_handlers/tiff_io.rs)
(`TiffCloner::clone_image_data` and `clone_dng_bigtable_data`) copies each
strip's data to the output writer without any cap on cumulative bytes
written. Both the strip **offsets** and the strip **byte counts** are
attacker-controlled TIFF IFD fields. A file whose 50 000 strip offsets all
point to byte 0, each with a `byte_count = source_len`, produces
`O(strip_count × source_len)` output — 4.9 GB from a 293 KB input in the
reported PoC.

### Vulnerable pattern (three copies)

Strip path — [tiff_io.rs:824-849](sdk/src/asset_handlers/tiff_io.rs#L824):

```rust
with_order!(so_entry.value_bytes.as_slice(), self.endianness, |src| {
    for cnt in sbcs.iter() {
        let so: u64 = /* attacker-controlled */;
        asset_reader.seek(SeekFrom::Start(so))?;
        let mut data_reader = asset_reader.take(*cnt);   // *cnt attacker-controlled
        std::io::copy(&mut data_reader, &mut self.writer)?;   // no cumulative cap
    }
});
```

Same shape in the tile branch (~line 925) and in
`clone_dng_bigtable_data` (~line 1042). Any of the three can drive the
output buffer to gigabytes from a small input.

## Fix

Introduce a running byte counter and cap it against the actual source
stream length. Legitimate TIFF strip / tile / BigTable data lives inside
the file, so `sum(byte_counts) <= source_len` is a genuine format
invariant — the cap rejects only crafted or corrupt inputs.

Applied identically to all three copy loops:

```rust
let source_len = stream_len(asset_reader)?;
let mut copied: u64 = 0;

with_order!(so_entry.value_bytes.as_slice(), self.endianness, |src| {
    for cnt in sbcs.iter() {
        let so: u64 = /* … */;

        copied = copied.checked_add(*cnt).ok_or_else(||
            Error::InvalidAsset("TIFF strip byte counts overflow".to_string()))?;
        if copied > source_len {
            return Err(Error::InvalidAsset(
                "TIFF strip byte counts exceed source length".to_string()));
        }

        asset_reader.seek(SeekFrom::Start(so))?;
        let mut data_reader = asset_reader.take(*cnt);
        std::io::copy(&mut data_reader, &mut self.writer)?;
    }
});
```

- `checked_add` prevents the counter itself from wrapping (defence against
  a follow-on `u64` overflow attempt with `u64::MAX` byte counts).
- `copied > source_len` fires **before** the next `std::io::copy`, so the
  output writer never grows past `source_len` bytes on the malicious path.
- Existing per-entry parsing (`sbc_entry.value_count`, `safe_vec`) already
  bounds the number of iterations; this fix bounds cumulative bytes.

### Sites touched

- [tiff_io.rs:824](sdk/src/asset_handlers/tiff_io.rs#L824) — `clone_image_data` strip path
- [tiff_io.rs:925](sdk/src/asset_handlers/tiff_io.rs#L925) — `clone_image_data` tile path
- [tiff_io.rs:1042](sdk/src/asset_handlers/tiff_io.rs#L1042) — `clone_dng_bigtable_data`

## Tests Added

Three regression tests in `tiff_io::tests` using a helper that builds
`IfdClonedEntry` values directly (no fixture files required):

| Test | Payload | Validates |
|---|---|---|
| `clone_image_data_rejects_strip_byte_count_amplification` | 100-byte source; 10 strips × 200 bytes = 2000-byte sum | `Err(Error::InvalidAsset(msg))` with `msg` containing `"exceed source length"` |
| `clone_image_data_rejects_tile_byte_count_amplification` | 100-byte source; 8 tiles × `u32::MAX/8` bytes | Same — the tile branch is guarded identically |
| `clone_image_data_accepts_valid_strip_byte_counts` | 100-byte source; 5 strips × 10 bytes = 50-byte sum | Happy path succeeds — regression guard against an over-tight future cap |

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/tiff_io.rs
+++ b/sdk/src/asset_handlers/tiff_io.rs
@@ -599,6 +599,34 @@ where
     c2pa_mode: bool,
 }
 
+/// Accumulates one strip/tile/BigTable byte count into `copied` and rejects the
+/// tag once the running total exceeds the source stream length.
+///
+/// TIFF strip/tile/BigTable data lives inside the source file, so a legitimate
+/// `sum(byte_counts)` is at most the file size. A crafted file can instead list
+/// many entries whose offsets all point at the same bytes and whose counts sum
+/// to far more than the file, making the clone re-copy the input over and over
+/// and grow the in-memory output to gigabytes (memory amplification → OOM
+/// abort; ~4.9 GB from a 293 KB input in the reported PoC). Capping the
+/// cumulative copy length at the source length stops that while leaving
+/// well-formed files untouched.
+///
+/// `kind` labels the error ("strip", "tile", "BigTable"). The error is returned
+/// via `Result` so the caller propagates it with `?`. Note that `with_order!`
+/// inlines its body into a `match` arm — it is not a real closure — so that `?`
+/// (like a bare `return`) exits `clone_image_data` directly, not just a closure.
+fn accumulate_copy_len(copied: &mut u64, cnt: u64, source_len: u64, kind: &str) -> Result<()> {
+    *copied = copied
+        .checked_add(cnt)
+        .ok_or_else(|| Error::InvalidAsset(format!("TIFF {kind} byte counts overflow")))?;
+    if *copied > source_len {
+        return Err(Error::InvalidAsset(format!(
+            "TIFF {kind} byte counts exceed source length"
+        )));
+    }
+    Ok(())
+}
+
 impl<T: Read + Write + Seek> TiffCloner<T> {
     pub fn new(endianness: Endianness, big_tiff: bool, writer: T) -> Result<TiffCloner<T>> {
         let bo = ByteOrdered::runtime(writer, endianness);
@@ -1462,14 +1490,7 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
                             _ => return Err(Error::InvalidAsset("invalid TIFF strip".to_string())),
                         };
 
-                        copied = copied.checked_add(*cnt).ok_or_else(|| {
-                            Error::InvalidAsset("TIFF strip byte counts overflow".to_string())
-                        })?;
-                        if copied > source_len {
-                            return Err(Error::InvalidAsset(
-                                "TIFF strip byte counts exceed source length".to_string(),
-                            ));
-                        }
+                        accumulate_copy_len(&mut copied, *cnt, source_len, "strip")?;
 
                         let dest_offset = self.writer.stream_position()?;
                         dest_offsets.push(dest_offset);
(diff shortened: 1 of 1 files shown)


sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -838,6 +847,15 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
                             _ => return Err(Error::InvalidAsset("invalid TIFF strip".to_string())),
                         };
 
+                        copied = copied.checked_add(*cnt).ok_or_else(|| {
+                            Error::InvalidAsset("TIFF strip byte counts overflow".to_string())
+                        })?;
+                        if copied > source_len {
```
tmathern: Around L834, there is a macro. This all is in the macro. When the macro is expanded, this can't be circumvented?
(The issue I am thinking about is the behavior of the closure inside the macro: what gets exited, when? Do we exit the whole function, what the marco expands to, the closure only? Please double check.)

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -838,6 +847,15 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
                             _ => return Err(Error::InvalidAsset("invalid TIFF strip".to_string())),
                         };
 
+                        copied = copied.checked_add(*cnt).ok_or_else(|| {
```
tmathern: Those check are repetitive in multiple locations (see belowaround L961 too). Are they refactorable into a function?

ssanthosh: Refactored the code to remove redundancy with https://github.com/contentauth/c2pa-rs/pull/2282/commits/36ed2ab443c3ead74c9ec9b4104c5cdfe1f45d3c

sdk/src/asset_handlers/tiff_io.rs
ssanthosh: Verified that it cannot be circumvented. `return Err(...)/? `exit clone_image_data directly, not a closure. The existing amplification tests prove this — they assert `clone_image_data`(...) returns Err, which would be impossible if return only left a closure.

tmathern (approved):

sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -1453,6 +1490,8 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
                             _ => return Err(Error::InvalidAsset("invalid TIFF strip".to_string())),
                         };
 
+                        accumulate_copy_len(&mut copied, *cnt, source_len, "strip")?;
```
mpgopala: Does it make sense to call `accumulate_copy_len` _before_ the computation of `so`?

mpgopala (approved):

ok-nick (approved):

sdk/src/asset_handlers/tiff_io.rs
ssanthosh: Tes, Done with https://github.com/contentauth/c2pa-rs/pull/2282/commits/5e47484a697ce6d581568f4c8d5a238fd3aa6ffa

caiopensrc:
- #2348

PR #2284, opened 2026-07-10
fix: Harden against integer underflow attacks in  ID3 v2.3 frame decoder

ssanthosh:
### The vulnerability
Parsing a crafted audio asset (MP3/FLAC) could **abort the process**. The `id3` crate's ID3v2.3 frame decoder (`src/stream/frame/v3.rs`) computes `content_size - 4` unconditionally when a frame sets the COMPRESSION flag. A frame declaring COMPRESSION with a body smaller than the 4-byte decompressed-size prefix makes that subtraction underflow (`usize`) to ~`u64::MAX`, and the following `vec![0; read_size]` aborts with "capacity overflow" (~293 KB input → multi-GB allocation in the reported PoC). Because c2pa-rs hands untrusted input to `id3::Tag::read_from2` when reading/writing manifests in audio, this is reachable from ordinary read paths.

### Approach — fix the root cause upstream (updated)

Investigation showed the crate **already fixes this exact bug for ID3v2.4** (`v4.rs`, `content_size.saturating_sub(4)`, upstream commit `d54ee7b6`) but the identical **ID3v2.3** decoder was simply never patched — in `1.17.0` *and* on current upstream `main`. So the clean fix is a one-liner in the crate, not a validator in c2pa-rs.

The crate's ID3v2.3 decoder computed content_size - 4 unchecked, so a crafted compressed frame underflowed to ~u64::MAX and aborted the process on allocation. The fix ([polyfloyd/rust-id3#163](https://codeberg.org/polyfloyd/rust-id3/pulls/163)) is released in 1.17.1 (saturating_sub(4) in v3.rs, verified against the tag).

Upgrade id3 crate to 1.17.1 to fix the issue.

The existing regression test (read_cai_store_malicious_id3v23_compression_does_not_crash) ensures the fix on ID3 crate works fine.

## Checklist


sdk/src/asset_handlers/id3_helper.rs
```diff
@@ -110,6 +112,141 @@ pub(crate) fn is_c2pa_mime_type(mime_type: &str) -> bool {
...
+        return None;
+    }
+
+    let mut hdr_buf = [0u8; 10];
+    if input_stream.read_exact(&mut hdr_buf).is_err() {
+        let _ = input_stream.rewind();
+        return None;
+    }
+
+    let header = match ID3V2Header::parse_from_bytes(&hdr_buf) {
```
tmathern: Are there (edge) cases where some values in the header should change this behavior? Do some version need special handling?

sdk/src/asset_handlers/id3_helper.rs
```diff
@@ -110,6 +112,141 @@ pub(crate) fn is_c2pa_mime_type(mime_type: &str) -> bool {
...
+        }
+    }
+
+    reader
+        .seek(SeekFrom::Start(tag_start))
+        .map_err(|_| Error::InvalidAsset("could not rewind ID3 tag body".to_string()))?;
+    Ok(())
+}
+
+/// Reads an ID3v2 tag from `input_stream`, pre-validating the frame table
```
tmathern: Are there versions of ID3 & friends to take into account in this reading logic? Flags that change behavior?

sdk/src/asset_handlers/id3_helper.rs
ssanthosh: Yes — good catch. The **tag-header flags byte** does change how the frame table is laid out, and the original validator ignored it. Two ID3v2.3 tag flags matter:

- **Extended header** (`0x40`): frames don't start right after the 10-byte header; an extended header comes first. The `id3` crate (v1.17.0, `src/stream/tag.rs`) reads and skips it (synchsafe-decoded size) before parsing frames.
- **Unsynchronisation** (`0x80`): the crate de-unsynchronises the body (drops `0x00` after `0xFF`) before parsing, so raw byte offsets don't match declared frame sizes.

Version handling itself is intentional: only **v2.3** is validated. v2.2 has no compression flag, and v2.4 uses `saturating_sub` upstream — neither can hit the `content_size - 4` underflow. Unsupported versions and any tag the crate rejects (e.g. unknown flag bits) fall through to `Tag::read_from2` and surface as "no tag".

Handled Extender header and Unsynchronisation flags with https://github.com/contentauth/c2pa-rs/pull/2284/commits/398b11b5e1340f2972560510188cc8445b1b4d00

sdk/src/asset_handlers/id3_helper.rs
ssanthosh: Same root issue, now fixed with https://github.com/contentauth/c2pa-rs/pull/2284/commits/398b11b5e1340f2972560510188cc8445b1b4d00. `validate_id3v2_3_frames` now takes the tag flags byte and, for v2.3:
- seeks past the **extended header** (matching the crate's synchsafe size decode), and
- **de-unsynchronises** the frame region into a buffer before walking it when the unsync flag is set.

tmathern (approved):
Please also confirm with @mauricefisher64 and @ok-nick !

sdk/src/asset_handlers/id3_helper.rs
ok-nick, line 1: It seems like we are re-implementing a lot of what the `id3` crate is already doing for us. From my understanding, in the `id3` crate [this line](https://codeberg.org/polyfloyd/rust-id3/src/commit/2326e0b80768d4ec2f50f26cb71424675b69ede4/src/stream/frame/v3.rs#L51) needs to change from:
```rust
content_size - 4
```
to:
```rust
content_size.saturating_sub(4)?
```
Is my understanding correct? If that's the case I would prefer we use a fork and submit a PR upstream.

ssanthosh: Yes. Done — this is the cleaner root-cause fix. Investigation confirmed the `id3` crate already fixes this exact underflow for **ID3v2.4** (`src/stream/frame/v4.rs`, `content_size.saturating_sub(4)`, added in upstream commit `d54ee7b6`) but the identical **ID3v2.3** decoder (`v3.rs`) was never touched — in `v1.17.0` *and* current `main`. So bumping the crate can't help; the v2.3 path needs the same one-liner.

tmathern (commented):
Do we wait for a release of the dependency?

Cargo.toml
```diff
@@ -26,3 +26,13 @@ strip = true  # Automatically strip symbols from the binary.
...
 
+# Temporary: the `id3` crate (v1.17.0, and current upstream `main`) still performs
+# an unchecked `content_size - 4` in its ID3v2.3 frame decoder
+# (src/stream/frame/v3.rs) — a compressed frame with body < 4 bytes underflows to
+# ~u64::MAX and aborts the process on the following allocation. The fix already
+# exists for ID3v2.4 (v4.rs, `saturating_sub`) but was never applied to v2.3.
+# Point at a fork that applies the same one-liner to v3.rs until it is released
+# upstream: https://codeberg.org/ssanthosh/rust-id3/pulls/1
+[patch.crates-io]
+id3 = { git = "https://codeberg.org/ssanthosh/rust-id3", rev = "4d8abd485a277e4f1b7b1fdeb4ea936936fd1f82" }
```
ok-nick: I believe crates.io will ignore this patch when actually publishing, although it will be fixed for local builds. Maybe that will give upstream some time to accept the change. I think we should move forked crates to the contentauth org. @scouten-adobe thoughts?

ssanthosh: I am trying to get the Security Researcher team to reject this vulnerability as a not valid one as it is a vulnerability on the third party library like ID3 and not on C2PA SDK

ssanthosh: @ok-nick : I have raised a PR to the upstream ID3 crate from my work : https://codeberg.org/polyfloyd/rust-id3/pulls/163 and we can update ID3 once it is merged and released

ssanthosh:
@tmathern / @ok-nick : I have updated the ID3 crate to fix the issue as the new ID3 crate includes the upstream fix that I have done. Please review

tmathern (approved):

ok-nick (approved):

caiopensrc:
- #2389

PR #1865, opened 2026-02-18
feat: Implement remote signing for CAWG `X509CredentialHolder`

BadrTad:
## Changes in this pull request
This PR implements CAWG remote signing (as requested in #1777 ) by introducing a new type `raw_signer:: RemoteRawSigner`

The feature could be tested by enabling the "remote_signing" feature during compilation. and remote signers created with this new method `create_signer::from_remote_url`

The remote_signing is supported only for openssl crypto provider.

The code snippet that converts X509 certificates from PEM to DER in raw_signature::openssl crate was also refactored as a macro under `raw_signature::openssl::cert_chain::cert_chain_to_der!` 

The remote signer could be created by providing a setting configuration to  `Context` struct or by instantiating  `X509CredentialHolder` from a `openssl::signers::remote_signer::RemoteRawSigner`.

Tests use the `httpmock` crate to create mock servers that responds with valid signature using a (local) signer instantiated with `create_signer::from_keys`. 

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #2285, opened 2026-07-10
fix: convert remote signer certificate to DER

### LINKED ISSUE #1903: Remote signer certificate fails validation because it is never converted to DER
### Component

c2pa-rs

### If Other, please specify

_No response_

### Version

0.76.1

### Platform

macOS

### What happened?

Creating a remote signer from a settings.json fails with a "the certificate is invalid" error. I believe this is caused by the certificate PEM never being converted to DER. When it hits [this check](https://github.com/contentauth/c2pa-rs/blob/355d27a90f1296360a9de51d359bfa281c2af0b6/sdk/src/crypto/cose/certificate_profile.rs#L85) the cert needs to be in DER in memory, and the signing operation therefore fails.

When using a local signer, PEM to DER conversion happens in the [signer](https://github.com/contentauth/c2pa-rs/blob/c9154776ae68478cd8e0113f359b99610ed36b5c/sdk/src/crypto/raw_signature/openssl/signers/rsa_signer.rs#L67). This isn't possible for the remote signer, as the above validation checks occur before the data packet is sent to the signer for signature.

A fix could potentially go [here](https://github.com/contentauth/c2pa-rs/blob/c9154776ae68478cd8e0113f359b99610ed36b5c/sdk/src/settings/signer.rs#L121) to parse the certificate JSON and convert to DER - similar to what happens in the local signer. I have a patch I could put a PR in for, but wanted to check this isn't already something you were working on or had a neater solution for.

Here is the remote signer settings snippet:
```
  "signer": {
    "remote": {
      "url": "www.example.com/sign",
      "alg": "ps256",
      "sign_cert": "<c2pa test ps256 cert>"
    }
```

and here is the rough code I'm using:
```
    // Parse settings
    let settings = Settings::new().with_json(include_str!("<path to settings json>"));

    // Build Context
    let context = Context::new().with_settings(settings);

    // Create Builder from Context
    let manifest = json!({
        "claim_generator_info": [{
            "name": "Local Debug",
            "version": "0.1.0"
        }]
    });

    let mut builder = Builder::from_context(context)
        .with_definition(manifest);

    // Signing
    builder.save_to_stream("image/jpeg", &mut input, &mut output);
```

Remote signer should pass certificate validation checks.

dcondrey:

A remote signer configured via `[signer.remote]` failed validation with "the certificate is invalid". `SignerSettings::c2pa_signer` stored `sign_cert` as raw PEM, but the certificate profile check runs before the remote call and needs DER. The local signer converts PEM→DER internally, so only the remote path was affected.

The `Remote` arm now parses `sign_cert` into a DER chain via `Pem::iter_from_buffer`, as the `rust_native` signers do; malformed PEM returns `Error::BadParam`. `test_make_remote_signer` asserts `certs()` returns DER and fails against the old code.

Overlaps #2075, which fixes the same bug and predates this. Adopted its `reserve_size` calculation, which sums the DER chain rather than using the PEM string length. This adds the regression test and an actionable parse error — happy to close in favour of #2075 if you'd rather take that one.

Fixes #1903 (still reproducible on `main`; #1865 was closed unmerged).

## Checklist

### DIFF
--- a/sdk/src/settings/signer.rs
+++ b/sdk/src/settings/signer.rs
@@ -16,6 +16,7 @@ use std::sync::Arc;
 use c2pa_raw_crypto::{signer_from_private_key, RawSigner, RawSignerError, SigningAlg};
 use http::Request;
 use serde::{Deserialize, Serialize};
+use x509_parser::{error::PEMError, pem::Pem};
 
 use crate::{
     create_signer,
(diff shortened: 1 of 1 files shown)


mpgopala:
Created https://github.com/contentauth/c2pa-rs/pull/2414 to address this issue.

PR #2286, opened 2026-07-10
fix: add timestamp when signing via save_to_stream

### LINKED ISSUE #2143: Some APIs of Builder do not add the timestamp
### Component

c2pa-rs

### If Other, please specify

Builder

### Version

v0.84.0

### Platform

Linux

### What happened?

c2pa_builder_sign() uses builder.sign() which calls maybe_add_timestamp().
However c2pa_builder_sign_context() calls builder.save_to_stream() which doesn't call maybe_add_timestamp() and hence doesn't add the timesatmp.

Using context should use the signer /settings from context but the rest of the builder behavior should be consistent across APIs.

dcondrey:

`Builder::save_to_stream()` built the store directly and skipped `maybe_add_timestamp()`, so context-based signing via `c2pa_builder_sign_context()` produced manifests without the timestamp `Builder::sign()` adds. It now builds the claim, adds a timestamp when the context signer exposes a time authority, then builds the store from that claim.

New test `timestamp_assertion_save_to_stream` fails against the old code.

Fixes #2143

## Checklist


dcondrey:
Superseded by #2437, which fixes the same bug more thoroughly (also covers `save_to_file` and async signing).

Issue #2266, opened 2026-07-07
crJSON Output b64 string missing second '


lbrooks:
### Component

c2patool

### Version

v0.89.0

### Platform

Linux

### What happened?

When outputting the validation result to crJson, all the b64 output string values are `"<field>": "b64'<data>"`

Example:
```
"pad": "b64'AAAAAAAAAAAA"
```

### What did you expect to happen?

The value should have a terminating `'` as well.

Example:
```
"pad": "b64'AAAAAAAAAAAA'"
```

gpeacock:
I expected the b64' prefix to have a closing brace as well, but that is not how it is documented in the crJson spec, and, as far as I can tell, this notation is unique to crJson.  https://spec.c2pa.org/specifications/specifications/2.4/crJSON/crjson-format.html

PR #2287, opened 2026-07-10
fix: emit empty redacted_assertions for claim.v2 with no redactions

### LINKED ISSUE #2267: crJSON Output missing redacted_assertions within claim.v2
### Component

c2patool

### If Other, please specify

_No response_

### Version

v0.89.0

### Platform

Linux

### What happened?

Assets without any `redacted_assertions` within the `claim.v2` Assertion are missing the `redacted_assertion` field with a value of an empty array.

According to the spec:
> If either of the gathered_assertions or redacted_assertions fields are not present, then the corresponding JSON field shall be present, and their respective values shall be an empty array.

There should be a `redacted_assertions` field with a value of an empty array.

dcondrey:

A `claim.v2` with no redactions omitted `redacted_assertions` entirely.

specs-core moved this rule out of crJSON Claims/General and into the `claim.v2` **Required** list, where `gathered_assertions` and `redacted_assertions` are both specified as "set to an empty array if it's missing in the source" (c2pa-org/specs-core#2345, for c2pa-org/specs-core#2344). `crJSON.schema.json` matches: `definitions/claim`, which `claim.v2` `$ref`s, requires both fields, while `definitions/claimV1` leaves `redacted_assertions` optional.

So this is v2-only and v1 output is unchanged. The published 2.4 HTML still carries the older version-agnostic wording in §3.5.1; specs-core is ahead of it here.

`gathered_assertions` already behaved this way; `redacted_assertions` didn't.

Rebased onto latest main, and the b64 delimiter half of this PR is dropped — #2266 landed separately as #2333.

## Checklist

Fixes #2267

### DIFF
--- a/sdk/src/crjson.rs
+++ b/sdk/src/crjson.rs
@@ -302,13 +302,14 @@ impl<'a> CrJsonExporter<'a> {
             alg: Some(claim.alg().to_string()),
             alg_soft: claim.alg_soft().map(|s| s.to_string()),
             title: claim.title().map(|s| s.to_string()),
-            redacted_assertions: claim.redactions().and_then(|r| {
-                if r.is_empty() {
-                    None
-                } else {
-                    Some(r.to_vec())
-                }
-            }),
+            redacted_assertions: if is_v1 {
+                claim
+                    .redactions()
+                    .and_then(|r| if r.is_empty() { None } else { Some(r.to_vec()) })
+            } else {
+                // crJSON 3.5.1: absent redactions are still serialised, as an empty array.
+                Some(claim.redactions().map(|r| r.to_vec()).unwrap_or_default())
+            },
             metadata: claim.metadata().and_then(|m| {
                 if m.is_empty() {
                     None
--- a/sdk/tests/crjson/created_gathered.rs
+++ b/sdk/tests/crjson/created_gathered.rs
@@ -106,6 +106,16 @@ fn test_created_and_gathered_assertions_separated() -> Result<()> {
         created_assertions.len()
     );
 
+    // crJSON 3.5.1: redacted_assertions is present as an empty array when nothing is redacted
+    let redacted_assertions = claim_v2["redacted_assertions"]
+        .as_array()
+        .expect("redacted_assertions should be present as an array");
+
+    assert!(
+        redacted_assertions.is_empty(),
+        "redacted_assertions should be empty when no assertions are redacted"
+    );
+
     // Find the created assertion - should be in created_assertions
     let has_created_ref = created_assertions.iter().any(|assertion_ref| {
         if let Some(url) = assertion_ref.get("url") {
--- a/sdk/tests/crjson/schema_compliance.rs
+++ b/sdk/tests/crjson/schema_compliance.rs
@@ -455,12 +455,13 @@ fn test_claim_v2_required_fields() -> Result<()> {
                 "claim.v2.gathered_assertions must be an array"
             );
         }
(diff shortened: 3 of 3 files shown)


gpeacock:
Will the crJson spec be updated to reflect this change? The text and examples in the spec only use a 64' prefix.

sdk/src/crjson.rs
```diff
@@ -54,10 +54,15 @@ struct Base64Bytes(Vec<u8>);
...
     fn serialize<S: serde::Serializer>(&self, s: S) -> std::result::Result<S::Ok, S::Error> {
-        s.serialize_str(&format!("b64'{}", base64::encode(&self.0)))
+        s.serialize_str(&b64_str(&self.0))
     }
 }
 
+/// Formats bytes as a crJSON `b64'<base64>'` delimited string.
+fn b64_str(bytes: impl AsRef<[u8]>) -> String {
+    format!("b64'{}'", base64::encode(bytes.as_ref()))
+}
```
ok-nick: This was addressed on main. Would you mind updating the PR to latest?

ok-nick (changes requested):

sdk/src/crjson.rs
dcondrey: Done — rebased onto latest main and dropped the b64 delimiter changes entirely, since #2333 covers them.

What's left is just the empty `redacted_assertions` for `claim.v2` (#2267), which is still open on main.

sdk/src/crjson.rs
```diff
@@ -302,13 +302,14 @@ impl<'a> CrJsonExporter<'a> {
...
-                }
-            }),
+            redacted_assertions: if is_v1 {
+                claim
+                    .redactions()
+                    .and_then(|r| if r.is_empty() { None } else { Some(r.to_vec()) })
+            } else {
+                // crJSON 3.5.1: absent redactions are still serialised, as an empty array.
+                Some(claim.redactions().map(|r| r.to_vec()).unwrap_or_default())
+            },
```
ok-nick, line 312: The spec doesn't state the "empty array when no values present" is v2 claim specific. I think we should share the same behavior among all claim versions. See:
https://spec.c2pa.org/specifications/specifications/2.4/crJSON/crjson-format.html#_general

dcondrey: @ok-nick That's still in the published 2.4 HTML, but specs-core has since moved it. c2pa-org/specs-core#2345, for issue #2344 ("crJSON - claim.v2 - gathered_assertions & redacted_assertions required?"), dropped it from Claims/General and put both fields in the `claim.v2` Required list instead:

> `gathered_assertions` — array of zero or more hashed URI maps (same structure as created_assertions); set to an empty array if it's missing in the source
> `redacted_assertions` — array of sero or more JUMBF URI strings (references to redacted ingredient manifest assertions); set to an empty array if it's missing in the source

`crJSON.schema.json` matches — `definitions/claim` (which `claim.v2` `$ref`s) requires both fields; `definitions/claimV1` doesn't require `redacted_assertions`.

My PR description cited 3.5.1. Fixing that.

ok-nick: Ok that's fine, the behavior for claim v1 isn't changing in this PR.

ok-nick (approved):

ok-nick:
Thank you for the contribution!

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 30856475833](https://github.com/contentauth/c2pa-rs/actions/runs/30856475833).

PR #2300, opened 2026-07-13
chore: Pin docs.rs preflight nightly and add stable doc warnings check

scouten-adobe:
## Problem

The `docs_rs` preflight job in Tier 1A installed a floating `nightly` toolchain and denied **all** warnings (`RUSTDOCFLAGS: -Dwarnings`). That made CI subject to the Rust team's daily nightly changes: any new nightly lint became a hard, unscheduled CI failure — most recently a bunch of reactive work.

Two independent causes:
- **Floating `nightly`** pulls in new lints/behavior on the Rust team's daily cadence.
- **`-Dwarnings`** escalates any one of those into a hard failure — even though docs.rs itself does *not* build with `-Dwarnings`, so those warnings would never actually break our published docs.

## Change

**`docs_rs` (nightly — docs.rs fidelity)**
- Pin `nightly` → `nightly-2026-07-12`, the nightly docs.rs is currently running (see the version string at <https://docs.rs/about/builds>). Refresh periodically (e.g. at release time) to track what docs.rs uses.
- Replace blanket `-Dwarnings` with a scoped deny of `rustdoc::broken_intra_doc_links` + `rustdoc::invalid_html_tags` — the doc-correctness lints that would actually produce broken published docs, and which don't churn across nightlies.
- Move the c2patool doc build out (it never needed nightly).

**`docs_stable` (new — strict enforcement)**
- Runs on `stable` (advances ~every 6 weeks, predictable) with blanket `-Dwarnings` across `c2pa`, `c2pa-c-ffi`, and `c2patool`.
- Uses plain `cargo doc` and omits `--cfg docsrs`, so the nightly-only `doc_cfg` feature gated in `sdk/src/lib.rs` stays inert and stable can compile the docs.

Net: nightly still verifies "docs.rs will build our docs" but only breaks on real doc-correctness bugs; strict no-warnings enforcement lives on a toolchain whose cadence we control.

## Note

The nightly pin will drift as docs.rs advances its own toolchain. Simplest cadence is to bump it at release time. A follow-up could add a scheduled workflow that scrapes the current docs.rs nightly and opens a PR when the pin falls behind.


.github/workflows/tier-1a.yml
```diff
@@ -316,17 +320,59 @@ jobs:
...
           version: "1.0.4"
 
+      # docs.rs does not build with `-Dwarnings`, so a new nightly warning
+      # would not actually break our published docs. Deny only the doc-
+      # correctness lints that would (broken links / bad HTML). Strict, blanket
+      # warning enforcement lives in the `docs_stable` job below instead.
       - name: Preflight c2pa-rs docs.rs
         env:
-          RUSTDOCFLAGS: -Dwarnings
+          RUSTDOCFLAGS: -D rustdoc::broken_intra_doc_links -D rustdoc::invalid_html_tags
```
ok-nick: We can remove this flag (and as done below) since it's just a warning, it won't affect the docs.rs build.

ok-nick (approved):

.github/workflows/tier-1a.yml
scouten-adobe: Good call. Removed.

tmathern (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 30856487793](https://github.com/contentauth/c2pa-rs/actions/runs/30856487793).

PR #2304, opened 2026-07-14
ci: Publish binaries for each release-candidate build

scouten-adobe:
## Problem

The first bug in the new open-source release process: too many downstream projects consume our **pre-built binary libraries** and can't easily build from source. Under the current process, release candidates are never published in *any* form (crates.io *or* GitHub-release binaries), so during a breaking train's ≥3-business-day bake those consumers have nothing to validate against.

## Fix

Give each RC **build** a *prerelease* GitHub release with binaries attached, while preserving the invariant that **RCs never reach crates.io**.

### Branch/version model
- RC branch is renamed to `0.N.0-rc` (dot-less) — the build number now belongs to each **build tag** (`-rc.1`, `-rc.2`, …), not the branch.
- All `*-rc.*` CI-trigger globs become `*-rc*`, and the PR-base `contains(…, '-rc.')` conditions become `-rc`, so the dot-less branch still matches the Tier suites and upstream-first check.
- The crates.io publish guard in `release.yml` matches `-rc` instead of `-rc.`.

### New workflows
- **`release-rc.yml`** — cuts a numbered build: sets the `-rc.N` versions on the branch (`c2pa`/`c2pa-c-ffi` and `c2patool` bumped in lockstep on the build number via `cargo set-version`), commits, and pushes the `c2pa-v…-rc.N` / `c2patool-v…-rc.N` tags **with the PAT** (so the tag-driven binary workflows actually fire). Guarded to run only on a `*-rc` branch.
- **`c2patool-release.yml`** — builds the `c2patool` CLI binaries + SBOMs on any `c2patool-v*` tag (moved out of `release.yml`). `-rc.` tags produce prereleases.

### Tag-driven binaries
Binary builds are now entirely tag-driven for **both** crates (`library-release.yml` already handled `c2pa-v*`), so RC tags produce the exact same artifacts as a real release. `release.yml` no longer builds binaries itself — it just runs `release-plz` publish and lets the resulting tags drive the binary workflows.

### Automation
- `release-train-cut.yml` creates the `0.N.0-rc` branch and then **auto-cuts the first build (`-rc.1`)** by dispatching `release-rc.yml`. Subsequent builds (`-rc.2`, …) are a manual `release-rc.yml` dispatch after bugfixes are cherry-picked during the bake.
- **`release-train-cut.yml` also now auto-advances `main` to the next `0.(N+1).0-dev` cycle** — committed directly to `main` (no PR), replacing the manual step done by hand in #2303. It runs *after* the RC branch is cut, so the candidate is taken from pre-bump `main`. The tracking issue no longer asks for this. *(Scoping note: this is a self-contained step folded into this PR because it edits the same `cut` job; happy to split it into its own PR if preferred.)*

### Docs
`docs/release-process.md` and `docs/support-tiers.md` updated (branching table, version-numbering + worked example, new *Release-candidate builds* section, auto-advance note, branch-protection bypass note, glob patterns).

## Notes / follow-ups
- **crates.io is untouched for RCs** — this publishes GitHub-release binaries only, per the design decision confirmed for this change.
- Real-release behavior is preserved: `library-release.yml` keeps its existing `prerelease: true`; `c2patool-release.yml` marks only `-rc.` tags as prereleases (matching c2patool's prior real-release behavior).
- **`RELEASE_PLZ_TOKEN` must be on the branch-protection bypass list** for `main` (next-dev commit) and RC branches (`-rc.N` commits), and needs `actions:write` for `gh workflow run`. Documented in the branch-protection section.
- **Backport:** per plan, this should be backported to the current RC branch — which also means renaming `0.90.0-rc.1` → `0.90.0-rc` (the `release-rc.yml` branch guard expects the dot-less name).
- Validated with `actionlint` (new/changed workflows clean; only pre-existing info-level notes remain elsewhere).

### DIFF
--- a/.github/workflows/library-release.yml
+++ b/.github/workflows/library-release.yml
@@ -1,10 +1,13 @@
 name: Publish c2pa-library binaries
 
-# Triggered by `c2pa-v*` tags. Under the open-source release process these tags
-# are created by release-plz on a release-line branch (never on `main`, and
-# never on a release-candidate branch -- RCs are not published and therefore
-# never produce a `c2pa-v*` tag). Git tags are repo-global, so no branch filter
-# is needed here.
+# Triggered by `c2pa-v*` tags, which come from two places (see
+# docs/release-process.md):
+#   * release-plz, on a release-line branch -> a real tag like `c2pa-v0.90.0`.
+#   * release-rc.yml, when a release-candidate build is cut -> a prerelease tag
+#     like `c2pa-v0.90.0-rc.1`.
+# Either way the binaries are built here; the GitHub release is always marked as
+# a prerelease (pre-1.0). Nothing here publishes to crates.io. Git tags are
+# repo-global, so no branch filter is needed.
 
 on:
   push:
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -76,7 +76,7 @@ The convention:
 
 * **`main` always carries the *next* release's version with a `-dev` suffix** — e.g. `0.91.0-dev`. Because `main` is never published, the `-dev` prerelease is purely a label that says "work in progress toward 0.91.0."
 * **Cutting the train** for `0.N.0` produces the release-candidate branch `0.N.0-rc` (dropping the numeric suffix from the branch name; the number belongs to each *build*). Its builds are versioned `0.N.0-rc.1`, `0.N.0-rc.2`, … and, while never published to crates.io, are tagged and get prerelease GitHub-release binaries. On promotion the line becomes `0.N.0`.
-* **Right after the cut, `main` moves to `0.(N+1).0-dev`** so ongoing development is always numbered ahead of the line that's baking.
+* **Right after the cut, `main` moves to `0.(N+1).0-dev`** so ongoing development is always numbered ahead of the line that's baking. This bump is committed to `main` automatically by [`release-train-cut.yml`](#release-train-cut) as part of the cut — no separate PR.
 * **`c2patool` follows the same pattern on its own numbering** (its own next minor), independent of `c2pa`. RC build numbers are kept in lockstep across the crates, so `-rc.N` identifies one coherent candidate.
 
 Worked example (the first train, which is also the transition onto this convention):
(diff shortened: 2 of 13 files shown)


.github/workflows/library-release.yml
```diff
@@ -1,10 +1,13 @@
...
-# never on a release-candidate branch -- RCs are not published and therefore
-# never produce a `c2pa-v*` tag). Git tags are repo-global, so no branch filter
-# is needed here.
+# Triggered by `c2pa-v*` tags, which come from two places (see
+# docs/release-process.md):
+#   * release-plz, on a release-line branch -> a real tag like `c2pa-v0.90.0`.
+#   * release-rc.yml, when a release-candidate build is cut -> a prerelease tag
+#     like `c2pa-v0.90.0-rc.1`.
+# Either way the binaries are built here; the GitHub release is always marked as
+# a prerelease (pre-1.0). Nothing here publishes to crates.io. Git tags are
```
tmathern, line 9: Just to make sure: we have pre-releases and release candidates?

docs/release-process.md
```diff
@@ -27,9 +27,9 @@ Most changes never wait for the train: anything additive ships fairly quickly by
...
+| `0.(x+1).0-rc` (release-candidate branch) | A transient breaking candidate, cut from `main`, that bakes before promotion. Its name ends in `-rc` so it is validated by CI but **never** matches a crates.io publish trigger. Individual **builds** cut from it (`-rc.1`, `-rc.2`, …) are tagged and get *prerelease* GitHub releases with binaries, but are never published to crates.io. | crates.io: **No**; GitHub-release binaries: **yes** |
 
-The branch-name conventions are also the publish guard: the publish workflow only ever runs on `stable` and `v0.*`, and the `-rc.`-named candidate matches neither, so a candidate can never be published by construction.
+The branch-name conventions are also the crates.io publish guard: the publish workflow only ever runs on `stable` and `v0.*`, and the `-rc`-named candidate branch matches neither, so a candidate can never be published to crates.io by construction. (Its build tags carry the `-rc.` prerelease marker and drive binary-only GitHub releases; see [RC builds](#release-candidate-builds).)
```
tmathern, line 32: All the workflows I think look mostly, if not only, on branch references. So in the first few runs, make sure the rule `Its build tags carry the `-rc.` prerelease marker and drive binary-only GitHub releases` is properly followed.

tmathern (approved):

.github/workflows/library-release.yml
scouten-adobe: "Pre-release" is a specific term in GitHub parlance. Our RC builds (and actually all pre-1.0 releases) are tagged as pre-release in GitHub. (See screen shot below from our most recent tagged release.)

Screenshot 2026-07-14 at 10 06 13

scouten-adobe: Open question: Should the non-RC but pre 1.0 releases be called "pre-release?" It's what we've been doing so far (so IMHO not directly relevant to this PR), but worthy of separate debate.

ale-adobe (approved):

caiopensrc:
- #2305

PR #2306, opened 2026-07-14
fix: Correct tag, commit, and changelog for tag-driven binary releases

scouten-adobe:
Three related defects in the tag-driven binary release step, all visible on the mislabeled `c2pa-v0.90.0` release that a `c2pa-v0.90.0-rc.2` build produced. Applies to both [`library-release.yml`](https://github.com/contentauth/c2pa-rs/blob/main/.github/workflows/library-release.yml) and [`c2patool-release.yml`](https://github.com/contentauth/c2pa-rs/blob/main/.github/workflows/c2patool-release.yml).

## 1. Release published under the wrong name (dropped `-rc.N`)

`library-release.yml` re-derived the tag from a built zip's filename with a regex assuming a bare `X.Y.Z`:

```sh
sed -E 's/.*\/(c2pa-v[0-9]+\.[0-9]+\.[0-9]+)-.*\.zip/\1/'
```

`c2pa-v0.90.0-rc.2-<target>.zip` → `c2pa-v0.90.0`, burning the tag reserved for the post-bake final release. **Fix:** use `github.ref_name` (the exact ref that triggered the run) with a `c2pa-v*` guard — matching how `c2patool-release.yml` already resolves its tag. Never reconstruct a version from a filename.

## 2. Release pointed at `main`, not the tag's commit

Because of #1, `action-gh-release` created a *new* tag `c2pa-v0.90.0` and defaulted its commit to the default branch (`main`). Fixing the name means it attaches to the existing rc tag at the right commit; **as a safety net**, `target_commitish: ${{ github.sha }}` is now pinned so a release can never be created against the wrong commit.

## 3. Changelog cited an unrelated crate's tag

The auto-notes showed `Full Changelog: c2patool-v0.26.72...c2pa-v0.90.0`. Tags are repo-global and interleaved (`c2pa-v*`, `c2patool-v*`, `c2pa-c-ffi-v*`), and GitHub's automatic "previous tag" is the chronologically previous tag *regardless of crate*. **Fix:** compute the previous **same-crate** tag (`git/matching-refs` + `sort -V`) and pass it explicitly as `previous_tag_name` to the generate-notes API, using the result as the release body.

## Notes
- Artifacts now download to `dist/` so the generated `release-notes.md` is never uploaded as a release asset.
- `prerelease` handling is unchanged (library: always prerelease pre-1.0; c2patool: only `-rc.` tags).
- For a real (promotion) release, this sets GitHub-generated notes scoped to the previous same-crate tag. If we'd rather preserve release-plz's `CHANGELOG.md`-based body on promotion releases, that's an easy follow-up (only generate notes when the release doesn't already exist) — called out separately to keep this fix race-tolerant/idempotent.
- The erroneous `c2pa-v0.90.0` release+tag are being deleted separately; the correct `c2pa-v0.90.0-rc.2` tag remains.
- Still open (not this PR): no `c2patool-v0.28.0-rc.2` tag/release was produced for that RC build — worth investigating.

Validated with `actionlint` (new steps clean).

### DIFF
--- a/.github/workflows/c2patool-release.yml
+++ b/.github/workflows/c2patool-release.yml
@@ -95,6 +95,36 @@ jobs:
       - name: List artifacts
         run: ls -lh dist
 
+      # Generate the release notes ourselves, scoped to the previous tag of THIS
+      # crate. Tags are repo-global and interleaved across crates (c2pa-v*,
+      # c2patool-v*, c2pa-c-ffi-v*), and GitHub's automatic "previous tag" is the
+      # chronologically previous tag regardless of crate, producing nonsense
+      # ranges. Picking the previous same-crate tag keeps the changelog within
+      # this crate's history.
+      - name: Generate release notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          TAG="${{ github.ref_name }}"
+          PREFIX="c2patool-v"
+          repo="${{ github.repository }}"
+          matching=$(gh api --paginate "repos/$repo/git/matching-refs/tags/$PREFIX" \
+                       --jq '.[].ref' | sed 's#refs/tags/##')
+          sorted=$(printf '%s\n%s\n' "$matching" "$TAG" | sort -V -u)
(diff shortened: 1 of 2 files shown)


caiopensrc:
- #2307

PR #2315, opened 2026-07-17
ci: Automate train promotion so `-rc.N` is stripped before reaching stable

scouten-adobe:
## Problem

Promoting a baked release-candidate branch was a hand-run `git merge` of `0.N.0-rc` into `stable`. That merge carried the candidate's **prerelease** version (`0.N.0-rc.M`) onto `stable`, and `release-plz release-pr` then aborted — as it just did for the 0.90.0 train ([run](https://github.com/contentauth/c2pa-rs/actions/runs/29547621229/job/87783227903)):

```
local package `c2pa` has a greater version (0.90.0-rc.3) with respect to the
registry package (0.89.3), but the git tag c2pa-v0.90.0-rc.3 exists.
```

release-plz's guard matches a git tag against the **exact local version**, and `release-rc.yml` had already created `c2pa-v0.90.0-rc.3` for the prerelease binaries. So the RC tag looked like a botched publish, and the release line was stuck. This recurs **every train** — and the fix must not delete the RC tags (they own the prerelease binary releases).

## Fix

New **`release-promote.yml`** (`workflow_dispatch` on the RC branch) automates the promotion the docs previously described as a manual step:

1. **Snapshots** the outgoing `stable` as `v0.<old>` (retiring line, for backports).
2. **Strips `-rc.N` → final `0.N.0`** on a `promote/…` branch *before* it reaches `stable`. release-plz then looks for `c2pa-v0.N.0` (doesn't exist) — no collision, **no RC tags deleted**.
3. **Generates the changelog** with release-plz (see below).
4. **Opens a release PR** into `stable`, labeled `release`. Merging it publishes — so publishing stays a deliberate, reviewed merge and the in-repo changelog is never skipped (the failure mode we hit on #2313).

### Changelog handling (the part most worth reviewing)

release-plz fights us two ways here, both handled **non-destructively**:

- It **won't touch a version already set ahead of the registry** → the workflow temporarily resets versions to the last published release, lets release-plz generate, then restores the forced final versions.
- It **baselines the changelog against the highest matching tag** (a `-rc.N` build tag) → the workflow deletes `-rc.N` tags from its *ephemeral checkout only* (never pushed; remote tags untouched) so the baseline is the last real release.

release-plz computes each crate's bump from its own commits and **can't know the train's forced-minor, lockstep-across-crates policy** (e.g. c2patool with no commits of its own would compute `0.26.73`, not `0.27.0`). So `normalize_promotion_changelog.py` retitles each crate's newest section to the forced final version, or inserts a header-only section for a crate with no commits of its own. Tested against all three cases (retitle / insert / already-final), including the c2pa-c-ffi case where the changelog trails its published version.

## Notes

- Draft — the changelog dance is the fiddly part; would appreciate a close look there. An alternative I considered was moving the RC binary tags out of the `{crate}-v*` namespace release-plz keys on; happy to discuss if you'd prefer that direction.
- Docs updated (`docs/release-process.md`): promotion step, a new **Release promotion** automation subsection, and the branch-protection bypass list.
- Requires the `RELEASE_PLZ_TOKEN` identity to be able to create `v0.*` branches.

### DIFF
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -1,0 +1,228 @@
+name: Promote release candidate
+
+# Promotes a baked release-candidate branch (`0.N.0-rc`) onto the `stable`
+# release line -- the deliberate, manual final step of a breaking train's bake
+# (see docs/release-process.md, "Track 2 -- The breaking train").
+#
+# WHY THIS EXISTS
+#
+# Promotion used to be a hand-run `git merge` of the RC branch into `stable`.
+# That merge carried the candidate's *prerelease* version (`0.N.0-rc.M`) onto
+# `stable`, and release-plz then aborted its `release-pr` run with:
+#
+#     local package `c2pa` has a greater version (0.90.0-rc.3) with respect to
+#     the registry package (0.89.3), but the git tag c2pa-v0.90.0-rc.3 exists.
+#
+# release-plz's guard matches a git tag against the EXACT local version, and
+# `release-rc.yml` had already created `c2pa-v0.90.0-rc.3` for the prerelease
+# binaries. So the RC tag looked to release-plz like a botched publish, and the
+# release line was stuck every train.
+#
+# Stripping `-rc.N` to the final `0.N.0` BEFORE it reaches `stable` avoids the
+# collision entirely -- release-plz then looks for `c2pa-v0.N.0`, which does not
+# exist -- WITHOUT deleting any RC tags.
+#
+# WHAT IT DOES
+#
+#   1. Snapshots the outgoing `stable` as a `v0.<old>` branch, so the retiring
+#      line stays available for backports.
+#   2. Cuts a promotion branch from the RC branch and strips `-rc.N` -> final
+#      versions (`cargo set-version`, which also fixes intra-workspace deps).
+#   3. Generates the final-version changelog with release-plz (see the changelog
+#      step for how it works around release-plz's quirks).
+#   4. Opens a PR from the promotion branch into `stable`, labeled `release`.
+#
+# Merging that PR is what publishes: the push to `stable` triggers release.yml
+# (`release-plz release`), which publishes the crates, creates the real
+# `c2pa-v0.N.0` / `c2patool-v0.M.0` tags, and drives the binary builds. Because
+# the promotion PR already carries BOTH the final versions and the changelog,
+# publishing is a reviewed, deliberate PR merge -- not a side effect of a raw
+# branch merge, and the in-repo changelog is never skipped.
+#
+# The `release` label matters twice: it exempts the PR from the upstream-first
+# check (the version bump legitimately originates on the release line), and it
+# is the label release-plz already recognizes for release PRs.
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  workflow_dispatch:
+
+jobs:
+  promote:
+    name: Promote release candidate
+    runs-on: ubuntu-latest
+    steps:
+      - name: Refuse to run outside a release-candidate branch
+        run: |
+          # An RC branch is named `<version>-rc` (no numeric suffix). Guard
+          # against dispatching against main / stable / a v0.x line.
+          case "${{ github.ref_name }}" in
+            *-rc) echo "Promoting RC branch ${{ github.ref_name }}." ;;
+            *)
+              echo "::error::release-promote must be dispatched on a *-rc branch, not '${{ github.ref_name }}'."
+              exit 1
+              ;;
+          esac
+
+      - name: Checkout RC branch
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          ref: ${{ github.ref_name }}
+          fetch-depth: 0
+          token: ${{ secrets.RELEASE_PLZ_TOKEN }}
+
+      - name: Install Rust toolchain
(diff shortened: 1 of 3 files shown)


.github/workflows/release-promote.yml
ok-nick, line 1: Looks like https://github.com/release-plz/release-plz/issues/2691 would simplify this work if implemented. Can you confirm? Should we track it?

.github/workflows/release-promote.yml
```diff
@@ -0,0 +1,228 @@
...
+          cargo set-version -p c2patool "${{ steps.ver.outputs.fin_tool }}"
+
+          # Normalize each published crate's newest changelog section so its
+          # header version matches the forced final version (rewriting the
+          # header release-plz just wrote, or inserting a header-only section if
+          # release-plz found nothing releasable for that crate).
+          python3 .github/scripts/normalize_promotion_changelog.py \
+            "CHANGELOG.md:c2pa:${{ steps.ver.outputs.old_c2pa }}:${{ steps.ver.outputs.fin_c2pa }}:${PRE_C2PA}" \
+            "c2pa_c_ffi/CHANGELOG.md:c2pa-c-ffi:${{ steps.ver.outputs.old_c2pa }}:${{ steps.ver.outputs.fin_c2pa }}:${PRE_FFI}" \
+            "cli/CHANGELOG.md:c2patool:${{ steps.ver.outputs.old_tool }}:${{ steps.ver.outputs.fin_tool }}:${PRE_TOOL}"
```
ok-nick, line 195: I don't understand this normalization step. If `release-plz` is writing the wrong versions here it seems like a much deeper problem or that we are doing something wrong?

.github/workflows/release-promote.yml
```diff
@@ -0,0 +1,228 @@
...
+          git commit -am "chore: promote ${{ github.ref_name }} to c2pa ${FIN_C2PA} / c2patool ${FIN_TOOL}"
+          git push -f origin "$BRANCH"
+
+          gh pr create \
+            --base stable \
+            --head "$BRANCH" \
+            --label release \
+            --title "chore: release c2pa ${FIN_C2PA} / c2patool ${FIN_TOOL}" \
+            --body "$(cat <<EOF
+          Promotion of release-candidate branch \`${{ github.ref_name }}\` (candidate \`${{ steps.ver.outputs.rc_c2pa }}\`) onto \`stable\`.
```
ok-nick, line 218: Should we handle if the PR already exists? Such as if this workflow needs to run twice in case of a problem?

.github/workflows/release-promote.yml
scouten-adobe: I don't think so, mostly because we are choosing not to release the RCs to crates.io; that complicates things in a way that the issue you linked to doesn't appear to resolve.

.github/workflows/release-promote.yml
scouten-adobe: To some extent, we're fighting `release-plz`'s natural choice of versioning with the release-train workflow. Consider this hypothetical scenario when we cut the next release train (0.91, successor to the just-released 0.90): `c2pa` itself has breaking changes, so it would naturally be bumped to 0.91 when the release train is cut, but `c2patool` doesn't have any such changes. It would instead remain at 0.27.next when we want (according to our process) to have it bump up.

A variation on this happened in this cycle since we had an effectively equivalent 0.89.3 release of `c2pa` just before we started this process.

This normalization fires in a situation such as that.

.github/workflows/release-promote.yml
scouten-adobe: Oy. Yeah, that could happen. I think in that scenario, my vote would be to manually delete the first PR. This feels complicated enough as it is.

.github/workflows/release-promote.yml
tmathern: To some extent, we're fighting release-plz's natural choice of versioning with the release-train workflow. <= Seems we have a habit of doing that. Maybe worth a release-plz feature/config request?

ok-nick: I agree with @tmathern. The underlying issue is that we don't have an automated way of detecting breaking changes in our CLI, which is complicated to do. We play it safe by always doing a minor version bump even if that's overly conservative. I think it would be reasonable for `release-plz` to support always bumping the minor instead of patch in this case.

Maybe in place of that we can use [`release-plz set-version`](https://release-plz.dev/docs/usage/set-version) to explicitly set the version? It seems like this is the natural choice if you want to override `release-plz`'s decision.

.github/workflows/release-promote.yml
ok-nick: I'm okay with that.

scouten-adobe:
Closing in favor of #2636.

The landscape changed under this PR: the c2patool split (#2617, merged) removed the 3-crate world this was built around, and #2617 added `release-pr.yml` and reshaped the release workflows. The approach also shifted — instead of building a promotion PR with a pre-generated changelog (the `normalize_promotion_changelog.py` dance here), #2636 automates a **force-push onto `stable`** that resets the candidate version so the existing `release-pr.yml` opens the reviewed release PR, and it moves RC build tags out of the `c2pa-v*` namespace so release-plz baselines the changelog correctly on its own. That removes the guard collision and the changelog normalization this PR worked around, so none of the code here carries forward.

Thanks @ok-nick and @tmathern for the review — the `release-plz set-version` suggestion lives on in #2636 as the manual override on the release PR when a policy-breaking train under-bumps.

PR #2335, opened 2026-07-21
chore(sdk): Mark CLAIM_REVIEW deprecated

mpgopala:
`stds.schema-org` was deprecated in c2pa spec 2.0 version. Mark all members of `stds.schema-org` as deprecated.

## Checklist

### DIFF
--- a/sdk/src/assertions/labels.rs
+++ b/sdk/src/assertions/labels.rs
@@ -176,7 +176,7 @@ pub const CLAIM_REVIEW: &str = "stds.schema-org.ClaimReview";
 /// Label prefix for a Creative Work assertion. Deprecated in C2PA 2.0 spec.
 #[deprecated(
     since = "0.91.0",
-    note = "This attribute is deprecated from c2pa spec version 2.0"
+    note = "This attribute is deprecated from C2PA spec version 2.0. Will be deleted on or after 2026-12-01."
 )]
 pub const CREATIVE_WORK: &str = "stds.schema-org.CreativeWork";
 
--- a/sdk/src/assertions/labels.rs
+++ b/sdk/src/assertions/labels.rs
@@ -162,7 +162,13 @@ pub const IPTC_PHOTO_METADATA: &str = "stds.iptc.photo-metadata";
 ///
 /// See [Use of Schema.org - C2PA Technical Specification](https://spec.c2pa.org/specifications/specifications/2.3/specs/C2PA_Specification.html#_use_of_schema_org).
 #[doc(hidden)]
+#[deprecated(
+    since = "0.91.0",
+    note = "This attribute is deprecated from C2PA spec version 2.0. Will be deleted on or after 2026-11-09."
+)]
 pub const SCHEMA_ORG: &str = "schema.org";
+
+pub(crate) const SCHEMA_ORG_INTERNAL: &str = "schema.org";
 
 /// Label prefix for a Claim Review assertion. Deprecated since C2PA 2.0 spec.
 #[deprecated(
--- a/sdk/src/assertions/labels.rs
+++ b/sdk/src/assertions/labels.rs
@@ -167,9 +167,11 @@ pub const SCHEMA_ORG: &str = "schema.org";
 /// Label prefix for a claim review assertion.
 ///
 /// Label prefix for a creative review assertion. Deprecated in c2pa 2.0 spec.
+#[deprecated(note = "This attribute is deprecated from c2pa spec version 2.0")]
 pub const CLAIM_REVIEW: &str = "stds.schema-org.ClaimReview";
 
 /// Label prefix for a creative work assertion. Deprecated in c2pa 2.0 spec.
+#[deprecated(note = "This attribute is deprecated from c2pa spec version 2.0")]
 pub const CREATIVE_WORK: &str = "stds.schema-org.CreativeWork";
 
 /// Label prefix for a timestamp assertion.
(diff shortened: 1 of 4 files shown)


tmathern (commented):
Is there no deprecation syntax?

mpgopala:
> Is there no deprecation syntax?

I made this similar to CREATIVE_WORK. 
While these are spec-deprecated, We should still be able to read if we come across one which is not yet deprecated in Rust.

scouten-adobe (changes requested):
Please use the Rust deprecation syntax as described here: https://github.com/contentauth/c2pa-rs/blob/main/docs/deprecation-policy.md#rust

mpgopala:
> Please use the Rust deprecation syntax as described here: https://github.com/contentauth/c2pa-rs/blob/main/docs/deprecation-policy.md#rust


@scouten-adobe : We cannot deprecate these as we should still be able to read these set by someone signing using an older c2pa version. Only writes need to be deprecated. So, for now, it can only be documented.

scouten-adobe:
> > Please use the Rust deprecation syntax as described here: https://github.com/contentauth/c2pa-rs/blob/main/docs/deprecation-policy.md#rust
> 
> @scouten-adobe : We cannot deprecate these as we should still be able to read these set by someone signing using an older c2pa version. Only writes need to be deprecated. So, for now, it can only be documented.

I'd like very much to have this name marked deprecated in the public API surface so as to discourage anyone from using it. We could use a _different_ non-public constant to support the legacy reading cases.

mpgopala:
@scouten-adobe @tmathern 
Marked both the attributes as deprecated. Interestingly, those attributes were never used in our code. Any use of CLAIM_REVIEW and CREATIVE_WORK happened by actually defining those strings in tests.

gpeacock (approved):

sdk/src/assertions/labels.rs
```diff
@@ -166,10 +166,18 @@ pub const SCHEMA_ORG: &str = "schema.org";
 
 /// Label prefix for a claim review assertion.
 ///
-/// See [Claim review - C2PA Technical Specification](https://spec.c2pa.org/specifications/specifications/2.3/specs/C2PA_Specification.html#_claim_review).
+/// Label prefix for a creative review assertion. Deprecated in c2pa 2.0 spec.
```
scouten-adobe:
```suggestion
/// Label prefix for a claim review assertion. Deprecated since C2PA 2.0 spec.
```

sdk/src/assertions/labels.rs
```diff
@@ -166,10 +166,18 @@ pub const SCHEMA_ORG: &str = "schema.org";
...
+#[deprecated(
+    since = "0.91.0",
+    note = "This attribute is deprecated from c2pa spec version 2.0"
```
scouten-adobe:
```suggestion
    note = "This attribute is deprecated from C2PA spec version 2.0. Will be deleted on or after 2026-12-14."
```

sdk/src/assertions/labels.rs
```diff
@@ -166,10 +166,18 @@ pub const SCHEMA_ORG: &str = "schema.org";
...
+)]
 pub const CLAIM_REVIEW: &str = "stds.schema-org.ClaimReview";
 
-/// Label prefix for a creative work assertion.  Deprecated.
+/// Label prefix for a creative work assertion. Deprecated in c2pa 2.0 spec.
```
scouten-adobe: This was a proper name in the schema.org parlance.


```suggestion
/// Label prefix for a Creative Work assertion. Deprecated in C2PA 2.0 spec.
```

sdk/src/assertions/labels.rs
```diff
@@ -166,10 +166,18 @@ pub const SCHEMA_ORG: &str = "schema.org";
...
+#[deprecated(
+    since = "0.91.0",
+    note = "This attribute is deprecated from c2pa spec version 2.0"
```
scouten-adobe:
```suggestion
    note = "This attribute is deprecated from C2PA spec version 2.0. Will be deleted on or after 2026-12-01."
```

scouten-adobe (changes requested):
Very minor tweak requested. Please use all caps when referring to C2PA.

scouten-adobe (approved):

scouten-adobe:
Hold the phone. I gave you the wrong dates!

sdk/src/assertions/labels.rs
```diff
@@ -162,14 +162,26 @@ pub const IPTC_PHOTO_METADATA: &str = "stds.iptc.photo-metadata";
 ///
 /// See [Use of Schema.org - C2PA Technical Specification](https://spec.c2pa.org/specifications/specifications/2.3/specs/C2PA_Specification.html#_use_of_schema_org).
 #[doc(hidden)]
+#[deprecated(
+    since = "0.91.0",
+    note = "This attribute is deprecated from C2PA spec version 2.0. Will be deleted on or after 2026-11-09."
+)]
 pub const SCHEMA_ORG: &str = "schema.org";
```
ok-nick, line 169: We still support reading pre-2.x manifests. It might be deprecated in the latest spec, but we support all spec versions.

mpgopala: For reading we use SCHEMA_ORG_INTERNAL which is not a `pub` member. Deprecating this is so that no client of c2pa sdk should use this label going forward as it is deprecated.
Done as recommended by @scouten-adobe [here](https://github.com/contentauth/c2pa-rs/pull/2335#issuecomment-5052567964).

ok-nick: OK I see, that's fine. I prefer using the same constant and `#[allow(deprecated)]` where we need, although no strong opinions.

ok-nick (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 30856484923](https://github.com/contentauth/c2pa-rs/actions/runs/30856484923).

PR #2336, opened 2026-07-21
feat: Add binding to take last error in C FFI

tmathern:
Currently we can set a last error, we can read a last error, but we can not clear it. 
Being able to clear it would help to fine-tune error handling in bindings (notably Python) since an error that was handled and might have been therefore "recovered" could be cleared. This would reset the error slot.

Also small refactor to reuse existing code/integrate new code.

## Checklist


c2pa_c_ffi/src/c_api.rs
```diff
@@ -506,6 +506,33 @@ pub unsafe extern "C" fn c2pa_error_set_last(error_str: *const c_char) -> c_int
...
+///
+/// # Errors
+/// Never fails. Returns an empty string (and writes code 0) when no error is set.
+///
+/// # Safety
+/// The returned value MUST be released by calling `c2pa_free`
+/// and is no longer valid after that call.
+/// If `code` is non-NULL it must point to a writable `int`.
+#[no_mangle]
+pub unsafe extern "C" fn c2pa_error_take(code: *mut c_int) -> *mut c_char {
```
gpeacock, line 528: I'm curious why we want this to return an empty string when there's nothing to take vs returning null . I thought the intent of this is to make it clear when there is no previous error to return.

gpeacock (commented):
it is important that callers know this should never be used to know if there was an error - always check the return value first! I'm unclear in exactly where/how this would be used.
I don't conceptually have a problem with it. But I've seen c_ffi users who don't understand that they always need to check return values and try to check for errors only by callingj c2pa-error and expecting to always be cleared if there was no error.

c2pa_c_ffi/src/c_api.rs
tmathern: c2pa_error also returns an empty string on no error. I figured we want the same contract here? Since c2pa_error peeks, and that one takes+clears?

(Also in other non-error related calls, a null return may mean check the error... which coming from that function is a bit a chicken-egg problem then).

tmathern:
> it is important that callers know this should never be used to know if there was an error - always check the return value first! I'm unclear in exactly where/how this would be used. I don't conceptually have a problem with it. But I've seen c_ffi users who don't understand that they always need to check return values and try to check for errors only by callingj c2pa-error and expecting to always be cleared if there was no error.

The reason I was thinking of adding this is to "clear" the sticky error slot. Because the last error, once set, is always there. In a binding, like in Python, I can retrieve that error and handle it. Then execution continues. But if the execution would check an error message only (for reasons), I would potentially see a sticky one. This is mostly to "clear" the slot once we are sure an error was handled.

gpeacock:
> Then execution continues. But if the execution would check an error message only (for reasons), I would potentially see a sticky one. This is mostly to "clear" the slot once we are sure an error was handled.

It is the "for reasons" that worries me. Anything that tries to read c2pa_error() except in response to an error from a c2pa function that was just called, is going to run into problems, since methods that don't return errors never clear the error. It will always mess things up. 
I'd rather provide some alternative, like registering a callback to hold the last error whenever we set it.

tmathern:
> > Then execution continues. But if the execution would check an error message only (for reasons), I would potentially see a sticky one. This is mostly to "clear" the slot once we are sure an error was handled.
> 
> It is the "for reasons" that worries me. Anything that tries to read c2pa_error() except in response to an error from a c2pa function that was just called, is going to run into problems, since methods that don't return errors never clear the error. It will always mess things up. I'd rather provide some alternative, like registering a callback to hold the last error whenever we set it.

How would a callback work here? All objects would need to define one to get "their" errors?

The reason specifically for me here, is that if I create an object in Python, I would like to clear the previous error slate to start from a clean slate, with no sticky error. (Because we don't return specific error codes which indicate an error, we return roughly mostly ok/not ok encoded in some way, either 0/-1 or null/pointer adress). 

Another case is in unit tests: In normal runtime it is fine to have a sticky one because it is checked after an error, but when I plant errors in tests (using, ironically, set last error), I pollute the error state and I can't clear it without what feels trickery (setting an empty string as error) as part of the test teardown. That can be annoying because it breaks test isolation easily, and it is found only when test run order changes. 

So... The thing I am doing right now mostly in negative tests is setting an empty string in set last error instead to reset error statest. Is this the intended way to clear error state maybe then? (It feels... Wrong). 

Eventually, I think we should switch to returning error codes on the calls, and maybe remove set_last_error use. Then it is also easier to add more thread safety around erro handling at that layer. (But that would be for another day/PR, since that clearly is a very breaking change).

PR #2341, opened 2026-07-21
feat: Add `try_*` APIs for `with_archive` on Builder, `with_stream`, `with_manifest_data_and_stream`, `with_fragment` on Reader to propagate errors

tmathern:

For some bindings using the C FFI, on error cases, it can be ambiguous to determine if a pointer passed in is still own by the caller or if the function took ownership. 

This PR suggests a change to propagate errors differently in the C FFI:
- Introduce try_* APIs: `try_with_archive` on Builder, `try_with_stream`, `try_with_manifest_data_and_stream`, `try_with_fragment` on Reader. Those APIs return Self and a potential error (if any) as result. The pick of those 4 APIs is not random, because it is those which can be ambiguous, as the C FFI can fail in the function body in a way that a caller can not easily determine the pointer ownership because of our own pointer tracking. The goal is to make the contract here clearer/more deterministic so a caller can rely more on return values (and has less errors cases to check).
- Use those APIs in the C FFI to handle errors: `reconfigure_or_restore` either replaces a pointer, or restores the original one so the caller can free it. 

## Checklist

### DIFF
--- a/c2pa_c_ffi/src/c_api.rs
+++ b/c2pa_c_ffi/src/c_api.rs
@@ -66,15 +66,16 @@ unsafe fn is_safe_buffer_size(size: usize, ptr: *const c_uchar) -> bool {
 /// Runs a fallible, value-consuming operation on the value
 /// behind an untracked pointer.
 ///
-/// * On success the result is tracked in a fresh allocation and its pointer is
-///   returned; the input allocation is freed and its pointer is consumed.
-/// * On error the input value is restored into its allocation and re-tracked, the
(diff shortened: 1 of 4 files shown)


c2pa_c_ffi/src/c_api.rs
```diff
@@ -14,6 +14,7 @@
 use std::{
     os::raw::{c_char, c_int, c_uchar, c_void},
     sync::Arc,
+    mem::ManuallyDrop,
```
tmathern, line 17: Any better name idea?

PR #2344, opened 2026-07-22
fix: builder style c_ffi_api functions will now consistently consume the self parameter

gpeacock:
This is for builder model methods that pass in self and return Result<Self>. 
It ensures that all steps are done correctly since this is a common model

c2pa_reader_with_stream and four similar functions (with_manifest_data_and_stream,
with_fragment, builder_with_definition, builder_with_archive) left the input pointer
tracked and reusable if a secondary argument (format, stream, fragment, manifest_json)
failed validation, but consumed it if the underlying operation failed instead

untrack_or_return!/_null!/_int! now take ownership of the pointer directly instead of
just untracking it, and are called before any other validation, so every failure path
consumes the input uniformly via ordinary scope-exit drop.

Removed ~10 test assertions that freed a just-consumed pointer and checked it was
untracked. They only verified the old inconsistent behavior, and the check is
inherently unreliable in a parallel test binary: another concurrently-running test
can allocate a new tracked pointer at that exact freed address before the check runs.

### DIFF
--- a/c2pa_c_ffi/src/cimpl/macros.rs
+++ b/c2pa_c_ffi/src/cimpl/macros.rs
@@ -350,35 +350,34 @@ macro_rules! arc_tracked {
     }};
 }
 
-/// Untrack a pointer from the registry (ownership transfer from C to Rust).
+/// Untrack a pointer from the registry and take ownership of the pointee
+/// (ownership transfer from C to Rust), yielding it by value.
 ///
-/// Validates the pointer is tracked with the correct type, then removes it
-/// from the registry without running cleanup. Call this *before* `Box::from_raw()`
-/// in any FFI function that consumes a tracked pointer, so the registry doesn't
-/// hold a stale entry that would cause a double-free or false leak warning.
-///
-/// After this macro succeeds, the caller owns the allocation and must drop it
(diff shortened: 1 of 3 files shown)


c2pa_c_ffi/src/cimpl/macros.rs
```diff
@@ -403,6 +403,46 @@ macro_rules! untrack_or_return_null {
...
+///     std::ptr::null_mut()
+/// )
+/// ```
+///
+/// On `Err`, the boxed value is dropped via Rust's ordinary `Drop` — the pointer
+/// was already removed from the registry by the untrack step, so no explicit free call
+/// is needed (or correct: the pointer is no longer tracked, so `cimpl_free` would just
+/// report an error) at that point.
+#[macro_export]
+macro_rules! consume_or_return {
```
tmathern: That change looks good! Thanks!

tmathern (commented):
All changes in c2pa_c_ffi look good to me.

But what about the changes in `sdk/src/`? They look unrelated, maybe strays?

tmathern (approved):

tmathern:
LGTM!

PR #2345, opened 2026-07-22
docs: Add experimental features policy

scouten-adobe:
Sets forth a policy under which third-party / community contributions can be accepted into the SDK without active maintenance from the Adobe CAI team.


ok-nick (commented):
How do experimental features affect versioning? If there is a breaking/minor/patch in them?

docs/experimental-features.md
ok-nick, line 1: How do experimental features affect versioning? What if there is a breaking/minor/patch in them? I'm also curious how you feel if we named them "unstable" features to better match Rust's model.

scouten-adobe: Ooh, good point. I just learned that `cargo-semver-checks` has special handling for features with an `unstable-` or `unstable_` prefix. I'll be revising shortly.

docs/experimental-features.md
```diff
@@ -0,0 +1,76 @@
...
+
+The Content Authenticity Initiative (CAI) SDK is an evolving project, and we welcome contributions from the wider community. From time to time, an external contributor may propose a new feature that we believe is valuable but that the CAI team is not able to commit to maintaining as part of the core SDK.
+
+Rather than reject such contributions outright, we may accept them as **experimental features**. This document describes what that designation means, the conditions a contribution must satisfy to qualify, and the support expectations that apply on both sides.
+
+This policy complements the [Deprecation policy](deprecation-policy.md) and [Support tiers](support-tiers.md). Experimental features sit outside the stability guarantees described in those documents.
+
+## What is an experimental feature?
+
+An experimental feature is functionality that has been accepted into the SDK source tree but is **not** covered by the SDK's normal stability and maintenance commitments. We accept experimental features when we judge the feature to be of value to the community but cannot commit to maintaining it ourselves.
```
tmathern, line 11: Add also that they are probably not covered by the deprecation policy and its timelines?

docs/experimental-features.md
```diff
@@ -0,0 +1,76 @@
...
+>
+> The `unstable_` prefix is deliberate. [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) — which we run on every release-line PR (see the [release process](release-process.md#semver-checks)) — by default ignores features named `unstable`, `nightly`, `bench`, or `no_std`, and any feature whose name begins with `_`, `unstable-`, or `unstable_`. Naming experimental flags `unstable_<name>` therefore makes our tooling automatically treat them as exempt from breaking-change detection, so changing or removing an experimental API never forces a version bump on its own.
+
+### 2. Public API changes are gated
+
+Any changes to the public API surface must be visible **only** when the experimental feature flag is enabled. With the flag disabled, the public API must be identical to a build that does not include the feature at all (for example, by gating the affected items with `#[cfg(feature = "unstable_<name>")]`).
+
+### 3. New dependencies are optional and gated
```
tmathern, line 34: How do we handle security or licensing notice for such experimental features?

docs/experimental-features.md
```diff
@@ -0,0 +1,76 @@
...
+- A brief (a few sentences) description of the feature's behavior.
+- Contact information for support, including **at least one GitHub username** of a maintainer or sponsor who can answer questions about the feature.
+
+## Stability and support expectations
+
+By accepting a feature as experimental, the CAI team makes **no** stability guarantees about it:
+
+- **Unstable by design.** Any APIs or behaviors gated by an experimental feature flag are considered unstable and may be altered or removed at any time, in any release, without following the [Deprecation policy](deprecation-policy.md).
+- **Best-effort maintenance.** The CAI team will make a best effort to keep experimental features building and working as the primary SDK evolves, but reserves the right to disable or remove an experimental feature if it cannot be made compatible with the primary SDK code.
+- **Community-supported.** Day-to-day support for an experimental feature is primarily the responsibility of the contributor and the contacts listed in the registry, not the CAI team.
```
tmathern, line 66: What happens if the contact needs to handover a feature? Or if the contact can't continue supporting it? There might be a need for some process here.

tmathern (changes requested):
There are a few questions that probably deserve to be clarified in the docs.

docs/experimental-features.md
```diff
@@ -0,0 +1,76 @@
...
+
+The `unstable_` flag prefix makes this automatic: [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) ignores `unstable_`-prefixed features by default, so a change confined to an experimental feature is never flagged as a break and never triggers a breaking (`0.x.0`) bump. Experimental APIs are therefore free to change at will – including breaking changes, at any time, in any release – **provided that the change does not affect any API in the public/stable surface** (a default build, or any non-experimental feature). See [Semver checks](release-process.md#semver-checks) in the release process.
+
+## Promotion or removal
+
+An experimental feature is not expected to remain experimental forever. Over time, an experimental feature may be:
+
+- **Promoted** to a fully supported feature if it proves valuable and the CAI team is able to take on its maintenance. At that point it becomes subject to the normal stability and deprecation guarantees.
+- **Removed** if it falls out of maintenance, cannot be kept compatible with the primary SDK, or is no longer of sufficient value to justify its presence in the tree.
```
gpeacock: No deprecation policy for these, once they are removed, they are removed.

docs/experimental-features.md
gpeacock: What promises are we making about actually building these? Do we include them in some kind of all-features build? What happens when one of them doesn't build anymore? Do we stop building them all? Will this mess up anyplace we are already building with all-features?
Will we enforce cargo fmt and clippy rules on these? If so, are we responsible to update all these when rust version changes require updates?
Maybe we need a policy that says if one of these fails to pass, we remove it from builds until it is updated.

docs/experimental-features.md
```diff
@@ -0,0 +1,76 @@
...
+
+By accepting a feature as experimental, the CAI team makes **no** stability guarantees about it:
+
+- **Unstable by design.** Any APIs or behaviors gated by an experimental feature flag are considered unstable and may be altered or removed at any time, in any release, without following the [Deprecation policy](deprecation-policy.md).
+- **Best-effort maintenance.** The CAI team will make a best effort to keep experimental features building and working as the primary SDK evolves, but reserves the right to disable or remove an experimental feature if it cannot be made compatible with the primary SDK code.
+- **Community-supported.** Day-to-day support for an experimental feature is primarily the responsibility of the contributor and the contacts listed in the registry, not the CAI team.
+
+### Exempt from release-number calculations
+
+Because experimental features are gated behind a non-default flag and excluded from the public API of a default build, changing or removing one is **not** considered a breaking change under the [Deprecation policy](deprecation-policy.md), and it does **not** affect the crate's version number.
```
mpgopala: > it does **not** affect the crate's version number.
It would be difficult for users of SDK to actually use these experimental features if they do not affect version number when something is changed in them. Why not subject these also to the same norms as other features? (back ports/breaking release etc)? I agree we might need some support from the support team of such features. But, without that, the feature itself may not be useful.
Looking at our non-experimental features, do we have a process to build with various features turned on/off and make sure the build is OK? May be we need some policy there too?

docs/experimental-features.md
scouten-adobe: We address that in line 68 below. (We reserve the right to remove an experimental feature at any time if it falls out of maintenance.)

docs/experimental-features.md
scouten-adobe: Addressed in the latest push. Flags now use the `unstable_` prefix, which `cargo-semver-checks` (and, by extension, our release tooling) excludes by default — so a change confined to an experimental feature never forces a minor/breaking (`0.x.0`) bump; at most it rides a patch (`0.x.y`) release. To keep opt-in users informed, those changes are routed into a dedicated **Experimental** changelog section via an `experimental` commit scope.

On naming: I kept the *concept* name "experimental" but adopted `unstable_` as the flag prefix to match Rust's model and the tooling. See the new "Exempt from breaking-change calculations" section.

docs/experimental-features.md
scouten-adobe: Done — there's now an explicit statement up front that experimental features are **not** covered by the Deprecation policy or its timelines, and can be changed or removed in any release with no deprecation window.

docs/experimental-features.md
scouten-adobe: Good question. Added a new condition 6, "Same licensing and dependency review as the rest of the SDK": experimental code and any new dependencies get the same license-compatibility and security review as everything else in the tree — the exemption is from the team's *maintenance* commitment, not from the project's legal/supply-chain requirements. I also added a support-expectations bullet: a security or licensing problem the registered contact doesn't address promptly is grounds for immediately disabling or removing the feature, rather than waiting out any window.

docs/experimental-features.md
scouten-adobe: Added a "Handing off or losing a contact" subsection: contacts are kept current by PR (a contact who can no longer support a feature hands it off by naming a replacement), and a feature left with **no responsive contact** becomes an immediate removal candidate — building on the removal rule @scouten-adobe pointed to at line 68.

docs/experimental-features.md
scouten-adobe: Agreed — now stated explicitly (no deprecation policy or window; once removed, it's gone), both up front in "What is an experimental feature?" and in the "Exempt from breaking-change calculations" section.

docs/experimental-features.md
scouten-adobe: Great set of questions — added a "Build and CI expectations" section that answers them directly:

- **Building them:** a new **non-blocking** workflow, `experimental-features.yml`, builds/lints/tests the SDK with every `unstable_` feature enabled, on PRs and daily. It's advisory only (`continue-on-error`, not a required check), so it never blocks a merge.
- **When one stops building (incl. Rust-version churn / fmt / clippy):** we notify the registered contact; if it isn't fixed promptly, the feature is **disabled/removed** rather than holding up the SDK — essentially the policy you proposed.
- **The all-features concern is real:** our required jobs (`tier-1a` get-features, `beta-preflight` clippy, `tier-1b` udeps, canary, and `docs.rs`) currently build `--all-features` and so *do* pull experimental features into the merge gate today. Separating them out (compute a non-experimental feature set) is tracked in #2362.

docs/experimental-features.md
scouten-adobe: Fair concern — and I corrected the wording, which was too strong. An experimental change **can** still ride an ordinary patch (`0.x.y`) release; what it never does is force a minor/breaking (`0.x.0`) bump. And it's recorded in a dedicated **Experimental** changelog section (via the `experimental` commit scope), so anyone who opted into an `unstable_` feature gets a clear, human-readable signal of what changed between releases. What the exemption removes is the semver *guarantee* (and the automatic bump), not the paper trail — which mirrors Rust's own treatment of unstable/nightly features.

On the build-matrix half of your question: the new non-blocking `experimental-features.yml` exercises the `unstable_` features, and separating them from the required `--all-features` jobs is tracked in #2362.

docs/experimental-features.md
```diff
@@ -0,0 +1,99 @@
...
+
+A registered contact is the feature's lifeline. Contacts are kept current by pull request: a contact who can no longer support a feature should hand it off by updating the registry to name a replacement. If a feature is left with **no responsive contact** – nobody who answers questions or repairs breakage – it becomes an immediate candidate for removal under the rule above; the CAI team is not obligated to adopt an orphaned experimental feature.
+
+## Registry of experimental features
+
+The following table lists the experimental features currently present in the SDK. See [Conditions for acceptance](#conditions-for-acceptance) for what each entry must include.
+
+| Feature | Cargo flag | Description | Contact |
+| -- | -- | -- | -- |
+| _(none yet)_ | | | |
```
tmathern: Except there is one in main already I think? Didn't we gat something recently?

tmathern (approved):

docs/experimental-features.md
gpeacock: I'm pretty sure the all-features thing will bite us, so we definitely need to follow up on that soon! If we get a LOT of unstable features it could be a lot to deal with.

One other related point, how will this impact generated rust docs? I think currently we build them with all features and flag feature gated apis. Does that mean all these will show up in our docs? What if the docs build doesn't work because of it?

gpeacock (approved):
I left an additional question about how these features show up and impact our generated docs builds.

tmathern (approved):

mpgopala (approved):

docs/experimental-features.md
ok-nick: If we can avoid having unstable APIs **change** the behavior of existing APIs, and rather have them **only** be opt-in APIs, we can prevent potential issues with `--all-features`. I'd suggest we document this explicitly.

On the note of docs, I think it's fine but we should make sure it explicitly states unstable APIs. Rust docs does this nicely in `std`, see `simd` example [here](https://doc.rust-lang.org/std/simd/index.html).

ok-nick: If you take a look at the `simd` module I posted, you can also see that every unstable API in the docs is associated with a specific ticket to give context about why it's unstable and what's needed to make it stable, if applicable. I think we should adopt that policy here.

scouten-adobe: Good questions — this is now spelled out in the **[Build and CI expectations](https://github.com/contentauth/c2pa-rs/blob/docs/experimental-features-policy/docs/experimental-features.md#build-and-ci-expectations)** section. The short version:

**What we promise about building them.** Only that they build cleanly and pass `fmt` / `clippy` / their own tests *at contribution time*. Keeping them green after that is best-effort and is primarily the registered contact's responsibility, not ours. We make no standing promise that an experimental feature stays buildable.

**Are they in an all-features build?** Not in any *required* one anymore — that was the real risk you flagged. The Tier 1A/1B/2 suites, the beta preflight, and the cross-repo canary all used `--all-features`, which pulled experimental features into the merge gate. They now compute the feature set from `cargo metadata` and filter out `unstable_`-prefixed features, so experimental breakage can't fail those gates. Instead, a dedicated **non-blocking** workflow (`experimental-features.yml`) builds/lints/tests with every `unstable_` feature enabled, on PRs and daily. It's advisory only (`continue-on-error`, not a required check). The one intentional exception is the `docs.rs` build (and its Tier 1A preflight): we *do* still publish experimental API docs, clearly marked experimental.

**Will it mess up places we already build with `--all-features`?** That's exactly what the above fixes — those jobs no longer see `unstable_` features. (`c2pa-c-ffi` and `c2patool` doc builds keep `--all-features` since neither crate has experimental features.)

**`fmt` / `clippy`, and toolchain-churn upkeep?** Enforced at contribution time, and the non-blocking job keeps running `clippy -Dwarnings` on them. But we are **not** on the hook to fix them when a Rust upgrade breaks them — that's the contact's job.

**What happens when one stops building?** Your proposed policy, adopted verbatim: the advisory job goes red, we ping the registered contact, and if it isn't fixed promptly we **disable it** (drop it from the build, and ultimately remove it from the tree) rather than hold up the SDK. We don't stop building the others — disabling the broken one keeps the advisory signal useful for the rest.

Every experimental feature also now needs a standing discussion issue (e.g. #2368 for `unstable_builder_filter`) where exactly these hosting/maintenance questions get hashed out per-feature.

caiopensrc:

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2345-to-stable origin/stable
cd .worktree/backport-2345-to-stable
git switch --create backport-2345-to-stable
git cherry-pick -x 8a4228729193b1e97a1abd69b6a986555281f824
```

PR #2349, opened 2026-07-23
fix: Reject loopback and DNS-rebinding hosts in did:web resolution (CAI-10364)

mpgopala:
## Summary

- Reject `did:web` domains that are bare IPv4/IPv6 literals, `localhost`, or subdomains of known DNS-rebinding services (`nip.io`, `sslip.io`) — closes an SSRF path where a malicious `did:web` identifier could force resolution to point at loopback, link-local, or internal RFC-1918 addresses (e.g. cloud metadata endpoints).
- Rename the test suite's mock resolution domain from the literal `"localhost"` to a dedicated `did_web::TEST_MOCK_DOMAIN` (`"did-web-mock.test"`), removing the `#[cfg(test)]` carve-out that previously exempted `"localhost"` from the guard during `cargo test`. The guard now runs identically in test and production builds, and its rejection of `did:web:localhost` is itself covered by a test.

## Changes

- `sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs`
  - `to_url`: strip `[`/`]` before IP-literal parsing, add case-insensitive suffix check against `localhost` / `nip.io` / `sslip.io`.
  - Tests: new assertions for IP literals, `localhost` (both cases), `nip.io`/`sslip.io` subdomains (including case-insensitivity and a non-boundary-substring negative case), plus mock-server tests updated to use `TEST_MOCK_DOMAIN` instead of `localhost`.

## Test plan

- [x] `cargo test -p c2pa --lib did_web` — 6/6 pass
- [x] `cargo test -p c2pa --lib identity::` — 76/76 pass (1 unrelated ignore)
- [x] `cargo clippy -p c2pa --lib --tests` — clean

### DIFF
--- a/sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
+++ b/sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
@@ -198,8 +198,27 @@ pub(crate) fn to_url(did: &str) -> Result<String, DidWebError> {
     // Reject bare IPv4/IPv6 literals — did:web requires a DNS domain name.
     // Domain may include a %3A-encoded port (e.g. "192.168.1.1%3A8080"), so
     // strip the port suffix before checking.
-    let host_part = domain_name.split("%3A").next().unwrap_or(domain_name);
-    if host_part.parse::<std::net::IpAddr>().is_ok() {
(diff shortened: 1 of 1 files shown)


scouten-adobe (approved):

caiopensrc:
- #2353

PR #2351, opened 2026-07-23
chore: with_stream should be public

alextrnnn:
The deprecation note recommends to use a method not publicly visible:
https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/ingredient.rs#L911


gpeacock (changes requested):
Why do you need this? Ingredient creation should only be done via the Builder apis and my intent is to deprecate direct ingredient creation.

scouten-adobe:
Closing for lack of activity.

PR #2356, opened 2026-07-23
chore: Characterize BMFF timed-media track-Merkle verification path

scouten-adobe:
## What

Adds `sdk/tests/bmff_timed_media_merkle.rs`, the first end-to-end coverage of the **timed-media, track-based Merkle** branch of `BmffHash::verify_stream_hash` ([`bmff_hash.rs:1387`](https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/assertions/bmff_hash.rs#L1387)) — the path that enumerates tracks and reads individual samples via the third-party `mp4` crate (`Mp4Reader::read_header` / `tracks()` / `read_sample()`).

This path is **verify-only legacy BMFF-v2 interop**: every current signer emits `fixed_block_size` / mdat-indexed Merkle maps and routes to `validate_merkle_maps_mdat_boxes` instead, so the track path had no direct test.

## Why

Prep for a follow-up PR that replaces the `mp4` crate with a hardened native sample reader (addresses CAI-12277 / VULN-35208, the crafted-MP4 `emsg` integer-underflow DoS, plus additional panic vectors in the same crate). These tests are the **behavioral equivalence oracle** for that swap: they must pass unchanged afterward.

## Coverage

Built with hand-assembled MP4s through the public API only (no `mp4` internals, no binary fixture):

- **Sample-offset matrix (positive):** single/multi-sample, single/multi-chunk, `stco` vs `co64`, fixed vs variable `stsz`, multi-run `stsc` `first_sample` expansion.
- **Error surface (negative):** tampered sample → `HashMismatch`; unknown track `local_id`; empty `stsc`; `samples_per_chunk == 0`.

No production code changes; the `mp4` dependency is untouched in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2357, opened 2026-07-23
fix: Replace mp4 crate with hardened native BMFF sample reader (CAI-12277)

scouten-adobe:
Fixes the crafted-MP4 integer-underflow DoS in CAI-12277 / VULN-35208 by **removing the third-party `mp4` crate** rather than forking it.

> **Stacked on #2356** (the characterization tests). Review/merge that first; this PR's base will retarget to `main` once #2356 lands. The diff here is just the swap.

## Background

`c2pa-rs` used the `mp4` crate (v0.14.0, unmaintained 3+ years) in exactly one place: the timed-media, track-based Merkle verification path, to enumerate tracks and read sample bytes. The reported underflow lived in the crate's `EmsgBox::read_box`; building coverage for #2356 surfaced **additional** panic vectors in the same crate reachable from the public `verify_stream_hash` API (e.g. an `unwrap()` on a missing `stts` entry). A `[patch.crates-io]` fork would not protect downstream consumers of the published `c2pa` crate, so we remove the dependency instead — killing this bug and the whole class of future `mp4`-crate parser CVEs.

## What changed

- New hardened `pub(crate)` sample reader in `asset_handlers/bmff_io.rs` (`BmffSampleReader`), reusing the module's existing box-header primitives. It parses only `tkhd` / `stsc` / `stco` / `co64` / `stsz` and never touches `emsg` / `stts` / `ctts` / `stsd` (where the crate's panics lived).
- All offset/size arithmetic is checked:
  - sample-table entry counts bounded against box size before allocating (memory amplification);
  - `samples_per_chunk == 0` and `chunk_id == 0` rejected (no divide-by-zero / index underflow);
  - intra-chunk accumulation and final offset use `checked_add`;
  - a sample read is bounds-checked against stream length **before** the buffer is allocated.
  - Malformed input → `Error::InvalidAsset`, never a panic.
- `verify_stream_hash`'s timed-media branch now uses `BmffSampleReader`; the free `stsc_index` helper became a method.
- Dropped `mp4 = "0.14.0"` from `sdk/Cargo.toml` (and `Cargo.lock`).
- Removing `use mp4::*` surfaced `box_list_to_user_exclusions` relying on the re-exported `mp4::Result` alias; it now uses `crate::Result`.

## Testing

- The #2356 characterization suite **passes unchanged** against the native reader — behavioral equivalence across the sample-offset matrix and error surface.
- Added: the CAI-12277 `emsg` regression (crafted asset now returns an error, previously panicked) and two memory-amplification guards (oversized sample read; oversized chunk-table entry count).
- Full `c2pa` lib suite green with `--features file_io` (848 passed), including all BMFF sign/verify and fragmented tests.


sdk/src/assertions/bmff_hash.rs
```diff
@@ -1390,10 +1393,9 @@ impl BmffHash {
                 let track_to_bmff_merkle_map = self.split_bmff_merkle_map(bmff_merkle.clone())?;
 
                 reader.rewind()?;
-                let buf_reader = BufReader::new(reader);
-                let mut mp4 = mp4::Mp4Reader::read_header(buf_reader, size)
+                let media = BmffSampleReader::from_stream(reader)
                     .map_err(|_e| Error::InvalidAsset("Could not parse BMFF".to_string()))?;
-                let track_count = mp4.tracks().len();
+                let track_count = media.tracks().len();
```
tmathern, line 1412: Does the verification run always, or only when there are tracks?
That hits only tracks we could read. This confounds the notion of tracks and timed media.

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,489 @@ pub enum BmffError {
...
+/// Returns `Ok(None)` if the track lacks the sample tables required to read
+/// samples.
+fn parse_trak<R: Read + Seek + ?Sized>(
+    reader: &mut R,
+    start: u64,
+    end: u64,
+) -> Result<Option<SampleTrack>> {
+    let (tkhd_start, tkhd_end) = match find_box(reader, start, end, BoxType::TkhdBox)? {
+        Some(r) => r,
+        None => return Ok(None),
```
tmathern, line 3149: Why does it return None and not an error? And error would make it easier to match on to determine what happened here (Error vs no track).

Because right now we can't distinguish a legitimate error against "skip this".

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,489 @@ pub enum BmffError {
...
+/// Returns `(fixed_sample_size, sample_sizes, sample_count)`.
+fn parse_stsz<R: Read + Seek + ?Sized>(
+    reader: &mut R,
+    start: u64,
+    end: u64,
+) -> Result<(u32, Vec<u32>, u32)> {
+    reader.seek(SeekFrom::Start(start))?;
+    let (_version, _flags) = read_box_header_ext(reader)?;
+    let sample_size = reader.read_u32::<BigEndian>()?;
+    let sample_count = reader.read_u32::<BigEndian>()?;
```
tmathern, line 3334: Doesn't those still need to be checked against limits, like stream sizes for instance (or whatever we have at hand)?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1466,14 +1460,16 @@ impl BmffHash {
                                 e.insert(hasher_enum);
                             }
 
-                            if let Ok(Some(sample)) = &mp4.read_sample(track_id, sample_id) {
+                            if let Ok(Some(sample)) =
```
tmathern: That is where I think additional error matching may shine to report errors. Gives us more flexibility to distinguish cases and be more precise in handling of things, and what we report as error vs what can be acceptable. Potentially, if we need that level of precision.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1390,10 +1393,9 @@ impl BmffHash {
                 let track_to_bmff_merkle_map = self.split_bmff_merkle_map(bmff_merkle.clone())?;
 
                 reader.rewind()?;
-                let buf_reader = BufReader::new(reader);
-                let mut mp4 = mp4::Mp4Reader::read_header(buf_reader, size)
+                let media = BmffSampleReader::from_stream(reader)
```
tmathern: I wonder if this will cause many small reads (lots of seeks on disk), the previous version seemed to read more at once to avoid that.

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,489 @@ pub enum BmffError {
...
+/// `entries_start`, would read past `end` (guards against memory-amplification),
+/// returning the count as a `usize` when it fits.
+fn bounded_entry_count(
+    declared: u32,
+    entries_start: u64,
+    end: u64,
+    entry_size: u64,
+) -> Result<usize> {
+    let available = end.saturating_sub(entries_start);
+    let max_entries = available / entry_size;
```
tmathern, line 3241: Divide by zero risk!

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,489 @@ pub enum BmffError {
...
+    reader.seek(SeekFrom::Start(pos))?;
+    let header = BoxHeaderLite::read(reader)?;
+    let content_start = reader.stream_position()?;
+
+    let box_end = pos
+        .checked_add(header.size)
+        .ok_or_else(|| Error::InvalidAsset("BMFF box size overflow".to_string()))?;
+
+    if box_end > end || box_end < content_start {
+        return Err(Error::InvalidAsset(
```
tmathern, line 3105: This is relatively strict, probably rightfully so.
But let's make sure we test our own test assets could go through this flow (thinking especially of some mov files).

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,489 @@ pub enum BmffError {
...
+    /// the `mp4` crate's `stsc_index`.
+    pub(crate) fn stsc_index(&self, sample_id: u32) -> Result<usize> {
+        if self.stsc.is_empty() {
+            return Err(Error::InvalidAsset("BMFF has no stsc entries".to_string()));
+        }
+
+        for (i, run) in self.stsc.iter().enumerate() {
+            if sample_id < run.first_sample {
+                return if i == 0 {
+                    Err(Error::InvalidAsset("BMFF no sample not found".to_string()))
```
tmathern: That error message reads weird to me, I don't understand the error (without looking at the code).

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,489 @@ pub enum BmffError {
...
+                } else {
+                    Ok(i - 1)
```
tmathern: Protect against udnerflow?

tmathern (changes requested):
Let's at least mitigate the divide by zero risk.

scouten-adobe:
Thanks Tania — rebased onto main (which now includes #2369) and worked through all the points. Summary:

**Blocking**
- **Divide-by-zero in `bounded_entry_count`** — guarded `entry_size == 0` (all callers pass a fixed non-zero size, but the helper no longer assumes it).

**Bounds / DoS**
- **`stsz` sample_count vs limits** — the fixed-size `stsz` branch didn't cap `sample_count`, and `verify_stream_hash` loops `1..=sample_count`, so a crafted value could drive an enormous loop. `from_stream` now rejects a track whose `sample_count` exceeds the stream length. Added `oversized_stsz_sample_count_is_rejected`.

**Error vs skip / precision** (your `read_child`/`1463` comments)
- The sample read is now a `match` distinguishing `Ok(Some)` / `Ok(None)` (absent → `HashMismatch("Merkle location not found")`) / `Err` (malformed → `InvalidAsset("BMFF sample read failed: ..")`) — landed in #2369 and carried into the native reader here.
- Documented `parse_trak`'s deliberate contract: a *missing* sample-table box means "not a readable sample track" and is skipped (`Ok(None)`) so real hint/metadata tracks don't fail the parse; a *present-but-malformed* box is a hard `Err`.

**tracks vs timed-media** (`1398`)
- The `track_count > 0` gate now has an `else` that errors — a moov with Merkle boxes but no readable track no longer silently passes (fixed + characterized in #2369).

**Perf** (`1396`)
- Wrapped the reader in a `BufReader` for parsing and the per-sample reads, restoring the batched-read behavior of the old path. (Note the previous `mp4` code also seeked per sample; the buffering mainly helps the many small header/table reads.)

**Clarity**
- Reworded `"BMFF no sample not found"` → `"BMFF sample precedes first stsc entry"`.
- Rewrote `stsc_index` to track the running match instead of `i - 1` / `len - 1`, so it's self-evidently underflow-free.

**Over-strictness** (`3079`)
- Added `native_reader_parses_real_assets`: parses `video1.mp4`, `c.mov`, and `BigBuckBunny_320x180.mp4` through the native reader and reads a sample, so the strict bounds are proven not to reject our real fixtures.

Full lib suite green with `--features file_io` (894); fmt + clippy clean. Ready for another look.

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -2834,6 +2834,518 @@ pub enum BmffError {
...
+                % run.samples_per_chunk;
+
+        let mut intra_chunk_offset: u64 = 0;
+        for i in first_sample_in_chunk..sample_id {
+            intra_chunk_offset = intra_chunk_offset
+                .checked_add(self.sample_size(i)? as u64)
+                .ok_or_else(|| {
+                    Error::InvalidAsset("BMFF intra-chunk offset overflow".to_string())
+                })?;
+        }
```
dcondrey: This sum sits inside `verify_stream_hash`'s `for sample_id in 1..=sample_cnt`, so every sample re-walks the ones before it in its chunk. With a single `stsc` run covering the whole track that's quadratic in sample count — the new `sample_count <= stream_len` cap bounds the outer loop but not this one.

Fixed-size samples don't need the walk at all:

```suggestion
        let mut intra_chunk_offset: u64 =
            (sample_id - first_sample_in_chunk) as u64 * self.fixed_sample_size as u64;
        if self.fixed_sample_size == 0 {
            for i in first_sample_in_chunk..sample_id {
                intra_chunk_offset = intra_chunk_offset
                    .checked_add(self.sample_size(i)? as u64)
                    .ok_or_else(|| {
                        Error::InvalidAsset("BMFF intra-chunk offset overflow".to_string())
                    })?;
            }
        }
```

`fixed_sample_size` is 0 in the variable-size case, so the product is 0 and the loop runs as before. That path still walks, but it needs real `stsz` entries so the box size bounds it.

Merkle tests still pass locally.

scouten-adobe:
Good catch, thanks @dcondrey — that inner walk is indeed quadratic over a single-`stsc`-run track, and the `sample_count <= stream_len` cap only bounds the outer loop. Applied in `16b0e6a5`: fixed-size samples now compute `(sample_id - first_sample_in_chunk) * fixed_sample_size` directly (via `checked_mul` to match the surrounding checked-arithmetic style), and only the variable-size case walks — where the `stsz` box size already bounds the entry count. `read_sample` is O(1) for the common fixed-size path now. Merkle tests still green (19 integration + full lib with `--features file_io`).

mauricefisher64 (approved):

tmathern (approved):

caiopensrc:

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2357-to-stable origin/stable
cd .worktree/backport-2357-to-stable
git switch --create backport-2357-to-stable
git cherry-pick -x d8b5f1d8e66ee0895037a6bc027fbb35a2efd248
```

caiopensrc:

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2357-to-stable origin/stable
cd .worktree/backport-2357-to-stable
git switch --create backport-2357-to-stable
git cherry-pick -x d8b5f1d8e66ee0895037a6bc027fbb35a2efd248
```

PR #2359, opened 2026-07-24
fix: Harden against integer overflow attacks in exclusion subsets checks in BMFF hash processing

ssanthosh:
### The vulnerability
`BmffHash::verify_self()` in `sdk/src/assertions/bmff_hash.rs` checked whether exclusion subsets overlap with an unchecked add:

```rust
if subsets[i].offset + subsets[i].length > subsets[i + 1].offset { … }
```

`SubsetMap.offset` and `SubsetMap.length` are `u64` values deserialized from **attacker-controlled CBOR** (the `BmffHash` assertion inside a C2PA manifest). When `offset + length` exceeds `u64::MAX`, the add overflows and, in overflow-checked builds, **panics with "attempt to add with overflow" (process abort, exit 101)**. Verifying any BMFF asset (MP4/MOV/HEIC/AVIF) whose manifest carries a crafted `BmffHash` assertion crashes the host application. `verify_self` is a structural assertion check reached during validation, so **no valid signature is required**.

### The fix
Replace the unchecked add with `checked_add`. A subset whose `offset + length` overflows `u64` cannot fit the addressing space, so it is inherently malformed/overlapping — return the **existing** `Error::C2PAValidation(ASSERTION_BMFFHASH_MALFORMED)` instead of panicking. 

### Tests added (`bmff_hash_tests`)
- `verify_self_rejects_subset_offset_length_overflow` — a subset with `offset + length` overflowing `u64` returns `C2PAValidation` instead of panicking. **Fails without the fix** (the test profile has overflow checks on, so the old code aborts).
- `verify_self_rejects_overlapping_subsets` — a genuine (non-overflowing) overlap is still rejected (guards against weakening the check).
- `verify_self_accepts_valid_non_overlapping_subsets` — positive test: valid ordered, non-overlapping subsets still pass (no over-rejection).

## Checklist

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -1198,7 +1198,17 @@ impl BmffHash {
                     }
                     // check that the subsets do not overlap
                     for i in 0..subsets.len() - 1 {
-                        if subsets[i].offset + subsets[i].length > subsets[i + 1].offset {
+                        // `offset` and `length` are attacker-controlled u64s from
+                        // CBOR; a sum that overflows u64 cannot fit the addressing
+                        // space, so treat it as malformed rather than panicking on
+                        // the add (overflow-checked builds abort otherwise).
+                        let subset_end = subsets[i]
+                            .offset
+                            .checked_add(subsets[i].length)
+                            .ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


tmathern (approved):

caiopensrc:
- #2360

Issue #2327, opened 2026-07-20
Report `manifest.inaccessible` error code on inaccessible manifests


ok-nick:
The SDK currently adds `manifest.inaccessible` when failing to fetch a remote manifest: 
https://github.com/contentauth/c2pa-rs/blob/69ab20fb03e7d448ea1933551ee539b1f7b27c45/sdk/src/ingredient.rs#L682-L692
This doesn't work because the spec introduced a requirement that the ingredient must contain both the `active_manifest` and `validation_results` fields or neither. In this case, we have the validation code `manifest.inaccessible`, but we don't have the manifest, see:
https://github.com/contentauth/c2pa-rs/blob/69ab20fb03e7d448ea1933551ee539b1f7b27c45/sdk/src/assertions/ingredient.rs#L437-L443
Thus, when remote manifest fetch fails or parsing the stream fails, the invariant is broken and signing will always fail with the ambiguous error above.

In #2176 we change the behavior to return the error directly. This is more informative but now moves the error from `Builder::sign` to `Builder::add_ingredient_from_stream`. When the spec clarifies the usage of `manifest.inaccessible` this issue should be readdressed.

lrosenthol:
@ok-nick Feel free to file an issue in the C2PA repo

However, I **CANNOT** find this anywhere in the spec.  Please cite where you are seeing this requirement:

>This doesn't work because the spec introduced a requirement that the ingredient must contain both the active_manifest and validation_results fields or neither. In this case, we have the validation code manifest.inaccessible, but we don't have the manifest, see:

ok-nick:
> In a v3 ingredient assertion with no `activeManifest` field, the `validationResults` field shall not be present.

https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_v3_ingredient_assertions

PR #2367, opened 2026-07-27
feat: add the `builder.ignore_ingredient_errors` setting

ok-nick:
Adds a `builder.ignore_ingredient_errors` setting to ignore parsing/remote manifest fetch/etc. errors and adding the ingredient anyways.

This was split from #2176 and is a result of #2327.

### DIFF
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -1485,7 +1485,12 @@ impl Ingredient {
             };
 
         // set validation status from result and log
-        ingredient.update_validation_status(result, Some(manifest_bytes), &validation_log, &context)?;
+        ingredient.update_validation_status(
+            result,
+            Some(manifest_bytes),
+            &validation_log,
+            &context,
+        )?;
 
         // create a thumbnail if we don't already have a manifest with a thumb we can use
         #[cfg(feature = "add_thumbnails")]
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -660,6 +660,7 @@ impl Ingredient {
         result: Result<Store>,
         manifest_bytes: Option<Vec<u8>>,
         validation_log: &StatusTracker,
+        context: &Context,
     ) -> Result<()> {
         match result {
             Ok(store) => {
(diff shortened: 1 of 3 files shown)


tmathern (approved):
I am ambivalent to it, but it is better than failing and asking users to detect (and figure out) which ingredient failed to hand craft something.

Should be OK to have until we have a fix in the spec to better report that... Wdyt @gpeacock ?

sdk/src/ingredient.rs
```diff
@@ -660,6 +660,7 @@ impl Ingredient {
         result: Result<Store>,
         manifest_bytes: Option<Vec<u8>>,
         validation_log: &StatusTracker,
+        context: &Context,
```
gpeacock, line 663: We should probably be storing the context in the ingredient if we need it for methods like this. Normally for a single test I'd just pass in a boolean flag or settings if we need multiple settings. I'm ok to leave it like this since it is a private method not called very much, but we should consider adding context to Ingredient for this later.

sdk/src/settings/builder.rs
```diff
@@ -591,6 +591,16 @@ pub struct BuilderSettings {
...
+    /// Whether to ignore errors encountered while loading or validating an ingredient's
+    /// manifest (e.g. a remote manifest that couldn't be fetched, or an invalid file format).
+    ///
+    /// The default value is false.
+    ///
+    /// <div class="warning">
+    /// When enabled, an ingredient whose manifest can't be loaded or validated is returned
+    /// with no validation status or validation results set, instead of failing.
+    /// </div>
+    pub ignore_ingredient_errors: bool,
```
gpeacock, line 603: this should be something like return_ingredient_errors as a new option since we have never supported it. I think it should default to false. Then we fix the sdk to not complain if there's a validation_status with no manifest.

the real option is something like ingredient_errors_in_validation_status.

gpeacock (changes requested):
If we ignore ingredient_errors we need should (eventually at least) report them in validation status. Otherwise we should be returning the actual errors here directly, not just return Ok(()

PR #1, opened 2022-05-23
Generate full cert chain

scouten-adobe:
## Changes in This Pull Request
[from @mauricefisher64] Rather than a single self-signed certificate, generate a cert chain from root -> intermediate -> signer to simulate expected certificates. This change will allow us to disable self signed certs in a future PR.

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #2369, opened 2026-07-27
fix: Reject timed-media BMFF Merkle maps that verify against no track

scouten-adobe:
Fixes two latent bugs in the timed-media, track-based Merkle branch of `BmffHash::verify_stream_hash`, with characterization tests locking in the corrected behavior. Lands **before** #2357 (the native sample-reader swap) so those lines are well-covered and the fixes are captured against the current `mp4`-backed code.

## Bugs fixed

1. **Silent pass when there are no tracks.** The per-track verification was gated on `track_count > 0` with no `else`. A `moov` carrying C2PA Merkle boxes but exposing no readable track therefore passed verification *without checking anything*. It now fails with a clear error. (This is the "confounds tracks and timed media" concern from @tmathern's review of #2357.)

2. **Read error masqueraded as a location miss.** The sample read used `if let Ok(Some(sample)) = read_sample(..) { .. } else { Err("Merle location not found") }`, collapsing a genuine read/parse failure and a legitimately-absent sample into one error. They're now distinguished:
   - absent sample → `HashMismatch("Merkle location not found")`
   - read/parse failure → `InvalidAsset("BMFF sample read failed: ..")`

   Also fixes the `"Merle"` → `"Merkle"` typo.

## Tests

Extends the `#2356` characterization suite (`sdk/tests/bmff_timed_media_merkle.rs`):
- `no_tracks_with_merkle_map_is_rejected` — locks in fix #1.
- `missing_chunk_offset_reports_location_miss` / `unreadable_sample_reports_read_error` — lock in fix #2's two distinct outcomes.
- `valid_sha512_verifies` — covers the SHA-512 hasher branch.

No behavior change for valid assets. Full `c2pa` lib suite green with `--features file_io` (888 passed); fmt + clippy clean.

## Follow-up

#2357 (native sample reader) will rebase on this and carry the same call-site error handling forward into the native reader.

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -1476,18 +1476,31 @@ impl BmffHash {
                                 e.insert(hasher_enum);
                             }
 
-                            if let Ok(Some(sample)) = &mp4.read_sample(track_id, sample_id) {
-                                let h = chunk_hash_map.get_mut(&chunk_id).ok_or(
-                                    Error::HashMismatch(
-                                        "Bad Merkle tree sample mapping".to_string(),
-                                    ),
-                                )?;
-                                // add sample data to hash
-                                h.update(&sample.bytes);
-                            } else {
-                                return Err(Error::HashMismatch(
-                                    "Merle location not found".to_owned(),
-                                ));
+                            match mp4.read_sample(track_id, sample_id) {
+                                Ok(Some(sample)) => {
+                                    let h = chunk_hash_map.get_mut(&chunk_id).ok_or(
+                                        Error::HashMismatch(
+                                            "Bad Merkle tree sample mapping".to_string(),
+                                        ),
+                                    )?;
+                                    // add sample data to hash
+                                    h.update(&sample.bytes);
+                                }
+                                // The sample's tables place it outside the asset:
+                                // a genuine Merkle location miss.
+                                Ok(None) => {
+                                    return Err(Error::HashMismatch(
+                                        "Merkle location not found".to_owned(),
+                                    ));
+                                }
+                                // A read/parse failure is distinct from a missing
+                                // sample and is reported as a malformed asset rather
+                                // than masqueraded as a location miss.
+                                Err(e) => {
+                                    return Err(Error::InvalidAsset(format!(
+                                        "BMFF sample read failed: {e}"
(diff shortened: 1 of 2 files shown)


sdk/src/assertions/bmff_hash.rs
```diff
@@ -1476,18 +1476,31 @@ impl BmffHash {
...
+                                Ok(None) => {
+                                    return Err(Error::HashMismatch(
+                                        "Merkle location not found".to_owned(),
+                                    ));
+                                }
+                                // A read/parse failure is distinct from a missing
+                                // sample and is reported as a malformed asset rather
+                                // than masqueraded as a location miss.
+                                Err(e) => {
+                                    return Err(Error::InvalidAsset(format!(
```
tmathern, line 1500: All others above relating to Merkle trees are HashMismatch. Should this reported then as `InvalidAsset` or `HashMismatch` too?

What defines the line between HashMismatch and InvalidAsset when the asset is... invalid (that the invalidity is caused by something else than the Merkle tree?)?

tmathern (approved):

sdk/src/assertions/bmff_hash.rs
mauricefisher64: I think when we can detect it is a media parsing error then it is an invalid asset.

tmathern: Why does an error in the Merkle tree not always mean that the asset is invalid? Because some are recoverable?

mauricefisher64: What do you mean by error in the Merkle tree?  If the file cannot be read that is a file error.  If the values that are pulled from the C2PA data  in the fragment do not validate against the Merkle tree in the manifest that is a hash mismatch.  If there is actually and error in the Merkle tree that is a manifest error.

tmathern: Ok, that was where I was confused, the Merkle tree is entirely ours!

mauricefisher64: Yes, stored in the manifest at generation time.

mauricefisher64 (approved):

ok-nick (approved):

caiopensrc:
- #2400

Issue #2317, opened 2026-07-17
timeStamp.trusted is emitted outside the verify_trust guard, asserting a trusted timestamp for claim-v1 manifests with no trust check performed


contrario:
Component: c2pa-rs (SDK implementation, not the C2PA spec)
Version / commit reviewed: main @ f8f26657162a4177746946234f8b071fae298569 (2026-07-15)
Type: implementation logic (positive assertion emitted without the check that backs it)
Scope (do not drop): claim-version-1 active manifests only

Summary

For a claim-v1 manifest the timestamp trust check is skipped by design, but
c2pa-rs emits timeStamp.trusted into the success bucket anyway. The success
code is emitted outside the if verify_trust { ... } guard, so on the v1 path the
SDK affirms the timestamp certificate is trusted having performed no trust check.

The skip is defensible -- that is not the bug

sdk/src/store.rs:1952 passes rc.version() != 1 as the verify_trust argument,
preceded (:1951) by the verbatim comment, including the source typo, "no trust
checks for leagacy timestamps." Declining to trust-check a legacy v1 timestamp for
backward compatibility is a reasonable choice. The skip is not the defect.

Precise defect (source-level, control-flow trace)

In sdk/src/crypto/time_stamp/verify.rs, inside the per-timestamp loop:

:513 -- digest-match check; on match, TIMESTAMP_VALIDATED success; on mismatch,
continue.
:535 -- if verify_trust { opens the trust block.
:561 and :581 -- the only trust-failure exits, both continue (each skips the
success emission below).
:583 -- the trust block closes.
:585-591 -- TIMESTAMP_TRUSTED is emitted
(.validation_status(TIMESTAMP_TRUSTED).success(...) at :590-591), at
function-body depth, i.e. outside the block that closed at :583.

Because the emission sits outside the guard and the guard's only trust-failure exits
are continue, the success emission is reached unconditionally on the digest-matched
path whenever verify_trust == false: the entire 535-583 block is skipped and
control passes straight to :585. This is a source-level consequence of the branch
structure, not a runtime measurement. For a claim-v1 manifest verify_trust is
false (sdk/src/store.rs:1952), so timeStamp.trusted is written to the success
bucket with no trust evaluation. This is a positive assertion of trust with no
evaluation behind it -- distinct from a missing informational signal; the success
code is present and is false.

Confirmation (source-derived; runtime check suggested, not performed)

The control flow above establishes the defect from source; no asset is required to
[...]

PR #2371, opened 2026-07-29
fix(sdk): only emit timeStamp.trusted when a trust check ran

0xdorian-sm:
Towards #2317. **Opening this as a concrete option, not to pre-empt the spec
question** @scouten-adobe raised there on 17 July — if @lrosenthol's answer
prescribes a different response for a V1 claim with no trust list, I am happy to
rework or close this.

`verify_time_stamp()` emitted the `timeStamp.trusted` success status at
function-body depth, outside the `if verify_trust { ... }` block. The block's
only trust-failure exits are `continue`, so when `verify_trust == false` the
whole block is skipped and control falls straight through to the success
emission. The SDK therefore asserted a trusted timestamp certificate having
performed no trust check.

`Store` derives that argument as `rc.version() != 1`
(`sdk/src/store.rs:1952`, "no trust checks for leagacy timestamps"), so the path
is reachable for any claim-v1 active manifest. For v2 claims `verify_trust` is
true and the check runs, so behaviour there is unchanged.

This moves the emission inside the guard, which is the conservative reading:
a v1 manifest gets `timeStamp.validated` and **no trust claim in either
direction**, rather than a positive claim with nothing behind it. It asserts
nothing new, so it should be compatible with whatever the spec answer turns out
to be — if the correct response is a distinct "trust not checked" status, that
is a new status code and a larger change, and I did not want to invent one.

The skip itself is left alone: declining to trust-check a legacy timestamp is a
deliberate choice and is not what this touches.

The two other call sites passing `verify_trust = false` literally
(`sdk/src/assertions/timestamp.rs:117` and
`sdk/src/crypto/time_stamp/http_request.rs:67`) run at signing time with
`CertificateTrustPolicy::passthrough()` and a `StatusTracker` that is discarded,
so they were not user-visible and are unaffected.

## Testing

`cargo test -p c2pa --lib`: 904 passed, 6 failed — and the **same 6 fail on
`main` without this change**, so there is no regression from it. Those 6 are
environmental: I built with `--no-default-features --features
rust_native_crypto,file_io,add_thumbnails,http_ureq`, which leaves out the async
HTTP resolver, so the three `did_web::resolve` tests fail with
`AsyncHttpResolverNotImplemented` and the three `identity::validator` tests that
depend on them fail in turn.

Because those CAWG tests cannot run in that configuration, they do **not**
answer the question I asked on the issue: whether
`identity/claim_aggregation/ica_signature_verifier.rs:735`, which consumes
`TIMESTAMP_TRUSTED` as a boolean, is reachable with `verify_trust == false`. If
it is, this change also removes a trust decision resting on an unchecked
timestamp — but I could not verify that and am not claiming it.

Reported by @contrario, whose control-flow trace was accurate; I confirmed it
still held on `main` and checked the surrounding call sites before changing
anything.

## Checklist

### DIFF
--- a/sdk/src/crypto/time_stamp/verify.rs
+++ b/sdk/src/crypto/time_stamp/verify.rs
@@ -540,15 +540,18 @@ pub fn verify_time_stamp(
                 last_err = TimeStampError::Untrusted;
                 continue;
             }
-        }
 
-        log_item!(
-            "",
-            format!("timestamp cert trusted: {}", &common_name),
-            "verify_time_stamp"
-        )
-        .validation_status(TIMESTAMP_TRUSTED)
-        .success(&mut current_validation_log);
+            // Only assert trust when a trust check actually ran. Emitting this
+            // outside the guard claims a trusted timestamp on the
+            // verify_trust == false path, where nothing was evaluated.
+            log_item!(
+                "",
+                format!("timestamp cert trusted: {}", &common_name),
+                "verify_time_stamp"
+            )
+            .validation_status(TIMESTAMP_TRUSTED)
+            .success(&mut current_validation_log);
+        }
 
         // If we find a valid value, we're done.
         validation_log.append(&current_validation_log);


contrario:
Answering the ica_signature_verifier.rs:735 question you flagged (and the
17 July thread) — I traced it on 5be5c1e. Short version: verify_trust is false
on that path, but :735 does not turn that into a trust decision, so this PR is a
status-bug fix and not more than that.

Both ICA call sites pass verify_trust = false literally:
ica_signature_verifier.rs:116 (sync) and :209 (async), each with
CertificateTrustPolicy::passthrough() and a fresh, single-use StatusTracker
(:109 / :199). validate_cose_tst_info forwards the flag unchanged
(sigtst.rs:107 -> verify_time_stamp at :166), so the out-of-guard emission does
fire here, and it is not v1-scoped — it is every CAWG identity timestamp.

But the :735 consumer sits inside handle_time_stamp_error (:714), which is
called only from the Err(_err) arms (:122, :216). In verify_time_stamp,
TIMESTAMP_TRUSTED is written (verify.rs:550) immediately before return Ok(tst)
(:555); the Err return (:559) is reached only when every loop iteration took a
continue, leaving the tracker with TIMESTAMP_MISMATCH / TIMESTAMP_UNTRUSTED,
never TIMESTAMP_TRUSTED. The tracker is fresh per check_signature, so
has_status(TIMESTAMP_TRUSTED) is false whenever handle_time_stamp_error runs.
The *ok = true at :737 is therefore not reachable via TIMESTAMP_TRUSTED; this PR
does not remove a trust decision, and I would not describe it as doing so.

The residual on the Ok path — a CAWG credential accepted without a TSA trust
check — is the deliberate verify_trust = false / passthrough behaviour already
tracked as CAI-8847, orthogonal to this emission fix.

This is a source read, not a runtime result: the CAWG identity tests don't run
in that crypto feature set for me either, so the same caveat applies as to the
original report. The fix as written (emission inside the guard) matches the
conservative option and leaves the deliberate skip alone.

mauricefisher64:
Our decision is to support V1 timestamps as to not break existing manifests.

PR #2379, opened 2026-07-30
feat: new builder method to combine filtering of actions and ingredients

cdmurph32:
filter_functions should not remove edited actions that point to kept ingredients." We need `c2pa.edited` actions to survive when they point at an ingredient we want to keep, even if the caller's action-level predicate wouldn't otherwise keep that action.

## Checklist


sdk/src/builder.rs
```diff
@@ -1169,6 +1155,78 @@ impl Builder {
         Ok(self)
     }
 
+    /// Retains actions and ingredients together in one atomic step.
```
```suggestion
    /// Retains actions and ingredients together in one call.
```
(It makes multiple calls to other things, so itself it's not atomic)

tmathern: (Also update the other APIs' docs maybe to point to this combined one)

sdk/src/builder.rs
```diff
@@ -1118,23 +1118,9 @@ impl Builder {
...
-                    for uri in uris {
-                        if let Some(label) = assertion_label_from_uri(&uri.url()) {
-                            let (_, idx) = parse_positional_label(&label);
-                            if let Some(id) = pre_filter_ids.get(idx) {
-                                keep_set.insert(id.clone());
-                            }
-                        }
-                    }
-                }
+                // Relationship-agnostic: any referenced ingredient is kept whatever its
```
tmathern: Do we need this comment if it roughly says "NOOP"?

sdk/src/builder.rs
```diff
@@ -1169,6 +1155,78 @@ impl Builder {
         Ok(self)
     }
 
+    /// Retains actions and ingredients together in one atomic step.
+    ///
+    /// Calling [`Builder::filter_actions`] then [`Builder::filter_ingredients`] separately can
```
tmathern: I would summarize by explaining the order of calls is important and why

sdk/src/builder.rs
```diff
@@ -1169,6 +1155,78 @@ impl Builder {
...
+    pub fn filter_actions_and_ingredients<FA, FI>(
+        &mut self,
+        mut keep_action: FA,
+        mut rescue_ingredient: FI,
+    ) -> Result<&mut Self>
+    where
+        FA: FnMut(&Action) -> bool,
+        FI: FnMut(&Ingredient) -> bool,
+    {
+        // Positional order matters for resolving `parameters.ingredients` positional refs; this
```
```suggestion
        // Snapshot to keep positional order as ordering reference
```

sdk/src/builder.rs
```diff
@@ -1169,6 +1155,78 @@ impl Builder {
...
+        // must be snapshotted before any pruning happens.
```
```suggestion
```

sdk/src/builder.rs
```diff
@@ -1169,6 +1155,78 @@ impl Builder {
...
+            .iter()
+            .map(Ingredient::effective_id_internal)
+            .collect();
+
+        // Evaluate `rescue_ingredient` up front, exactly once per ingredient, so an action
+        // referencing a to-be-rescued ingredient can be protected before any action is dropped.
+        let rescued_ids: HashSet<String> = self
+            .definition
+            .ingredients
+            .iter()
```
tmathern: definitions.ingredients is accessed twice with `iter()`. Maybe this can be refactored a bit to spare one of the iter() calls? (I don't thing this ingredient list gets mutated in between the two calls).

sdk/src/builder.rs
```diff
@@ -1169,6 +1155,78 @@ impl Builder {
...
+        FA: FnMut(&Action) -> bool,
+        FI: FnMut(&Ingredient) -> bool,
+    {
+        // Positional order matters for resolving `parameters.ingredients` positional refs; this
+        // must be snapshotted before any pruning happens.
+        let pre_filter_ids: Vec<String> = self
+            .definition
+            .ingredients
+            .iter()
+            .map(Ingredient::effective_id_internal)
```
tmathern: I did not realize before because I think this can only be hit through using only the Rust SDK when the instance_id is somehow not set (or rather I did not see it been hit otherwise), but do we need to worry about side-effects if an effective id is "None" When the list is empty?

Because of https://github.com/contentauth/c2pa-rs/blob/f5d112f81a70d39aa123346b332149f65b8bfc1c/sdk/src/ingredient.rs#L287 and https://github.com/contentauth/c2pa-rs/blob/f5d112f81a70d39aa123346b332149f65b8bfc1c/sdk/src/ingredient.rs#L252

cdmurph32: I'm going to switch to index based.

tmathern (approved):
https://github.com/contentauth/c2pa-rs/blob/main/docs/experimental-features.md#registry-of-experimental-features will need an update for the new function

caiopensrc:

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2379-to-stable origin/stable
cd .worktree/backport-2379-to-stable
git switch --create backport-2379-to-stable
git cherry-pick -x 12f9ac6b6661871c03e16756310af09511d56e26
```

PR #2380, opened 2026-07-30
feat(sdk): Read support for c2md (JUMBF data) manifests

### LINKED ISSUE #549: Read support for c2md (JUMBF data) manifests
The 2.x spec say we must pass through c2md manifests.  Writing is not required.

ssanthosh:
Adds read support for standard C2PA Manifests that use the `c2md` JUMBF type
UUID (`63326D64-0011-0010-8000-00AA00389B71`). Per C2PA spec §11.2.2, Manifest
Consumers shall accept standard manifests specified with this type (claim
generators shall not create them). Previously the manifest-store reader only
accepted `c2ma`/`c2um` superboxes and silently skipped a `c2md` manifest, so it
was never read or validated.

Writing is intentionally not implemented — the spec forbids generators from
creating `c2md` manifests.

Fixes #549.

## Tests added

- `test_from_jumbf_reads_c2md_standard_manifest` — builds a normal `c2ma`
  manifest, rewrites its superbox type UUID to `c2md`, and asserts the store
  still reads the manifest (claim count preserved). Confirmed to fail without
  the fix (0 claims read) and pass with it.

## Checklist


sdk/src/store.rs
```diff
@@ -1264,9 +1264,12 @@ impl Store {
             let cai_store_box = store_box.super_box();
             let cai_store_desc_box = cai_store_box.desc_box();
 
-            // ignore unknown boxes per the spec
+            // ignore unknown boxes per the spec. A c2md manifest is a standard
+            // manifest that consumers shall accept (C2PA spec 11.2.2), so it is
+            // read the same as a c2ma manifest.
             if cai_store_desc_box.uuid() != CAI_UPDATE_MANIFEST_UUID
                 && cai_store_desc_box.uuid() != CAI_MANIFEST_UUID
+                && cai_store_desc_box.uuid() != CAI_MANIFEST_C2MD_UUID
```
mauricefisher64, line 1274: I don't think this is necessary.  Do you actually hit this?  If you look at jumbf/boxes.rs in read_super_box_impl we just skip unknown boxes.  Is that not happening?

tmathern: +1, how do you land here?

ssanthosh: @mauricefisher64 / @tmathern : `read_super_box_impl` does parse the c2md box — its "skip unknown boxes" match arm keys on the low-level JUMBF BoxType (https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/jumbf/boxes.rs#L2471), and a c2md manifest is an ordinary jumb superbox, so it's parsed via the BoxType::Jumb arm, not skipped. c2md is a content-type UUID in the jumd description box and never reaches that match.

The manifest is dropped one layer up, at the manifest-type gate in from_jumbf_impl (if uuid != c2um && != c2ma { continue }) — so read_super_box reads it, but the store discards it before it becomes a Claim.
Hence, the above changes are required otherwise the c2md manifest is silently dropped and its provenance lost.

Regression test added proves the above issue that after flipping a manifest's superbox UUID c2ma → c2md, read_super_box shows the c2md superbox is parsed, but Store::from_jumbf returns 0 claims without this change and 1 with it.

mauricefisher64: Huh, I would expect that from_jumbf would return 0 since we should have ignored it.

mauricefisher64: Remove this change and I think you are OK.  The code is testing that only ordinary and update manifests pass through.  All others are skipped.  We want c2md manifests to be skipped also not added to the Store.

ssanthosh: Added changes to preserve c2md manifest on round trip as well as when adding an update manifest with https://github.com/contentauth/c2pa-rs/pull/2380/commits/803b5f298238c76af057d7cd2f0520f4b6644d8b

mauricefisher64 (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31401081228](https://github.com/contentauth/c2pa-rs/actions/runs/31401081228).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2461 |

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31401109512](https://github.com/contentauth/c2pa-rs/actions/runs/31401109512).

| Target | Status |
|--------|--------|
| `stable` | :heavy_minus_sign: Skipped (PR already exists) |

PR #2388, opened 2026-07-31
fix: C2PA 2.3 reporting

### LINKED ISSUE #2316: signingCredential.ocsp.skipped is defined, classified, and unit-tested, but emitted from no validation path
Component: c2pa-rs (SDK implementation, not the C2PA spec)
Version / commit reviewed: main @ f8f26657162a4177746946234f8b071fae298569 (2026-07-15)
Type: implementation / API-completeness (observability gap, not a security defect)

Summary

c2pa-rs defines five OCSP status codes but emits only four. The fifth,
SIGNING_CREDENTIAL_OCSP_SKIPPED (signingCredential.ocsp.skipped), is defined,
carries a doc comment, is classified Informational by log_kind(), and appears
in a serialization unit test, yet it is emitted from no validation path. On the
default DoNotFetch path (ocsp_fetch = false) the OCSP routine returns an empty
default response and logs no status, so a caller reading validation_results()
cannot distinguish "OCSP not checked" from "OCSP checked, certificate not revoked."

The default is correct -- this is not a request to change it

Not making outbound network calls on manifest-controlled input is good
engineering. SECURITY.md:27 lists network calls driven by C2PA manifest input as
a known problematic characteristic, tracked under #1765. ocsp_fetch = false as the
default is right. This issue is not about the default. It is about the missing
informational signal on that path: the SDK already anticipated the signal (the
code exists) but never emits it.

Relationship to #1489 (merged) -- distinct arm, not a re-tread

#1489 ("Don't log OCSP status if the result cannot be validated") addresses the
inaccessible / error arm: when an OCSP response is fetched but cannot be
validated, do not log a status from it. That is a different state from the one this
issue is about. The state here is skipped -- the DoNotFetch arm, where no OCSP
request is ever made. #1489 does not reach this arm, and the suggested fix below is
consistent with #1489's principle rather than against it: the informational signal
proposed here is one the SDK can stand behind (it knows it deliberately did not
check), unlike the unvalidated response #1489 correctly stopped logging.

The five states, as shipped

Emitted from a validation path:

revoked (SIGNING_CREDENTIAL_REVOKED): sdk/src/crypto/cose/ocsp.rs:133, :266
notRevoked (SIGNING_CREDENTIAL_NOT_REVOKED): sdk/src/crypto/cose/ocsp.rs:151, :284
inaccessible (SIGNING_CREDENTIAL_OCSP_INACCESSIBLE): sdk/src/crypto/cose/ocsp.rs:435
unknown (SIGNING_CREDENTIAL_OCSP_UNKNOWN): sdk/src/crypto/ocsp/mod.rs:344

Emitted from no validation path:

skipped (SIGNING_CREDENTIAL_OCSP_SKIPPED)

skipped is the sole defined OCSP status that is never emitted from a validation
path.

Precise defect (source-level, no assets needed)

sdk/src/validation_results.rs:647 -- SIGNING_CREDENTIAL_OCSP_SKIPPED is
defined with a doc comment to the effect that the validator chose not to perform
an online OCSP check.
sdk/src/validation_results.rs:1115 -- log_kind() classifies it
Informational (grouped with INACCESSIBLE and the timestamp-informational
codes).
sdk/src/assertions/ingredient.rs:1114 -- its only other occurrence is a
serialization unit test (#[test] fn test_serialization), which uses the code
as add_informational_val fixture data via its string literal. This is test
fixture construction, not a validation-path emission.
sdk/src/crypto/cose/ocsp.rs:32-33 -- the import list brings in REVOKED,
NOT_REVOKED, and INACCESSIBLE, but not SKIPPED; the code is not in scope in
the module that would emit it.
sdk/src/crypto/cose/ocsp.rs:184 -- the OcspFetchPolicy::DoNotFetch arm
returns Ok(OcspResponse::default()) and logs no status.
sdk/src/crypto/ocsp/mod.rs:49-59 -- the Default impl sets revoked_at = None
(:54), which is structurally identical to a successful check that found no
revocation.

Net effect: the "we did not check" case is represented by the absence of a code,
in a struct shaped identically to the "we checked, it is clean" case. The one code
that would disambiguate the two exists but is never reached.

Runtime reproduction (optional)

Read any C2PA-signed asset with default settings (ocsp_fetch = false) and dump
validation_results(). Observe that no OCSP code appears in any bucket
(informational included), and the returned OcspResponse has revoked_at = None.
Then confirm via source grep that signingCredential.ocsp.skipped is referenced
only at its definition, its log_kind classification, and one serialization test.

Suggested fix

Emit the SIGNING_CREDENTIAL_OCSP_SKIPPED informational code on the DoNotFetch
arm (sdk/src/crypto/cose/ocsp.rs:184, which already returns
Ok(OcspResponse::default()); add the informational log entry there). Roughly one
line at the emission site, no behavior change to the network default.

Scope / impact

Observability only. A caller that reads validation_results() to decide whether
revocation was evaluated currently gets no signal on the default path. The missing
skipped signal is what a caller would need to know the online check did not run.

mauricefisher64:
Support for multiple independent named trust lists.
Validation of soft-binding assertions
Support for specVersion and trust_list_uri
Support for soft binding algorithm list
Unified trust list model (manifest, CAWG, TSA) are configured in a common manor
More detailed reporting for trust sources
New settings support to automatically handle multiple trust anchors
Per trust list EKU configuration
Hardening of trust checks

## Checklist


sdk/src/crypto/cose/certificate_trust/openssl.rs
```diff
@@ -24,70 +24,53 @@ pub(crate) fn check_certificate_trust(
...
 
         let store = builder.build();
 
+        // try trust anchors
         let mut store_ctx = X509StoreContext::new()?;
         if store_ctx.init(&store, cert.as_ref(), &cert_chain, |f| f.verify_cert())? {
-            Ok(TrustAnchorType::User)
-        } else {
-            Err(CertificateTrustError::CertificateNotTrusted)
+            return Ok((TrustAnchorType::Signer, anchor_set.trust_anchor_uri.clone()));
```
tmathern: Here we try anchors, in rust_native we check/use them. But it is different there:
```
 return Ok((
    anchor_set.trust_anchor_type,
    anchor_set.trust_anchor_uri.clone(),
));
```
Why is it different? (The first param, `anchor_set.trust_anchor_type`)

sdk/src/store.rs
```diff
@@ -163,13 +165,34 @@ impl Store {
...
-        if let Some(ta) = &settings.trust.trust_anchors {
-            let _v = store.add_trust(ta.as_bytes());
-        }
+        // Add all of the trust anchors
+        if let Some(anchors) = &settings.trust.anchors {
+            for anchor in anchors {
+                let trust_list_uri = match &anchor.trust_uri {
+                    Some(uri) => uri.to_string(),
+                    None => format!(
+                        "https:://cai.org/unknown_trust_list_{}",
```
tmathern:
```suggestion
                        "https://cai.org/unknown_trust_list_{}",
```
One colon too much!

sdk/src/spec_versions.rs
```diff
@@ -0,0 +1,24 @@
...
+// each license.
+
+/// C2PA Generator Version
+pub const C2PA_GENERATOR_VERSION: &str = "2.2.0";
+/// C2PA Validator Version
+pub const C2PA_VALIDATOR_VERSION: &str = "2.3.0";
+
+// C2PA TSA Trust URI
+#[allow(dead_code)]
+pub const C2PA_TSA_TRUST_URI: &str = "https://c2pa.org//trustlist-tsa";
```
tmathern:
```suggestion
pub const C2PA_TSA_TRUST_URI: &str = "https://c2pa.org/trustlist-tsa";
```
(Typo I think)

sdk/src/spec_versions.rs
```diff
@@ -0,0 +1,24 @@
...
+/// C2PA Trust URI
+#[allow(dead_code)]
+pub const C2PA_TRUST_URI: &str = "https://c2pa.org//trustlist";
```
tmathern:
```suggestion
pub const C2PA_TRUST_URI: &str = "https://c2pa.org/trustlist";
```
(Typo I think)

sdk/src/validation_results.rs
```diff
@@ -117,11 +118,36 @@ pub struct ValidationResults {
...
+    /// The version of the specification against which the validation was performed (SemVer formatted string).
+    #[serde(rename = "specVersion", skip_serializing_if = "Option::is_none")]
+    spec_version: Option<String>,
+
+    /// URI to the trust list that was used to validate manifests signing certificate.
+    #[serde(rename = "trustListUri", skip_serializing_if = "Option::is_none")]
+    trust_list_uri: Option<String>,
+
+    /// URI to the trust list use to validate the time-stamp.
+    #[serde(rename = "timestampTrustListUri", skip_serializing)]
```
tmathern, line 130: Can you have a rename and `skip_serializing`? Isn't skip-serializing always skipping it?

sdk/src/settings/mod.rs
```diff
@@ -56,6 +56,47 @@ pub(crate) trait SettingsValidate {
...
+    CAWG,
+}
+
+#[cfg_attr(
+    feature = "json_schema",
+    derive(schemars::JsonSchema),
+    schemars(default)
+)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct TrustAnchor {
```
tmathern, line 79: Are all those required fields? What happens if one is missing from the (JSON or TOML) config but not the others? How does it behave, does it define defaults?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -384,6 +384,9 @@ pub struct BmffHash {
     #[serde(skip_serializing)]
     url: Option<UriT>, // deprecated in V2 and not to be used
 
+    #[serde(rename = "sequenceNumber", skip_serializing_if = "Option::is_none")]
```
tmathern: Did this sneak in from another PR/fix/branch?

sdk/src/spec_versions.rs
```diff
@@ -0,0 +1,24 @@
...
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+/// C2PA Generator Version
+pub const C2PA_GENERATOR_VERSION: &str = "2.2.0";
+/// C2PA Validator Version
+pub const C2PA_VALIDATOR_VERSION: &str = "2.3.0";
```
tmathern, line 18: Just to be sure because it is a bit surprising reading through: We generate v2.2.0, but validate against v2.3.0?

cli/src/main.rs
```diff
@@ -247,6 +249,10 @@ enum Commands {
         /// URL or path to file containing configured EKUs in Oid dot notation
         #[arg(long = "trust_config", env="C2PATOOL_TRUST_CONFIG", value_parser = parse_resource_string)]
         trust_config: Option<TrustResource>,
+
+        /// URI to used to identify the trust list
```
tmathern:
```suggestion
        /// UR used to identify the trust list
```

sdk/src/crypto/cose/certificate_trust_policy.rs
```diff
@@ -23,13 +23,17 @@ use x509_parser::{extensions::ExtendedKeyUsage, pem::Pem};
...
+    #[default]
+    Signer,
+
+    /// Time Stamping Authority (TSA) provided trust anchor.
+    TSA,
 
-    /// User provided trust anchor.
-    User,
+    /// CAWG provided trust anchor.
+    CAWG,
```
tmathern, line 39: In the CLI tool the, since it's set to anchor type "Signer", will there be a need for the sibling param to configure CAWG trust?

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -320,12 +320,14 @@ impl IdentityAssertion {
...
-                }
-
-                if let Some(pa) = &settings.cawg_trust.user_anchors {
-                    let _ = ctp.add_user_trust_anchors(pa.as_bytes());
+                if let Some(ta) = &settings.cawg_trust.anchors {
+                    for anchor in ta {
+                        let _ = ctp.add_trust_anchors(
+                            anchor.trust_anchors.as_bytes(),
+                            anchor.trust_uri.as_deref().unwrap_or(""),
+                            crate::crypto::cose::TrustAnchorType::CAWG,
```
tmathern: Why not import it on top instead of using a fully qualified name?

sdk/src/settings/mod.rs
```diff
@@ -56,6 +56,47 @@ pub(crate) trait SettingsValidate {
...
+// Kind of Trust list represented
+#[cfg_attr(feature = "json_schema", derive(schemars::JsonSchema))]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum TrustKind {
+    // Certificate signing trust list
+    Signer,
+    // Time Stamping Authority trust list
+    TSA,
+    // Certificate Authority Working Group trust list
```
tmathern, line 69: Wait, is this another "CAWG" than Creator Assertions Working Group (https://cawg.io/)? (I assumed for CAWG comments we were talking about the cawg,io CAWG).

sdk/tests/fixtures/test_settings.json
```diff
@@ -1,7 +1,13 @@
...
+    "anchors": [
```
tmathern, line 4: The renaming will be a breaking change... ?

sdk/src/validation_results.rs
mauricefisher64: Leaving this way to follow the valdation_time case.  It may be serialized one day.

sdk/src/crypto/cose/certificate_trust/openssl.rs
mauricefisher64: You are correct

sdk/src/crypto/cose/certificate_trust_policy.rs
mauricefisher64: CAWG does not currently have the requirement to report which trust list contained the certificate.  Also there is no way in the CLI to configure just CAWG trust at the moment.  If CAWG needs this feature a custom list can be added to the CLI.  The SDK supports it and does not care.  The whole point of the changes is that you can multiple independent trust lists.  The validation code will just report which list a certificate was found in.  How you configure the lists is up to the SDK user.

sdk/src/settings/mod.rs
mauricefisher64: Don't know what you mean by that.  I will rename this variable to make it more clear what it is for.

sdk/src/spec_versions.rs
mauricefisher64: Pasting from C2PA page tried to be too smart and added the "/" to escape for me.  Good catch.

mauricefisher64:
@tmathern Thanks for the pre-code review review :)

sdk/src/settings/mod.rs
mauricefisher64: If there is no trust uri we generate one.  If there is no kind it is a signer trust list.

tmathern (approved):

sdk/src/claim.rs
```diff
@@ -459,6 +463,7 @@ impl Claim {
             created_assertions: Vec::new(),
             gathered_assertions: None,
             context: None,
+            spec_version: Some(C2PA_GENERATOR_VERSION.to_string()),
```
gpeacock: Maybe we should be defaulting this to None until we have passed conformance, and even then, the app needs to set this I think, can the API claim confomance?

sdk/src/identity/tests/validation_method/continue_when_possible.rs
```diff
@@ -676,7 +676,7 @@ async fn pad1_invalid() {
 
     assert_eq!(
         log.description,
-        "signing certificate trusted, found in User trust anchors"
+        "signing certificate trusted, found in [https://cai.org/unknown_tl] trust anchors"
```
gpeacock: nice

sdk/src/settings/mod.rs
```diff
@@ -73,16 +115,14 @@ pub struct Trust {
     /// Verifying trust is REQUIRED by the CAWG spec. This option should only be used for development or testing.
     /// </div>
     pub(crate) verify_trust_list: bool,
-    /// List of additional user-provided trust anchor root certificates as a PEM bundle.
-    pub user_anchors: Option<String>,
```
gpeacock, line 71: So this would be a breaking change to all existing apps that have trust list settings, right?
I think we need to honor the old settings as best we can and deprectate them, instead of outright replacing them

gpeacock (changes requested):
I don't think we can change remove support for existing settings without deprecation time. If the plan is to wait for the next breaking release, we could deprecate some of this on a non-breaking release. Otherwise I think we should continue to support the old format while adding the new and letting it work both ways for a while.

Also I think the default specVersion should be None. I'm not positive that we should ever  set it in the SDK unless the SDK can be conformant outside of an app. Anyway, the safe thing, lacking clear rules, would be let the SDK keep it as None, and let apps set it.

cli/src/main.rs
```diff
@@ -247,6 +251,10 @@ enum Commands {
         /// URL or path to file containing configured EKUs in Oid dot notation
         #[arg(long = "trust_config", env="C2PATOOL_TRUST_CONFIG", value_parser = parse_resource_string)]
         trust_config: Option<TrustResource>,
+
+        /// UR used to identify the trust list
+        #[arg(long = "trust_list_uri", env = "C2PATOOL_TRUST_LIST_URI")]
+        trust_list_uri: Option<String>,
```
ok-nick, line 257: This can be confusing since c2patool only expects one trust anchor list and they are likely to be merged with multiple lists. I think we should leave this arg out of here for now.

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -521,38 +521,52 @@ pub fn verify_time_stamp(
...
+                Some(signing_time),
+            ) {
+                Err(_) => {
+                    log_item!(
+                        "",
+                        format!("timestamp cert untrusted: {}", &common_name),
+                        "verify_time_stamp"
+                    )
+                    .validation_status(TIMESTAMP_UNTRUSTED)
+                    .informational(&mut current_validation_log);
```
ok-nick, line 535: Can we add the err from this case to the log item for context?

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -521,38 +521,52 @@ pub fn verify_time_stamp(
...
-        }
+        } else {
+            // this is the 1.x backward compatibility case, since there were no trust lists for timestamps
+            log_item!(
+                "",
+                format!("legacy timestamp cert trusted: {}", &common_name),
+                "verify_time_stamp"
+            )
+            .validation_status(TIMESTAMP_TRUSTED)
+            .success(&mut current_validation_log);
```
ok-nick, line 565: Should we report trusted? I prefer if we don't say anything about the trust state of the timestmap in this case.

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -320,12 +322,14 @@ impl IdentityAssertion {
...
-                    let _ = ctp.add_trust_anchors(ta.as_bytes());
-                }
-
-                if let Some(pa) = &settings.cawg_trust.user_anchors {
-                    let _ = ctp.add_user_trust_anchors(pa.as_bytes());
+                if let Some(ta) = &settings.cawg_trust.anchors {
+                    for anchor in ta {
+                        let _ = ctp.add_trust_anchors(
+                            anchor.trust_anchors.as_bytes(),
+                            anchor.trust_uri.as_deref().unwrap_or(""),
```
ok-nick, line 328: > /// URI identifier for the trust list.  If not is present a unique identifier will be generated.

Should we be generating a unique URI here?

sdk/src/settings/mod.rs
```diff
@@ -184,21 +223,23 @@ impl Default for Trust {
...
-            trust.user_anchors = Some(
-                String::from_utf8_lossy(include_bytes!(
+
+            trust.anchors = Some(vec![TrustAnchor {
+                trust_anchors: String::from_utf8_lossy(include_bytes!(
                     "../../tests/fixtures/certs/trust/test_cert_root_bundle.pem"
                 ))
                 .into_owned(),
-            );
+                trust_uri: Some("https://cai.org/unknown_tl".to_string()),
```
ok-nick: We don't own this site, can we use `https://contentauthenticity.org/` or an arbitrary test URI?

sdk/src/store.rs
```diff
@@ -163,13 +165,34 @@ impl Store {
...
-        if let Some(ta) = &settings.trust.trust_anchors {
-            let _v = store.add_trust(ta.as_bytes());
-        }
+        // Add all of the trust anchors
+        if let Some(anchors) = &settings.trust.anchors {
+            for anchor in anchors {
+                let trust_list_uri = match &anchor.trust_uri {
+                    Some(uri) => uri.to_string(),
+                    None => format!(
+                        "https://cai.org/unknown_trust_list_{}",
```
ok-nick: We shouldn't default to an arbitrary site here.

sdk/src/store.rs
```diff
@@ -163,13 +165,34 @@ impl Store {
...
+                        extfmt::Hexlify(&hash_by_alg(
+                            "sha256",
+                            anchor.trust_anchors.as_bytes(),
+                            None
+                        ))
+                    ),
```
ok-nick, line 186: IMO I think we should require a URI to be specified by the user when adding a trust list, they can choose to hash the list if they'd like.

sdk/src/spec_versions.rs
```diff
@@ -0,0 +1,24 @@
...
+
+// C2PA TSA Trust URI
+#[allow(dead_code)]
+pub const C2PA_TSA_TRUST_URI: &str = "https://c2pa.org/trustlist-tsa";
+/// C2PA Trust URI
+#[allow(dead_code)]
+pub const C2PA_TRUST_URI: &str = "https://c2pa.org/trustlist";
```
ok-nick, line 25: These URLs don't actually point to anything, I don't see them defined in the spec either?

sdk/src/settings/mod.rs
ok-nick: Yes agreed, a lot of code depends on this right now.

tmathern (commented):
I think it _could_ go to main, into a backport it may need some extra care (or accept the backport to be a breaking changes one).
(See also https://github.com/contentauth/c2pa-rs/pull/2388#discussion_r3731804475)

sdk/src/crypto/cose/certificate_profile.rs
```diff
@@ -54,6 +54,29 @@ pub fn check_end_entity_certificate_profile(
...
+    if ctp.has_mandatory_ekus() {
+        if let Ok(Some(BasicExtension { value: eku, .. })) = signcert.extended_key_usage() {
+            if !ctp.verify_mandatory_ekus(eku) {
+                log_item!(
+                    "",
+                    "certificate missing mandatory EKU",
+                    "check_certificate_profile"
+                )
+                .validation_status(SIGNING_CREDENTIAL_INVALID)
+                .failure_no_throw(validation_log, CertificateProfileError::InvalidEku);
```
tmathern, line 67: Why is this a non-fatal failure? Because it's a pre-flight check maybe?

sdk/src/crypto/cose/certificate_profile.rs
```diff
@@ -54,6 +54,29 @@ pub fn check_end_entity_certificate_profile(
...
+            }
+        } else {
+            log_item!(
+                "",
+                "certificate missing mandatory EKU",
+                "check_certificate_profile"
+            )
+            .validation_status(SIGNING_CREDENTIAL_INVALID)
+            .failure_no_throw(validation_log, CertificateProfileError::InvalidEku);
```
tmathern, line 76: Why is this a non-fatal failure? Because it's a pre-flight check maybe?

sdk/src/crypto/cose/certificate_trust_policy.rs
```diff
@@ -352,44 +393,90 @@ impl CertificateTrustPolicy {
...
+    pub fn clear_ekus(&mut self) {
+        self.additional_ekus.clear();
+        self.mandatory_ekus.clear();
     }
 
     /// Remove all trust anchors, private credentials, and EKUs previously
     /// configured.
     pub fn clear(&mut self) {
-        self.trust_anchor_ders.clear();
+        self.trust_anchors.clear();
```
tmathern, line 430: `clear_ekus` clears both additional_ekus and mandatory_ekus, Why does this one, `clear` only clear additional_ekus and not everything? Are the mandatory ones unclearable through this?

sdk/src/store.rs
```diff
@@ -166,23 +168,44 @@ impl Store {
...
+                // Anchors are pre-validated when loading the settings
+                let _v = store.add_trust_anchor(
+                    anchor.trust_anchors.as_bytes(),
+                    &trust_list_uri,
+                    trust_list_type,
+                );
 
-        if let Some(pa) = &settings.trust.user_anchors {
-            let _v = store.add_user_trust_anchors(pa.as_bytes());
+                if let Some(al) = &anchor.allowed_list {
```
tmathern, line 203: No `anchor.trust_config` here to read or other trust settings?

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -319,23 +320,26 @@ impl IdentityAssertion {
...
+                            anchor.trust_kind.clone().into(),
+                        );
 
-                if let Some(pa) = &settings.cawg_trust.user_anchors {
-                    let _ = ctp.add_user_trust_anchors(pa.as_bytes());
-                }
+                        if let Some(tc) = &anchor.trust_config {
+                            // override default EKUs
+                            ctp.clear_ekus();
+                            ctp.add_valid_ekus(tc.as_bytes());
```
tmathern: Why override default EKUs? 
Also, this is in a loop, so not additive: last one wins since previous ones would be wiped.

sdk/src/store.rs
```diff
@@ -166,23 +168,44 @@ impl Store {
...
+                };
+
+                let trust_list_type = match anchor.trust_kind {
+                    TrustListKind::Manifest => TrustAnchorType::Manifest,
+                    TrustListKind::TSA => TrustAnchorType::TSA,
+                    TrustListKind::CAWG => TrustAnchorType::CAWG,
+                };
+
+                // Anchors are pre-validated when loading the settings
+                let _v = store.add_trust_anchor(
```
tmathern, line 196: Ignoring the return value, can we miss errors/validation cases here?

sdk/src/store.rs
```diff
@@ -166,23 +168,44 @@ impl Store {
...
+                    anchor.trust_anchors.as_bytes(),
+                    &trust_list_uri,
+                    trust_list_type,
+                );
 
-        if let Some(pa) = &settings.trust.user_anchors {
-            let _v = store.add_user_trust_anchors(pa.as_bytes());
+                if let Some(al) = &anchor.allowed_list {
+                    let _v = store.add_trust_allowed_list(al.as_bytes());
```
tmathern, line 204: Ignoring the return value, can we miss errors/validation cases here?

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -520,38 +520,52 @@ pub fn verify_time_stamp(
...
+                        format!(
+                            "timestamp cert trusted: {}, trust list: {}",
+                            &common_name, &trust_uri
+                        ),
+                        "verify_time_stamp"
+                    )
+                    .validation_status(TIMESTAMP_TRUSTED)
+                    .set_trust_list_uri(&trust_uri)
+                    .success(&mut current_validation_log);
+                    break;
```
tmathern: `last-err` is set in some branches, and is set to a default value:
https://github.com/contentauth/c2pa-rs/blob/20041db425519e8b882f4fdf8fe5943bea6cbfdc/sdk/src/crypto/time_stamp/verify.rs#L115

If we break in success, to be safe, shouldn't that error state be reset too? So it's not stale? Especially since that variable seem to be reused in quite some fallthroughs.

sdk/src/settings/mod.rs
mauricefisher64: @gpeacock @ok-nick I added back compat for these two fields.   Please review.

mauricefisher64: See updated PR

sdk/src/crypto/time_stamp/verify.rs
mauricefisher64: We don't really care which err was the last error since what we really use is  current_validation_log.  current_validation_log will contain the last_error.  If we break on success there will be no error values.  Notice that current_validation_log is cleared each loop.

sdk/src/crypto/cose/certificate_profile.rs
mauricefisher64: It is.  Fatal errors are one where we can no longer continue processing and have to stop.  Here we keep going and find all the errors in the manifest..

sdk/src/crypto/cose/certificate_trust_policy.rs
mauricefisher64: That is the case.   See how they are used.  Mandatory ekus are only applied in special circumstances.

sdk/src/identity/identity_assertion/assertion.rs
mauricefisher64: That is the whole point of per anchor set EKU lists.  There is a top level eku list and per anchor set eku set.   It is not that the last one.  What the loop is doing is replacing the default EKU set when there is a custom EKU set listed for that particular anchor set.   This is so that CAWG can have the general EKU behavior but when validating a CAWG anchor use is CAWG defined EKUs.

sdk/src/crypto/cose/certificate_profile.rs
mauricefisher64: Same as above.  We track all errors in a single pass.

sdk/src/store.rs
mauricefisher64: Good catch let me fix.

mauricefisher64: Now we propagate config too.

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -374,8 +374,7 @@ fn check_stapled_ocsp_response(
 
         // make sure this is an OCSP signing EKU
         let mut new_ctp = ctp.clone();
-        new_ctp.clear_ekus();
-        new_ctp.add_valid_ekus(OCSP_OID_STR.as_bytes()); // ocsp signing EKU
+        new_ctp.add_mandatory_ekus(OCSP_OID_STR.as_bytes()); // ocsp signing EKU
```
dcondrey, line 385: `clear_ekus()` is dropped here but kept in `fetch_and_check_ocsp_response` below (line 510) — was that intentional?

Without it, `additional_ekus` still holds the configured C2PA set, which includes `id-kp-documentSigning` (`valid_eku_oids.cfg`). So for a responder certificate that has no `ocspSigning` EKU, `has_allowed_eku` finds documentSigning and returns `Some`, and `check_certificate_profile` doesn't return `Err`. The new mandatory-EKU check logs via `failure_no_throw` with no `return Err`, so the `.is_err()` gate below stays false and the stapled response is accepted. On `main` the same certificate is rejected, because `clear_ekus()` left only the OCSP OID in the allow list.

`time_stamp/verify.rs:501` lost `clear_ekus()` in the same commit, which has the same effect for the TSA profile check.

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -520,38 +519,52 @@ pub fn verify_time_stamp(
...
+                        format!(
+                            "timestamp cert trusted: {}, trust list: {}",
+                            &common_name, &trust_uri
+                        ),
+                        "verify_time_stamp"
+                    )
+                    .validation_status(TIMESTAMP_TRUSTED)
+                    .set_trust_list_uri(&trust_uri)
+                    .success(&mut current_validation_log);
+                    return Ok(tst);
```
dcondrey, line 554: This return skips `validation_log.append(&current_validation_log)`. The legacy branch just below and the fall-through at the end of the function both append before returning, and `current_validation_log` is function-local, so on the trusted path everything logged into it is dropped — `timeStamp.validated`, `timeStamp.trusted`, and anything `check_end_entity_certificate_profile` wrote at line 505 (now non-fatal, so `.is_err()` doesn't stop us getting here).

Side effect worth noting: `ValidationResults::timestamp_trust_list_uri` is only set when `from_store` sees a `timeStamp.trusted` status (`validation_results.rs:174-177`), so with a trusted timestamp that field never populates.

Looks like it just needs the same `append` as the branch below. `git log` shows this was a `break` before 6ac4587f, where the append still ran.

CLAUDE.md
```diff
@@ -0,0 +1,129 @@
+# CLAUDE.md
+
```
gpeacock: Did you intend to include this?

cli/src/assets/c2patool.iconset/icon_128x128.png
gpeacock, line 1: Were did these icons come from and why are we adding them to c2patool? I don't object but seems out of context with everything else here. Maybe a separate pr?

cli/src/zero_wm.rs
```diff
@@ -0,0 +1,385 @@
+use std::{
```
gpeacock: looks cool but what is this and how would we use it?

cli/src/mac_sign.rs
```diff
@@ -0,0 +1,492 @@
+use std::ptr::{self, NonNull};
+
+use objc2_core_foundation::{
```
gpeacock: is all this hooked into to c2patool somewhere? it looks like some work on secure enclave got mixed in. I like it, but maybe we need to move this into a crate so it can be used by other apps as well?

sdk/src/asset_handlers/bmff_io.rs.avif
gpeacock, line 1: if these are test assets shouldn't they go in fixtures?

sdk/src/utils/range_iterator.rs
```diff
@@ -0,0 +1,41 @@
...
+
+
+/// Splits `range` into consecutive sub-ranges of at most `size` elements.
+///
+/// Every sub-range has exactly `size` elements except possibly the last,
+/// which holds the remainder. Yields nothing for an empty range.
+///
+/// # Panics
+/// Panics if `size == 0`.
+fn sub_ranges(range: Range<usize>, size: usize) -> impl Iterator<Item = Range<usize>> {
```
gpeacock: I like this, where do we use it?

CLAUDE.md
mauricefisher64: No, I did not do that.

cli/src/assets/c2patool.iconset/icon_128x128.png
mauricefisher64: I did not add those either

cli/src/zero_wm.rs
mauricefisher64: Weird, looks like random files got added.

sdk/src/utils/range_iterator.rs
mauricefisher64: Wow it looks like my experiments stuff got added.

mauricefisher64:
Let me fix this PR.  Some stuff got added that should not be there.

mauricefisher64:
Replaced by clean PR 2545

PR #2372, opened 2026-07-29
chore: release

caiopensrc:
## 🤖 New release

* `c2pa`: 0.90.3 -> 0.90.4 (⚠ API breaking changes - see changelog for why this wasn't a version bump)
* `c2pa-c-ffi`: 0.90.3 -> 0.90.4
* `c2patool`: 0.27.3 -> 0.27.4

### ⚠ `c2pa` breaking changes

```text
--- failure feature_missing: package feature removed or renamed ---

Description:
A feature has been removed from this package's Cargo.toml. This will break downstream crates which enable that feature.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#cargo-feature-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.47.0/src/lints/feature_missing.ron

Failed in:
  feature experimental_builder_filter in the package's Cargo.toml

--- failure inherent_method_missing: pub method removed or renamed ---

Description:
A publicly-visible method or associated fn is no longer available under its prior name. It may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.47.0/src/lints/inherent_method_missing.ron

Failed in:
  Builder::filter_actions, previously in file /tmp/.tmpEFRQl8/c2pa/src/builder.rs:990
  Builder::filter_ingredients, previously in file /tmp/.tmpEFRQl8/c2pa/src/builder.rs:1088
  Ingredient::effective_id, previously in file /tmp/.tmpEFRQl8/c2pa/src/ingredient.rs:276
```

<details><summary><i><b>Changelog</b></i></summary><p>

## `c2pa`

<blockquote>

## [0.90.4](https://github.com/contentauth/c2pa-rs/compare/c2pa-v0.90.3...c2pa-v0.90.4)

_04 August 2026_

### Added

* New builder method to combine filtering of actions and ingredients ([#2379](https://github.com/contentauth/c2pa-rs/pull/2379)) ([#2401](https://github.com/contentauth/c2pa-rs/pull/2401))

### Documented

* Doc how to generate bench fixtures (backport #2366) ([#2370](https://github.com/contentauth/c2pa-rs/pull/2370))

### Fixed

* Make URI checks for (data)boxes when going through redactions mo… ([#2404](https://github.com/contentauth/c2pa-rs/pull/2404))
[...]

PR #2390, opened 2026-08-01
fix: Make URI checks for (data)boxes when going through redactions more strict

tmathern:
Match the databox by its exact final label segment, not just a contain segment.
This ensures the search looks for more exact things, and prevents redaction URIs errors not found due to confusing URIs.

Test-wise:
- end-to-end (sign then sign again to redact previous things) in builder.rs;
- tests for redaction in claim.rs, also to make sure the comparison rules are stable (since depending in other parts of the SDK behavior).

## Checklist


tmathern:
Note: same emscripten build failure on https://github.com/contentauth/c2pa-rs/pull/2372

sdk/src/claim.rs
```diff
@@ -1811,24 +1811,39 @@ impl Claim {
...
-            }) {
+            // Compare the assertion label strictly using label and instance.
+            let target = Claim::label_with_instance(&label, instance);
+            if let Some(index) = self
+                .assertion_store
+                .iter()
+                .position(|x| Claim::label_with_instance(&x.label_raw(), x.instance()) == target)
+            {
                 self.assertion_store.remove(index);
                 return Ok(());
```
dcondrey, line 1840: `assertion_label_from_link` returns `label_parts[0]` for non-thumbnail labels, truncating at the first `__`.

A stored label of `com.example.a__b` parses to `("com.example.a", 0)`, so `target` is `com.example.a` and the redaction returns `AssertionRedactionNotFound`. `ends_with` matched it before.

```suggestion
        // Delete assertion or databox
        if assertion_uri.contains(ASSERTION_STORE) {
            // Compare the assertion label strictly, against the URI's final segment.
            let target = to_normalized_uri(assertion_uri)
                .rsplit('/')
                .next()
                .unwrap_or_default()
                .to_string();
            if let Some(index) = self
                .assertion_store
                .iter()
                .position(|x| Claim::label_with_instance(&x.label_raw(), x.instance()) == target)
            {
                self.assertion_store.remove(index);
                return Ok(());
```

tmathern: Good catch on the truncation. 

There's a second component here though: The cases you describe are only reachable if a malformed assertion label gets here, and currently it can't due to the way the SDK works e2e/normalization.

`redact_assertion` only runs on claims from `Store::from_jumbf_with_context` (via `add_ingredient_data`), and both `get_assertion_from_jumbf_store` (store.rs) and `Assertion::from_assertion_data` (assertion.rs) normalize the box label on the way in. A box named `com.example.a__b` is stored as `label_raw() == "com.example.a"`. 

That normalization is what the comparison here relies on: we only redact from previously signed manifests, so the stored side is always normalized, while the redaction URI is (potentially) arbitrary caller input (even though it should have been read from existing manifests to be given in as aprameters).

Normalizing both sides is what keeps them comparable.

Such a crafted label can't reach here, since it fails at signing with `AssertionMissing { url: "com.example.a__b" }` (`add_assertion_impl` stores raw, `assertion_hashed_uri_from_label` looks up normalized). If the SDK renormalizing labels at signing is a concern, please consider filing another bug. Redaction follows the current SDK behavior and relies on it.

I've added tests to catch if things would change and bread the e2e redaction flows:

- `test_redact_assertion_instance_label_end_to_end` covers your case directly through sign -> load -> redact, both directions, against a store holding `CreativeWork` and `CreativeWork__1`.
- `test_signing_label_normalization` fails if signing starts accepting these labels, which also indicates redaction will need an update. 
- `test_redact_assertion_suffix_label_is_not_a_match` verifies we don't match on things we should not match on (that the check is not too lax).

sdk/src/claim.rs
```diff
@@ -1811,24 +1811,43 @@ impl Claim {
...
         // cannot redact action assertions per the spec
         // cannot redact hash bindings
-        let (label, _instance) = Claim::assertion_label_from_link(assertion_uri);
+        let (label, instance) = Claim::assertion_label_from_link(assertion_uri);
         if label.starts_with(assertions::labels::ACTIONS) || label.starts_with("c2pa.hash.") {
             return Err(Error::AssertionInvalidRedaction);
         }
 
-        // delete assertion or databox
+        // A redaction URI may only target this claim.
```
ok-nick: The function docs state:
>/// Redact an assertion from a prior claim.

However, the new comment contradicts it with:
>// A redaction URI may only target this claim.

sdk/src/claim.rs
```diff
@@ -1811,24 +1811,43 @@ impl Claim {
...
+            // with both sides normalized so they are comparable.
+            // `assertion_label_from_link` truncates the URI's label at `__`.
+            // Currently, the stored side is always normalized on ingest, and signing rejects
+            // a non-numeric `__` suffix, so a manifest we read here can not hold such a label.
+            let target = Claim::label_with_instance(&label, instance);
+            if let Some(index) = self
+                .assertion_store
+                .iter()
+                .position(|x| Claim::label_with_instance(&x.label_raw(), x.instance()) == target)
+            {
```
ok-nick, line 1838: 👍

sdk/src/builder.rs
```diff
@@ -8423,6 +8423,276 @@ mod tests {
...
 
+    /// End-to-end redaction against a claim that went through sign -> load,
+    /// so the assertion store is the loader-normalized one that
+    /// `redact_assertion` sees when looking for assertions to redact.
+    ///
+    /// Runs against an ingredient holding `com.example.test` and `com.example.test__1`:
+    ///   1. redacting the `__1` URI must leave the base assertion, and
+    ///   2. redacting the un-suffixed URI must leave the `__1` assertion.
+    #[test]
+    fn test_redact_assertion_instance_label_end_to_end() {
```
ok-nick, line 8434: I'd prefer to keep e2e tests in sdk/tests with the other integration tests rather than in the module itself. Not sure if we are already doing this anyways though.

ok-nick (approved):

sdk/src/builder.rs
tmathern: I wasn't sure either... Builder.rs also already has some e2e tests. (I prefer them close to the source they test e2e, in my mind here it is clearly Builder being tested?!?).

sdk/src/claim.rs
tmathern: Fixed!

cdmurph32 (approved):
LGTM

caiopensrc:

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2390-to-stable origin/stable
cd .worktree/backport-2390-to-stable
git switch --create backport-2390-to-stable
git cherry-pick -x f02f2da40c3cbc3bf4d136c70075a2a6e78134be
```

PR #2399, opened 2026-08-03
feat: Add core soft binding resolution API support (Decoupled spec)

gpeacock:
Implements the registry, resolution query client, and a caller-driven Reader::with_soft_binding() for the C2PA "Decoupled" soft binding spec. This is an update from ok-nick/soft-binding-api branch; local watermark/fingerprint embedding and extraction are deliberately out of scope here

### DIFF
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -461,6 +461,114 @@ impl Reader {
         Ok(self)
     }
 
+    /// Resolve and validate a manifest for `stream` using a soft binding value the caller
+    /// has already obtained (e.g. via a proprietary local watermark/fingerprint decoder, or
+    /// the caller's own call to [`crate::soft_binding::resolution_api::query_by_content`]).
+    ///
+    /// This implements the caller-driven half of the C2PA
+    /// ["Decoupled" soft binding](https://spec.c2pa.org/specifications/specifications/2.2/softbinding/Decoupled.html)
+    /// resolution flow: it looks up `alg` in
+    /// [`Settings.soft_binding.algorithm_registry`](crate::settings::SoftBinding::algorithm_registry),
+    /// queries each of its registered resolution API endpoints via `byBinding`, and — for
+    /// the best match found — constructs the manifest's fetch URL and hands it to the same
+    /// remote-manifest fetch path used for an XMP-embedded provenance reference. If
+    /// fetching is disabled ([`Verify::remote_manifest_fetch`](crate::settings::Verify::remote_manifest_fetch))
+    /// or unsupported (the `fetch_remote_manifests` feature is off), this surfaces as
+    /// [`Error::RemoteManifestUrl`], exactly as it would for an XMP remote manifest
+    /// reference, so the caller can fetch it out-of-band and pass the bytes to
+    /// [`with_manifest_data_and_stream`](Self::with_manifest_data_and_stream) instead.
+    ///
+    /// The SDK never uploads asset content on the caller's behalf as part of this call —
+    /// only the small, already-known `value` is sent in the query.
+    ///
+    /// # Arguments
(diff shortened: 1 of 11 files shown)


sdk/src/soft_binding/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+//! Support for the C2PA soft binding algorithm list registry and the
+//! [Soft Binding Resolution API](https://spec.c2pa.org/specifications/specifications/2.2/softbinding/Decoupled.html)
+//! ("Decoupled" soft binding).
+//!
+//! This module deliberately does not perform any local watermark/fingerprint
+//! embedding or extraction; callers are expected to obtain a soft binding
+//! `alg` and `value` themselves (e.g. via a proprietary decoder) and pass it
+//! to [`crate::Reader::with_soft_binding`].
+
+pub mod algorithm_list;
```
tmathern, line 23: I like the idea of this being in its own crate/mod, so it is reusable and accessible from the other crates.

sdk/src/reader.rs
```diff
@@ -461,6 +461,114 @@ impl Reader {
...
+                (None, None) => std::cmp::Ordering::Equal,
+            }
+        });
+
+        let Some((endpoint, best_match)) = best else {
+            return Ok(self);
+        };
+
+        let manifest_endpoint = best_match.endpoint.as_deref().unwrap_or(endpoint);
+        let manifest_url =
```
tmathern, line 555: This will likely need URL validation (like is it HTTP(S) or any other allowed scheme?).

tmathern: ps: As most often in the SDK, we also will want to have no redirects here I guess.

sdk/src/reader.rs
```diff
@@ -461,6 +461,114 @@ impl Reader {
...
+                .await
+            };
+            // Best-effort: one endpoint failing shouldn't abort the others.
+            if let Ok(query_result) = result {
+                all_matches.extend(query_result.matches.into_iter().map(|m| (endpoint, m)));
+            }
+        }
+
+        let best = all_matches.into_iter().max_by(|(_, a), (_, b)| {
+            match (a.similarity_score, b.similarity_score) {
```
tmathern, line 542: On the similarity: maybe here there should be a cap, because docs say: "where 0 is the weakest possible match and 100 is the strongest possible match.". So it there are 0 scores here, they might still be taken into account for best matches...

sdk/src/soft_binding/resolution_api.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// Endpoint of a Soft Binding Resolution API from which the C2PA Manifest may be
+    /// obtained. If absent, the C2PA Manifest is available from the same endpoint the
+    /// query was sent to, using the `/manifests` endpoint.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub endpoint: Option<String>,
+    /// An integer score in the range (0-100) representing the strength of match, if
+    /// appropriate, where 0 is the weakest possible match and 100 is the strongest
+    /// possible match.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub similarity_score: Option<u8>,
```
tmathern, line 58: Does min and max accepted values need to be explicitly enforced?

sdk/src/reader.rs
```diff
@@ -461,6 +461,114 @@ impl Reader {
...
+    /// to validate.
+    #[async_generic]
+    pub fn with_soft_binding(
+        self,
+        alg: &str,
+        value: &[u8],
+        format: &str,
+        stream: impl Read + Seek + MaybeSend,
+    ) -> Result<Self> {
+        use crate::soft_binding::{resolution_api, SoftBindingList};
```
tmathern, line 503: Import on top?

tmathern (commented):
Looks promising!

PR #2401, opened 2026-08-03
feat: new builder method to combine filtering of actions and ingredients (#2379)

cdmurph32:

* feat: new builder method to combine filtering of actions and ingredients

Filter functions should not remove edited actions that point to kept ingredients. We need `c2pa.edited` actions to survive when they point at an ingredient we want to keep, even if the caller's action-level predicate wouldn't otherwise keep that action.

- Fix comments
- Cross-link filter_actions/filter_ingredients docs to the new combined method, and explain the call-order gap it closes.
- Combine two separate iterations over definition.ingredients into one pass computing pre_filter_ids and rescued_ids together.
- Change filter_actions_and_ingredients to rescue based on index rather than effective_id, as ingredients may have neither instance_id or label.

* docs: Update experimental features for new function

(cherry picked from commit 12f9ac6b6661871c03e16756310af09511d56e26)

### DIFF
--- a/docs/experimental-features.md
+++ b/docs/experimental-features.md
@@ -1,0 +1,104 @@
+# Experimental features policy
+
+The Content Authenticity Initiative (CAI) SDK is an evolving project, and we welcome contributions from the wider community. From time to time, an external contributor may propose a new feature that we believe is valuable but that the CAI team is not able to commit to maintaining as part of the core SDK.
+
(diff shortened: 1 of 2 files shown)


tmathern (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 30868508136](https://github.com/contentauth/c2pa-rs/actions/runs/30868508136).

Issue #2074, opened 2026-04-20
Bug: Certificate Parsing in Remote Signing Flow


trufo-ai:
### Component

c2pa-rs

### Version

v0.80.0

### Platform

Linux

### What happened?

Last week at the IPTC meeting, Brendan ran into the error `Error: the certificate is invalid` when running `c2patool`. After some investigation, we found that the remote signing component was incorrectly parsing the certificate (more details in the PR commit message). At the time, we provided Brendan with a fix to `c2pa-rs` so that it would work.

### What did you expect to happen?

The remote signer component should parse the PEM cert chain to DER, just like the local signer component does.

trufo-ai:
relevant PR: https://github.com/contentauth/c2pa-rs/pull/2075

trufo-ai:
I also heard from Brendan that Badr (EBU) may already have a fork in which this bug has been resolved.

scouten:
:white_check_mark: Jira issue https://jira.corp.adobe.com/browse/CAI-12731 is successfully created for this GitHub issue.

mpgopala:
Created https://github.com/contentauth/c2pa-rs/pull/2414. The PR needed few corrections and tests.

PR #2414, opened 2026-08-05
fix(sdk): Parse PEM cert chain to DER for RemoteSigner

mpgopala:
## Summary

- `SignerSettings::c2pa_signer()`'s `Remote` variant was storing the raw PEM
  text bytes as a single "cert" and sizing `reserve_size` off PEM text length,
  instead of parsing the chain into individual DER-encoded certificates like
  the `Local` variant does via `create_signer::from_keys`.
- Now uses `crypto::cert_chain_pem_to_der` to parse `sign_cert` into DER certs
  (also handles JSON-escaped `\n` PEM), and computes `reserve_size` from the
  summed DER cert lengths.

## Issue

https://github.com/contentauth/c2pa-rs/issues/2074

### DIFF
--- a/sdk/src/settings/signer.rs
+++ b/sdk/src/settings/signer.rs
@@ -124,13 +124,17 @@ impl SignerSettings {
                 tsa_url,
                 referenced_assertions: _,
                 roles: _,
-            } => Ok(Box::new(RemoteSigner {
-                url,
-                alg,
-                reserve_size: 10000 + sign_cert.len(),
-                certs: vec![sign_cert.into_bytes()],
-                tsa_url,
-            })),
+            } => {
+                let certs = cert_chain_pem_to_der(sign_cert.as_bytes())?;
+                let reserve_size = 10000 + certs.iter().map(|c| c.len()).sum::<usize>();
+                Ok(Box::new(RemoteSigner {
+                    url,
+                    alg,
+                    reserve_size,
+                    certs,
+                    tsa_url,
+                }))
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


sdk/src/settings/signer.rs
```diff
@@ -524,5 +528,113 @@ pub mod tests {
...
+            )
+            .unwrap();
+
+        let signer_settings = settings.signer.expect("signer settings should be present");
+        assert!(signer_settings.c2pa_signer().is_err());
+    }
+
+    #[test]
+    fn test_make_remote_signer_no_pem_blocks_yields_empty_certs() {
+        // No BEGIN/END markers at all: the underlying parser treats this as
```
tmathern, line 618: This is a correct behavior? One we intent to keep @mauricefisher64 @ok-nick ?

mpgopala: @scouten-adobe ^^

scouten-adobe: This isn't new risk introduced by this PR — `cert_chain_pem_to_der` (sdk/src/crypto/mod.rs:30) already has this exact behavior for the `Local` variant via `create_signer::from_keys`, so this PR is just making `Remote` consistent with existing, already-shipped semantics. And downstream, sdk/src/cose_sign.rs:129 explicitly checks `certs.first()` and returns `Error::CoseNoCerts` if the list is empty — so a misconfigured `sign_cert` doesn't silently produce an unsigned/invalid manifest, it fails at sign time with a clear error rather than at settings-construction time. That's a reasonable answer to give in the thread: the behavior is safe, just deferred to a later point than some reviewers might expect.

(Perhaps worth a follow-up issue to move the verification earlier?)

scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31613224336](https://github.com/contentauth/c2pa-rs/actions/runs/31613224336).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2481 |

PR #2421, opened 2026-08-05
fix: Preserve identity assertions in split-signing paths

### LINKED ISSUE #2055: Bug: Identity assertions silently dropped in `sign_embeddable` and `sign_data_hashed_embeddable`

scouten-adobe:
## Summary

Fixes #2055 — identity (`cawg.identity`) assertions were silently dropped by the split-signing APIs.

`IdentityAssertionSigner::dynamic_assertions()` (and `AsyncIdentityAssertionSigner`) drained their internal list via `split_off(0)`, so they only returned the configured identity assertions on the **first** call. The split-signing paths query `dynamic_assertions()` more than once per manifest — once to reserve placeholder slots and again to write the assertion content — so the second call saw an empty list and the identity assertion was silently dropped. Only `Builder::sign` → `save_to_stream` (which queries exactly once) worked.

## Changes

- **Root-cause fix:** store the registered `IdentityAssertionBuilder`s behind `Arc` and hand out shared clones from `dynamic_assertions()` instead of draining, so every query within a signing operation sees the same set. This makes `Builder::placeholder` + `Builder::sign_embeddable` (both modes) and the fragmented `save_to_folder` path work correctly. The registered assertions are now retained after signing (the doc comment on `add_identity_assertion` is updated accordingly).

- **No more silent drops on the data-hashed path:** `Builder::data_hashed_placeholder` cannot reserve space for dynamic assertions (it has no signer), so `Store::get_data_hashed_embeddable_manifest` and `Store::sign_manifest` now return an error when a signer exposes dynamic assertions without reserved placeholder slots — rather than emitting a manifest silently missing the identity assertion. The error directs callers to the `placeholder()` / `sign_embeddable()` workflow instead.

## Tests

New regression tests in `sdk/src/identity/tests/embeddable_signing.rs`:

- `sign_embeddable_includes_identity_assertion` — the `placeholder()` + `sign_embeddable()` path now produces a manifest containing `cawg.identity`. Verified to **fail** against the old draining behavior with the exact symptom from the issue (`cawg.identity assertion missing`).
- `data_hashed_embeddable_rejects_identity_signer` — the data-hashed path now returns an error instead of silently dropping the identity assertion.
- `dynamic_assertions_are_not_drained` — the signer returns its assertions on every call.

Existing identity, builder, and store test suites pass; `wasm32-unknown-unknown` still compiles; `cargo +nightly fmt --all -- --check` is clean.

## Notes / for reviewer discussion

- This is a **draft** — the behavior change on the data-hashed path (silent drop → error) is the conservative choice suggested in the issue. An alternative would be to extend `data_hashed_placeholder` to accept the signer so it can reserve space, but that changes the public API; happy to pursue that instead if preferred.
- Both sync and async signers are fixed symmetrically.

### DIFF
--- a/sdk/src/identity/builder/identity_assertion_signer.rs
+++ b/sdk/src/identity/builder/identity_assertion_signer.rs
@@ -11,7 +11,7 @@
 // specific language governing permissions and limitations under
 // each license.
 
-use std::sync::RwLock;
+use std::sync::{Arc, RwLock};
 
 use c2pa_raw_crypto::{RawSigner, SigningAlg};
 
@@ -28,7 +28,7 @@ use crate::{
 pub struct IdentityAssertionSigner {
     signer: Box<dyn RawSigner + Send + Sync>,
     cert_chain: Vec<Vec<u8>>,
-    identity_assertions: RwLock<Vec<IdentityAssertionBuilder>>,
+    identity_assertions: RwLock<Vec<Arc<IdentityAssertionBuilder>>>,
 }
 
 impl IdentityAssertionSigner {
@@ -67,17 +67,21 @@ impl IdentityAssertionSigner {
     /// Add an [`IdentityAssertionBuilder`] to be used when signing the
     /// next [`Manifest`].
     ///
-    /// IMPORTANT: When [`sign()`] is called, the list of
-    /// [`IdentityAssertionBuilder`]s will be cleared.
+    /// The registered assertions are retained after signing so that the same
+    /// signer can be used across every signing path, including the split
+    /// signing paths ([`Builder::placeholder`] + [`Builder::sign_embeddable`])
+    /// that query [`dynamic_assertions()`] more than once per manifest.
     ///
     /// [`Manifest`]: crate::Manifest
-    /// [`sign()`]: Self::sign
+    /// [`Builder::placeholder`]: crate::Builder::placeholder
+    /// [`Builder::sign_embeddable`]: crate::Builder::sign_embeddable
+    /// [`dynamic_assertions()`]: Signer::dynamic_assertions
     pub fn add_identity_assertion(&mut self, iab: IdentityAssertionBuilder) {
         #[allow(clippy::unwrap_used)]
         let mut identity_assertions = self.identity_assertions.write().unwrap();
         // TO DO: Replace with error handling in the very unlikely case of a panic here.
 
-        identity_assertions.push(iab);
+        identity_assertions.push(Arc::new(iab));
     }
 }
 
@@ -105,16 +109,18 @@ impl Signer for IdentityAssertionSigner {
 
     fn dynamic_assertions(&self) -> Vec<Box<dyn DynamicAssertion>> {
         #[allow(clippy::unwrap_used)]
-        let mut identity_assertions = self.identity_assertions.write().unwrap();
+        let identity_assertions = self.identity_assertions.read().unwrap();
         // TO DO: Replace with error handling in the very unlikely case of a panic here.
 
-        let ia_clone = identity_assertions.split_off(0);
-        let mut dynamic_assertions: Vec<Box<dyn DynamicAssertion>> = vec![];
-
-        for ia in ia_clone.into_iter() {
-            dynamic_assertions.push(Box::new(ia));
-        }
-
-        dynamic_assertions
+        // Hand out shared clones instead of draining the list. Several signing
+        // paths (notably `Builder::placeholder` + `Builder::sign_embeddable`)
+        // call this method more than once per manifest – once to reserve
+        // placeholder slots and again to write the assertion content. Draining
+        // on the first call left later calls empty, which silently dropped the
+        // identity assertion (see issue #2055).
+        identity_assertions
+            .iter()
+            .map(|ia| Box::new(Arc::clone(ia)) as Box<dyn DynamicAssertion>)
+            .collect()
     }
 }
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -2293,9 +2293,10 @@ impl Store {
             ));
         };
 
-        // Write dynamic assertions only if placeholders were added during placeholder generation.
-        // We check if the dynamic assertion labels exist in the claim - if not, placeholders
-        // weren't added and we should skip writing to avoid size mismatches.
+        // Write dynamic assertions when the caller reserved placeholder slots for
+        // them (via `add_dynamic_assertion_placeholders`). If a signer exposes
+        // dynamic assertions but no placeholders were reserved, fail loudly rather
+        // than silently dropping them (see issue #2055).
         let dynamic_assertions = signer.dynamic_assertions();
         if !dynamic_assertions.is_empty() {
             // Check if placeholders exist for these dynamic assertions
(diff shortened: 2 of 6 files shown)


sdk/src/store.rs
```diff
@@ -2293,47 +2293,41 @@ impl Store {
...
         let dynamic_assertions = signer.dynamic_assertions();
         if !dynamic_assertions.is_empty() {
-            // Check if placeholders exist for these dynamic assertions
-            let has_placeholders = {
-                dynamic_assertions
-                    .iter()
-                    .all(|da| pc.assertion_hashed_uri_from_label(&da.label()).is_some())
-            };
-
-            if has_placeholders {
```
gpeacock, line 2308: I think all you need here is to return an error if has_placeholders false. We still need to check if they exist. I guess that will be caught if when you call write_dynamic_assertions since there's nothing to replace. But I'm not sure what the errors message is there or if it will make any sense.

sdk/src/identity/builder/identity_assertion_signer.rs
```diff
@@ -28,7 +28,7 @@ use crate::{
 pub struct IdentityAssertionSigner {
     signer: Box<dyn RawSigner + Send + Sync>,
     cert_chain: Vec<Vec<u8>>,
-    identity_assertions: RwLock<Vec<IdentityAssertionBuilder>>,
+    identity_assertions: RwLock<Vec<Arc<IdentityAssertionBuilder>>>,
```
tmathern, line 31: Given those are pub structs, will this be a breaking change for users/callers, due to the Arc in the chain of types?

sdk/src/store.rs
```diff
@@ -2423,37 +2417,18 @@ impl Store {
...
-                    self.write_dynamic_assertions(&dynamic_assertions, &mut preliminary_claim)?;
-                } else {
-                    self.write_dynamic_assertions_async(
-                        &dynamic_assertions,
-                        &mut preliminary_claim,
-                    )
-                    .await?;
-                }
-            }
+        // The data-hashed placeholder (`get_data_hashed_manifest_placeholder`) does not
```
gpeacock, line 2438: we really need to deprecate the builder.data_hashed_placeholder() method. none of this code should still exist!

gpeacock (approved):
I like the arc solution.
I'm unsure about what error will be returned by write_dynamic_assertions when it fails. We may want to be more specific with the error here.

sdk/src/identity/builder/identity_assertion_signer.rs
scouten-adobe: Not a breaking change. `identity_assertions` is a private field, so the `Arc` never appears in any public signature: `add_identity_assertion` still takes `IdentityAssertionBuilder` by value (the `Arc::new` wrap is internal), and `dynamic_assertions()` still returns `Vec<Box<dyn DynamicAssertion>>`. The new `impl DynamicAssertion for Arc<IdentityAssertionBuilder>` is purely additive and can't conflict downstream (it's our trait, so external crates can't implement it for these types anyway).

`cargo semver-checks` (HEAD vs `main`, default and `--all-features`) reports "no semver update required". And since the whole `identity` module is `#[doc(hidden)]`, it's outside the documented stable surface regardless.

One thing worth flagging as a *behavioral* (not API) change: `sign_data_hashed_embeddable` with an identity signer now returns an error instead of silently emitting a manifest with the identity dropped. That path never produced a correct result, so nothing correct relied on it — but noting it for completeness.

sdk/src/store.rs
scouten-adobe: Good call — traced it, and without the check `write_dynamic_assertions` fails with a bare `Error::NotFound`. Restored an explicit guard in `sign_manifest` that names the offending assertion label(s) and points at the fix:

`no placeholder slots were reserved for dynamic assertions [cawg.identity]; call add_dynamic_assertion_placeholders() before signing`

Covered by a new direct test (`test_sign_manifest_errors_when_dynamic_placeholders_missing`). Pushed in 7ded89d3.

stupside:
@scouten-adobe Thank you for the fix !

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) failed to backport this pull request in [workflow run 31429955496](https://github.com/contentauth/c2pa-rs/actions/runs/31429955496).

| Target | Status |
|--------|--------|
| `stable` | :x: Failed |

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2421-to-stable origin/stable
cd .worktree/backport-2421-to-stable
git switch --create backport-2421-to-stable
git cherry-pick -x b0a3c2bd9e526fd0634f205f91b0e8e9543e43a1
```

</details>

PR #2426, opened 2026-08-05
fix: Treat an unreachable ingredient manifest as absent when building a claim

scouten-adobe:

`Ingredient::add_to_claim` loads the bytes behind an ingredient's `manifest_data` reference with `?`, so a reference whose bytes cannot be resolved fails claim generation — `Builder::to_archive` (and signing) returns `Error::ResourceNotFound`. This happens for a legitimately unreachable manifest, e.g.:

- a cloud/remote manifest that was never fetched (offline), or
- an ingredient that was serialized to JSON — which `#[serde(skip)]`s its resource store — and later reloaded, leaving `manifest_data_ref()` set but the bytes gone.

`Ingredient::manifest_data()` already treats this case gracefully (returns `None`). This makes `add_to_claim` consistent: if the referenced bytes cannot be loaded, it logs at `debug` and treats the ingredient as having no manifest (the existing `None` arm) instead of erroring.

### Checklist

- [x] Added a test (`builder::tests::test_to_archive_tolerates_unreachable_ingredient_manifest`) reproducing the dangling-reference scenario.
- [x] `cargo +nightly fmt --all -- --check` clean; clippy clean on the changed crate.
- [x] `ingredient::` (18) and `builder::tests` (92) suites pass — no regressions from the behavior change.

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -9392,17 +9392,11 @@ mod tests {
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
 
-    // An ingredient that references a manifest whose bytes are unavailable (e.g. a cloud-only
-    // manifest that was never fetched, or an ingredient reloaded without its resource store) must
-    // be treated as if it had no manifest, rather than failing archiving/signing with a
-    // "resource not found" error. `manifest_data_ref()` returns the dangling reference while
-    // `manifest_data()` already returns `None`.
-    #[test]
-    fn test_to_archive_tolerates_unreachable_ingredient_manifest() {
-        let mut builder = Builder::default()
-            .with_definition(simple_manifest_json())
-            .unwrap();
-
+    // Build an ingredient whose `manifest_data` reference points at bytes that are not present in
+    // any resource store: the dangling-reference scenario (a cloud-only manifest that was never
+    // fetched, or an ingredient reloaded without its resource store). `manifest_data_ref()` returns
+    // the dangling reference while `manifest_data()` already returns `None`.
+    fn unreachable_manifest_ingredient() -> Ingredient {
         let mut ingredient = Ingredient::new_v2("cloud-parent", "image/jpeg");
         ingredient.set_relationship(Relationship::ParentOf);
         ingredient.set_active_manifest("urn:c2pa:cloud-manifest");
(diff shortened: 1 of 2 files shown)


sdk/src/ingredient.rs
```diff
@@ -1155,12 +1155,24 @@ impl Ingredient {
...
-        let (active_manifest, claim_signature) = match self.manifest_data_ref() {
-            Some(resource_ref) => {
-                // get the c2pa manifest bytes
-                let manifest_data = get_resource(&resource_ref.identifier)?;
-
+        // this is how any existing claims are added to the new store.
+        //
+        // A `manifest_data` reference whose bytes cannot be resolved (e.g. a cloud/remote manifest
+        // that was never fetched, or an ingredient reloaded without its resource store) is treated
+        // as if the ingredient had no manifest, mirroring `Ingredient::manifest_data`, rather than
```
tmathern:
```suggestion
        // as if the ingredient had no manifest, like in `Ingredient::manifest_data`, rather than
```

sdk/src/ingredient.rs
```diff
@@ -1155,12 +1155,24 @@ impl Ingredient {
...
+        // failing claim generation.
+        let manifest_data = self.manifest_data_ref().and_then(|resource_ref| {
+            get_resource(&resource_ref.identifier)
```
tmathern: Maybe this needs error matching, to be more targeted: maybe similar to here: https://github.com/contentauth/c2pa-rs/blob/869d633dc7ee9382afb96eb6cf357f1a4f63601e/c2pa_c_ffi/src/error.rs#L121
And let others still propagate? (For NotFound and ResourceNotFound out of those 3, since I think those are the ones we attempt to cover here. Not sure if others are relevant here or not, but greater precisions might be better here, and error matching is here to be leveraged).

sdk/src/builder.rs
```diff
@@ -9392,6 +9392,34 @@ mod tests {
         assert!(reader_json.contains("thumbnail.ingredient"));
     }
 
+    // An ingredient that references a manifest whose bytes are unavailable (e.g. a cloud-only
+    // manifest that was never fetched, or an ingredient reloaded without its resource store) must
+    // be treated as if it had no manifest, rather than failing archiving/signing with a
+    // "resource not found" error. `manifest_data_ref()` returns the dangling reference while
+    // `manifest_data()` already returns `None`.
+    #[test]
+    fn test_to_archive_tolerates_unreachable_ingredient_manifest() {
```
tmathern: Can we add the following tests, so we have more behavioral e2e coverage:
- create such an archive, use it as Builder, sign;
- create such an archive, add it as ingredient to a Builder, sign.

tmathern (commented):
- Will this change have impacts on things that consult the manifest data ref, like here: https://github.com/contentauth/c2pa-rs/blob/869d633dc7ee9382afb96eb6cf357f1a4f63601e/sdk/src/ingredient.rs#L1261-L1269 (for thumbnails, and especially thumbnails in redaction)? This feels like a difference between what got resolved, and the reference it was resolved from. Do checks like that need an update then, or is it still valid to do them that way? 
- Any potential side-effect on the validations when doing this? (Asking because of the other recent set of errors where either 2 fields need to be here together, or not here at all, so wondering if a clean up would be needed).

tmathern:
(Also you may need to consider adding a `backport` label here).

sdk/src/ingredient.rs
scouten-adobe: Applied in `1cf013f2` — folded the "like in `Ingredient::manifest_data`" wording into the reworked comment.

sdk/src/ingredient.rs
scouten-adobe: Done in `1cf013f2`. `add_to_claim` now matches `Error::NotFound | Error::ResourceNotFound(_)` and treats only those as an absent manifest; any other error (e.g. from a resource resolver) propagates instead of being swallowed by `.ok()`. Left out `MissingDataBox` since `get_resource` can't produce it on this path.

sdk/src/builder.rs
scouten-adobe: Added two e2e tests in `1cf013f2`: `test_sign_reloaded_archive_tolerates_unreachable_ingredient_manifest` (archive → reload as `Builder` → sign) and `test_sign_with_added_archive_ingredient_tolerates_unreachable_ingredient_manifest` (archive → add as ingredient to a `Builder` → sign); both sign successfully. Extracted a shared `unreachable_manifest_ingredient()` helper.

scouten-adobe:
Withdrawn. Other work is in progress with a better solution to this problem.

PR #2428, opened 2026-08-06
fix: Gate lopdf's rayon feature by real thread availability

klynsaha:

Enabling `pdf` (`dep:lopdf`) unconditionally pulls in lopdf's default features, which include `rayon`. On targets without real OS thread support (e.g. single-threaded WASM/Emscripten builds with no pthread support), rayon's global thread pool construction calls `std::thread::spawn` and panics when spawn fails - there's no graceful degradation.

lopdf already ships a serial fallback for both of its rayon call sites (`object_stream.rs`, `reader.rs`), gated behind `#[cfg(not(feature = "rayon"))]`. There was previously no way for a consumer to reach that path, since `c2pa`'s `lopdf` dependency didn't set `default-features = false`, so `rayon` came along for free with `pdf` regardless of target.

## Fix

`lopdf` now opts out of its default features and is re-declared with its defaults restored only under:

```toml
[target.'cfg(any(not(target_arch = "wasm32"), target_feature = "atomics"))'.dependencies]
```

- Native builds: always have real threads, so this is always true - identical behaviour to before.
- WASM builds compiled with atomics (threaded WASM, e.g. -C target-feature=+atomics): also true - identical behaviour to before.
- Single-threaded WASM (the case that was panicking): false, so lopdf falls back to its existing serial parsing path instead of touching rayon at all.

This restores today's behaviour everywhere it currently works and changes behaviour only for the one target family where it was actually broken. 

## Checklist

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -122,7 +122,7 @@ img-parts = "0.4.0"
 iref = { version = "3.2.2", features = ["serde"] }
 jfifdump = "0.6.0"
 log = "0.4.30"
-lopdf = { version = "0.44.0", optional = true }
+lopdf = { version = "0.44.0", optional = true, default-features = false }
 lazy_static = "1.5.0"
 memchr = "2.8.1"
 nom = "7.1.3"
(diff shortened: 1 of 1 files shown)


tmathern:
Can you elaborate more on: "This restores today's behaviour everywhere it currently works and changes behaviour only for the one target family where it was actually broken." ? Especially where it is currently broken?

klynsaha:
> Can you elaborate more on: "This restores today's behaviour everywhere it currently works and changes behaviour only for the one target family where it was actually broken." ? Especially where it is currently broken?

`lopdf` uses rayon for multithreading, which is on by default. It is broken for single-threaded WASM build. With this PR, existing behaviour stays the same for all other builds/environments.

tmathern (approved):
Pending a second approval.

scouten-adobe (changes requested):
PR actually looks reasonable to me, but I'm blocking merge until GitHub outage is cleared and we can verify all platforms.

scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31141324806](https://github.com/contentauth/c2pa-rs/actions/runs/31141324806).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2439 |

PR #2429, opened 2026-08-06
fix: remove `validationResults` requires `activeManifest` when serializing v3 ingredients

### LINKED ISSUE #2327: Report `manifest.inaccessible` error code on inaccessible manifests

ok-nick:
Remove `validationResults` requirement that `activeManifest` must exist when serializing v3 ingredients. This is an alternative to #2176 and #2367, pending spec approval.

- Closes #2327


gpeacock (approved):

tmathern (approved):

sdk/src/assertions/ingredient.rs
ok-nick, line 1: There's another issue that should be addressed here.
https://github.com/contentauth/c2pa-rs/blob/ae964ed6dc7ff7a94f4a02fc7f0d23ef7a461bb5/sdk/src/ingredient.rs#L728-L747
When there is a hard error, such as failing to parse an ingredient (as addressed in #2176), there is nothing in the validation log and thus nothing surfaced to the user about the problem.

gpeacock (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 32741225775](https://github.com/contentauth/c2pa-rs/actions/runs/32741225775).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2537 |

Issue #2430, opened 2026-08-06
Harden default HTTP resolver against SSRF to non-global IPs in manifest URLs


scouten-adobe:
## Background

CAI-12574 (SSRF via HTTP redirect) was resolved by disabling redirect-following in the default HTTP resolver path (no `allowed_network_hosts` configured). See `sdk/src/context.rs` `resolver()` / `resolver_async()`.

That fix closes the **redirect** vector. This issue tracks a separate, pre-existing residual identified during that work.

## Residual issue

In the **default configuration** (no `allowed_network_hosts` set), `is_uri_allowed` allows all hosts. An attacker who controls a manifest URL (e.g. via `dcterms:provenance` XMP) can therefore point it **directly** at an internal or cloud-metadata endpoint — no redirect required:

```
http://169.254.169.254/latest/meta-data/
http://127.0.0.1:.../...
http://10.x / 192.168.x / 172.16.x ...
```

Remote manifest fetch is on by default (`verify.remote_manifest_fetch = true`), so any app using default settings (including c2patool) will issue this request.

## Options to evaluate

1. **Block cloud-metadata / link-local by default** (e.g. `169.254.0.0/16`, IMDS endpoints). These are never legitimate manifest hosts; low backwards-compat risk. Keeps `localhost`/intranet working.
2. **Block all non-global IPs by default** (loopback, RFC1918, link-local, ULA, metadata). Most secure, but breaks `localhost`/intranet manifest fetching, local dev, and some existing httpmock-based tests (which use `127.0.0.1`) unless carefully gated (e.g. `#[cfg(test)]` allowances or a dev opt-in).
3. Make the IP policy configurable, with a safe default.

## Prior art in this repo

`sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs` `to_url()` already rejects IP literals and `localhost`/`nip.io`/`sslip.io` for `did:web` resolution (CAI-10364). A similar safe-IP check could be applied at the generic-resolver layer for manifest fetches.

## Notes

- `sdk/src/settings/signer.rs` `RemoteSigner::sign` also uses `SyncGenericResolver::with_redirects()`. That URL comes from trusted signer config rather than an attacker-controlled manifest, so it is lower risk, but should be considered as part of any resolver-hardening pass.

Related: CAI-12574, CAI-10364.

mpgopala:
Fixed with #2683

PR #2433, opened 2026-08-06
fix(sdk)!: Block redirect SSRF via per-hop validation; add allow_redirects (CAI-12574)

scouten-adobe:
## What & why

Fixes **CAI-12574** (HackerOne #3784091): SSRF via HTTP redirect. A remote-manifest / OCSP / timestamp / `did:web` URL (some of which come from untrusted content) could answer with a `302` pointing at an internal or cloud-metadata endpoint (e.g. `169.254.169.254`), and the SDK would follow it.

## Design: `core.allow_redirects` with per-hop validation

Rather than disable redirects (which breaks legitimate CDN redirects) or block internal hosts outright (which breaks internal PKI / localhost dev — see @gpeacock), the SDK now **follows redirects but validates each hop**:

- **`allow_redirects = true` (default)** — redirects are followed, **except** a hop whose target is a non-globally-routable address (loopback, RFC1918, link-local/cloud-metadata, IPv6 ULA/link-local, CGNAT, …), which is rejected with `HttpResolverError::RedirectTargetDisallowed`. Redirects to public hosts still work.
- **`allow_redirects = false`** — no redirect is followed (`HttpResolverError::RedirectDisallowed`).

The check is on redirect **targets**, not the initial request, so a URL that *directly* names an internal host (enterprise OCSP/TSA on a private address, a `localhost` dev server) is still fetched. Defaulting to `true` keeps the reported PoC blocked (the `302 → 169.254.169.254` hop is rejected) while almost no existing caller breaks.

`core.allowed_network_hosts` is unchanged and is re-checked on every hop.

## Implementation

- New `RedirectResolver` follows redirects at the SDK layer (HTTP clients pinned to no auto-redirect), resolves relative `Location`s via the `url` crate, preserves method/body, caps at 10 hops (`TooManyRedirects`), and re-issues each hop through the wrapped resolver so the allow-list applies per hop.
- Reuses the `host_is_non_global` classifier (normalizes host, rejects obfuscated/malformed numeric forms — per @tmathern) for redirect targets.
- Resolver stack composes concrete wrapper types and boxes once — fixes the earlier WASM/WASI `Arc<dyn>` `Send`/`Sync` errors.

## Tests

- `RedirectResolver` unit tests (follow-to-public, block-to-internal, disabled, hop cap, relative-`Location` resolution).
- Host classifier unit tests (v4/v6/mapped/obfuscated/names).
- Context + `Store::fetch_remote_manifest` integration: default allows a direct internal host; default blocks a redirect to `169.254.169.254`; `allow_redirects = false` refuses redirects.

## CI

Green locally: fmt, clippy (incl. `--all-features`, `--all-targets`), doc build (warnings denied), `wasm32-unknown-unknown` and `wasm32-wasip1` checks, and the lib/doc tests. (The three `store::tests` PNG-fixture failures are pre-existing on `main` — they need the fixture/file-I/O feature.)

## Notes

- `CHANGELOG.md` is left to release-plz (not hand-edited).
- Resolved-IP filtering to defeat DNS rebinding remains tracked in #2430.


gpeacock (commented):
This looks like it will do the job. But do we know if our own services are doing redirects? Do we know if there are redirects involved in timestamp services or ocsp checks people are using? This won't impact just remote manifest requests. No one is expecting c2patool to suddenly break because we did a security update, but this will effectively break them. What are the impacts to web or mobile clients? Will the allowed redirects need to be changed depending on region? 

I agree that a production service using our SDK should be configured to be safe from these attacks, but should be be blocking a developer using c2patool by default?

I think the implementation looks good but I have serious questions about making this the default in the SDK. Maybe we should override that by default in c2patool? IDK

sdk/src/http/restricted.rs
```diff
@@ -281,6 +284,127 @@ pub(crate) fn is_uri_allowed(patterns: &[HostPattern], uri: &Uri) -> bool {
...
+
+/// Returns true if `uri`'s host must be blocked under the default (no allow-list) network policy.
+///
+/// See [`DefaultNetworkGuard`] for the rationale and the DNS-rebinding caveat.
+pub(crate) fn host_denied_by_default_policy(uri: &Uri) -> bool {
+    let Some(host) = uri.host() else {
+        return false;
+    };
+
+    // `Uri::host()` may surround an IPv6 literal with brackets; strip them before parsing.
```
tmathern: I would throw in a normalization of the host (covers the malformed URIs case).

sdk/src/http/mod.rs
```diff
@@ -492,4 +492,31 @@ pub mod tests {
         redirect.assert_calls(1);
         target.assert_calls(1);
     }
+
+    // A resolver built for the default (no allow-list) path must not follow redirects: it returns
+    // the 3xx response itself rather than fetching the redirect target. This is the SSRF hardening
+    // for CAI-12574 – the SDK cannot inspect intermediate redirect hops, so it never follows them.
```
tmathern: Might depend on the resolver. Turns out reqwest may be able to do something here with a custom policy: https://docs.rs/reqwest/latest/reqwest/redirect/struct.Policy.html | https://docs.rs/reqwest/latest/src/reqwest/redirect.rs.html#102-109.

The Attempt struct (https://docs.rs/reqwest/latest/reqwest/redirect/struct.Attempt.html) can let you add checks.

sdk/src/settings/mod.rs
tmathern, line 411: Note for another time: Would be nice to make those policies reusable when using `with_resolver` (if not already done).

CHANGELOG.md
```diff
@@ -9,6 +9,23 @@ As of December 2025 and until the 1.0.0 version is released, the CAI team will o
 
 ## [Unreleased]
 
+### Security
+
+* Harden the default HTTP resolver against SSRF via redirects and internal hosts (CAI-12574). The
```
tmathern:
```suggestion
* Harden the default HTTP resolver on read against SSRF via redirects and internal hosts (CAI-12574). The
```
(Makes it more precise)

sdk/src/http/restricted.rs
```diff
@@ -281,6 +284,127 @@ pub(crate) fn is_uri_allowed(patterns: &[HostPattern], uri: &Uri) -> bool {
     false
 }
 
+/// Resolver wrapper that enforces the SDK's default (no allow-list) SSRF policy.
+///
+/// When [`Core::allowed_network_hosts`] is not configured, the SDK still must not connect to
```
tmathern: Might also want to reject malformed addresses

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
```diff
@@ -384,6 +384,20 @@ mod tests {
...
         };
 
+        // The default context blocks requests to non-globally-routable hosts (SSRF hardening,
+        // CAI-12574). These hermetic tests point `set_proxy` at a loopback mock server, so they
+        // must explicitly allow that host – which is also how a deployment would reach an internal
+        // did:web host. Allowing a host also routes through `RestrictedResolver`, bypassing the
+        // default guard.
+        fn context_allowing(server: &MockServer) -> Context {
+            Context::new()
+                .with_settings(format!(
```
tmathern: I think this is worthy of an example/doc/howto, because someone developing locally a client and/or its server might need that, since the server might run locally on localhost during dev.

sdk/src/identity/validator.rs
```diff
@@ -114,19 +114,41 @@ mod tests {
             &server.url("/"),
         );
 
+        // The default resolver now blocks non-globally-routable hosts (SSRF hardening,
+        // CAI-12574). The proxy above points did:web resolution at this loopback mock server, so
+        // the host must be allow-listed for resolution to reach it. Callers that build their
+        // `Settings` from thread-local state pick this up; those that build a fresh `Settings`
```
tmathern: thread local settings are deprecated!

sdk/src/identity/claim_aggregation/w3c_vc/did_web.rs
tmathern: (And this relates to @gpeacock's point too)

scouten-adobe:
Thanks for the reviews — pushed a substantial revision.

**@gpeacock** — you were right that forcing strict-by-default would break real users, and it hit more than remote manifests: OCSP, timestamps, and `did:web` all flow through the same resolver, and enterprise PKI often puts OCSP/TSA on internal hosts. The fix is now a `core.network_security` policy (`off` / `default` / `strict`, default `default`). The default only stops **redirect following** — it does *not* block directly-named internal hosts — so internal PKI and `localhost` dev keep working. `strict` is the opt-in production lockdown; `off` restores the old redirect-following behavior. Your specific questions (do our services/TSAs/OCSP redirect? regional redirects?) no longer gate correctness for most users, since directly-reachable internal/public hosts are unaffected and only redirect *chasing* is disabled by default.

**@tmathern** — thanks, all addressed:
- **Normalize host + reject malformed addresses** — the strict guard (renamed `NonGlobalHostGuard`) now normalizes the host (strips IPv6 brackets and a trailing FQDN dot, lower-cases) and fails closed on host-less URIs and obfuscated/malformed numeric forms (`2130706433`, `127.1`, `0x7f…`).
- **ureq / `Location` header** — good catch adjacent to your reqwest-policy note: the ureq resolver was dropping all headers except `Content-Type`, so the redirect target was invisible. It now forwards all response headers (matching reqwest), which is what lets `RedirectGuard` surface the target.
- **reqwest custom redirect `Policy`** — noted for a future iteration. We deliberately kept per-hop-follow out of this PR (the team chose the policy knob over Approach C); the client-agnostic `RedirectGuard` gives a clear error today, and a reqwest `Policy::custom` per-hop validator is a good enhancement to layer on later.
- **Reusable policies with `with_resolver`** — agreed, filed as a follow-up thought; the guards are composable wrappers (`RedirectGuard`, `NonGlobalHostGuard`, `RestrictedResolver`) so exposing them for custom resolvers is straightforward later.
- **CHANGELOG "on read"** — applied.
- **Local-dev how-to** — added a "Local development and internal hosts" section (with a runnable example) to the `http` module docs, and reverted the `did:web`/validator test allow-list workarounds (the default policy allows loopback now), which also removed the deprecated thread-local settings usage you flagged.

**@ok-nick** — added you since you've worked on the redirect handling here; would value your eyes on the `RedirectGuard` approach and the ureq header-forwarding change in particular.

Still deferred (documented): resolved-IP filtering to defeat DNS rebinding, tracked in #2430.

tmathern (changes requested):
Builds fail...

sdk/src/http/mod.rs
```diff
@@ -327,6 +364,29 @@ pub enum HttpResolverError {
...
+    #[error(
+        "HTTP redirect to \"{location}\" (from \"{uri}\") was not followed: the SDK does not follow \
+         redirects under the current network_security policy. Set core.network_security = \"off\" to \
+         allow redirects, or add the target host to core.allowed_network_hosts and request it directly."
+    )]
+    RedirectDisallowed {
+        /// The URI that returned the redirect response.
+        uri: String,
+        /// The redirect target from the response `Location` header (or `"(unknown)"` if absent).
+        location: String,
```
ok-nick: Can this be an optional instead?

sdk/src/http/ureq.rs
```diff
@@ -41,8 +41,11 @@ pub mod sync_impl {
...
                 .version(response.version());
 
-            if let Some(content_type) = response.headers().get(header::CONTENT_TYPE) {
-                builder = builder.header(header::CONTENT_TYPE, content_type);
+            // Forward all response headers. Preserving headers such as `Location` and
+            // `Content-Length` lets downstream wrappers (e.g. the redirect guard) and callers see
+            // the full response, matching the behavior of the reqwest resolver.
+            for (name, value) in response.headers().iter() {
+                builder = builder.header(name, value);
             }
```
ok-nick, line 49: 👍

CHANGELOG.md
ok-nick, line 1: Will this conflict with automated changelog generation?

sdk/src/settings/mod.rs
```diff
@@ -440,6 +463,55 @@ impl SettingsValidate for Core {
...
+    /// [`HttpResolverError::UriDisallowed`].
+    ///
+    /// Recommended for production services that process untrusted content. Note that this check is
+    /// applied to the request URI; filtering on the *resolved* IP address (to defeat DNS names that
+    /// point at internal addresses) is tracked in
+    /// <https://github.com/contentauth/c2pa-rs/issues/2430>.
+    ///
+    /// [`HttpResolverError::UriDisallowed`]: crate::http::HttpResolverError::UriDisallowed
+    Strict,
+}
```
ok-nick: I'm not totally convinced we need this functionality in the SDK. I'm just not sure how many people will be using it. If there are security concerns you can set an allowed list to avoid sending arbitrary requests. Otherwise you can define your own HTTP resolver.

I have a feeling introducing this functionality will expose new types of vulnerabilities associated with how and how much we filter, etc..

scouten-adobe:
Pushed a significant revision reflecting the discussion with @ok-nick.

**Redirect handling is now a single boolean, `core.allow_redirects`, with per-hop validation** (rather than the earlier `network_security` enum that disabled redirects):

- `true` (default): follow redirects, but reject any hop whose target is a private/internal address (`RedirectTargetDisallowed`). Public redirects still work.
- `false`: don't follow redirects at all (`RedirectDisallowed`).

The internal-address check applies to redirect **targets**, not the initial request, so directly-named internal hosts (internal PKI OCSP/TSA, localhost dev) keep working — which is the balance @gpeacock was after. The reported PoC (`302 → 169.254.169.254`) is still blocked at the hop, and the default is `true` so existing redirect-reliant callers don't break.

Mechanically this is the SDK-level redirect loop @tmathern pointed at (clients pinned to no auto-redirect; each hop re-issued through the wrapped resolver, so `allowed_network_hosts` is re-checked per hop; relative `Location`s resolved via the `url` crate; 10-hop cap).

Also in this push:
- Fixed all the CI failures — the WASM/WASI `Arc<dyn AsyncHttpResolver>` `Send`/`Sync` errors (now composing concrete wrapper types and boxing once), the broken/private intra-doc links, and the `expect_used` clippy lint.
- Reverted `CHANGELOG.md` (release-plz-generated — my mistake to touch it earlier).
- Updated to latest `main`.

Green locally: fmt, clippy (`--all-features --all-targets`), doc build (warnings denied), `wasm32-unknown-unknown` + `wasm32-wasip1` checks, lib/doc tests.

@ok-nick @tmathern — re-requesting review on the new `RedirectResolver` (the per-hop loop and method/body handling are the parts most worth a look).

sdk/src/http/ureq.rs
```diff
@@ -41,8 +41,11 @@ pub mod sync_impl {
                 .status(response.status())
                 .version(response.version());
 
-            if let Some(content_type) = response.headers().get(header::CONTENT_TYPE) {
-                builder = builder.header(header::CONTENT_TYPE, content_type);
+            // Forward all response headers. Preserving headers such as `Location` and
+            // `Content-Length` lets downstream wrappers (e.g. the redirect guard) and callers see
```
tmathern, line 45: `Content-Length` is (roughly) a honor system, if it is checked reused later, it needs to be scrutinized (aka, clamped).

sdk/src/http/restricted.rs
```diff
@@ -281,6 +284,269 @@ pub(crate) fn is_uri_allowed(patterns: &[HostPattern], uri: &Uri) -> bool {
...
+        .map_err(|e| HttpResolverError::Other(Box::new(e)))?;
+
+    target
+        .as_str()
+        .parse::<Uri>()
+        .map_err(|e| HttpResolverError::Http(http::Error::from(e)))
+}
+
+/// Builds the follow-up request for a redirect, preserving the original method and body and copying
+/// headers other than `Host` (which the client recomputes for the new authority).
```
tmathern: If they would be set: Is it OK to not drop Authorization and Cookies headers too here? They likely would be needed for remote storage, just asking for due dilligence.

sdk/src/http/mod.rs
```diff
@@ -34,10 +34,46 @@
 //! Network requests may also be issued during the signing process, such as when
 //! [`SignerSettings::Remote`] is specified.
 //!
+//! # Network security policy (SSRF hardening)
```
tmathern, line 37: This is only a read policy, not for signing (right?). If so, this must be made clearer in the docs (so in the future there is no confusion between read/write(sign)).

For instance, the code below would still stay as is?
https://github.com/contentauth/c2pa-rs/blob/22ca5f29468db2b7dc3c064c6ae605502f17a8a1/sdk/src/http/mod.rs#L205-L207

sdk/src/http/mod.rs
```diff
@@ -34,10 +34,46 @@
...
+//! reach internal or cloud-metadata endpoints (SSRF – CAI-12574). By default the SDK follows
+//! redirects **except** when a redirect target is a non-globally-routable (internal) address, which
+//! is rejected as [`HttpResolverError::RedirectTargetDisallowed`]. This is controlled by
+//! [`Core::allow_redirects`]; set it to `false` to refuse redirects entirely (reported as
+//! [`HttpResolverError::RedirectDisallowed`]). See [`Core::allowed_network_hosts`] for an explicit
+//! host allow-list.
+//!
+//! ## Local development and internal hosts
+//!
+//! The redirect check applies to redirect *targets*, not the initial request, so a request that
```
tmathern: This should be more prominent in the comment, because it is also an accepted risk (you could still put an IPv4 url here directly and reach it).

sdk/src/http/mod.rs
```diff
@@ -289,7 +325,11 @@ impl AsyncHttpResolver for AsyncGenericResolver {
 }
 
 /// An error that occurs during sync/async HTTP resolver resolution.
+///
+/// This enum is `#[non_exhaustive]`: match it with a wildcard (`_`) arm so that future error
```
tmathern: Is this comment really needed here?

sdk/src/http/mod.rs
```diff
@@ -327,6 +367,54 @@ pub enum HttpResolverError {
...
+    /// An HTTP redirect response was received, but redirect following is disabled because
+    /// [`allow_redirects`] is `false`.
+    ///
+    /// This is expected when a remote-manifest, OCSP, timestamp, or `did:web` URL responds with a
+    /// 3xx redirect and the caller has turned redirect following off. To follow redirects (to
+    /// public hosts) again, set `core.allow_redirects = true`.
+    ///
+    /// [`allow_redirects`]: crate::settings::Core::allow_redirects
+    #[error(
+        "HTTP redirect to \"{location}\" (from \"{uri}\") was not followed because \
```
tmathern, line 415: This might also log the URLs. Is this fine?
(Means you might log attacker content).

sdk/src/http/mod.rs
```diff
@@ -327,6 +367,54 @@ pub enum HttpResolverError {
...
+    ///
+    /// The redirect target is a loopback, private (RFC 1918), link-local / cloud-metadata,
+    /// IPv6 unique-local/link-local, or similar internal address. Redirects to public hosts are
+    /// followed normally; only internal targets are blocked. A URL that *directly* names an
+    /// internal host (rather than being redirected to one) is not affected by this and is still
+    /// fetched.
+    ///
+    /// [`allow_redirects`]: crate::settings::Core::allow_redirects
+    #[error(
+        "HTTP redirect to \"{location}\" (from \"{uri}\") was rejected: the target is a private or \
```
tmathern, line 437: This might also log the URLs. Is this fine?

sdk/src/http/mod.rs
```diff
@@ -327,6 +367,54 @@ pub enum HttpResolverError {
...
+    )]
+    RedirectTargetDisallowed {
+        /// The URI that returned the redirect response.
+        uri: String,
+        /// The internal redirect target from the response `Location` header.
+        location: String,
+    },
+
+    /// The request exceeded the maximum number of HTTP redirects the SDK will follow.
+    #[error("too many HTTP redirects while resolving \"{uri}\"")]
```
tmathern, line 448: This might also log the URLs. Is this fine?

sdk/src/http/restricted.rs
```diff
@@ -281,6 +284,269 @@ pub(crate) fn is_uri_allowed(patterns: &[HostPattern], uri: &Uri) -> bool {
...
+    }
+}
+
+fn ipv4_is_non_global(ip: Ipv4Addr) -> bool {
+    let [a, b, ..] = ip.octets();
+
+    ip.is_unspecified()          // 0.0.0.0
+        || a == 0                // 0.0.0.0/8 "this network"
+        || ip.is_loopback()      // 127.0.0.0/8
+        || ip.is_private()       // 10/8, 172.16/12, 192.168/16
```
tmathern, line 545: 👍

sdk/src/context.rs
```diff
@@ -1072,6 +1096,106 @@ mod tests {
         let _resolver = context.resolver_async();
     }
 
+    // The default policy (`allow_redirects = true`) must NOT block direct requests to internal
```
tmathern, line 1099: This must be made clear in docs that this is on purpose (it is also an accepted risk/trade-off).

scouten-adobe:
Thanks @tmathern — addressed the actionable items in the latest push:

- **Redirect follower now drops `Authorization` / `Cookie` / `Proxy-Authorization`** so a cross-origin redirect can't leak credentials to the target.
- **Attacker-influenced URLs are sanitized before going into error messages** — control chars escaped (no log injection via a crafted URL or redirect `Location`) and truncated. Applies to `UriDisallowed`, `RedirectDisallowed`, `RedirectTargetDisallowed`, `TooManyRedirects`.
- **Docs clarified**: the policy governs the SDK's HTTP resolvers (remote manifest, OCSP, timestamp, `did:web` — read *and* the sign-path timestamp/OCSP fetches), but **not** the remote-signer transport (`RemoteSigner::sign` uses its own client with an operator-configured URL). Also made the accepted trade-off explicit: a URL that *directly* names an internal host is fetched by design; only redirect *targets* are validated.
- Trimmed the redundant `#[non_exhaustive]` comment; added tests for the header dropping and URL sanitization.

Filed follow-ups for the rest so they don't block this security fix:
- Content-Length preallocation clamp on manifest fetch → CAI-13290
- Remote-signer transport not going through the policy → CAI-13291
- Backend-native reqwest redirect policy → CAI-13292
- Reusable policies for custom `with_resolver` resolvers → CAI-13293

Also merged latest `main` and the earlier build failures (WASM/WASI/doc/clippy) are fixed — CI should be green now. Could you re-review / clear the earlier "changes requested" when you get a chance?

sdk/src/http/mod.rs
```diff
@@ -492,4 +616,31 @@ pub mod tests {
         redirect.assert_calls(1);
         target.assert_calls(1);
     }
+
+    // A resolver built for the default (no allow-list) path must not follow redirects: it returns
+    // the 3xx response itself rather than fetching the redirect target. This is the SSRF hardening
+    // for CAI-12574 – the SDK cannot inspect intermediate redirect hops, so it never follows them.
```
tmathern: This comment may be outdated.

sdk/src/http/restricted.rs
```diff
@@ -281,6 +284,276 @@ pub(crate) fn is_uri_allowed(patterns: &[HostPattern], uri: &Uri) -> bool {
...
+    host.to_ascii_lowercase()
+}
+
+/// Returns true if `host` looks like a numeric IPv4 address in a non-standard/obfuscated form (all
+/// digits and dots, or a `0x`/`0X` hex form). Standard addresses are handled by `IpAddr::parse`
+/// before this is called; this only catches forms that parse failed on.
+fn looks_like_obfuscated_ip(host: &str) -> bool {
+    let digits_and_dots = !host.is_empty() && host.bytes().all(|b| b.is_ascii_digit() || b == b'.');
+
+    digits_and_dots || host.starts_with("0x") || host.starts_with("0X")
```
tmathern: (Deferrable) You could have hex at any position, not only start.

tmathern (approved):

scouten-adobe:
Thanks @tmathern — addressed your two latest comments (and merged latest `main`):

- **`restricted.rs` obfuscated-IP check**: now catches a `0x` hex octet in *any* position, not just the first (e.g. `127.0x1`, `0x7f.0x0.0x0.0x1`), in addition to the decimal/octal forms. Added regression cases.
- **`mod.rs` test-helper comment**: it was stale — post-redesign the SDK *does* follow and validate redirect hops via `RedirectResolver`. Reworded it to say what the helper actually checks: that the underlying HTTP client doesn't auto-follow redirects (the precondition for per-hop validation).

tmathern (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31808513568](https://github.com/contentauth/c2pa-rs/actions/runs/31808513568).

PR #2437, opened 2026-08-06
fix: Builder.save_to_stream was not adding timestamp assertions CAI-21732

gpeacock:
Updated so that save_to_stream just gets the signer from the context and calls sign, so they are now identical.

This also fixes save_to_file but does not change sign_data_hashed embeddable or sign_embeddable

those methods have never done all the things that sign does - like adding thumbnails

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -3198,8 +3198,8 @@ impl Builder {
     /// This provides a simpler alternative to [`sign()`](Self::sign) when you want to use
     /// the context's configured signer rather than providing an explicit signer.
     ///
-    /// **Note**: This method is only available for synchronous signing. For async signing,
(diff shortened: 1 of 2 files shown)


tmathern (approved):

tmathern:
(WIll this need to be backported?)

tmathern:
(I added and removed labels to trigger the jobs again. Seems this was in the 1% of PRs whose jobs did not have recovered...).

ok-nick (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31226874342](https://github.com/contentauth/c2pa-rs/actions/runs/31226874342).

PR #2438, opened 2026-08-07
fix: Ensure inception action is only auto-added once

tmathern:
Ensure inception action is only auto-added once even if there are multiple action assertions.
Currently, with Builder intents, an inception action may get added to all action assertions, not only to one. This fix attempts to prevent that by make sure only one auto-inception action is added maximum.

## Checklist

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -1755,37 +1755,13 @@ impl Builder {
             }
         }
 
-        // A manifest may carry more than one actions assertion
-        // (a created-list assertion (with inception action), and a gathered-list assertion)
-        // Auto-inception action must still be only added exactly once.
-        let actions_assertions = || {
-            definition
-                .assertions
-                .iter()
-                .enumerate()
-                .filter(|(_, a)| parse_label(a.label()).0.starts_with(Actions::LABEL))
-        };
-        let has_inception = actions_assertions().any(|(_, a)| {
-            let Ok(actions) = a.to_assertion::<Actions>() else {
-                return false;
-            };
-            actions
-                .actions()
-                .iter()
-                .any(|act| matches!(act.action(), c2pa_action::CREATED | c2pa_action::OPENED))
-        });
-        let inception_target_index = (!has_inception)
-            .then(|| {
-                actions_assertions()
-                    .find(|(_, a)| a.created())
-                    .or_else(|| actions_assertions().next())
-                    .map(|(index, _)| index)
-            })
-            .flatten();
-
         let mut found_actions = false;
+        // A manifest may carry more than one actions assertion (a created-list assertion, which
+        // holds the inception action, and a gathered-list assertion). Only the first one may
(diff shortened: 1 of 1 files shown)


cdmurph32 (approved):
LGTM

sdk/src/builder.rs
```diff
@@ -1962,17 +1997,22 @@ impl Builder {
...
+    /// but only one of them may carry the inception (created/opened) action.
     fn add_auto_actions_assertions_settings(
         &self,
         ingredient_map: &HashMap<String, (&Relationship, HashedUri)>,
         actions: &mut Actions,
+        allow_inception: bool,
     ) -> Result<()> {
         let settings = self.context.settings();
         // https://spec.c2pa.org/specifications/specifications/2.3/specs/C2PA_Specification.html#_mandatory_presence_of_at_least_one_actions_assertion
         let auto_created = settings.builder.actions.auto_created_action.enabled;
```
gpeacock, line 2004: We should deprecate these two settings. We now do the same thing with intent and I don't know what happens when intent and auto actions collide. That's not for this PR, just something to think about.

sdk/src/builder.rs
```diff
@@ -1844,7 +1872,11 @@ impl Builder {
...
                     // are resolved to their hashed URIs.
-                    self.add_actions_assertion_settings(&ingredient_map, &mut actions)?;
+                    self.add_actions_assertion_settings(
+                        &ingredient_map,
+                        &mut actions,
+                        inception_target_index == Some(assertion_index),
+                    )?;
 
                     add_assertion(&mut claim, &actions, manifest_assertion.created())
                 }
```
gpeacock, line 1860: We can flag that we already found the first actions assertion and return an error if we find any others with inception. Set a local inception_allowed  to true at the top, set it to false if we find an actions assertion, and return an error from add_actions_assertion if it finds an actions assertion with inception_allowed false. That should handle both the order and the duplications without the needing to determine the index up front. 

If the user has already added an inception out of order, we would add an inception to the first and then error out when we find the second one, so  I think that handles all the cases we need.

gpeacock (changes requested):
This would work, but we also need to assure that the action assertion with inception is the first actions assertion in assertions!

sdk/src/builder.rs
tmathern: Sold!

sdk/src/builder.rs
tmathern: Agreed. Remember we hit cases where it would duplicate the inception action too...

gpeacock (approved):
Thanks this looks good

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31203845681](https://github.com/contentauth/c2pa-rs/actions/runs/31203845681).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2444 |

PR #2445, opened 2026-08-07
fix: Reject BMFF Merkle map location that overflows u32 chunk index (CAI-12884)

scouten-adobe:
## Summary

- CAI-12884 (HackerOne VULN-36815) reports a crafted MP4 whose C2PA merkle `uuid` box declares `BmffMerkleMap.location = u32::MAX`. In `BmffHash::verify_stream_hash_with_progress`'s leaf-hash finalization step, `chunk_bmff_mm.location as u32 + 1` overflows and panics ("attempt to add with overflow"), since `location` is a `usize` deserialized straight from attacker-controlled CBOR.
- Note this is **not** related to the recently-removed `mp4` crate (CAI-12277): the panic site is c2pa-rs's own code in `sdk/src/assertions/bmff_hash.rs`, in the step that runs *after* sample reading succeeds. I confirmed by reproduction that it's still live on current `main`.
- Fix: convert `location` to `u32` with a checked `try_from`, and use `checked_add(1)` for the chunk-index computation, so a malformed/out-of-range `location` is rejected with `Error::HashMismatch` instead of aborting the process.
- Added a regression test (`location_u32_max_does_not_panic`) reproducing the reporter's exact PoC value and asserting a clean rejection.

## Test plan

- [x] `cargo test -p c2pa --test bmff_timed_media_merkle` — all 23 tests pass, including the new regression test
- [x] `cargo test -p c2pa --lib --features file_io` — 941 passed
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `cargo clippy -p c2pa --lib --test bmff_timed_media_merkle` — clean

### DIFF
--- a/sdk/tests/bmff_timed_media_merkle.rs
+++ b/sdk/tests/bmff_timed_media_merkle.rs
@@ -1111,3 +1111,92 @@ fn track_without_chunk_offsets_is_skipped() {
 fn track_without_stsz_is_skipped() {
     assert_dropped_track_rejected(b"stsz");
 }
+
+// --- Regression: CAI-12884 / VULN-36815 -------------------------------------
+//
+// A crafted C2PA merkle `uuid` box declaring `BmffMerkleMap.location =
+// u32::MAX` drove an integer overflow panic in this crate's own
+// `verify_stream_hash_with_progress`: `chunk_bmff_mm.location as u32 + 1`
+// overflows when the (attacker-controlled, CBOR-deserialized) `location`
+// casts to `u32::MAX`. Unlike CAI-12939, this bug has nothing to do with the
+// removed `mp4` crate -- it lives entirely in `bmff_hash.rs`, in the
+// "finalize leaf hashes" step that runs after sample reading succeeds.
+
+/// Builds the same `uuid` merkle box as [`build_merkle_uuid_box`], but with a
+/// `location` value encoded as a full CBOR `u32` (major type 0, additional
+/// info 26) instead of the small-int encoding `put_uint` produces, so
+/// `location` can be set to values as large as `u32::MAX`.
+fn build_merkle_uuid_box_with_u32_location(
+    unique_id: u64,
+    local_id: u64,
+    location: u32,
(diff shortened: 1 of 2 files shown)


sdk/tests/bmff_timed_media_merkle.rs
```diff
@@ -1111,3 +1111,92 @@ fn track_without_chunk_offsets_is_skipped() {
...
+}
+
+/// A `location = u32::MAX` merkle box must be rejected with an error, not
+/// panic. (Before the checked conversion, this input panicked with an
+/// integer-overflow abort at `bmff_hash.rs:1508`.)
+#[test]
+fn location_u32_max_does_not_panic() {
+    let track = TrackSpec {
+        track_id: 1,
+        stsc: vec![StscEntry {
```
tmathern, line 1161: Assuming this  is attacker controlled, are there other places that could benefit from similar hardening in the same file? 
I see there is 1 other chunk_id computation in the file, would it benefit from hardening too? Here: https://github.com/contentauth/c2pa-rs/blob/d49f624a06004277abcf1f3996cfbb9c20f6cbf8/sdk/src/assertions/bmff_hash.rs#L1457-L1458.

tmathern (commented):
Note: Panic vs. wrap-around behavior depends on [overflow-checks behavior](https://doc.rust-lang.org/cargo/reference/profiles.html#overflow-checks), whose default flag value depends on release vs debug build too (release by default wraps, debug/dev panics by default to catch such things early).

ps: We don't set any in our Cargo.toml files (or I didn't see it).

tmathern (approved):
Might need to be merged with https://github.com/contentauth/c2pa-rs/pull/2454, or at least coordinated, since same file near same location.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31729686571](https://github.com/contentauth/c2pa-rs/actions/runs/31729686571).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2500 |

PR #2446, opened 2026-08-07
feat: Add related assertions field and validation to c2pa.actions

### LINKED ISSUE #2394: RelatedAssertions support
The 2.4 spec has a relatedAssertions field that can be added to an action. This requires JUMBF Urils which the current api does not provide. So we either need access to jumbf uris or some kind of identifier mapping.

In scope:
* Validation for relatedAssertions in Actions assertion
* New field in Actions assertion API

 
~~--~~
GitHub issue: <https://github.com/contentauth/c2pa-rs/issues/2394>
Labels: {{{}accepted{}}}, `feature`

*Created manually to backfill the GJSS GitHub→Jira outage. External Issue URL is set so the bridge can reconcile this issue once restored.*

ok-nick:
Adds the new 2.4 `Action::related_assertions` field and the corresponding validation for it.

- Closes #2394

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -2656,7 +2656,7 @@ impl Claim {
 
                         if assertion.is_none() {
                             log_item!(
-                                label.clone(),
+                                url.clone(),
                                 format!("relatedAssertions reference could not be resolved within the current manifest: {url}"),
                                 "verify_actions"
                             )
@@ -2676,7 +2676,7 @@ impl Claim {
                         let base = labels::base(&target_label);
                         if base == labels::ACTIONS || base == labels::INGREDIENT {
                             log_item!(
-                                label.clone(),
+                                url.clone(),
                                 format!("relatedAssertions must not reference an actions or ingredient assertion: {url}"),
                                 "verify_actions"
                             )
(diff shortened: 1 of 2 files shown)


sdk/src/claim.rs
```diff
@@ -2615,6 +2615,82 @@ impl Claim {
                     }
                 }
 
+                // 2.f if the action’s parameters field exists and contains a relatedAssertions field
+                if let Some(related_assertions) = action.related_assertions() {
+                    // 2.f.i if the value of relatedAssertions is not an array with at least one element,
+                    //       the claim shall be rejected with a failure code of assertion.action.malformed
+                    if related_assertions.is_empty() {
+                        log_item!(
+                            label.clone(),
```
gpeacock, line 2624: we want the full jumbf uri to the action if possible here, it really helps in debugging.

gpeacock (changes requested):
we should pass the actions jumbf uri, not just the label. we use this to know where the problem was.

sdk/src/claim.rs
ok-nick: The `label` here is actually the JUMBF URI taken from other parts of the code. However, I've changed the other part of the validation to report the hashed URI that was wrong instead of the hashed URI to the actions assertion.

scouten-adobe (approved):

gpeacock (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33415856327](https://github.com/contentauth/c2pa-rs/actions/runs/33415856327).

PR #2447, opened 2026-08-07
feat: Add digital source type field to ingredient

### LINKED ISSUE #2077: Feature + Bug: Digital Source Type in Ingredients (§18.16.12.3 in version 2.4) is not supported
### Component

c2pa-rs

### If Other, please specify

_No response_

### Version

v0.80.0

### Platform

Linux

### What happened?

Because `digitalSourceType` is not in the `V3_FIELDS` array, `c2patool` will reject any manifest that contains `digitalSourceType` in lieu of an active manifest in an ingredient.

This behavior is expected, because this usage of `digitalSourceType` is new feature in version 2.4. However, it is marked as a bug because of the practical implications:

1. The labeling requirements of the EU AI Act go into effect in August 2026. Similar legislation is pending in the US and in China.
2. We are recommending C2PA as the go-to method of compliance with labeling requirements (I assume that this is in line with the strategic goals of the C2PA / CAI).
3. For the vast majority of companies, they are not able to move their entire content workflow into a conformant TOE. This would require either going through the Conformance Program themselves, or ensuring that 100% of their software and their vendors are conformant.
4. Since compliance risk requires them to label ALL (not SOME) of their AI content, the only practical solution seems to be to inject C2PA before publication and add the gathered AI labels there (not created, but perhaps referenced by an interim CAWG identity).
5. In "normal" C2PA circumstances, the `digitalSourceType` attached to the `c2pa.created` action is the main indicator, but `c2pa.created` cannot be in a gathered assertion, so an alternative "standard" method is needed.
6. What has been recommended (by @lrosenthol) is that `digitalSourceType:trainedAlgorithmicMedia` is added to the ingredient using the new version 2.4 feature, along with a gathered `c2pa.ai-disclosure` assertion.
7. Adding `digitalSourceType` to the ingredient results in all existing Validator Products outright rejecting the asset (no information shown, just "Invalid"), because `c2pa-rs` does not expect this field in `c2pa.ingredient.v3`.
8. Requesting that existing Validator Products support this seems difficult, especially since conformance is still pegged to version 2.2, but is a potential solution.

If the above is true (and please correct me if I am wrong!), then this "bug" in practice may push companies away from C2PA-based solutions for AI labeling compliance. Depending on how much this distracts from C2PA's main focus on cryptography-backed provenance, we may or may not want to invest more effort into clearing up this use case.

ok-nick:
Adds the new 2.4 `Ingredient::digital_source_type` field and the corresponding validation for it.

- Closes #2077


tmathern (approved):

scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33402757353](https://github.com/contentauth/c2pa-rs/actions/runs/33402757353).

PR #2450, opened 2026-08-07
fix: relax field restriction requirements on ingredients

ok-nick:
Assertions may contain any additional unknown fields. We should not require that any unknown field must exist in the set of required and optional fields defined in the CDDL. This keeps assertions forward compatible with the spec.

> Since the addition of optional fields can be done while maintaining backwards compatibility, such fields may be added to an existing assertion’s schema without a change to the version number.

https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_versioning_2

### DIFF
--- a/sdk/src/assertions/ingredient.rs
+++ b/sdk/src/assertions/ingredient.rs
@@ -1137,6 +1137,36 @@ pub mod tests {
     }
 
     #[test]
+    fn test_unknown_field_does_not_break_decoding() {
+        let ingredients = [
+            Ingredient::new("title", "image/jpeg", "instance_id", None),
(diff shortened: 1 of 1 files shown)


gpeacock (approved):
This feels so wrong but it must be right

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31398241457](https://github.com/contentauth/c2pa-rs/actions/runs/31398241457).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2458 |

PR #2452, opened 2026-08-07
feat(sdk): Add jumbf_io::remove_jumbf_from_stream

### LINKED ISSUE #2451: jumbf_io has no stream form of remove, so a manifest store cannot be removed without file_io
### Describe your idea

**Setup**

I run a C2PA host with no filesystem: a wasm verifier in the browser, and a signer that streams. Assets never land on disk, so everything goes through `CAIRead` and `CAIReadWrite`, and `file_io` is off.

**What I need**

To remove an existing manifest store from an asset I only have as a stream.

The case I hit is box hashing an asset that is already signed. The existing manifest has to come out before I hash, or embedding the new one shifts the `caBX` box and the hash I computed is not the one a verifier recomputes. It is also the operation `remove_jumbf_from_file` documents for turning an embedded manifest into a remote one.

**Why I cannot**

`jumbf_io` publishes memory, stream and file forms of load and save, but only the file form of remove:

| | memory | stream | file |
|---|---|---|---|
| load | `load_jumbf_from_memory` | `load_jumbf_from_stream` | `load_jumbf_from_file` |
| save | `save_jumbf_to_memory` | `save_jumbf_to_stream` | `save_jumbf_to_file` |
| remove | | | `remove_jumbf_from_file` |

`remove_jumbf_from_file` is gated on `file_io` and takes a `&Path` I do not have.

The capability exists. Every asset handler implements `CAIWriter::remove_cai_store_from_stream`, and the file form is a thin dispatch over the path equivalent. It is just not reachable from outside the crate, and `get_caiwriter_handler` is `pub(crate)`, so I cannot dispatch it myself either. `jumbf_io::tests::test_jumbf` does exactly that dispatch internally, because it wants this operation and there is no public function to call.

**The ask**

`remove_jumbf_from_stream`, the sibling of `save_jumbf_to_stream`, dispatching the same way.

Target platform: all, and any build without `file_io` in particular.

stupside:
Fixes #2451.

`jumbf_io` publishes memory, stream and file forms of load and save, but only the file form of remove:

| | memory | stream | file |
|---|---|---|---|
| load | `load_jumbf_from_memory` | `load_jumbf_from_stream` | `load_jumbf_from_file` |
| save | `save_jumbf_to_memory` | `save_jumbf_to_stream` | `save_jumbf_to_file` |
| remove | | | `remove_jumbf_from_file` |

`remove_jumbf_from_file` is gated on `file_io`, so with that feature off there is no public way to remove a manifest store at all. The capability is already there: every asset handler implements `CAIWriter::remove_cai_store_from_stream`, and the file form is a thin dispatch over the path equivalent. Only the public entry point was missing.

This adds the sibling, dispatching through `get_caiwriter_handler` exactly as `save_jumbf_to_stream` does and returning `Error::UnsupportedType` for a format with no writer:

```rust
pub fn remove_jumbf_from_stream(
    asset_type: &str,
    input_stream: &mut dyn CAIRead,
) -> Result<()>
```

The doc comment refers to `remove_jumbf_from_file` in backticks rather than as an intra-doc link, since that item does not exist in a build without `file_io` and the link would dangle there.

## Checklist

## Tests

`test_jumbf`, the helper behind the per-format `test_streams_*` tests, already wanted this operation and had no public function to call, so it resolved the handler itself:

```rust
let handler = get_caiwriter_handler(asset_type).unwrap();
let mut removed = Cursor::new(Vec::new());
handler
    .remove_cai_store_from_stream(&mut writer, &mut removed)
    .unwrap();
```

It now calls `remove_jumbf_from_stream`, so the new function is covered for every format that helper runs: jpeg, png, webp, tiff, svg, mp3, wav, avi, mp4, heif, heic, avif, jxl and c2pa. That replaces the hand-rolled dispatch rather than adding a parallel test for it.

`test_remove_jumbf_from_stream_unsupported_type` covers the dispatch failure, which the per-format tests cannot reach.

`cargo test -p c2pa --lib jumbf_io` is green (21 tests), as are `cargo fmt` and `cargo clippy --all-targets`. `cargo check --no-default-features` passes, since being available without `file_io` is the point of the function.

### DIFF
--- a/sdk/src/jumbf_io.rs
+++ b/sdk/src/jumbf_io.rs
@@ -265,6 +265,26 @@ pub fn save_jumbf_to_stream(
     }
 }
 
+/// reads an asset of asset_type from input_stream, removes the manifest store and writes the
+/// result to output_stream
+///
+/// This is the stream equivalent of `remove_jumbf_from_file`, and the only form available
+/// without the `file_io` feature.
+///
+/// returns Unsupported type or errors from remove_cai_store_from_stream
+pub fn remove_jumbf_from_stream(
+    asset_type: &str,
+    input_stream: &mut dyn CAIRead,
+    output_stream: &mut dyn CAIReadWrite,
+) -> Result<()> {
+    match get_caiwriter_handler(asset_type) {
+        Some(asset_handler) => {
+            asset_handler.remove_cai_store_from_stream(input_stream, output_stream)
+        }
+        None => Err(Error::UnsupportedType),
+    }
+}
+
 /// writes the jumbf data in store_bytes into an asset in data and returns the newly created asset
 pub fn save_jumbf_to_memory(asset_type: &str, data: &[u8], store_bytes: &[u8]) -> Result<Vec<u8>> {
     let mut input_stream = Cursor::new(data);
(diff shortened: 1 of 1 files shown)


mauricefisher64:
Working on new standard asset I/O APIs so are not updating these APIs

PR #2453, opened 2026-08-09
fix: Harden against unbounded memory allocation attack in JPEG XL parser

ssanthosh:
`parse_all_boxes` (`sdk/src/asset_handlers/jpegxl_io.rs`) trusted a JPEG XL box's declared `total_size` (attacker-controlled, read straight from the file header) without checking it against the file's actual length. `JxlBoxInfo::end()`/`data_size()` derive a read-buffer size directly from `total_size`, and `find_xmp_data()`/`find_jumb_data()` size a zero-filled allocation from that value — so a tiny file (e.g. 44 bytes) could declare a multi-GB box and force a multi-GB allocation. In scope per `SECURITY.md`'s "large multiple of input size" carve-in.

## The fix

`parse_all_boxes` now rejects (`Error::InvalidAsset`) any box whose `offset + total_size` (computed with the `saturating_add` the function already used) exceeds the file length, before that `JxlBoxInfo` is ever returned to a caller. This matches the convention used by every other format handler in this codebase for the same class of problem — BMFF's box-tree builder, RIFF, PNG, and TIFF all reject a declared size that exceeds the file rather than silently clamping it and continuing.

## Tests added

- `test_box_info_end_and_data_size_trust_declared_size` — boundary case: a box whose declared size lands exactly at EOF parses normally (not rejected).
- `test_find_xmp_data_rejects_oversized_header` — reproduces the report: an `xml ` box lies about its size; `find_xmp_data` now returns `None` instead of attempting an oversized allocation.
- `test_find_jumb_data_rejects_oversized_header` — same for the C2PA `jumb` path via the 32-bit size field.
- `test_find_jumb_data_rejects_oversized_largesize_header` — same, via the 64-bit ISOBMFF `largesize` form (size field `== 1` + 8-byte size), which can declare sizes far beyond `u32::MAX` from a ~58-byte file.
- `test_parse_all_boxes_overflow_safe` (existing, updated) — an `offset + total_size` computation that would overflow `u64` must now be rejected outright (previously tolerated as an accepted side effect of `saturating_add`).

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/jpegxl_io.rs
+++ b/sdk/src/asset_handlers/jpegxl_io.rs
@@ -199,7 +199,14 @@ fn parse_all_boxes(reader: &mut dyn CAIRead) -> Result<Vec<JxlBoxInfo>> {
                 let next_pos = if info.total_size == 0 {
                     file_len
                 } else {
-                    info.offset.saturating_add(info.total_size)
+                    // A box can never legitimately extend past the file.
+                    let box_end = info.offset.saturating_add(info.total_size);
+                    if box_end > file_len {
+                        return Err(Error::InvalidAsset(
+                            "JPEG XL box size extends beyond asset bounds".to_string(),
+                        ));
+                    }
+                    box_end
                 };
 
                 if boxes.len() >= MAX_JXL_BOX_COUNT {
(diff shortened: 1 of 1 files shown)


sdk/src/asset_handlers/jpegxl_io.rs
```diff
@@ -113,19 +113,22 @@ impl JxlBoxInfo {
...
+    /// cannot legitimately extend beyond the bytes actually present, and an
+    /// unclamped value here would let a tiny file declare an enormous box
+    /// size, driving a huge zero-filled allocation in callers that size a
+    /// buffer from `data_size`/`end` (e.g. `find_xmp_data`, `find_jumb_data`).
     fn end(&self, file_len: u64) -> u64 {
         if self.total_size == 0 {
             file_len
         } else {
-            self.offset.saturating_add(self.total_size)
+            self.offset.saturating_add(self.total_size).min(file_len)
```
tmathern: Why clamp and not reject the asset with an error (out of bound reads)?

ssanthosh: Clamping (not rejecting) is intentional: the vulnerability is specifically the allocation size (safe_vec zero-fills before any read happens), and clamping caps that to what the file can actually supply, closing the DoS directly. It doesn't risk misreading corrupted data as valid — the recovered bytes still have to pass real validation downstream (String::from_utf8 for XML, Store::from_jumbf's BMFF box parsing for JUMB), so a genuinely malicious/corrupt oversized header still ends up failing. The upside of clamping over rejecting outright: a file where a box's declared size is merely wrong (e.g. a producer miscalculation) but the real XMP/JUMB payload is still fully present still gets recovered, instead of failing the whole asset over one bad header field.

tmathern: What is the behavior of other similar cases in other formats? Do they clamp or reject?

ssanthosh: I have checked other file formats like BMFF, TIFF, PNG etc., where all of them reject such cases with `InvalidAsset` asset. Hence, reverting the fix to reject such assets with `InvalidAsset` error to be consistent with other file formats handling.
@tmathern / @mpgopala : Please check the updated fix

tmathern (approved):

mpgopala (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31828064731](https://github.com/contentauth/c2pa-rs/actions/runs/31828064731).

PR #2454, opened 2026-08-09
fix: Harden against local_id truncation issue in accessing BMFF merkle maps

ssanthosh:
`BmffHash`'s timed-media Merkle verification (`sdk/src/assertions/bmff_hash.rs`) panicked on a crafted MP4 whose Merkle assertion declared a `MerkleMap.local_id` (an untrusted `usize` from the manifest) greater than `u32::MAX`. `split_bmff_merkle_map` keyed its output map by the full `local_id`, but the verify loop looked the group back up using the real track's `u32` `track_id` widened to `usize` — a value that only coincidentally equals `local_id` when the latter already fits in `u32`. Once `local_id > u32::MAX` (but truncates to a real track id), the two keys diverge and `HashMap`'s panicking `Index` (`[]`) crashed with "no entry found for key" — an unrecoverable panic from untrusted, attacker-controlled input in both debug and release builds. In scope per `SECURITY.md`'s welcome "parsing errors that cause crashes" category.

**Fix**: Look up by `mm.local_id` — the exact key it was inserted under — instead of re-deriving a value via the real track's id. This removes the only place the two identifiers were conflated, with no truncation needed anywhere. Also replaced the panicking `[]` index with `.get(...).ok_or(Error::HashMismatch(...))`, matching the non-panicking style already used for every other attacker-influenced lookup in this function (e.g. the existing `mm.local_id as u32` track lookup a few lines above).

## Tests added
- `timed_media_oversized_local_id_matching_real_track_does_not_panic` — builds a real single-track MP4 (reusing existing `build_timed_media`/`tm_assertion` test helpers) with `local_id = (1usize << 32) | 1`, which truncates to the real track id (1). 

## Checklist

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -1502,9 +1502,16 @@ impl BmffHash {
                             }
                         }
 
+                        // Look up by `mm.local_id`, the key this group was actually
+                        // inserted under in `split_bmff_merkle_map` (not `track_id`,
+                        // which only coincidentally matches it).
+                        let chunk_bmff_mms = track_to_bmff_merkle_map
+                            .get(&mm.local_id)
+                            .ok_or(Error::HashMismatch("Merkle location not found".to_owned()))?;
+
                         // finalize leaf hashes
                         let mut leaf_hashes = Vec::new();
-                        for chunk_bmff_mm in &track_to_bmff_merkle_map[&(track_id as usize)] {
+                        for chunk_bmff_mm in chunk_bmff_mms {
                             // `location` is attacker-controlled (deserialized from the
                             // file's uuid merkle box), so both the u32 conversion and the
                             // +1 must be checked rather than wrapping/panicking.
@@ -1525,7 +1532,7 @@ impl BmffHash {
                             }
                         }
 
-                        for chunk_bmff_mm in &track_to_bmff_merkle_map[&(track_id as usize)] {
+                        for chunk_bmff_mm in chunk_bmff_mms {
                             if chunk_bmff_mm.location >= leaf_hashes.len() {
                                 return Err(Error::HashMismatch(
                                     "BmffMerkleMap location exceeds leaf hash count".to_string(),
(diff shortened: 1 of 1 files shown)


tmathern (approved):
Might need to be merged with https://github.com/contentauth/c2pa-rs/pull/2445, or at least coordinated, since same file near same location.

tmathern:
Note: Panic vs. wrap-around behavior depends on [overflow-checks behavior](https://doc.rust-lang.org/cargo/reference/profiles.html#overflow-checks), whose default flag value depends on release vs debug build too (release by default wraps, debug/dev panics by default to catch such things early).

ps: We don't set any in our Cargo.toml files (or I didn't see it).

tmathern:
ps: wasm build job seems to be complaining about overflows too (it's in the test code):
```
error: this arithmetic operation will overflow
    --> sdk/src/assertions/bmff_hash.rs:3089:34
     |
3089 |         let oversized_local_id = (1usize << 32) | 1; // truncates to the real track id, 1
     |                                  ^^^^^^^^^^^^^^ attempt to shift left by `32_i32`, which would overflow
```

mpgopala (approved):
Should it also change `track_id` to `usize` so the mismatch does not happen?

ssanthosh:
> Should it also change `track_id` to `usize` so the mismatch does not happen?

@mpgopala : I thought about the same but it turns out track_id (the tkhd box's track_ID field) are a 32-bit field in the file format of MP4/BMFF itself. A real track can never have an id that needs more than 32 bits.

ssanthosh:
> Might need to be merged with #2445, or at least coordinated, since same file near same location.

Merged https://github.com/contentauth/c2pa-rs/pull/2445 and resolved conflicts on top of the changes

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31732968712](https://github.com/contentauth/c2pa-rs/actions/runs/31732968712).

PR #2455, opened 2026-08-10
fix(sdk): Validate inputTo ingredients against manifest tampering

mpgopala:
## Summary

The manifest-store validator skipped hash validation entirely for ingredients with `relationship: inputTo`. In `ingredient_checks` an unconditional `continue` bailed out before the ingredient's manifest bytes were ever compared against the hash the active claim signed. As a result, an attacker with **no signing key** could take a legitimately-signed asset, edit the bytes of an `inputTo` ingredient's manifest — forging that source's provenance (author, copyright, origin, actions) — and the tool would still report `validation_state: Valid`, even though the active claim holds a signed `c2pa_manifest` hash that the edited ingredient no longer matches.

This violates C2PA spec §15.11.2.1:

> "When processing a standard manifest, a validator shall validate each ingredient (**regardless of the value of its relationship field**)..."

The spec's only relationship-based carve-out (§15.11.3.3) is narrow: suppress the `ingredient.unknownProvenance` informational code for `inputTo` **when no manifest is present** — it never exempts a present ingredient manifest from hash validation.

## Changes

`sdk/src/store.rs` — `ingredient_checks`:

- Removed the unconditional `continue` for `Relationship::InputTo`, so an `inputTo` ingredient with a `c2pa_manifest` now runs the same manifest-hash / claim-signature validation as `parentOf` / `componentOf`.
- Kept the spec-sanctioned narrow behavior: the `unknownProvenance` informational code is still suppressed for `inputTo` ingredients **that have no provenance** (moved into the `else if relationship != InputTo` branch).

## Tests

Added two regression tests plus a shared builder helper:

- `test_input_to_ingredient_is_validated` — a well-formed `inputTo` ingredient must emit `ingredient.manifest.validated` and produce no errors.
- `test_input_to_ingredient_hash_mismatch_detected` — an `inputTo` ingredient whose bytes no longer match its signed hash must surface `ingredient.manifest.mismatch` and fail validation.
- `build_asset_with_input_to_ingredient(corrupt_hash)` builds M1 (standalone signed manifest) → M2 (regular v2 manifest referencing M1 as `inputTo`), with an optional one-byte hash flip simulating post-signing tampering.

## Test plan

- [x] Both new tests pass with the fix.
- [x] Both new tests **fail** when the `continue` is temporarily restored — confirming they catch the vuln (the failing dump showed the tampered asset reporting all-success with zero ingredient-manifest checks).
- [x] `store::tests` — 93 passed / 0 failed; all `ingredient` tests pass.
- [x] `make fmt` + `make clippy` clean.

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -9283,8 +9283,8 @@ pub mod tests {
     /// Returns the asset mime format and the signed M2 bytes.
     fn build_asset_with_input_to_ingredient(corrupt_hash: bool) -> (&'static str, Vec<u8>) {
         use crate::{
-            hashed_uri::HashedUri, jumbf::labels::to_signature_uri,
-            utils::test::create_test_store, ClaimGeneratorInfo, ValidationResults,
+            hashed_uri::HashedUri, jumbf::labels::to_signature_uri, utils::test::create_test_store,
+            ClaimGeneratorInfo, ValidationResults,
         };
 
         let mut context = Context::new();
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -1635,11 +1635,6 @@ impl Store {
                 .failure_as_err(validation_log, e)
             })?;
 
-            // we don't care about InputTo ingredients
-            if ingredient_assertion.relationship == Relationship::InputTo {
-                continue;
-            }
-
             validation_log
                 .push_ingredient_uri(jumbf::labels::to_assertion_uri(claim.label(), &i.label()));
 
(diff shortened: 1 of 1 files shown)


sdk/src/store.rs
```diff
@@ -1635,11 +1635,6 @@ impl Store {
                 .failure_as_err(validation_log, e)
             })?;
 
-            // we don't care about InputTo ingredients
```
tmathern, line 1638: Why didn't we care? Is this "jsut" old code that never got updated?

mpgopala: That could shouldn't have existed. This looks to be some old code added with [this](https://github.com/contentauth/c2pa-rs/pull/1144/changes#diff-36f6b2e73f041f21b9185d49356237eea3e87d0c6361040152d2bdee13014a06). According to spec, this is wrong. We should not differentiate between ingredients. Please go through the PR description for more details.

tmathern (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) failed to backport this pull request in [workflow run 31570223533](https://github.com/contentauth/c2pa-rs/actions/runs/31570223533).

| Target | Status |
|--------|--------|
| `stable` | :x: Failed |

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2455-to-stable origin/stable
cd .worktree/backport-2455-to-stable
git switch --create backport-2455-to-stable
git cherry-pick -x ac938fe1d47808edcb34dc105df78e778cc6d146
```

</details>

PR #2460, opened 2026-08-10
fix: `created` and `kind` flags should roundtrip in Builder archives

tmathern:
Builder archives currently do no set again the kind and created flags when roundtripping (archive then reload). This enables that propagation, and also checks the assertion size limit.

Fixes bug introduced by #2173, which added the following to reader.rs:
```rust
  for assertion in manifest.assertions.iter() {
      builder.add_assertion(assertion.label(), assertion.value()?)?;
  }
```
add_assertion always pushes created: false, kind: None (Cbor)
This PR replaces it with direct AssertionDefinition construction that preserves assertion.created() and maps ManifestAssertionKind::Json correctly.

## Checklist


cdmurph32 (approved):
LGTM

sdk/src/reader.rs
```diff
@@ -1244,7 +1244,29 @@ impl Reader {
                     {
                         continue;
                     }
-                    builder.add_assertion(assertion.label(), assertion.value()?)?;
+                    // For archive roundtrip: keep created/gathered attribution and kind
```
gpeacock, line 1247: wow, yes, thank you! this is a great catch and scary we didn't catch it before!

sdk/src/builder.rs
```diff
@@ -9689,6 +9690,133 @@ mod tests {
         Ok(())
     }
 
+    #[test]
```
gpeacock, line 9704: it's very good to have these tests now

sdk/src/reader.rs
gpeacock: Maybe we add this to builder as pub(crate) for now - add_assertion_impl(label, value, kind, created) ?

sdk/src/builder.rs
```diff
@@ -934,23 +934,24 @@ impl Builder {
...
-            let original_label = assertion_def.label.clone();
-            (assertion_def.to_assertion()?, original_label)
-        } else {
-            (Actions::new(), Actions::LABEL_VERSIONED.to_string())
-        };
-
-        let actions = actions.add_action(action);
-
-        self.add_assertion(original_label, &actions)?;
+            Some(pos) => {
```
gpeacock, line 954: Add assertion already has this logic in it, so we could just update that code with this improvement and call add_assertion. - or we could call add_asssertion_impl as referenced below.

gpeacock (approved):
We definitely need this fix! But, of course there's more to it. We need to have a public way to add gathered assertions!

sdk/src/reader.rs
tmathern: I was wondering the same, but it felt like a bigger API surface to add? But since you suggest it too, happy to oblige! (It's better/less brittle).

sdk/src/builder.rs
tmathern: I'll go for an `add_asssertion_impl`. Likely to be in next changeset.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31432050977](https://github.com/contentauth/c2pa-rs/actions/runs/31432050977).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2464 |

PR #2462, opened 2026-08-10
fix: Harden against arbitrary file write via manifest resource path traversal in Reader::to_folder

ssanthosh:
`Reader::to_folder()` (`sdk/src/reader.rs`) — the public API behind `c2patool <asset> -o <folder>` — writes every claim's binary assertions and databoxes to disk using a path built by `uri_to_path()` (`sdk/src/utils/io_utils.rs`). That path is derived from a JUMBF `jumd` box label read directly from the asset (`db_desc_box.label()` in `Store::from_jumbf_impl`, embedded raw into a URI by `Claim::put_databox`/`to_databox_uri` with no character restriction), so it is fully attacker-controlled. `uri_to_path` performed no traversal sanitization — only `:`→`_` replacement and prefix stripping — so a crafted asset with a databox (or binary assertion) label containing `../` could write attacker-controlled bytes to an attacker-chosen path outside the output directory. `to_folder` has no validation/trust gate, so this triggers even for assets that fail C2PA validation. This is the extraction-side counterpart to the ingest-side fix in PR #2066/#2271 (`sanitize_archive_path`), which never covered this sink.

**Fix:** `uri_to_path` now runs its result through the existing `sanitize_archive_path` helper (reused, not reimplemented) before returning, on every code path, and its signature changed from `-> PathBuf` to `-> Result<PathBuf>` so callers can reject rather than silently write. `Reader::to_folder`'s two call sites (binary assertions and databoxes — both share the exact same bug, since both route through `uri_to_path`) now propagate that `Result` with `?`.

## Tests added
- `test_uri_to_path_rejects_traversal` (`io_utils.rs`) — direct coverage of `..` in the URI tail, `..` in the `manifest_label` prefix, a bare absolute path, and a bare backslash path, plus a positive case confirming legitimate labels still resolve to the expected path (not over-blocked).
- `test_to_folder_rejects_path_traversal_in_databox_label` (`reader.rs`) — full integration test: injects a `Claim` with a malicious databox label via `put_databox` (the exact function the untrusted-parsing path calls) and confirms `to_folder` rejects it and writes nothing outside the target folder.

## Checklist

### DIFF
--- a/sdk/src/utils/io_utils.rs
+++ b/sdk/src/utils/io_utils.rs
@@ -22,7 +22,7 @@ use std::{
 #[allow(unused)] // different code path for WASI
 use tempfile::{tempdir, Builder, NamedTempFile, SpooledTempFile, TempDir};
 
-use crate::{asset_io::rename_or_move, Error, Result};
+use crate::{asset_io::rename_or_move, utils::path_utils::sanitize_archive_path, Error, Result};
 
 // Replace data at arbitrary location and len in a file.
 // start_location is where the replacement data will start
@@ -273,13 +273,19 @@ pub(crate) fn tempdirectory() -> Result<TempDir> {
 }
 
 /// Convert a URI to a file path using PathBuf for better path handling.
+///
+/// The URI may carry a JUMBF label taken directly from an untrusted asset
+/// (see callers in `Reader::to_folder`, which writes the resulting path to
+/// disk), so the derived path is passed through `sanitize_archive_path`
+/// before being returned - this rejects `..`, absolute paths, and backslash
+/// components rather than silently stripping them.
 #[cfg(feature = "file_io")]
-pub fn uri_to_path(uri: &str, manifest_label: Option<&str>) -> PathBuf {
+pub fn uri_to_path(uri: &str, manifest_label: Option<&str>) -> Result<PathBuf> {
     let mut path_str = uri.replace(':', "_");
     if let Some(stripped) = path_str.strip_prefix("self#jumbf=") {
         path_str = stripped.to_owned();
     } else {
-        return PathBuf::from(path_str);
+        return sanitize_archive_path(&path_str).map(PathBuf::from);
     }
 
     let mut path = PathBuf::from(path_str);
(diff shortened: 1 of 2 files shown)


tmathern (approved):
If this is in used in c2patool, should there be also updated error handling there to handle the new errors/the result error type?

sdk/src/utils/io_utils.rs
```diff
@@ -273,13 +273,19 @@ pub(crate) fn tempdirectory() -> Result<TempDir> {
 }
 
 /// Convert a URI to a file path using PathBuf for better path handling.
+///
+/// The URI may carry a JUMBF label taken directly from an untrusted asset
+/// (see callers in `Reader::to_folder`, which writes the resulting path to
+/// disk), so the derived path is passed through `sanitize_archive_path`
+/// before being returned - this rejects `..`, absolute paths, and backslash
+/// components rather than silently stripping them.
```
tmathern: Doesn't it strip things though? The `.` are technically resolved and stripped.

tmathern (approved):

ssanthosh:
> If this is in used in c2patool, should there be also updated error handling there to handle the new errors/the result error type?

Checked this and there are no change needed. Reader::to_folder's public signature (Result<()>) didn't change; only its two internal uri_to_path calls now propagate via ?. c2patool already calls reader.to_folder(&output)? to handle error

sdk/src/utils/io_utils.rs
ssanthosh: Yes, components are silently dropped by sanitize_archive_path, not rejected. Only .., absolute paths, and backslashes are rejected. Updated the doc comment to say that explicitly instead of the blanket "rather than silently stripping them" claim.

ssanthosh:
@tmathern : Added a minor fix needed to handle windows path separator backslash after manifest label with https://github.com/contentauth/c2pa-rs/pull/2462/commits/1597b62faae90adb24c866aa1927549d040e7bce

tmathern (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31797525540](https://github.com/contentauth/c2pa-rs/actions/runs/31797525540).

PR #2231, opened 2026-06-10
feat!: Split out c2pa-raw-crypto, take 2

scouten-adobe:
Revive #2202. Draft while we review impact on other APIs.

PR #2465, opened 2026-08-10
fix: Preserve identity assertions in split-signing paths (backport #2421)

scouten-adobe:
Automated backport of #2421 to `stable`.

Upstream-first: this change already merged to `main`. The automated
[backport-action](https://github.com/korthout/backport-action) run failed
to cherry-pick cleanly, so this was done by hand:

- Cherry-picked merge commit b0a3c2bd onto `stable`.
- Resolved a conflict in `sdk/src/identity/builder/identity_assertion_signer.rs`:
  `stable` doesn't yet have the `cert_chain` field that main's version of the
  struct carries (added by an unrelated, not-yet-backported commit), so the
  struct/field-type change from #2421 was applied without introducing that
  field.
- Adapted the new regression test in
  `sdk/src/identity/tests/embeddable_signing.rs`, which used
  `c2pa_raw_crypto::signer_from_private_key` and the two-argument
  `X509CredentialHolder::from_raw_signer(signer, cert_chain_der)` — both from
  the c2pa-raw-crypto split (#2231), which hasn't been backported to `stable`.
  Switched to stable's `crate::crypto::raw_signature::
  signer_from_cert_chain_and_private_key` and the single-argument
  `from_raw_signer`, matching the existing pattern in
  `x509_credential_holder::tests::simple_case`.

No other functional changes vs. the original PR.

## Test plan

- [x] `cargo +nightly fmt --all -- --check` clean
- [x] `cargo clippy -p c2pa --lib --tests` clean
- [x] `cargo test -p c2pa --lib identity::` — 77 passed, 0 failed
- [x] `cargo test -p c2pa --lib store::` — same 3 pre-existing fixture
      failures (`libpng-test.png` not found) reproduced on unmodified
      `stable`; unrelated to this change

### DIFF
--- a/sdk/src/store.rs
+++ b/sdk/src/store.rs
@@ -2300,47 +2300,59 @@ impl Store {
             ));
         };
 
-        // Write dynamic assertions only if placeholders were added during placeholder generation.
-        // We check if the dynamic assertion labels exist in the claim - if not, placeholders
-        // weren't added and we should skip writing to avoid size mismatches.
+        // Write any dynamic assertions exposed by the signer. The caller
+        // (`Builder::sign_embeddable`) reserves matching placeholder slots via
+        // `add_dynamic_assertion_placeholders` before calling this, so the assertion
+        // content replaces those slots in place. This must reuse the same
+        // `dynamic_assertions()` result across the whole operation – draining it on an
+        // earlier call is what silently dropped identity assertions in issue #2055.
         let dynamic_assertions = signer.dynamic_assertions();
         if !dynamic_assertions.is_empty() {
-            // Check if placeholders exist for these dynamic assertions
-            let has_placeholders = {
-                dynamic_assertions
-                    .iter()
-                    .all(|da| pc.assertion_hashed_uri_from_label(&da.label()).is_some())
-            };
+            // Every dynamic assertion needs a reserved placeholder slot to replace.
+            // Surface any that are missing with an actionable message instead of
+            // letting `write_dynamic_assertions` fail later with an opaque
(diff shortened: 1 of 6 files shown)


ok-nick (approved):
Approving backport fixes.

sdk/src/store.rs
```diff
@@ -2430,37 +2442,18 @@ impl Store {
...
-            }
+        // The data-hashed placeholder (`get_data_hashed_manifest_placeholder`) does not
+        // reserve space for dynamic assertions, so there is no way to embed them here
+        // without breaking the size contract the caller already committed to. Fail loudly
+        // rather than silently dropping the assertions (see issue #2055); callers that need
+        // dynamic assertions should use Builder::placeholder() + Builder::sign_embeddable().
+        if !signer.dynamic_assertions().is_empty() {
+            return Err(Error::BadParam(
+                "signer has dynamic assertions (e.g. CAWG identity) that the data-hashed \
+                 embeddable workflow cannot represent; use Builder::placeholder() followed \
```
gpeacock, line 2450: I'm thinking we should deprecate the older embeddable apis now.

gpeacock (approved):

sdk/src/store.rs
scouten-adobe: Agree. Filed CAI-13289 for that. :-)

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31620257317](https://github.com/contentauth/c2pa-rs/actions/runs/31620257317).

PR #2467, opened 2026-08-11
fix: Avoid QoS runtime detection warning

tmathern:
- Avoids spawning a hash worker thread when there is nothing to overlap, which avoids the warning from Apple's Thread Performance Checker which reports a priority inversion during signing on iOS/iPadOS:
```
Thread running at User-interactive quality-of-service class waiting on a lower QoS thread running at Default quality-of-service class. 
```

The loop in `hash_stream_by_alg_with_progress` spawns a worker thread per chunk to read the next chunk while the current worker hashes. On Apple platforms (iOS) a thread created by `std::thread::spawn` runs at `QOS_CLASS_DEFAULT` (~default priority) and does not inherit the spawner's priority, so a user-interactive caller (higher priority) blocking on that worker (default priority is lower priority) gets flagged with the warning about the inversion of QoS/priority levels.

So, here we hash inline for "small" cases, and spawn the worker only on-demand. The read-ahead behaviour for ranges larger than `MAX_HASH_BUF` is unchanged (and hash output is unchanged too, as is the progress callback). Also, names the thread so it is easier to identify.

Note: Ranges larger than `MAX_HASH_BUF` still spawn a `QOS_CLASS_DEFAULT` thread and can still trigger the warning on iOS, but this change will spare the (runtime check) warning for smaller assets (which are likely to be more common cases than 256MB+ ones).

Note 2: It should not impact perf for larger files, since the larger-filesize path is kept as-is.

## Checklist

### DIFF
--- a/sdk/src/utils/hash_utils.rs
+++ b/sdk/src/utils/hash_utils.rs
@@ -14,6 +14,7 @@
 use std::{
     fs::File,
     io::{Cursor, Read, Seek, SeekFrom},
+    num::NonZeroUsize,
     ops::RangeInclusive,
     path::Path,
 };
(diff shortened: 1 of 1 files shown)


tmathern:
Important note: THis avoid the warning for smaller assets (<256MB) on iOS, but does not "fix" the warning. The warning may still appear in runtime checks for larger assets.

(The assumption is that the smaller assets case is more common on iOS than the large assets case).

sdk/src/utils/hash_utils.rs
```diff
@@ -230,6 +231,33 @@ where
     R: Read + Seek + ?Sized,
     F: FnMut(u32, u32) -> Result<()>,
 {
+    let max_hash_buf = NonZeroUsize::new(MAX_HASH_BUF)
```
scouten-adobe, line 234: Not a blocker, but this is dead-code-in-practice: `MAX_HASH_BUF` is a const that's always nonzero, so the error branch can never trigger. It's harmless defensive code and enables the clean `_impl` signature, but a plain `usize` parameter (skipping `NonZeroUsize`) could be a simpler way to accomplish the same testability goal.

scouten-adobe (approved):

sdk/src/utils/hash_utils.rs
tmathern: I pondered that. In both case (usize) or this without the error filtering (or unwrap) could trigger clippy complains. I also prefer to handle errors explicitly if possible, Rust lets you do that so nicely!

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31614583883](https://github.com/contentauth/c2pa-rs/actions/runs/31614583883).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2482 |

PR #2468, opened 2026-08-11
fix: Complete OCSP responder certificate path from signer x5chain

byh-trufo:
It is common for OCSP responses to include only the responder's certificate, not the certificate of the CA that issued it. That issuing CA is, per RFC 6960 section 4.2.2.2, the same CA that issued the C2PA leaf -- so it is already present in the leaf's certificate chain (x5chain), and using it for path building is standard practice. Currently c2pa-rs validates the responder using only the certificates embedded in the OCSP response, so when the trust store holds just the root, path validation fails and the response is silently discarded -- no status code is emitted at all.

Fix: when the response embeds exactly the responder certificate and its issuer matches the signer's issuing CA, append the signer's chain as untrusted path-building candidates; trust is still established solely by check_certificate_trust.

## Checklist

### DIFF
--- a/sdk/src/crypto/cose/ocsp.rs
+++ b/sdk/src/crypto/cose/ocsp.rs
@@ -11,9 +11,11 @@
 // specific language governing permissions and limitations under
 // each license.
 
+use asn1_rs::FromDer;
 use async_generic::async_generic;
 use chrono::{DateTime, Utc};
 use coset::{cbor::value::Value, CoseSign1, Label};
+use x509_parser::prelude::X509Certificate;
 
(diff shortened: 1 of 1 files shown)


scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31543975487](https://github.com/contentauth/c2pa-rs/actions/runs/31543975487).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2472 |

PR #2469, opened 2026-08-11
fix(sdk): Prevent panic on out-of-range timestamp in GeneralizedTime conversion

mpgopala:
Fixes a denial-of-service panic when reading a C2PA asset whose timestamp
token carries an out-of-range signing time.

The `From<DateTime<Utc>> for GeneralizedTime` conversion called `.expect()` on
`der::asn1::GeneralizedTime::from_system_time`, which fails for any date outside
der's supported **1970-9999** range. The PKCS#9 signing-time attribute inside an
RFC 3161 timestamp token is fully attacker-controlled (a `UTCTime` with `YY`
50-69 decodes to 1950-1969, and `GeneralTime` can encode any year). Because the
conversion was infallible, an out-of-range value panicked instead of erroring.

### Attack path

`Reader` / manifest validation → `Store::verify_time_stamp` → `verify.rs`
`timestamp_to_generalized_time(signed_signing_time)` → `.into()` →
`GeneralizedTime::from` → `.expect(...)` → **panic (DoS)**.

Reachable both when **verifying** an asset (any reader of a crafted manifest)
and, via `verify_after_sign`, when **signing** against a malicious/MITM'd TSA.

### Fix

- Replace the infallible `From<DateTime<Utc>>` with a fallible
  `TryFrom<DateTime<Utc>>` (`Error = der::Error`); out-of-range dates now return
  an error instead of panicking.
- `timestamp_to_generalized_time` maps out-of-range values to `None`, so a
  malformed signed signing-time is ignored and validation falls back to the
  DER-parsed `genTime` (guaranteed in range). Matches the existing `None`
  semantics — no behavior change for valid inputs.
- Remove a pointless `gt -> dt -> gt` roundtrip that hit the same panic path.
- Correct stale `1970-2255` comments to the actual `1970-9999` range.

### DIFF
--- a/sdk/src/crypto/asn1/mod.rs
+++ b/sdk/src/crypto/asn1/mod.rs
@@ -461,21 +461,21 @@ impl From<GeneralizedTime> for chrono::DateTime<chrono::Utc> {
     }
 }
 
-// impl From cannot return Result; conversions should never fail for valid dates
-#[allow(clippy::expect_used)]
-impl From<chrono::DateTime<chrono::Utc>> for GeneralizedTime {
-    fn from(dt: chrono::DateTime<chrono::Utc>) -> Self {
+// Fallible: der's GeneralizedTime only supports 1970-9999, so a DateTime outside
+// that range (e.g. a pre-1970 signing time attacker-supplied in a timestamp token)
+// must return an error rather than panic.
+impl TryFrom<chrono::DateTime<chrono::Utc>> for GeneralizedTime {
+    type Error = der::Error;
+
+    fn try_from(dt: chrono::DateTime<chrono::Utc>) -> Result<Self, Self::Error> {
         // Convert chrono to SystemTime, then to der's GeneralizedTime
         let system_time: std::time::SystemTime = dt.into();
 
-        // Create GeneralizedTime from SystemTime
-        // This should never fail for valid dates (der crate supports dates from 1970-2255)
-        let der_time = der::asn1::GeneralizedTime::from_system_time(system_time)
-            .expect("Failed to create GeneralizedTime from valid DateTime");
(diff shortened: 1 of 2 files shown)


scouten-adobe (approved):

sdk/src/crypto/asn1/mod.rs
```diff
@@ -461,21 +461,21 @@ impl From<GeneralizedTime> for chrono::DateTime<chrono::Utc> {
...
-    fn from(dt: chrono::DateTime<chrono::Utc>) -> Self {
+// Fallible: der's GeneralizedTime only supports 1970-9999, so a DateTime outside
+// that range (e.g. a pre-1970 signing time attacker-supplied in a timestamp token)
+// must return an error rather than panic.
+impl TryFrom<chrono::DateTime<chrono::Utc>> for GeneralizedTime {
+    type Error = der::Error;
+
+    fn try_from(dt: chrono::DateTime<chrono::Utc>) -> Result<Self, Self::Error> {
         // Convert chrono to SystemTime, then to der's GeneralizedTime
         let system_time: std::time::SystemTime = dt.into();
```
tmathern, line 472: Can't this time conversion panic too then?

mpgopala: It wont as try_from returns a Result and if the clients handles that. from only returns Self that cannot return Error.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 31565911771](https://github.com/contentauth/c2pa-rs/actions/runs/31565911771).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2474 |

PR #2471, opened 2026-08-11
fix: Reject BMFF update manifest without a paired original manifest

ssanthosh:
`BmffIO::read_cai` (`sdk/src/asset_handlers/bmff_io.rs`) sorts C2PA `uuid` boxes into `manifest_bytes`/`original_bytes`/`update_bytes` by `box_purpose`, then merges `original` + `update` for Update Manifest assets. It errored on "`original` present, `update` missing" but had no check for the reverse — an `update` box with no `original` silently fell through to returning `manifest_bytes` alone, dropping a still-signed manifest with no error.

Since `box_purpose` is excluded from the C2PA hash (by design, for update workflows), an 8-byte ASCII retag (`original` → `manifest`) reproduces this without breaking any hash check. Reproduced with the reporter's files: baseline reports `Valid`/2 manifests, retagged reports `Valid`/1 manifest — a signed manifest silently vanishes. In scope per `SECURITY.md`'s "Failure to detect invalid/untrusted status" category.

**Fix:** added the missing symmetric branch — `update_bytes` present without `original_bytes` now errors (`"update manifest without original manifest"`), mirroring the existing reverse check. `write_cai` already enforces both directions in one condition; `read_cai` was missing half of it.

## Tests added
- `test_read_cai_rejects_update_without_original` — reproduces the retagged state; confirmed fails without the fix, passes with it.
- `test_read_cai_rejects_original_without_update` — the pre-existing error path had no test.
- `test_read_cai_accepts_ordinary_manifest_alone` — positive case.

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/bmff_io.rs
+++ b/sdk/src/asset_handlers/bmff_io.rs
@@ -1937,6 +1937,13 @@ impl CAIReader for BmffIO {
                     "original manifest without update manifest".to_string(),
                 ));
             }
+        } else if c2pa_boxes.update_bytes.is_some() {
+            // `update` only ever exists paired with `original` (spec Annex
+            // A.5.3). Falling through here would silently drop a
+            // still-signed update manifest instead of erroring.
+            return Err(Error::C2PAValidation(
+                "update manifest without original manifest".to_string(),
+            ));
         }
 
         c2pa_boxes.manifest_bytes.ok_or(Error::JumbfNotFound)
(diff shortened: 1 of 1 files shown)


scouten-adobe (approved):
Reviewed the fix in context (the box-purpose sorting in `read_cai`, and the symmetric check in `write_cai` that it mirrors). LGTM — correct and well-scoped.

- The added `else if c2pa_boxes.update_bytes.is_some()` branch closes exactly the asymmetry: `read_cai` errored on `original` without `update`, but let `update` without `original` fall through to `manifest_bytes` alone, silently dropping a still-signed manifest. It now mirrors `write_cai`'s existing both-directions check.
- A structural invariant is the right layer here, since `box_purpose` is excluded from the C2PA hash and so can't be caught by hash validation.
- Tests are solid: all three drive `read_cai` end-to-end, dummy bytes are fine because the pairing check returns before `Store::from_jumbf`, and they also cover the previously-untested reverse path plus a positive case. Low regression risk (a legitimate asset never has `update` without `original`).

One non-blocking note: Minor consistency: `read_cai` reports these as `Error::C2PAValidation` while `write_cai` uses `Error::BadParam` — pre-existing, not introduced here, but worth aligning at some point.

I've filed an internal follow-up (CAI-13282, linked to VULN-36807) to review the broader `box_purpose` retag surface as defense-in-depth — this PR correctly closes the reported permutation; that ticket covers a systematic pass over the others.

tmathern (approved):

mpgopala (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 31732878967](https://github.com/contentauth/c2pa-rs/actions/runs/31732878967).

PR #2480, opened 2026-08-12
fix: Apply certificate validity checks to all intermediate CAs

ssanthosh:
### Vulnerability 1
When verifying a manifest's signing certificate chain, the validity-period check was applied only to the leaf (signing) certificate, never to intermediate CA certificates — in both crypto backends.

**OpenSSL backend** (`certificate_trust/openssl.rs`): `check_certificate_trust` set `X509VerifyFlags::NO_CHECK_TIME` for the whole chain whenever there was no trusted timestamp, disabling expiry checking for every cert including intermediates.

**`rust_native` backend** (`certificate_trust/rust_native.rs`): a related but distinct bug — the validity check lived *inside* the trust-anchor-search loop, which returns as soon as it finds a match, so only the cert that happened to match a trust anchor ever got checked; every cert between that match and the leaf was skipped, and the check only ran at all when a signing timestamp was present.

In both cases, the leaf itself was separately covered by `check_certificate_profile`, but that function only ever examines the end-entity cert — nothing else validated intermediates.

### Vulnerability 2

On the `rust_native_crypto` backend (the only backend usable on wasm), `check_chain_order` verified that each certificate's signature validated against the next certificate's public key, but never checked whether that next certificate was actually authorized to act as a CA (`basicConstraints.cA`, `keyUsage.keyCertSign`, `pathLenConstraint`). An attacker holding the private key for any ordinary, validly-issued end-entity certificate could use it to self-issue a forged certificate that would then cryptographically "chain" up to a real trust anchor. The OpenSSL backend is unaffected — `X509_STRICT` already enforces this.

## Fix for Vulnerability 1

- **`certificate_trust/openssl.rs`**: no longer disables time-checking chain-wide — leaves the check time unset when there's no trusted timestamp, so OpenSSL's default "check against now" applies to every cert in the chain (matching the leaf's own "no timestamp → valid now" fallback).
- **`certificate_trust/rust_native.rs`**: added one unconditional pass over the whole chain (checking validity against the timestamp, or "now" if absent) run *before* the trust-anchor search, instead of embedding the check inside the early-returning loop, and removed it from that loop.

### Fix for Vulnerability 2

Merged the CA-authorization check into `check_chain_order` (renamed to `check_chain_order_and_ca_authorization`) so both properties are verified together per issuing certificate, with the structural checks (`basicConstraints`/`keyUsage`/`pathLenConstraint`) factored into a small `check_ca_authorization` helper for readability.

## Tests added for Fix for Vulnerability 1

- `test_intermediate_validity_is_checked` (`certificate_trust_policy.rs`) — backend-agnostic (dispatches through whichever backend is compiled), using three new fixtures (`chain_trust_root.pub`, `chain_trust_valid.pub`, `chain_trust_expired_intermediate.pub`): a currently-valid intermediate is trusted; an expired intermediate is rejected even though its own leaf is still within its validity window.

### Tests added for Fix for Vulnerability 2

- `test_non_ca_intermediate_is_rejected` and `test_path_len_constraint_is_enforced`, using newly generated certificate fixtures. Confirmed both fail against the pre-fix code and pass with the fix, on both crypto backends.

## Checklist


sdk/src/crypto/cose/certificate_trust/rust_native.rs
```diff
@@ -61,6 +65,33 @@ pub(crate) fn check_certificate_trust(
...
+    let check_time = match signing_time_epoch {
+        Some(t) => t,
+        None => SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map_err(|_| CertificateTrustError::InternalError("system time invalid".to_string()))?
+            .as_secs() as i64,
+    };
+    let check_time = x509_parser::time::ASN1Time::from_timestamp(check_time)
+        .map_err(|_| CertificateTrustError::CertificateNotTrusted)?;
+    for cert in &full_chain {
```
tmathern, line 84: Does this do a *chain* or all manifests? What if you should stop early in the chain because you reached the logical end of said relevant chain?
(Asking because sometimes we talk about candidate paths, so I assume there are potentially multiple paths in a chain, depending on where you stop in the chain!).

sdk/src/crypto/cose/certificate_profile.rs
```diff
@@ -72,6 +72,24 @@ pub fn check_end_entity_certificate_profile(
...
 }
 
+/// Returns true if `alg` is one of the signature algorithms allowed by
+/// [§14.5.1, Certificate Profiles]. This is a chain-wide requirement, not
+/// specific to the end-entity certificate - callers checking intermediate CA
+/// certificates should use this directly rather than assuming algorithm
+/// strength is only a leaf concern.
+///
+/// [§14.5.1, Certificate Profiles]: https://spec.c2pa.org/specifications/specifications/2.3/specs/C2PA_Specification.html#_certificate_profiles
+pub(crate) fn is_allowed_certificate_signature_algorithm(alg: &Oid) -> bool {
```
tmathern: Since this is new, maybe already return a Result instead of a bool?

sdk/src/crypto/cose/certificate_trust/openssl.rs
```diff
@@ -31,6 +35,19 @@ pub(crate) fn check_certificate_trust(
         return Err(CertificateTrustError::CertificateNotTrusted);
     }
 
+    // Signature-algorithm strength (C2PA spec 14.5.1) applies to every cert
+    // in the chain, not just the leaf: `check_certificate_profile` only ever
```
tmathern: But do all callers pass through the whole chain all the time here then too?
And rust native crypto impl has the same behavior (aka the two agree here?).

sdk/src/crypto/cose/certificate_profile.rs
ssanthosh: I would prefer to keep it as bool as the three call sites (here, openssl.rs, rust_native.rs) each wrap it in a different error type/logging shape (CertificateProfileError + log_item! here, CertificateTrustError in the other two). A Result would just be discarded and re-wrapped at every call site, so bool is the simpler fit for a plain predicate.

sdk/src/crypto/cose/certificate_trust/rust_native.rs
ssanthosh: `full_chain` is a single chain — one COSE x5chain array (leaf + its signer-provided intermediates), never multiple manifests or candidate paths. It's checked in full rather than stopping at the anchor match because every cert in it has already been cryptographically proven genuine by `check_chain_order` above — no legitimate chain would carry a bad cert past that point, so checking it all costs nothing and closes an edge case.

sdk/src/crypto/cose/certificate_trust/openssl.rs
ssanthosh: Checked all three callers (verifier.rs, ocsp.rs, time_stamp/verify.rs) — all pass the complete chain for whatever cert is being validated, never a truncated one. The two backends do agree on outcome: rust_native incidentally re-checks the leaf's algorithm (harmless, since the leaf is already checked separately by check_certificate_profile in every caller before this runs), openssl skips it. Same accept/reject result either way;

sdk/src/crypto/cose/certificate_profile.rs
tmathern: The fact that they already wrap into different error is indicative of a potential evolution from a bool to a Result though. The predicate might be plain now, but since it's already rewrapped in different errors by different backend, it may be wise nevertheless to wrap in a result already. Makes it more evolvable.

sdk/src/crypto/cose/certificate_trust/openssl.rs
tmathern: If it's an incidental check... Maybe make it explicit? Incidental behavior can change as the code evolve. Would the current tests in this PR catch that change?

sdk/src/crypto/cose/certificate_profile.rs
ssanthosh: This is no longer applicable as https://github.com/c2pa-org/specs-core/issues/2427 votes to do the algorithm check only for the leaf/signing certificate and not for the entire chain. Hence, removed this code

sdk/src/crypto/cose/certificate_trust/openssl.rs
ssanthosh: This is no longer applicable as https://github.com/c2pa-org/specs-core/issues/2427 votes to do the algorithm check only for the leaf/signing certificate and not for the entire chain. Hence, removed this code

ssanthosh:
@tmathern / @mpgopala  : Updated the PR as https://github.com/c2pa-org/specs-core/issues/2427 votes to do the algorithm check only for the leaf/signing certificate and not for the entire chain.  Please check

tmathern:
> @tmathern / @mpgopala : Updated the PR as [c2pa-org/specs-core#2427](https://github.com/c2pa-org/specs-core/issues/2427) votes to do the algorithm check only for the leaf/signing certificate and not for the entire chain. Please check

Do you need to hold off until the vote is settled then?

ssanthosh:
> > @tmathern / @mpgopala : Updated the PR as [c2pa-org/specs-core#2427](https://github.com/c2pa-org/specs-core/issues/2427) votes to do the algorithm check only for the leaf/signing certificate and not for the entire chain. Please check
> 
> Do you need to hold off until the vote is settled then?

Both @mauricefisher64 and @lrosenthol are in favour of doing the check for hash algorithm only for the leaf signing certificate

sdk/src/crypto/cose/certificate_trust/rust_native.rs
tmathern: Is this the same behavior as the openssl counterpart?

sdk/src/crypto/cose/certificate_trust/openssl.rs
```diff
@@ -46,10 +46,14 @@ pub(crate) fn check_certificate_trust(
     verify_param.set_flags(X509VerifyFlags::X509_STRICT)?;
     verify_param.set_flags(X509VerifyFlags::PARTIAL_CHAIN)?; // allow intermediates to be on anchor list
 
+    // Without a trusted timestamp, leave the check time unset so OpenSSL
+    // checks every cert in the chain (not just the leaf) against the
```
tmathern: `every cert in the chain (not just the leaf)` => Isn't openssl "just" checking the path to a cert?

mpgopala (approved):

sdk/src/crypto/cose/certificate_trust/rust_native.rs
ssanthosh: Yes, same behavior — openssl gets whole-chain expiry checking from its built-in path validation; rust_native needed the explicit loop since this backend doesn't have an equivalent.

sdk/src/crypto/cose/certificate_trust/openssl.rs
ssanthosh: In this case, chain and path refer to the same since `chain_der` is always the signer's own linear x5chain with nothing beyond it.  Updated the comment to make it clear with https://github.com/contentauth/c2pa-rs/pull/2480/commits/3b450dc7b56acc3cba6f512779e578e4030a03b5

tmathern (approved):

sdk/src/crypto/cose/certificate_trust/rust_native.rs
```diff
@@ -61,6 +62,29 @@ pub(crate) fn check_certificate_trust(
...
```
dcondrey, line 84: This makes the validity window chain-wide, but `check_chain_order` below (line 164) still only verifies that each certificate is signed by the next one — no `basicConstraints`/`cA`, no `keyUsage.keyCertSign`, no `pathLenConstraint` — and `check_end_entity_certificate_profile` is applied to `certs[0]` alone (`verifier.rs:179`).

So on this backend any valid end-entity certificate can act as an issuer: `[forged_leaf, real_leaf, real_intermediate, ...]` passes `check_chain_order`, passes this new validity loop, and the anchor search then matches on the real intermediate. The OpenSSL backend rejects that during path validation (`X509_V_ERR_INVALID_CA`), so the two backends disagree — and `rust_native` is the only one compiled for wasm (`tier-1a.yml` builds it `--no-default-features --features rust_native_crypto`).

Is the `cA` check meant to be in scope here, or a follow-up? Asking because the PR title and the vuln it closes read as covering intermediate CAs generally.

ssanthosh: @dcondrey : Good catch and thanks for the review comment. Event though initially the PR was intended for the reported Vulnerability issue, I have addressed the above ask as part of https://github.com/contentauth/c2pa-rs/pull/2480/commits/1aee32246702a71cf0a621446c73230b199da924 as it closes another possible vulnerability gap. Please check

mauricefisher64:
> > > @tmathern / @mpgopala : Updated the PR as [c2pa-org/specs-core#2427](https://github.com/c2pa-org/specs-core/issues/2427) votes to do the algorithm check only for the leaf/signing certificate and not for the entire chain. Please check
> > 
> > 
> > Do you need to hold off until the vote is settled then?
> 
> Both @mauricefisher64 and @lrosenthol are in favour of doing the check for hash algorithm only for the leaf signing certificate

It was decided and the text is changed to be more explicit in the upcoming 2.5 spec.

mauricefisher64 (approved):

tmathern (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33077806581](https://github.com/contentauth/c2pa-rs/actions/runs/33077806581).

PR #2118, opened 2026-05-06
feat: Support CAWG callback signing via c_ffi

gpeacock:
Replace AsyncRawSigner with AsyncSigner throughout AsyncX509CredentialHolder and AsyncIdentityAssertionSigner, using AsyncSignerWrapper from cose_sign.rs to bridge into the COSE signing path. Rename from_async_raw_signer to from_async_signer. Simplify test setup to use async_test_signer() directly.

Apply the same treatment to the sync side: replace hardcoded Box<dyn Signer + Send + Sync> with BoxedSigner in X509CredentialHolder, IdentityAssertionSigner, and IdentityAssertionBuilder, fixing WASM build failures. Add cfg-conditional Send/Sync handling to IdentityAssertionBuilder matching the pattern used by its async counterpart.

Add IdentityAssertionSigner::from_cawg_x509 as a convenience constructor for combining a C2PA signer with a CAWG X.509 identity assertion signer.

Add c2pa_create_cawg_signer(
takes C2paSigner pointers plus null-terminated arrays for referenced_assertions and roles, delegating to from_cawg_x509."

Expose SignerWrapper and
AsyncSignerWrapper as pub(crate) for use across the crate.

PR #2483, opened 2026-08-12
feat: Implement remote signing for CAWG X.509 identity assertions

BadrTad:

Implements remote (HTTP-based) signing for CAWG X.509 identity assertions, as requested in #1777, by filling in `SignerSettings::cawg_signer()`'s `Remote` arm (currently `todo!("Remote CAWG X.509 signing not yet supported")` on `main`).

This mirrors the existing plain c2pa-level `RemoteSigner`: a `RemoteRawSigner` implements `c2pa_raw_crypto::RawSigner` by POSTing the bytes-to-be-signed to the configured `url` and using the response body as the raw signature. `CawgX509IdentitySigner::from_remote_settings` wires this into the existing `CawgX509IdentitySigner`, reusing `cert_chain_pem_to_der` for certificate handling exactly as the `Local` case and the recently-fixed `RemoteSigner` (#2414) do.

Configuration is purely settings.toml-driven, e.g.:

```toml
[cawg_x509_signer.remote]
url = "https://example.com/sign-cawg-identity"
alg = "es256"
sign_cert = "..."
referenced_assertions = ["c2pa.actions"]
roles = ["creator"]
```

### Relationship to prior work

- This was previously implemented in #1865, which was closed with the note that "this has been implemented by @gpeacock recently" (referring to #2118). #2118 added CAWG signing via a **c_ffi callback**, letting host-language bindings (Python/Node/etc.) supply their own signer callback. That's a different mechanism serving a different use case (custom signing logic embedded in the host application) and doesn't cover the case this PR addresses: built-in, settings.toml-driven HTTP remote signing, matching how `signer.remote` already works for plain c2pa/claim signing. `SignerSettings::Remote::cawg_signer()` is still a `todo!()` on `main`, so the gap #1777 described remains open.
- Since #1865 was opened, the raw-signature layer was extracted into the external `c2pa_raw_crypto` crate (#2202/#2231), with a redesigned `RawSigner` trait (private-key-only; no cert chain, TSA, OCSP, or async variants — those are now SDK-side concerns). This PR is a fresh implementation against that current architecture rather than a rebase of #1865's commits, most of which targeted code paths that no longer exist in this repo.

### Testing

- Unit test verifying the settings-driven wiring (`test_make_cawg_remote_signer_from_settings`).
- A full round-trip test (`test_cawg_remote_signer_round_trip`): builds and signs a manifest with a CAWG identity assertion backed by a mocked remote endpoint (via `httpmock`, computing the mocked response dynamically from a reference local signer), reads the manifest back, and validates the identity assertion via `X509SignatureVerifier`.
- Verified locally against both crypto backends (`openssl` and `rust_native_crypto`), including the exact feature sets Tier 1A CI derives via `cargo metadata`, plus `cargo fmt`, `clippy -D warnings`, `cargo doc` with `RUSTDOCFLAGS=-Dwarnings`, doc-tests, `cargo check --locked`, and a `wasm32-unknown-unknown` compile check (including test-target compilation, to confirm the `httpmock`-based tests stay properly `#[cfg(not(target_arch = "wasm32"))]`-gated).

## Checklist

Opening as a draft to get an early signal on whether this approach (and its relationship to #2118) makes sense before finalizing.


gpeacock (approved):
I think this looks ok, but I'd like a second review on it to make sure.

BadrTad:
@gpeacock sure please let me know if any changes are required

scouten-adobe:
Thank you, @BadrTad! Incorporated into #2635 and merged.

PR #2387, opened 2026-07-31
chore: Count integration tests in Codecov coverage

scouten-adobe:
## Summary

The Tier 1A coverage-generation steps measured coverage with `cargo llvm-cov --lib`, which restricts to the library's own unit tests. Every test under `sdk/tests/` (integration tests) therefore contributed **nothing** to the Codecov number, understating coverage repo-wide.

This switches both coverage steps in `.github/workflows/tier-1a.yml` (the OpenSSL and the `rust_native_crypto` feature sets) from `--lib` to `--tests`, so all test targets — lib unit tests **plus** the integration tests under `sdk/tests/` — count toward coverage.

## Why

Surfaced while finishing #2357 (replacing the unmaintained `mp4` crate with a native BMFF sample reader): the BMFF timed-media Merkle verify path in `sdk/src/assertions/bmff_hash.rs` is well exercised by `sdk/tests/bmff_timed_media_merkle.rs`, yet it read as nearly uncovered on Codecov because integration tests were excluded from the measurement. That PR added lib-level unit tests as a stopgap; this is the real, repo-wide fix.

## Verification

Confirmed locally that the target selection is correct:

- `cargo test -p c2pa --lib ...` builds **1** test binary: `unittests src/lib.rs`.
- `cargo test -p c2pa --tests ...` builds **10** test binaries: the lib unit tests plus all 9 integration tests, including `tests/bmff_timed_media_merkle.rs`.

The `scouten/uncover-tests` filter step is retained for both variants, so test code is still stripped from the lcov and only product code is counted.

## Coverage baseline note

This is a one-time upward shift of the repo-wide coverage baseline (more tests now count). The shift is self-consistent: Codecov compares a PR head against its actual base commit, and an in-flight PR that merges main picks up both the higher baseline and the `--tests` workflow together. This PR itself only adds covered lines, so the `project` status rises and passes. Per maintainer decision, the `codecov.yml` gate (`target: auto`, `threshold: 1%`) is left unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2484, opened 2026-08-12
chore: Count integration tests in Codecov coverage (backport #2387)

scouten-adobe:
Manual backport of #2387 to `stable`.

The automated backport failed because `stable`'s `.github/workflows/tier-1a.yml` has diverged from `main` (it does not yet carry the `authorize` trust job or the reworked `get-features` step), so the cherry-pick hit context conflicts. The functional change itself is unchanged and applies cleanly: both coverage-generation steps switch from `--lib` to `--tests` so integration tests under `sdk/tests/` count toward Codecov, and the `scouten/uncover-tests` filter is retained for both variants.

Verified the resulting diff on `stable` is exactly the two intended line changes.

### DIFF
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -95,7 +95,7 @@ jobs:
           RUST_BACKTRACE: "1"
           FEATURES: ${{needs.get-features.outputs.openssl-features}}
         run: |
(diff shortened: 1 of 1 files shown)


scouten-adobe:
Closing. We'll be doing a new stable train soon, so no longer important.

PR #2352, opened 2026-07-23
ci: Trust same-repo PR branches instead of author_association

scouten-adobe:
## Problem

The CI trust gate in `tier-1a.yml` (and the AND-clause in `benchmarks.yml`) accepted only `author_association == 'MEMBER' || 'COLLABORATOR'` (plus dependabot and the `safe to test` / `check-release` labels).

GitHub only reports:
- **`MEMBER`** for **public** org members, and
- **`COLLABORATOR`** for **direct** repo collaborators.

A team member whose org membership is **private** and whose write access comes via a **team** is reported as **`CONTRIBUTOR`**. Their PRs had the *entire* CI suite skipped despite having write access — e.g. #2347 (@mpgopala), and the same would hit @ok-nick.

## Fix

Key trust off the **PR's head repository** instead of `author_association`. Pushing a branch to this repo already requires write access, so *"the branch lives in this repo"* is an accurate, maintenance-free proxy for *"the author is a trusted team member"* — and it works regardless of org-membership visibility. Fork PRs keep the `safe to test` / `check-release` label escape hatch.

### Changes
| File | Change |
|------|--------|
| `tier-1a.yml` | New `authorize` job outputs `trusted` (run the suite) and `same-repo` (repo secrets available, e.g. Codecov). Every gated job keys off it — replaces ~18 copies of the old 6-line condition. |
| `tier-1b.yml` | New `authorize` job combines trust with the existing release-line scope into one `run-suite` decision. |
| `tier-2.yml` | Trust folded into the existing inline scope condition (single job). |
| `benchmarks.yml` | Drop the now-redundant `author_association` inner clause; the `head.repo == repository` AND already gates it (benchmarks need AWS/S3/Codspeed secrets forks never receive). |
| `semver-checks.yml` | Gate on the same-repo rule (compiles PR code). |
| `check-no-patch-deps.yml` | Gate on the same-repo rule; `workflow_call` path is unaffected (non-PR event). |

### Deliberately unchanged
- **`upstream-first-check.yml`** — designed to *never skip* so its required check is always conclusive; runs only git/shell, no PR code execution.
- **`pr_title.yml`** — must run on fork PRs to give contributors title feedback; executes no repo code.
- **`backport.yml`** — `pull_request_target`, merged-only maintainer automation.
- **`beta-preflight.yml`** — not PR-triggered (dispatch/push only); its `author_association` reference is already commented out.

## Behavior after this change
[...]

PR #2485, opened 2026-08-12
ci: Trust same-repo PR branches instead of author_association (backport #2352)

scouten-adobe:
Manual backport of #2352 to `stable`.

The automated backport failed: `stable`'s workflow files had already diverged from `main` at the time of this PR (`tier-1a.yml` on `stable` still had the pre-#2300 combined `docs_rs` job, without the separate `docs_stable` job that `main` had already split out), so the cherry-pick hit a context conflict in `tier-1a.yml`.

Resolution: kept `stable`'s existing single combined `docs_rs` job structure (no `docs_stable` job — that split belongs to #2300, a separate not-yet-backported change) and applied only this PR's actual change to it: replacing the old `author_association`-based `if:` condition with the new `authorize` job + `needs.authorize.outputs.trusted` pattern, consistent with every other job in the file.

The other five workflow files this PR touched (`benchmarks.yml`, `check-no-patch-deps.yml`, `semver-checks.yml`, `tier-1b.yml`, `tier-2.yml`) cherry-picked without conflict.

Verified:
- No leftover conflict markers anywhere in `.github/workflows/`.
- All six changed workflow files parse as valid YAML.
- No stale `author_association` references remain in `tier-1a.yml` outside of the explanatory comment in the new `authorize` job.

Upstream-first: this change already merged to `main`.

### DIFF
--- a/.github/workflows/tier-1a.yml
+++ b/.github/workflows/tier-1a.yml
@@ -34,6 +34,40 @@ on:
       - '*-rc*'
 
 jobs:
+  # Decide once whether this run is trusted, and expose the result to every
+  # other job. A run is trusted when it is not a pull request (push, schedule,
+  # or manual dispatch), when the PR branch lives in this repository, or when a
+  # maintainer has applied the `safe to test` or `check-release` label to a
+  # fork PR.
+  #
+  # We key trust off the PR's head repository rather than `author_association`
+  # because GitHub only reports `MEMBER`/`COLLABORATOR` for public org members
+  # and direct collaborators. A private org member with team-granted write
+  # access shows up as `CONTRIBUTOR` and would otherwise be denied CI even
+  # though they can push branches here. Pushing a branch to this repository
+  # already requires write access, so "the branch is in this repo" is an
+  # accurate, maintenance-free proxy for "the author is a trusted team member".
+  authorize:
+    name: Authorize CI run
+    runs-on: ubuntu-latest
+    outputs:
+      # Full trust: run the entire CI suite.
+      trusted: ${{ steps.decide.outputs.trusted }}
+
+      # Run originates from this repository (not a fork), so repository secrets
+      # (e.g. the Codecov token) are available. Fork PRs never receive secrets,
+      # even when labeled `safe to test`, so secret-dependent steps must gate on
+      # this rather than on `trusted`.
+      same-repo: ${{ steps.decide.outputs.same-repo }}
+    steps:
+      - name: Determine trust level
+        id: decide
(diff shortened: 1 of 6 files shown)


.github/workflows/tier-1a.yml
```diff
@@ -34,6 +34,40 @@ on:
...
+    runs-on: ubuntu-latest
+    outputs:
+      # Full trust: run the entire CI suite.
+      trusted: ${{ steps.decide.outputs.trusted }}
+
+      # Run originates from this repository (not a fork), so repository secrets
+      # (e.g. the Codecov token) are available. Fork PRs never receive secrets,
+      # even when labeled `safe to test`, so secret-dependent steps must gate on
+      # this rather than on `trusted`.
+      same-repo: ${{ steps.decide.outputs.same-repo }}
```
ok-nick, line 61: Doesn't GitHub do this by default and if we manually approve workflows to run on forks it will. Do these gates prevent us from running the normal test suite on external PRs? If we added `pull_request_target` (always allow workflow to run on forks), then we would need extra layers of security?

.github/workflows/tier-1a.yml
```diff
@@ -34,6 +34,40 @@ on:
...
+  #
+  # We key trust off the PR's head repository rather than `author_association`
+  # because GitHub only reports `MEMBER`/`COLLABORATOR` for public org members
+  # and direct collaborators. A private org member with team-granted write
+  # access shows up as `CONTRIBUTOR` and would otherwise be denied CI even
+  # though they can push branches here. Pushing a branch to this repository
+  # already requires write access, so "the branch is in this repo" is an
+  # accurate, maintenance-free proxy for "the author is a trusted team member".
+  authorize:
+    name: Authorize CI run
```
dcondrey, line 51: Would it be worth putting `always()` on the dependents of this job?

14 of the 16 required checks on `stable` are now downstream of it, including five that had no `needs:` at all before this change (`docs_rs`, `cargo-check`, `cargo-lock`, `cargo_fmt`, `cargo-deny`). `Authorize CI run` isn't itself in the required-checks set, and per GitHub's docs a job skipped because a dependency failed "may not block merging" — their suggested remediation is exactly `always()` with `needs`.

Odds are low since this job is a single `echo` with no checkout, so it'd take a runner-allocation failure. But if it did fail, `Get features` and `cargo-semver-checks` would be the only required checks that actually ran, and this is the release line.

scouten-adobe:
Closing this since it's a backport and we're close to the next major-changes release.

PR #2488, opened 2026-08-12
fix: Reject BMFF sample offsets that point into unhashed exclusion boxes

ssanthosh:
`c2pa.hash.bmff.v3`'s default exclusion list (`/uuid`, `/ftyp`, `/mfra`, `/free`, `/skip`) unconditionally excludes `/free` and `/skip` box content from the hash, but nothing validated that `moov`'s `stco`/`co64` sample-chunk-offset tables (which *are* fully hashed) don't point into one of those excluded, unhashed regions.

## The fix

- **`bmff_io.rs`**: `bmff_to_jumbf_exclusions` (shared by both hash generation and verification) now tracks which exclusion ranges have no separate integrity coverage — everything except `/mdat`, which is independently checked via a Merkle tree when Merkle hashing is used. After building the exclusion list, it parses every `stco`/`co64` chunk-offset table under `moov` and rejects the file (`Error::InvalidAsset`) if any chunk offset lands inside one of those unverified ranges. This closes the gap for both signing (c2pa-rs now refuses to hash/sign such a file) and verification (a third-party-signed file exhibiting this pattern is rejected).
- Reused the existing `bounded_entry_count` helper to bound the new `stco`/`co64` parsing against a corrupted/oversized `entry_count`, matching the codebase's existing defense for the same field elsewhere (`parse_stco`/`parse_co64`).
- **`utils/test.rs`**: `write_bmff_placeholder_stream` spliced a C2PA placeholder box into a stream without adjusting `moov`'s existing `stco`/`co64` entries, leaving them pointing at stale (pre-splice) offsets. This was silently tolerated before (nothing checked chunk-offset correctness) but is now caught by the new validation. Added `shift_bmff_chunk_offsets` to correctly shift those offsets by the inserted placeholder length, matching what real BMFF embedding already does via `adjust_known_offsets`.

## Tests added

- `stco_pointing_into_free_box_is_rejected` (`bmff_io.rs`) — a minimal single-track MP4 with its one `stco` entry pointing into a `/free` box's payload must be rejected by `bmff_to_jumbf_exclusions`.
- `stco_pointing_into_mdat_is_accepted` (`bmff_io.rs`) — the same file with the entry pointing at `mdat` (the legitimate case) must still pass, proving the new check doesn't over-block ordinary assets.

## Checklist

### DIFF
--- a/sdk/src/asset_handlers/bmff_io.rs
+++ b/sdk/src/asset_handlers/bmff_io.rs
@@ -759,6 +759,11 @@ where
 
     let mut exclusions = Vec::new();
 
+    // Exclusion ranges with no other integrity coverage - i.e. everything
+    // except `/mdat`, which is separately checked via a Merkle tree when
+    // Merkle hashing is used. Sample offset tables must never point here.
+    let mut unverified_exclusions = Vec::new();
+
     for bmff_exclusion in bmff_exclusions {
         if let Ok(Some(box_token_list)) = bmff_tree.fetch(&bmff_exclusion.xpath) {
             for box_token in box_token_list {
@@ -859,12 +864,18 @@ where
 
                         let exclusion = HashRange::new(new_start, new_length);
 
+                        if bmff_exclusion.xpath != "/mdat" {
+                            unverified_exclusions.push(exclusion.clone());
+                        }
                         exclusions.push(exclusion);
                     }
                 } else {
                     // exclude box in its entirty
                     let exclusion = HashRange::new(exclusion_start, exclusion_length);
 
+                    if bmff_exclusion.xpath != "/mdat" {
+                        unverified_exclusions.push(exclusion.clone());
+                    }
                     exclusions.push(exclusion);
 
                     // for BMFF V2 hashes we do not add hash offsets for top level boxes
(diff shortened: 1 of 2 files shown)


sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -891,6 +904,71 @@ where
...
+        return Ok(());
+    }
+
+    let in_unverified_exclusion = |offset: u64| {
+        unverified_exclusions
+            .iter()
+            .any(|r| offset >= r.start() && offset < r.start() + r.length())
+    };
+
+    for xpath in [
```
tmathern, line 931: Is the hard coded list here complete for all BMFF-like formats?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -859,12 +864,18 @@ where
 
                         let exclusion = HashRange::new(new_start, new_length);
 
+                        if bmff_exclusion.xpath != "/mdat" {
```
tmathern, line 867: Suggestion: Instead of hardcoding and if-check, list what is covered by the Merkle tree check, and check if that case if hit here.
(Also consider making it a helper, since this is rechecked at multiple locations).

ssanthosh:
Closing this PR as we are rejecting this vulnerability as not a valid vulnerability

PR #2459, opened 2026-08-10
fix: Propagate created flag to metadata assertions

tmathern:
## Changes in this pull request
Propagate created flag to metadata assertions (propagates the value, does not set it or decides what to set: what goes in, gets set).

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #2489, opened 2026-08-12
fix: Propagate created flag to metadata assertions (backport #2459)

caiopensrc:
Automated backport of #2459 to `stable`.

Upstream-first: this change already merged to `main`. If the
cherry-pick did not apply cleanly, adapt it so it compiles and
passes tests on this branch.


tmathern:
Will be in next release train, not backported to v0.90.*

PR #2491, opened 2026-08-13
feat: Clean up asset_io, rename things, share common code, and add private registry through `Context`

gpeacock:
Opens up the AssetIO handler system for external customization and cleans up its internals along the way.

- Custom AssetIO handlers can now be registered via `Context::with_io_handler`
- `asset_io` trait hierarchy is now public and documented
- dead methods (`read_cai_store`, `get_object_locations`, `RemoteRefEmbed::embed_reference`) removed and boilerplate (`save_cai_store`/`remove_cai_store`) given default impls
- New `WriteXmp` trait factors out the duplicated read-XMP/merge-provenance/write-container logic shared by nearly every format handler's remote-manifest-URL support
- `RemoteRefEmbed` renamed to `RemoteManifestUrl` (and `embed_reference_to_stream` → `write_remote_manifest_url`) to match spec terminology; stream params standardized to `input_stream`/`output_stream` across all handlers
- `RemoteManifestUrl` gains `read_manifest_url` so custom handlers can control read-back, not just write
- `AssetBoxHash` gets its own minimal `BoxMap` type, decoupled from the `c2pa.hash.boxes` assertion's serializable `BoxMap`


c2pa_c_ffi/src/c_api.rs
```diff
@@ -2500,6 +2500,9 @@ pub unsafe extern "C" fn c2pa_builder_update_hash_from_stream(
 /// The returned value MUST be released by calling c2pa_free
 /// and it is no longer valid after that call.
 #[no_mangle]
+#[deprecated(
+    note = "Use c2pa_builder_composed_manifest() instead, so custom asset I/O handlers registered on the builder's Context are consulted."
```
scouten-adobe: Before this is merged, please add the proposed removal date (likely mid-November, assuming this lands before the September release).

sdk/src/asset_io.rs
```diff
@@ -11,21 +11,139 @@
...
+//! Traits for implementing a custom C2PA I/O handler for a file format.
+//!
+//! The SDK ships with built-in handlers for common formats (JPEG, PNG, TIFF, MP4,
+//! WAV, GIF, ...). This module is what you implement to add support for a format
+//! the SDK doesn't natively handle, and to register it via
+//! [`Context::with_io_handler`](crate::Context::with_io_handler).
+//!
+//! There are three mandatory traits:
+//!
+//! - [`CAIReader`] — read the C2PA manifest store and XMP from a stream.
```
scouten-adobe: If we're requiring read + write implementations for any supported family, I don't see the value in separate reader and writer traits. Why not unify on a single `AssetIO` trait?

scouten-adobe: (Yes, I know that's how we've done it historically. Given that this will be a public API surface, I think we should take the opportunity to rethink.)

sdk/src/asset_io.rs
```diff
@@ -11,21 +11,139 @@
...
 
+/// The kind of region a [`HashObjectPositions`] entry describes.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum HashBlockObjectType {
+    /// The C2PA manifest store.
     Cai,
+    /// An XMP metadata block.
     Xmp,
+    /// Any other region that should be hashed.
     Other,
```
scouten-adobe, line 147: What is the difference between `Other` and `OtherExclusion`? Could we instead have one exclusion type and maybe add a `String` or `&'static str` reason field for the exclusion?

sdk/src/asset_io.rs
```diff
@@ -34,13 +152,27 @@ impl fmt::Display for HashBlockObjectType {
...
+    pub htype: HashBlockObjectType,
 }
 
+/// Marker trait for a seekable, readable stream that can be sent across threads.
+///
+/// Blanket-implemented for any type that is `Read + Seek + Send` (or `Read + Seek`
+/// on targets without threading). Exists so it can be used in `dyn` position
+/// (`&mut dyn CAIRead`) — Rust trait objects can only name one non-auto trait, so
+/// this trait exists to bundle `Read` + `Seek` into one.
 pub trait CAIRead: Read + Seek + MaybeSend {}
```
scouten-adobe: I think the distinction between `CAIRead` (raw stream) and `CAIReader` (file format) is potentially confusing. Could we choose a different name for one or both of these traits? Same for write.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
 pub trait CAIReader: Sync + Send {
-    // Return entire CAI block as Vec<u8>
-    fn read_cai(&self, asset_reader: &mut dyn CAIRead) -> Result<Vec<u8>>;
+    /// Returns the raw C2PA JUMBF manifest store bytes from `input_stream`.
+    ///
+    /// Returns [`Error::JumbfNotFound`] if no manifest
+    /// store is present, or
+    /// [`Error::TooManyManifestStores`] if
+    /// more than one is detected.
+    fn read_cai(&self, input_stream: &mut dyn CAIRead) -> Result<Vec<u8>>;
```
scouten-adobe: I'd like to see this renamed to `read_c2pa_manifest_store` or something similar. `cai` was once an appropriate name, but I think we should remove that historical legacy. Same for write.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+        match (hinted, detected) {
+            (Some(h), Some(d)) if h == d => hint.to_string(),
+            (_, Some(d)) => d.to_string(),
+            (_, None) => hint.to_string(),
+        }
+    }
+
+    /// Reads the C2PA manifest store from `stream` via the registered reader for
+    /// `asset_type`, checking `fallback` if this registry has no match.
+    pub(crate) fn read_jumbf(&self, asset_type: &str, stream: &mut dyn CAIRead) -> Result<Vec<u8>> {
```
scouten-adobe: I realize this isn't part of the proposed public API surface, but if the purpose of this function is to read the _manifest store,_ then I think `read_jumbf` is subtly misleading. I was expecting to find more about understanding the JUMBF format than a block of bytes.

c2pa_c_ffi/src/c_api.rs
scouten-adobe: Applies to all deprecations introduced in this PR.

sdk/src/builder.rs
```diff
@@ -3523,8 +3549,36 @@ impl Builder {
...
+    /// consulted instead of only the built-in global registry.
+    ///
+    /// # Arguments
+    /// * `manifest_bytes` - The bytes of the manifest to convert.
+    /// * `format` - The format to convert to.
+    /// # Returns
+    /// * The bytes of the composed manifest.
+    /// # Errors
+    /// * Returns an [`Error`] if the manifest cannot be converted.
+    pub fn composed_manifest_with_context(
```
scouten-adobe: Hrm. Sorry to keep hanging up on naming, but I'd have expected to find a `&Context` parameter given the name `…_with_context`. Not sure what I'd suggest as an alternative. Happy to have a brainstorming session if that helps.

scouten-adobe (changes requested):
Overall, very happy with where you're going. Most of my comments are a nudge to think about naming as we do this refactoring.

scouten-adobe:
Some additional items flagged when I asked for an AI-assisted code review:

1. **RIFF/WebP bug** — `HandlerRegistry` now shares a single `RiffIO` instance across AVI/WAV/WebP, built with `riff_format = "avi"`. Signing a `.webp` with a remote manifest URL silently skips the WebP-specific VP8X chunk update. No existing test catches this.
2. **XMP-based lookups bypass `Context`** — `XmpInfo::from_source` (used by `Store::load_jumbf_from_stream`'s fallback, `Builder::sign`'s instance-id extraction, and `Ingredient::from_file_info`) goes through the legacy `CAI_READERS` map, which never sees handlers registered via `Context::with_io_handler`. Affects 3 public APIs.
3. **JPEG/PNG permission narrowing** — losing their bespoke `remove_cai_store`, these now fall through to a temp-file+rename default that silently drops the file to `0600`, discarding the original's group/world permissions.
4. **`format_to_mime` bypasses `Context`** at 3 call sites (`builder.rs:3206`, `builder.rs:1726`, `ingredient.rs:1292/1316`) even though a `Context`/`self` is already in scope — custom formats get a raw string instead of a real MIME type in manifest assertions.
5. `is_bmff_format` hardcodes a comparison against the built-in `"avif"` handler — no way for a third-party BMFF-family handler to opt into BMFF-style hashing.
6. `sniff_container_from_stream` has no extension point for custom handlers to contribute magic-byte detection (mitigated today since all in-tree callers pass explicit hints).
7. Legacy `CAI_READERS`/`CAI_WRITERS` maps in `jumbf_io.rs` still exist in parallel with the new `HandlerRegistry`, independently built from the same prototype list — the root cause of #2.
8. `remove_xmp_key` duplicates `add_xmp_key`'s trim/padding math instead of sharing it.
9. `mp3_io.rs`/`flac_io.rs` `write_xmp` bodies are now byte-identical (mostly pre-existing debt, made exact by this PR).

sdk/src/asset_io.rs
tmathern: Why `c2pa_manifest_store` and not just `manifest_store` as suffix?

sdk/src/jumbf_io.rs
```diff
@@ -78,158 +71,26 @@ lazy_static! {
...
+    static ref DEFAULT_HANDLER_REGISTRY: Arc<HandlerRegistry> = {
+        let mut reg = HandlerRegistry::new();
         for h in HANDLER_PROTOTYPES.iter() {
-            let types = h.supported_types();
-            if let Some(&container_id) = types.first() {
-                for t in types {
-                    map.insert(t.to_string(), container_id);
-                }
-            }
+            // One instance stands in for the whole family, matching how a custom handler
```
tmathern, line 81: But what when different asset types have the same handler family? Are they always compatible? E.g. avi and webp are riff. COuld there be races/collisions on lookup?

For instance, are check based on "exact" types still handled properly?

For isntance, checks like this:
https://github.com/contentauth/c2pa-rs/blob/f6d40b2b584b00545280fe69e388aecb558d879e/sdk/src/asset_handlers/riff_io.rs#L180

Maybe this warrants, per "family" of handlers, the ability to subtype them too? (Exactly to have such overrides).

tmathern: ps: BMFF in the future may have similar quirks (thinking of mov which is sometimes a tad special).

c2pa_c_ffi/src/c_api.rs
tmathern: Will we ship a version where the IO handler are automatically registered still, or will the registration lie on the caller already? Is this a phased deprecation rollout?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -14,9 +14,8 @@
 use std::{
     cmp::min,
     collections::HashMap,
-    fs::{File, OpenOptions},
-    io::{Cursor, Read, Seek, SeekFrom, Write},
```
tmathern, line 18: interesting that Cursor is gone!

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
         output_stream: &mut dyn CAIReadWrite,
-        embed_ref: RemoteRefEmbedType,
+        xmp: &str,
     ) -> Result<()>;
 }
 
-/// `ComposedManifestRefEmbed` is used to generate a C2PA manifest.  The
-/// returned `Vec<u8>` contains data preformatted to be directly compatible
-/// with the type specified in `format`.
+impl<T: WriteXmp + CAIReader> RemoteManifestUrl for T {
```
tmathern: If you have this, then you can't reimplement `RemoteManifestUrl`, correct (this is a public trait)? So a caller has to pick one and stick with it?https://github.com/contentauth/c2pa-rs/blob/f6d40b2b584b00545280fe69e388aecb558d879e/sdk/src/asset_io.rs#L575

tmathern: ps: I would maybe just expose functions with that logic, without binding them together here in this impl?

sdk/src/asset_io.rs
```diff
@@ -11,21 +11,139 @@
...
+    collections::HashMap,
+    ffi::OsStr,
     fmt, fs,
     io::{Cursor, Read, Seek, Write},
-    path::Path,
+    path::{Path, PathBuf},
+    sync::Arc,
 };
 
-use tempfile::NamedTempFile;
```
tmathern, line 20: make sure that isn't used by any leftover pub function, or it may cause trouble to callers.
Like this:
pub fn rename_or_move<P>(temp_file: NamedTempFile, asset_path: P) -> Result<()>
(Just outside the visible changeset).

(That note applies to any removed import, to make sure it isn't used, even as fully qualified name).

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+        }
+    }
+
+    /// Returns the MIME type for `format` (an extension or MIME type), derived from a
+    /// registered handler (via [`Self::mime_for`]) so a custom handler's format is
+    /// recognized. Returns `format` itself (trimmed and lowercased) when no registered
+    /// handler recognizes it.
+    pub fn format_to_mime(&self, format: &str) -> String {
+        match self.mime_for(format) {
+            Some(mime) => mime.to_string(),
```
tmathern, line 886: Is mime already case insensitive? If this is caller-registered, we should make sure casing is ignored.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+    ///
+    /// The default implementation derives pairs from [`supported_types`](Self::supported_types):
+    /// every extension-shaped entry (no `/`) is paired with the first MIME-shaped entry (the
+    /// first one containing `/`). That's correct when every entry refers to the same
+    /// underlying format (e.g. `["jpg", "jpeg", "image/jpeg"]`), which covers most handlers.
+    ///
+    /// Override this when `supported_types()` covers multiple distinct sub-formats that each
+    /// need their own MIME type (e.g. BMFF's `avif`/`heic`/`heif`/`mp4`/`mov`, all handled by
+    /// one `AssetIO` impl but each needing a different MIME type) — the default heuristic
+    /// can't tell those apart from a single flat list.
```
tmathern, line 333: Maybe then we should ask callers to also provide their maps?

sdk/src/asset_handlers/svg_io.rs
```diff
@@ -13,37 +13,22 @@
...
+    crypto::base64,
     error::{Error, Result},
-    utils::{
-        io_utils::{patch_stream, stream_len, tempfile_builder, ReaderUtils},
-        xmp_inmemory_utils::{self, MIN_XMP},
-    },
+    utils::io_utils::{patch_stream, stream_len, ReaderUtils},
 };
 
 static SUPPORTED_TYPES: [&str; 3] = ["svg", "application/svg+xml", "image/svg+xml"];
```
tmathern, line 34: How does this play with the new supported maps here:
https://github.com/contentauth/c2pa-rs/blob/f6d40b2b584b00545280fe69e388aecb558d879e/sdk/src/asset_io.rs#L404-L408

Since it seems this is a multiple mimetype/format case?

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+            handlers: HashMap::new(),
+            containers: HashMap::new(),
+            mimes: HashMap::new(),
+            fallback: None,
+        }
+    }
+
+    /// Creates an empty registry that consults `fallback` for any format it has no handler
+    /// for itself.
+    pub fn with_fallback(fallback: Arc<HandlerRegistry>) -> Self {
```
tmathern, line 713: Could we expose our default as fallback? Closes potentially the loop, and gives other implementers something to start from for their own implementations.

sdk/src/asset_io.rs
```diff
@@ -34,13 +152,27 @@ impl fmt::Display for HashBlockObjectType {
...
+///
+/// Returned by [`CAIWriter::get_object_locations_from_stream`] so the hashing system
+/// knows what to hash and what to exclude.
 #[derive(Debug, PartialEq)]
 pub struct HashObjectPositions {
-    pub offset: usize, // offset from beginning of file to the beginning of object
-    pub length: usize, // length of object
-    pub htype: HashBlockObjectType, // type of hash block object
+    /// Offset from the beginning of the asset to the beginning of the region.
+    pub offset: usize,
```
tmathern: I am always hesitant to use any type of the usize family because they depend on the underlying arch. Is it safe here?
(Same question for the length value).

tmathern: Bonus note: BoxMap also uses u64 (and not usize) too for the ranges...

https://github.com/contentauth/c2pa-rs/blob/f6d40b2b584b00545280fe69e388aecb558d879e/sdk/src/asset_io.rs#L523-L526

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+pub const C2PA_BOXHASH: &str = "C2PA";
+
+/// Describes one hashable region ("box") in an asset's container format.
+///
+/// Returned by [`AssetBoxHash::get_box_map`] to describe the byte range and name
+/// of each box, in the order they occur. This only describes *where* the boxes
+/// are — computing and recording hashes over these regions is the job of the
+/// `c2pa.hash.boxes` assertion, not this type.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct BoxMap {
```
tmathern, line 441: Maybe call it `HashableRegion` then?
Because of the start of the comment: `Describes one hashable region...`

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+            None => self
+                .fallback
+                .as_deref()
+                .and_then(|f| f.container_for(format)),
+        }
+    }
+
+    /// Returns whether `format` belongs to the same container family as `"avif"` (i.e. the
+    /// ISO BMFF family: MP4, MOV, HEIC, HEIF, AVIF, ...).
+    pub fn is_bmff_format(&self, format: &str) -> bool {
```
tmathern, line 855: Can't this be in the dedicated handler? I know BMFF is special, but why have the specialty be here too, in the generic asset handler?

tmathern: (This might not be solvable nicely, depending in which direction you want the calls/dependencies to point here).

tmathern: (Also why avif, what about the other formats? Is it OK to blanket them all this way, they don't have special format quirks (This could warrant subtyping or "sub-trait-ing"?)

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+    }
+
+    /// Like [`add_handler`](Self::add_handler), but takes an already-boxed handler. Useful
+    /// when the caller only has a `Box<dyn AssetIO>` (e.g. from another
+    /// [`AssetIO::get_handler`] call) rather than a concrete, sized type.
+    pub(crate) fn add_boxed_handler(&mut self, handler: Box<dyn AssetIO>) {
+        let handler: Arc<dyn AssetIO> = Arc::from(handler);
+        let types = handler.supported_types();
+        let Some(canonical) = types.first() else {
+            return;
```
tmathern, line 741: At some point, we should think about having debug logging or something. Because going through here will be a hard to debug issue with handlers, since it just says nothing and ignores the call.
(This also means, I think, we can't override a registered handler "later", or reconfigure a handler).

sdk/src/reader.rs
```diff
@@ -622,7 +625,7 @@ impl Reader {
 
     /// Returns a [Vec] of mime types that [c2pa-rs] is able to read.
     pub fn supported_mime_types() -> Vec<String> {
-        jumbf_io::supported_reader_mime_types()
+        Context::default().io().reader_mime_types()
```
tmathern, line 628: That's only the default handlers? What if other were registered with the new APIs?

sdk/src/builder.rs
```diff
@@ -685,7 +684,7 @@ impl Builder {
 
     /// Returns a [Vec] of MIME types that the API is able to sign.
     pub fn supported_mime_types() -> Vec<String> {
-        jumbf_io::supported_builder_mime_types()
+        Context::default().io().writer_mime_types()
```
tmathern, line 687: That's only the default handlers? What if other were registered with the new APIs?

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+    pub fn format_from_path<P: AsRef<Path>>(&self, path: P) -> Option<String> {
+        let ext = path.as_ref().extension()?.to_string_lossy();
+        Some(self.format_to_mime(&ext))
+    }
+
+    /// Resolves the format string to use for reading by combining a caller-supplied hint
+    /// with stream-based container detection.
+    ///
+    /// * If the hint maps to the same container family as the detected bytes, the hint is
+    ///   returned unchanged (it may be more specific, e.g. `"dng"` within the TIFF family).
```
tmathern, line 903: (DNG is another of those formats where I could see it sub-trait a generic tiff handler).

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+/// MIME type, e.g. `"jpg"` or `"image/jpeg"`).
+///
+/// Handlers are matched by exact format string, case-insensitively and with surrounding
+/// whitespace trimmed. Registering a handler for a format string already claimed by an
+/// earlier one overrides it ("last-registered wins").
+///
+/// An optional `fallback` registry is consulted only when this registry has no match for a
+/// format. This lets a small registry (e.g. a caller's custom handlers) cheaply extend a
+/// larger one (e.g. the SDK's built-in handlers) without copying it.
+pub struct HandlerRegistry {
```
tmathern, line 687: Add a Debug derivation to log contents maybe? It could come in handy for debugging.

sdk/src/store.rs
```diff
@@ -2765,9 +2775,9 @@ impl Store {
         // add a Merkle tree map for each init segment and its associated fragments
         for (i, init_path) in init_paths.iter().enumerate() {
             // make sure it is a supported BMFF format
-            match get_supported_file_extension(init_path.as_ref()) {
+            match context.io().supported_extension(init_path.as_ref()) {
                 Some(ext) => {
-                    if !is_bmff_format(&ext) {
+                    if !context.io().is_bmff_format(&ext) {
```
tmathern, line 2810: That explains why `is_bmff_format` is on asset_io.rs, but I'm ambivalent to it... Unfortunately I do not know where else to stick it either.

sdk/src/asset_io.rs
tmathern: Maybe this relates to changes like this one: https://github.com/contentauth/c2pa-rs/pull/2511 ?

tmathern (commented):
This is a nice refactoring! And a good abstraction as well to pull things out.

sdk/src/asset_handlers/README.md
```diff
@@ -35,7 +38,7 @@ At startup, each handler is instantiated, asked for its `supported_types()` (e.g
 | `save_jumbf_to_stream()` | Write JUMBF manifest to a stream |
 | `save_jumbf_to_memory()` | Write JUMBF manifest to in-memory bytes |
 | `save_jumbf_to_file()` | Write JUMBF manifest to a file path |
-| `remove_jumbf_from_file()` | Remove C2PA manifest from a file |
+| `remove_jumbf_from_file()` | Remove C2PA manifest from a file. **Deprecated** — unused within the SDK; slated for removal. |
```
ok-nick, line 41: If someone wants to write a remote manifest but the asset already contains an embedded manifest, can they still remove the embedded and write the remote?

sdk/src/asset_handlers/README.md
```diff
@@ -163,23 +179,58 @@ pub trait AssetPatch {
...
+**Don't implement `RemoteManifestUrl` directly.** Since writing a remote manifest URL is
+always "merge it into XMP, then write XMP", implement `WriteXmp` instead —
+`RemoteManifestUrl` has a blanket impl for any type that implements `WriteXmp` (and
+`CAIReader`, for reading the current XMP):
 
 ```rust
-pub trait RemoteRefEmbed {
-    fn embed_reference(&self, asset_path: &Path, embed_ref: RemoteRefEmbedType) -> Result<()>;
-    fn embed_reference_to_stream(
+pub trait WriteXmp {
```
ok-nick, line 215: I think `write_remote_manifest_url` should be a method of `WriteXmp`, which by default calls `write_xmp` with the remote manifest URL string. Writing a remote manifest URL is just sugar for writing XMP after all.

sdk/src/asset_handlers/README.md
```diff
@@ -163,23 +179,58 @@ pub trait AssetPatch {
...
-        embed_ref: RemoteRefEmbedType,
+        xmp: &str,
     ) -> Result<()>;
 }
 ```
 
-It embeds a remote manifest reference URL into the asset's XMP metadata. The `RemoteRefEmbedType` enum supports `Xmp`, `StegoS`, `StegoB`, and `Watermark` variants, though most handlers only implement `Xmp`.
+The blanket impl reads the current XMP via `CAIReader::read_xmp` (falling back to
+`MIN_XMP` if there is none), merges in the URL with `xmp_inmemory_utils::add_provenance`,
+and calls `write_xmp` with the finished string — every built-in handler that supports
```
ok-nick: Why would `CAIReader::read_xmp` return `MIN_XMP` rather than an option?

sdk/src/asset_handlers/riff_io.rs
```diff
@@ -393,39 +387,30 @@ impl AssetIO for RiffIO {
...
+    fn mime_type_map(&self) -> Vec<(String, String)> {
+        [
+            ("webp", "image/webp"),
+            ("wav", "audio/wav"),
+            ("avi", "video/avi"),
+        ]
+        .into_iter()
+        .map(|(ext, mime)| (ext.to_string(), mime.to_string()))
+        .collect()
+    }
```
ok-nick, line 398: Do we need a function for this? Can we just have an associated const defined on the trait (or struct) that returns a list of formats and mime types?

sdk/src/asset_io.rs
ok-nick: I actually prefer the separate reader/writer traits here, I'd say it's pretty conventional for this type of stuff in Rust. For example, the image crate, serde and all it's impls, csv, zip, quick-xml, and so on. There may be an implementor implementing their own format which supports only reading or only writing, but not both.

sdk/src/asset_io.rs
```diff
@@ -11,21 +11,139 @@
...
+//!     }
+//!
+//!     fn get_writer(&self, _asset_type: &str) -> Option<Box<dyn CAIWriter>> {
+//!         Some(Box::new(MyFormatReader))
+//!     }
+//!
+//!     fn supported_types(&self) -> &[&str] {
+//!         &["myformat", "application/x-myformat"]
+//!     }
+//! }
```
ok-nick, line 109: If it's in scope of this PR, I'd actually prefer we remove the `AssetIO` trait completely. I don't think we need a trait to get other traits. If we need reading and writing, we require `CAIReader + CAIWriter`. If we need XMP write, we add `+ XmpWrite` and so on where we need it.

sdk/src/asset_io.rs
```diff
@@ -11,21 +11,139 @@
...
+    utils::{
+        mime::normalize_format,
+        xmp_inmemory_utils::{add_provenance, extract_provenance, remove_provenance, MIN_XMP},
+    },
+    Error,
+};
 
+/// The kind of region a [`HashObjectPositions`] entry describes.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum HashBlockObjectType {
```
ok-nick: Something to consider, we can have methods such as `get_c2pa_range -> Result<Option<Range>>` and `get_xmp_range -> Result<Option<Range>>` implemented on the format handler itself rather than the current method which returns a list of potentially all, potentially some, or potentially none object types.

sdk/src/asset_io.rs
ok-nick: Agreed, we should use u64.

sdk/src/asset_io.rs
```diff
@@ -55,7 +187,9 @@ impl From<String> for Box<dyn CAIRead> {
 // that is required.  For example a function defined like this
 //  pub fn read<T>(&self, reader: &mut T) cannot currently accept
 // a CAIRead trait because it is not Sized (bound to a object).
-// This will likely change in a future version of Rust.
+// Needed because some third-party crates (e.g. `riff`, `id3`) expose generic
+// `fn read<T: Read + Seek>(reader: T)` APIs with an implicit `Sized` bound, so a
+// bare `&mut dyn CAIRead` can't be passed directly.
 pub(crate) struct CAIReadWrapper<'a> {
```
ok-nick: Alternatively we can add a blanket impl: `impl<T: CAIRead> Read for T` (same for seek) and remove this struct.

sdk/src/asset_io.rs
```diff
@@ -72,16 +206,18 @@ impl Seek for CAIReadWrapper<'_> {
...
 // Helper struct to create a concrete type for CAIReadWrite when
-// that is required. For example a function defined like this
-//  pub fn write<T>(&self, writer: &mut T) cannot currently accept
-// a CAIReadWrite trait because it is not Sized (bound to a object).
-// This will likely change in a future version of Rust.
-// go away in future revisions of Rust.
+// that is required. See [`CAIReadWrapper`] for why this is needed.
 pub(crate) struct CAIReadWriteWrapper<'a> {
     pub reader_writer: &'a mut dyn CAIReadWrite,
 }
```
ok-nick: Same here.

sdk/src/asset_io.rs
ok-nick: Or just `read_c2pa` 🙂, synonymous with `read_xmp`.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
-    #[allow(dead_code)]
-    fn read_cai_store(&self, asset_path: &Path) -> Result<Vec<u8>>;
+    // -- File-based operations --
+
+    /// Reads the C2PA manifest store from the file at `asset_path`.
+    ///
+    /// The default implementation opens `asset_path` and delegates to
+    /// [`get_reader`](AssetIO::get_reader)'s [`CAIReader::read_cai`].
+    fn read_cai_store(&self, asset_path: &Path) -> Result<Vec<u8>> {
+        let mut input_stream = fs::OpenOptions::new()
```
ok-nick, line 260: Do we need file-based operations at this level? Can the files be converted to streams at the top layer of the code (ideally on the builder/reader level)? Not sure if that's in scope of this PR.

sdk/src/asset_io.rs
ok-nick: This name is taken from the CDDL definition in the spec.  This struct is actually just a copy of the `BoxMap` struct already defined in the `assertions/box_hash.rs` code. We should remove this and use the one there.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+
+    /// Reads the remote manifest URL from the asset in `input_stream`, if any.
+    ///
+    /// The default implementation returns `None`. The blanket impl for
+    /// [`WriteXmp`] overrides this to extract the URL from XMP, since that's
+    /// where `write_remote_manifest_url` puts it. Implement this directly if
+    /// your format stores the remote manifest URL somewhere other than XMP.
+    fn read_manifest_url(&self, _input_stream: &mut dyn CAIRead) -> Option<String> {
+        None
+    }
```
ok-nick, line 595: Can this return a `Result<Option<String>>`? Since reading the stream will produce errors.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+    /// leaving everything else (including any unrelated XMP) untouched.
+    ///
+    /// Needed when re-saving with a mode that no longer points at a remote
+    /// manifest (e.g. embedding a full manifest after a prior sidecar/remote
+    /// save) — without this, a stale reference would be left behind.
+    ///
+    /// The default implementation returns [`Error::NotImplemented`]. The
+    /// blanket impl for [`WriteXmp`] overrides this to read the current XMP,
+    /// strip the reference, and write the result back.
+    fn remove_remote_manifest_url(
```
ok-nick, line 608: Maybe return a boolean to signify if it removed anything or not? It's a pattern I see often.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
-    fn embed_reference_to_stream(
+/// Writes a complete XMP packet into an asset stream, replacing any existing XMP.
+///
+/// This is the format-specific half of [`RemoteManifestUrl`]: given the final XMP
+/// string to write (already merged with the remote manifest URL), insert it into
+/// the asset's container format. Implement this instead of `RemoteManifestUrl`
+/// directly — a blanket impl derives `RemoteManifestUrl` from it, handling the
+/// "read current XMP, merge in the URL" part common to every format.
+pub trait WriteXmp {
+    /// Writes `xmp` into `input_stream`, producing `output_stream`.
```
ok-nick, line 627: Where is the equivalent `ReadXmp` trait?

sdk/src/asset_io.rs
ok-nick: I agree with @scouten-adobe . This should also stay consistent with the underlying function implemented on the asset handlers.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+
+    /// Looks up the full [`AssetIO`] handler for `format`, checking `fallback` if this
+    /// registry has no match.
+    pub fn handler(&self, format: &str) -> Option<&dyn AssetIO> {
+        let key = normalize_format(format);
+        match self.handlers.get(&key) {
+            Some(h) => Some(h.as_ref()),
+            None => self.fallback.as_deref().and_then(|f| f.handler(format)),
+        }
+    }
```
ok-nick, line 773: IMO, we should remove a lot of the convenience functions that defer to the equivalent methods on the underlying trait, it would simplify a lot of code and reduce the API surface. I think it's simple and explicit enough to call `foo.handler("png").reader()`.

sdk/src/asset_io.rs
```diff
@@ -108,163 +244,906 @@ impl Seek for CAIReadWriteWrapper<'_> {
...
+}
+
+/// Detects the container format of a stream by inspecting its leading bytes.
+///
+/// Reads a small header from the stream and matches it against magic signatures for each
+/// known container, then rewinds before returning. Returns `None` when the container cannot
+/// be identified from the bytes alone.
+fn sniff_container_from_stream<R: Read + Seek>(stream: &mut R) -> Option<&'static str> {
```
ok-nick, line 1048: Can these be a method associated with each asset handler rather than as it's own impl? We risk drifting the implementations.

sdk/src/builder.rs
ok-nick: I agree. I don't know if this belongs on the builder versus something like the context or the asset handler itself. Can't they access the `compose_manifest` method directly now that asset handlers are being exposed? The builder isn't involved here.

sdk/src/jumbf_io.rs
```diff
@@ -78,158 +71,26 @@ lazy_static! {
...
-    ///
-    /// The container ID is the first format string registered by that format's I/O handler.
-    /// Two format strings with the same container ID belong to the same handler family
-    /// (e.g. "dng", "tif", "image/tiff" all map to "tif").
-    static ref CONTAINER_MAP: HashMap<String, &'static str> = {
-        let mut map = HashMap::new();
+    /// The shared default [`HandlerRegistry`], seeded with the same built-in handlers as
+    /// `CAI_READERS`/`CAI_WRITERS`. This is what a fresh [`crate::Context`] falls back to
+    /// once its own (typically empty) set of custom handlers has been checked.
+    static ref DEFAULT_HANDLER_REGISTRY: Arc<HandlerRegistry> = {
```
ok-nick, line 78: Do we need this to be defined statically and have a fallback registry? Can't we initialize this on context construction since it's shared anyways?

sdk/src/asset_io.rs
scouten-adobe: +1. The `std::io` functions generally speak in `u64` or `i64` (for relative-position seeks).

sdk/src/builder.rs
gpeacock: The problem is that the previous public compose_manifest does not have a self parameter, so we can't use context on it. In hindsight, we should have passed an unused self. So, in order to not break compatibility, we need a new name here. But nick brings up a good point, we may just deprecate this method on builder and direct users to asset_io for this.

sdk/src/asset_io.rs
gpeacock: We kind of use c2pa and jumbf synonymously here. In practice what we read and write si only the jumbf c2pa data, so I think read_c2pa makes sense.

sdk/src/asset_io.rs
gpeacock: It used to be the same and I changed it. That's for 2 reasons. AssetIO should not be dependent on assertions if we want to make a separate crate for it. Also the box hash assertion has several other fields it needs that an asset cannot return.

sdk/src/asset_io.rs
gpeacock: changed to u64

sdk/src/jumbf_io.rs
gpeacock: I think we need to deprecate all the jumbf_io api since it is based on a static handler list, but it is a public api so we have to deprecate and not just remove it. 
There is some extra performance cost if we don't maintain a static registry for the new context based api, but it may not be a practical issue.

sdk/src/asset_io.rs
gpeacock: FYI, this was a pretty big change

tmathern:
@gpeacock I've been pondering over the cases of large files (especially videos). I think this API here would be very nice, and we'd need to expose it somehow in all the bindings we have. 

The use case I am thinking about: in large video files (1Gb+), we could then go fish for the manifest... over a network, using range requests (hints at having this in the js-based SDKs). I would suggest a new cocnept here: an AssetSource, that handles getting asset bytes.

I poked a little bit and what we need for that is already in here in main (use the BMMF io boxes to find/merkle trees, I got a quick and dirty POC working with that doing range requests using a stand-in Rust-backed server), but it is reaching deep down (deeper down than I think we should without this nicer API from here) in internals (that are also currently only reachable in Rust). If we could do something similar in js, through asset io handlers bindings with those APIs, it would be a nice way to get display manifest information for large files too (and without much network traffic at that)!

sdk/src/asset_io.rs
tmathern: `AssetIO should not be dependent on assertions if we want to make a separate crate for it.` 👍

sdk/src/asset_io.rs
tmathern: Also, the other (pragmatic?) case is that read and write may be implemented one after another: first read, then write. Then it comes handy to have a split, so the work to implement those can be split too!

sdk/src/asset_io.rs
gpeacock: do you mean make them public?

sdk/src/asset_handlers/README.md
gpeacock: It is today, but the Fonts handler doesn't support XMP and the spec doesn't require it. It seems like the spec is moving towards XMP only for this, but I don't think that's clear yet.

sdk/src/reader.rs
gpeacock: This doesn't take a self parameter, so it is stuck in non-dynamic handlers realm. I could make a breaking change and add self, or deprecate this and add a new _with_context() option, or direct new users to use reader.context().io() for this.

sdk/src/asset_io.rs
gpeacock: The handler can still register its own way of doing remote urls, but if they don't and they support xmp, we can default to the xmp way.

sdk/src/builder.rs
gpeacock: I can't fix this function without changing it since it doesn't take self as a parameter.  We will need to deprecate this along with the older jumbf_io apis when we make dynamic assets handlers public.

sdk/src/asset_io.rs
gpeacock: I'll do that at some point as part of better logging support. For now I'll try to avoid unreported errors.

sdk/src/asset_handlers/README.md
gpeacock: yes, this is just the file_io method, you can still use streams.

sdk/src/asset_handlers/README.md
gpeacock: This is referring to how to write xmp when there isn't already xmp, we return None if we are only reading.

sdk/src/asset_handlers/riff_io.rs
gpeacock: yeah, I think that would be better. This kind of a patch to not make new requirements on the handler api.

sdk/src/asset_io.rs
gpeacock: Yes but we still need an AssetIO object to register dynamically. This isn't just a trait, it is a trait object.

sdk/src/asset_io.rs
gpeacock: what would we use that for?

sdk/src/asset_io.rs
gpeacock: I got rid of these

sdk/src/asset_io.rs
gpeacock: These are here only for backward compatibility with tests and public jumbf_io methods. when we update and deprecate those, we can get rid of these.

sdk/src/asset_io.rs
gpeacock: That sounds reasonable. But I haven't done that yet.

sdk/src/asset_io.rs
gpeacock: I suppose we could, but we need to add the underlying info for that.

sdk/src/asset_io.rs
gpeacock: not a bad idea. it really depends on how often we use it.

sdk/src/asset_io.rs
gpeacock: Yes, I'm working on that. I think they need to work that way in the long term

tmathern (approved):

sdk/src/asset_io.rs
gpeacock: I don't have a full answer on this one, I think it was added to allow the asset handler to provide segments that should be excluded.

sdk/src/error.rs
```diff
@@ -324,37 +317,6 @@ pub enum Error {
 
     #[error("manifest failed validation with: {}", .0.failure_summary())]
     InvalidManifest(Box<ValidationResults>),
-
```
scouten-adobe, line 327: ❤️

scouten-adobe (approved):

sdk/src/error.rs
```diff
@@ -321,38 +316,7 @@ pub enum Error {
...
-    #[error("error parsing RIFF: {0}")]
-    RiffError(#[from] RiffError),
-
-    #[error("error parsing SVG: {0}")]
-    SvgError(#[from] SvgError),
-
-    #[error("error parsing TIFF: {0}")]
-    TiffError(#[from] TiffError),
+    InvalidManifest(Box<ValidationResults>),
 }
```
ok-nick: I added these so that we can start creating more granular errors from the foundation up since each parser can error for different reasons. I'd prefer if we head in that direction.

sdk/src/read_seek.rs
ok-nick: Do we need a file for this? Maybe it would make sense to store in the asset IO module next to the traits.

sdk/src/asset_io.rs
```diff
@@ -1147,6 +1163,7 @@ fn tempfile_builder<T: AsRef<OsStr> + Sized>(prefix: T) -> Result<NamedTempFile>
 ///
 /// If the rename is not possible due to cross volume references, the file will be copied to the
 /// final and then the temp file we be deleted.
+#[allow(dead_code)] // file-backed helper used only when file I/O is enabled
```
ok-nick: If these are only file_io, consider feature flagging the allow.

sdk/src/asset_io.rs
```diff
@@ -542,6 +474,75 @@ impl BoxMap {
...
+}
+
+/// What a permitted exclusion range represents, per spec §15.12.3.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ExclusionKind {
+    /// The C2PA Manifest Store itself, or padding.
+    ManifestOrPadding,
+    /// Asset metadata (EXIF/XMP/IPTC-equivalent) per spec §9.2.6.
+    AssetMetadata,
+}
```
ok-nick: It could also be an arbitrary exclusion?

sdk/src/asset_io.rs
```diff
@@ -542,6 +474,75 @@ impl BoxMap {
...
+        self.allowed_exclusions = allowed_exclusions;
+        self
+    }
+}
+
+/// What a permitted exclusion range represents, per spec §15.12.3.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ExclusionKind {
+    /// The C2PA Manifest Store itself, or padding.
+    ManifestOrPadding,
```
ok-nick: Maybe name this ManifestBox?

sdk/src/asset_io.rs
```diff
@@ -361,16 +286,17 @@ pub trait AssetIO: Sync + Send {
...
 
     /// Removes the C2PA manifest store from the file at `asset_path`.
     ///
-    /// The default implementation mirrors [`save_cai_store`](AssetIO::save_cai_store),
-    /// calling [`CAIWriter::remove_cai_store_from_stream`] instead.
-    fn remove_cai_store(&self, asset_path: &Path) -> Result<()> {
+    /// The default implementation mirrors [`save_c2pa_store`](AssetIO::save_c2pa_store),
+    /// calling [`C2paWriter::remove_c2pa`] instead.
+    #[allow(dead_code)] // file-backed helper; not required on all targets
+    fn remove_c2pa_store(&self, asset_path: &Path) -> Result<()> {
```
ok-nick: This is called `remove_c2pa_store` but then we also have `patch_c2pa_file` with a different suffix. I'd prefer if we stay consistent with the `file` suffix.

PR #2283, opened 2026-07-10
feat(experimental): Add structured-text asset handler (C2PA A.9)

dcondrey:
## Changes in this pull request

Adds a native in-tree asset handler for C2PA structured text (Appendix A.9 — armored `-----BEGIN C2PA MANIFEST----- / -----END C2PA MANIFEST-----`). No third-party crate dependency.

Proposed as an experimental feature under [docs/experimental-features.md](https://github.com/contentauth/c2pa-rs/blob/main/docs/experimental-features.md), gated behind the non-default `unstable_structured_text` flag. Discussion issue: #2377.

- `sdk/src/asset_handlers/structured_text_io.rs`, `text_common.rs` — the handler and shared codec. Single-line comment form for formats with a comment syntax, and the A.9.3.2 multi-line form inside leading front matter for Markdown. A block is recognized only when the text between the delimiters is a URL or `data:` URI (A.9.3), so an unterminated delimiter is "no manifest block" (A.9.5 step 4) and content that merely mentions the delimiters is left alone. Update manifests keep the original block offset; reads are bounded via `stream_len` + `read_to_vec`.
- `sdk/Cargo.toml`, `sdk/src/asset_handlers/mod.rs`, `sdk/src/jumbf_io.rs` — register behind `#[cfg(feature = "unstable_structured_text")]`.
- `sdk/src/utils/mime.rs`, `docs/supported-formats.md` — extension/MIME mappings and docs.
- `docs/experimental-features.md` — registry entry.

A note on front matter: A.9.3 requires the block to appear inside front matter for formats that support it. This treats that as applying when the file already opens with a `---`/`+++` fence, rather than creating front matter in a file that has none, since that would change how renderers read the document. A leading `---` in a YAML or TOML file is document syntax rather than front matter, so those keep the comment form.

Tests under the feature: markdown/YAML/TOML round-trip validating as `Trusted`, tamper detection, front matter placement and exclusion range (A.9.4 puts the fences outside it), CRLF round-trip (A.9.4 byte-exact line terminators), block-recognition edge cases, and a `test_get_supported_list` gate. 915 passed with `--features unstable_structured_text,file_io`; `clippy --all-features --all-targets -- -Dwarnings` and `cargo +nightly fmt --check` clean.

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

Issue #2424, opened 2026-08-05
Experimental feature: unstable_plain_text (plain-text asset handler, C2PA A.8)


scouten-adobe:
Opened to meet the requirements of the [experimental features policy](https://github.com/contentauth/c2pa-rs/blob/main/docs/experimental-features.md) for the **`unstable_plain_text`** experimental feature.

## Feature

- **Name:** Plain-text asset handler (C2PA A.8)
- **Cargo flag:** `unstable_plain_text`
- **APIs:** `TextIO` asset handler, registered for `text/plain` (`.txt`)
- **Owner / contact:** @erik-sv

Embeds and extracts a C2PA Manifest Store in `text/plain` assets per [C2PA 2.4 Appendix A.8](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#embedding_manifests_into_unstructured_text) (unstructured text), encoding the JUMBF manifest bytes as invisible Unicode variation selectors that preserve the visible text. Built on the [`c2pa-text`](https://crates.io/crates/c2pa-text) reference crate maintained by EncypherAI, added as an optional dependency activated only by this feature. Proposed in #2117.

## Status

Proposed as experimental. Acceptance conditions per the policy:

- Gated behind the non-default `unstable_plain_text` flag; no public API or behavior visible without it.
- The new `c2pa-text` dependency is `optional = true` and activated only via `dep:c2pa-text` on this feature; a default build does not compile or link it.
- Default builds are unchanged (extension/MIME resolution for `.txt` is gated behind the flag).
- Registry entry added in `docs/experimental-features.md`.
- This issue is the standing home for design discussion, change proposals, and any argument for promotion or removal.

Day-to-day support is community-supported by the contact above; the `c2pa-text` crate is maintained at https://github.com/encypherai/c2pa-text.

PR #2494, opened 2026-08-13
feat: Add native plain-text asset handler (A.8)

dcondrey:

Adds a native in-tree asset handler for C2PA plain text (Appendix A.8, `C2PATextManifestWrapper`, invisible Unicode variation selectors). **No third-party crate** — the only new dependency is `unicode-normalization`, for NFC.

Proposed as an experimental feature under [docs/experimental-features.md](https://github.com/contentauth/c2pa-rs/blob/main/docs/experimental-features.md), gated behind `unstable_plain_text`. Discussion issue: #2505.

- `sdk/src/asset_handlers/plain_text_io.rs` — handler and wire-format codec, ported directly from A.8's spec text: byte-to-variation-selector mapping, header framing, and the deterministic-padding formula with its fixed `0x00`/`0x10` decomposition.
- `sdk/Cargo.toml`, `sdk/src/asset_handlers/mod.rs`, `sdk/src/jumbf_io.rs` — register behind `#[cfg(feature = "unstable_plain_text")]`.
- `sdk/src/utils/mime.rs`, `docs/supported-formats.md` — `txt` → `text/plain`, gated.
- `docs/experimental-features.md` — registry entry.
- `sdk/src/builder.rs` — sign/tamper/NFD-input integration tests through the real two-pass signing flow.
- `sdk/src/store.rs` — `test_unsupported_type_without_external_manifest` used `unsupported_type.txt` as its "definitely unsupported" fixture. Registering `.txt` makes that feature-dependent, so the assertion is now feature-conditional (`UnsupportedType` off, `JumbfNotFound` on) rather than a scope change.

NFC: A.8 hashes NFC-normalized text with the wrapper excluded, but `AssetIO` has no hash-transform hook — it hashes raw bytes outside the exclusion for every handler alike. This handler NFC-normalizes visible text *before* embedding, so raw-byte hashing of what it writes is already NFC-hashing. Correct for anything this handler embeds; it does **not** repair a third-party `.txt` asset produced non-NFC, which falls back to raw-byte comparison like any other format — a limitation of `AssetIO`, not this handler.

Failure codes: A.8 defines `manifest.text.corruptedWrapper` / `manifest.text.multipleWrappers`; `AssetIO` has no channel for container-level codes, so both collapse to `JumbfNotFound`, mirroring #2283's precedent for A.9.

### Interop with `c2pa-text`

Checked against the published `c2pa-text` 3.0.0 crate (what #2117 depends on) in both directions: a wrapper it produces decodes correctly here, and a wrapper this handler produces decodes correctly there — magic, version, marker, and both variation-selector ranges match exactly.

**One divergence: `c2pa-text` 3.0.0 doesn't implement the padding decomposition in A.8's current working-group text** (the `0x00`/`0x10` split closing an ambiguity in the original wording — not yet in a numbered spec release; Erik Svilich authored it in the WG repo about two months before `c2pa-text` 3.0.0 shipped). The published crate pads with a different `(a, b)` split and `0xFF` instead. Doesn't affect exclusion sizing or extraction — **only** the byte-identical-wrapper guarantee that text was written for. Filed upstream as [encypherai/c2pa-text#12](https://github.com/encypherai/c2pa-text/pull/12); open, unaddressed as of this PR.

984 passed with `--features unstable_plain_text,file_io` (964 with the feature off, no regression); `clippy --all-targets -- -Dwarnings` and `cargo +nightly fmt --check` clean in both configurations.

## Checklist

[AI Disclosure: ai-assisted](https://github.com/dcondrey/ai-disclosure-badges/blob/main/docs/ai-assisted.md)


scouten-adobe (approved):

scouten-adobe:
@dcondrey thank you and this looks good at my first review. Asking for @gpeacock to also review it.

Meanwhile, would you please file a new experimental-feature issue with your contact info instead of reusing #2424?

scouten-adobe:
@dcondrey also, please run `cargo +nightly fmt` and push those fixes.

dcondrey:
@scouten-adobe opened #2505 and pushed the fix.  Thanks.

ChanduBodduluriTR:
Hello, Which plain-text implementation will become canonical? Is the intended path the native in-tree handler behind [unstable_plain_text (#2494)](https://github.com/contentauth/c2pa-rs/pull/2494), or the [c2pa-text-based handler (#2117)](https://github.com/contentauth/c2pa-rs/pull/2117)? Will either PR be superseded, and what are the criteria for removing the unstable designation?

dcondrey:
@ChanduBodduluriTR Your comment was answered in #2283

PR #2503, opened 2026-08-13
fix: Set allActionsIncluded per C2PA 2.4 spec when opening/re-saving without other changes (CAI-13222)

scouten-adobe:
## Summary

Implements [CAI-13222](https://jira.corp.adobe.com/browse/CAI-13222). Per the C2PA 2.4 spec's ["All actions included"](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_all_actions_included) section (§18.15.3):

> In the situation where a claim generator is opening an asset strictly for the purposes of recording its c2pa.opened action and then immediately re-saving it without making any other changes, the claim generator **shall** set `allActionsIncluded` to `true` to assert that no other actions were performed on the asset.

The `Actions::all_actions_included` field already existed and could be populated from a static `builder.actions.all_actions_included` setting, but nothing implemented this specific `shall` — a manifest built via `auto_opened_action` (or an explicit `c2pa.opened`-only intent) with no other recorded actions was left with `allActionsIncluded` omitted, which the spec says validators should interpret as "additional, unrecorded actions may have been performed" — the wrong, spec-noncompliant default for this exact case.

**Update:** per review comments on the Jira ticket flagging that the SDK can't fully verify "no other changes occurred" (it can only see changes that were recorded as an `Action`), this detection is **opt-in** rather than default-on:

* New setting `builder.actions.auto_all_actions_included` (default `false`). When enabled, if the resulting actions assertion is the manifest's sole/inception actions assertion and contains exactly one action (`c2pa.opened`), `allActionsIncluded` is forced to `true`.
* `Builder::add_actions_assertion_settings` determines `all_actions_included` after all actions (auto-inserted and settings-configured) have been merged into the assertion, rather than before — required so the "sole action" check sees the final list.
* A value the caller explicitly authored into the actions assertion data itself (as opposed to either settings default) is always respected and never overridden, regardless of `auto_all_actions_included`.
* Updated doc comments on `Actions::all_actions_included` and `ActionsSettings` to describe the field's semantics and this opt-in behavior accurately, per spec.

## Test plan

- [x] `cargo test -p c2pa --lib builder::` (101 tests passing)
- [x] `cargo test -p c2pa --lib assertions::actions::`
- [x] `cargo clippy -p c2pa --lib --tests`
- [x] `cargo check -p c2pa --lib --features json_schema`
- [x] Full `cargo test -p c2pa --lib` run (891 passing; 3 pre-existing, unrelated failures confirmed present on `main` too — missing `libpng-test.png` fixture in this sandboxed environment)

New/updated tests cover:
- `auto_all_actions_included` enabled + opening a `parentOf` ingredient with no other actions → `allActionsIncluded: true` (`test_builder_settings_auto_all_actions_included`)
- `auto_all_actions_included` **disabled** (default) + same opened-only scenario → field stays unset (`test_builder_settings_auto_opened`, updated)
- `auto_all_actions_included` enabled + opening plus an additional edit action → still not forced (`test_all_actions_included_not_forced_when_other_actions_present`)
- A caller-authored `allActionsIncluded: false` on an opened-only actions assertion is preserved even with `auto_all_actions_included` enabled (`test_all_actions_included_explicit_false_not_overridden`)
- Existing `test_builder_settings_all_actions_included` (static settings default, `c2pa.created` inception) still passes unchanged


sdk/src/builder.rs
```diff
@@ -1933,20 +1933,26 @@ impl Builder {
...
     /// * For more, see [Builder::add_auto_actions_assertions]
     ///
     /// Only the first actions assertion of a manifest may carry an inception (created/opened)
     /// action, so `allow_inception` must only be true for that assertion.
+    ///
+    /// Per the C2PA spec's [All actions included](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_all_actions_included)
+    /// section, a claim generator that opens an asset strictly to record its `c2pa.opened`
+    /// action and immediately re-saves it without making any other changes shall set
+    /// `allActionsIncluded` to `true`. When `builder.actions.auto_all_actions_included` is
+    /// enabled, this is detected once every action has been merged in, so it happens at the
```
tmathern: If others actions are merged it, it's not just opened anymore? (Clarify comment)

sdk/src/builder.rs
```diff
@@ -1992,7 +1998,33 @@ impl Builder {
...
+        if !all_actions_included_explicit {
+            let auto_all_actions_included = self
+                .context
+                .settings()
+                .builder
+                .actions
+                .auto_all_actions_included;
+
+            // A sole `c2pa.opened` action means the asset was opened only to record that
+            // action and immediately re-saved with no other changes, which the spec requires
```
tmathern: re-saved = added another manifest in the chain? (Clarify doc comment)

sdk/src/builder.rs
```diff
@@ -1992,7 +1998,33 @@ impl Builder {
...
+            // to be reported as `allActionsIncluded: true`. The SDK can only see changes that
+            // were recorded as actions, so this detection is opt-in via
+            // `auto_all_actions_included` rather than always-on.
+            actions.all_actions_included = if auto_all_actions_included
```
tmathern, line 2034: What happens if you had redactions configured? Are they still applied or consulted?

sdk/src/builder.rs
```diff
@@ -1992,7 +1998,33 @@ impl Builder {
...
+                && allow_inception
+                && actions.actions.len() == 1
```
tmathern: There is an `actions()` accessor.

sdk/src/builder.rs
```diff
@@ -4715,6 +4750,193 @@ mod tests {
         assert_eq!(actions.all_actions_included, Some(true));
     }
 
+    // With `auto_all_actions_included` enabled, opening an ingredient and re-saving with no
+    // other changes is the case the spec requires `allActionsIncluded` to be set to `true` for.
+    #[test]
+    fn test_builder_settings_auto_all_actions_included() {
```
tmathern, line 4774: Should also make sure that action goes into created assertions

sdk/src/builder.rs
scouten-adobe: Good catch, that was ambiguous — the check runs *after* everything else has been merged in specifically so it can tell whether it's still just opened. Reworded in c1add995: "a claim generator that opens an asset strictly to record its `c2pa.opened` action, with no other action recorded for that manifest, shall set `allActionsIncluded` to `true`", plus an explanation of why the check happens at the end rather than the start.

sdk/src/builder.rs
scouten-adobe: Reworded to avoid the spec's abstract "re-saves" phrasing and describe it in terms of this codebase: the condition is just "this manifest records nothing but the inception action" (i.e. `actions.actions()` ends up as a single `c2pa.opened` entry), not anything about the ingredient chain or manifest count. See c1add995.

sdk/src/builder.rs
scouten-adobe: Good question. Redactions (`ManifestDefinition::redactions`) are applied earlier, while ingredients are being added to the claim, and this function never looks at that field. A redaction is only visible to this check if the caller also adds a matching `c2pa.redacted` action to `actions.actions` (which the spec requires, and which every redaction test in this module already does) — if they don't, `auto_all_actions_included` could wrongly fire. Documented this explicitly in c1add995 as a caveat on the opt-in setting.

sdk/src/builder.rs
scouten-adobe: Fixed — now using `actions.actions()` instead of the raw field.

sdk/src/builder.rs
scouten-adobe: Added that check, and it caught a real bug: when there's no explicit actions assertion in the definition and one gets synthesized purely from `auto_*_action` settings (as in this test), `Builder::to_claim` was adding it via `add_assertion(&mut claim, &actions, false)` — i.e. as a *gathered* assertion, not *created*. Per spec §18.15.2 there shall be at least one actions assertion in `created_assertions`, and since this is the manifest's only actions assertion, that's what it should be. Fixed to pass `true`, and the test now asserts it via `claim.created_action_assertions()`.

tmathern (approved):

sdk/src/builder.rs
```diff
@@ -1918,8 +1918,9 @@ impl Builder {
             self.add_actions_assertion_settings(&ingredient_map, &mut actions, true)?;
 
             if !actions.actions().is_empty() {
-                // todo: add setting for created added actions
-                add_assertion(&mut claim, &actions, false)?;
+                // This is the manifest's only actions assertion, so per spec it must live in
+                // `created_assertions` rather than `gathered_assertions`.
+                add_assertion(&mut claim, &actions, true)?;
```
gpeacock: This means any actions we have defined in settings will automatically be forced into created actions. Typically those are inception assertions being added, but there could be others here.

gpeacock (approved):
I continue to assert that the allActionsIncluded flag serves no useful purpose, but we do have a lot of rules about it and this helps us comply with it!

sdk/src/builder.rs
scouten-adobe: Per discussion in stand-up, only force into created assertions if there's an inception action step.

scouten-adobe: Narrowed in f6cfaee4, per the stand-up resolution: only force into `created_assertions` when the synthesized assertion actually carries the `c2pa.created`/`c2pa.opened` inception action. Non-inception actions added via `builder.actions.actions` now stay gathered, same as before this PR. Added `test_builder_settings_actions_without_inception_stay_gathered` to cover it.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32292971129](https://github.com/contentauth/c2pa-rs/actions/runs/32292971129).

PR #2510, opened 2026-08-17
fix: Fall back to OCSP fetch when stapled response is unusable

mpgopala:
A malformed or otherwise unusable OCSP response stapled into the COSE signature's unsigned `rVals` header previously short-circuited straight to `Ok(OcspResponse::default())`, without ever reaching the configured online OCSP fetch. Since `rVals` is unsigned, an intermediary could append arbitrary non-DER bytes to an honestly-signed asset and silently suppress revocation checking even with `verify.ocsp_fetch = true`, flipping validation from Invalid/REJECT to Trusted/ACCEPT for a revoked signer.

`check_ocsp_status` now falls back to `fetch_and_check_ocsp_response` whenever the stapled response doesn't affirmatively resolve to revoked/not-revoked, exactly as if no staple were present at all. This also collapses a duplicated fetch-dispatch block, since both the "no staple" and "unusable staple" cases now share a single `fetch_policy` match instead of two copies of the same arm.

Adds a regression test that staples garbage bytes and asserts the fetch fallback actually fires (observed via SIGNING_CREDENTIAL_OCSP_INACCESSIBLE in the validation log). Also drops a redundant `async_signature(...)` override on `fetch_and_check_ocsp_response` that duplicated its sync signature verbatim.

### DIFF
--- a/sdk/src/crypto/cose/ocsp.rs
+++ b/sdk/src/crypto/cose/ocsp.rs
@@ -95,127 +95,123 @@ pub fn check_ocsp_status(
         }
     }
 
-    match get_ocsp_der(sign1) {
-        Some(ocsp_response_der) => {
-            let mut ocsp_log = StatusTracker::default();
-            let result = if _sync {
-                check_stapled_ocsp_response(
-                    sign1,
-                    &ocsp_response_der,
-                    data,
-                    ctp,
-                    tst_info,
-                    &mut ocsp_log,
-                    context.settings(),
+    // If a stapled response is present, it takes precedence over fetching --
+    // but only when it affirmatively resolves revocation status. A present
+    // but inconclusive staple (malformed DER, untrusted OCSP signer, wrong
+    // EKU, etc.) falls through to the `fetch_policy` dispatch below, exactly
+    // as if no stapled response were present at all, instead of silently
+    // treating the unusable staple as a clean result and skipping the
+    // configured online check.
+    if let Some(ocsp_response_der) = get_ocsp_der(sign1) {
+        let mut ocsp_log = StatusTracker::default();
+        let result = if _sync {
+            check_stapled_ocsp_response(
+                sign1,
+                &ocsp_response_der,
+                data,
+                ctp,
+                tst_info,
+                &mut ocsp_log,
+                context.settings(),
+            )
+        } else {
+            check_stapled_ocsp_response_async(
+                sign1,
+                &ocsp_response_der,
+                data,
+                ctp,
+                tst_info,
+                &mut ocsp_log,
+                context.settings(),
+            )
+            .await
+        };
+
+        // we only care about OCSP value log info if the result is OK
+        if let Ok(ocsp_response) = result {
+            if ocsp_log.has_status(validation_status::SIGNING_CREDENTIAL_REVOKED) {
+                log_item!(
+                    "",
+                    format!(
+                        "signing cert revoked: {}",
+                        ocsp_response.certificate_serial_num
+                    ),
+                    "check_ocsp_status"
                 )
+                .validation_status(SIGNING_CREDENTIAL_REVOKED)
+                .informational(validation_log);
+
+                return Err(CoseError::CertificateTrustError(
+                    CertificateTrustError::CertificateNotTrusted,
+                ));
+            }
+
+            // If certificate is confirmed not revoked, return success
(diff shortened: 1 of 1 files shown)


sdk/src/crypto/cose/ocsp.rs
```diff
@@ -95,127 +95,123 @@ pub fn check_ocsp_status(
...
-                    data,
-                    ctp,
-                    tst_info,
-                    &mut ocsp_log,
-                    context.settings(),
+    // If a stapled response is present, it takes precedence over fetching --
+    // but only when it affirmatively resolves revocation status. A present
+    // but inconclusive staple (malformed DER, untrusted OCSP signer, wrong
+    // EKU, etc.) falls through to the `fetch_policy` dispatch below, exactly
+    // as if no stapled response were present at all, instead of silently
```
tmathern, line 102: What about using SIGNING_CREDENTIAL_OCSP_SKIPPED:
https://github.com/contentauth/c2pa-rs/blob/eb877f68f3876ed5334263ccdd3c51982aa025f8/sdk/src/validation_results.rs#L647

Right now, if there would be a tampering attempt, nothing at all would even report it.

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -95,127 +95,123 @@ pub fn check_ocsp_status(
...
+    // treating the unusable staple as a clean result and skipping the
+    // configured online check.
+    if let Some(ocsp_response_der) = get_ocsp_der(sign1) {
+        let mut ocsp_log = StatusTracker::default();
```
tmathern, line 106: For instance, here this resets the log, could we keep something here? Or why there is a reset? (I wonder if we need/want to see that there was corruption, or if we rightfully ignore it.).

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -95,127 +95,123 @@ pub fn check_ocsp_status(
...
+                )
+                .validation_status(SIGNING_CREDENTIAL_NOT_REVOKED)
+                .informational(validation_log);
+
+                return Ok(ocsp_response);
+            }
+        }
+    }
+
+    match fetch_policy {
```
tmathern, line 167: This might also fire often on a tampered asset

tmathern (commented):
I wonder if we actually should log something somewhere (but where?) when there are such tampering attempts, because right now it would not even be visible anymore.

sdk/src/crypto/cose/ocsp.rs
mpgopala: The issue is that the attacker is corrupting the stapled response. The only way to know that it is corrupted is to get the OCSP fetch. If we skip the OCSP fetch, we would not be able to detect that the corruption was done.

sdk/src/crypto/cose/ocsp.rs
mpgopala: We are starting the log here and not resetting, correct? (By the way, this code already existed git diff is not showing the diff properly).
The issue is that someone could escape a revocation by just appending random data to the staple.

sdk/src/crypto/cose/ocsp.rs
mpgopala: I did not. get you. Can you please expand?

tmathern: Well, isn't a network fetch retriggered on failure? If we have an asset with failures, that would call out to the network? (The thing I am thinking about is repetition across claims and validations, especially since I think all those calls aren't cached).

sdk/src/crypto/cose/ocsp.rs
tmathern: Right, this is the ocsp log (we don't put that in validation logs later it seems). I was thinking of the general validation logs.

sdk/src/crypto/cose/ocsp.rs
tmathern: Can't this function here help to detect that? (Or is it a chicken-egg problem then):
https://github.com/contentauth/c2pa-rs/blob/5dcaaff1aa591224b6bd6da90d1f29d11ec2b2ba/sdk/src/crypto/ocsp/mod.rs#L61-L72

The revocation status needs the network, that function can tell if the staple was usable or not, no? That could let fine tune error handling here.

mpgopala:
The VULN is closed as not valid as spec does not mandate us to query OCSP response. So, closing this PR too.
@tmathern

PR #2511, opened 2026-08-17
feat: Make BMFF hash exclusion of /free and /skip boxes configurable

ssanthosh:
c2pa-rs excludes /free and /skip from the BMFF hash by default (spec-permitted, not spec-required). Add builder.bmff_hash_exclude_free_and_skip_boxes (default true, matching existing behavior) so signers who want the stronger guarantee can opt to fold these boxes into the hash instead.

`BmffHash::set_default_exclusions()` always excludes `/free` and `/skip` boxes from the BMFF/MP4 hard-binding hash. This is spec-permitted but not spec-required, and some signers may want the stronger guarantee of hashing this content instead.

Adds `builder.bmff_hash_exclude_free_and_skip_boxes` (default `true`, matching existing behavior). When set to `false`, `/free`/`/skip` content is folded into the hash rather than excluded.

- **`settings/builder.rs`**: new `bmff_hash_exclude_free_and_skip_boxes: bool` field on `BuilderSettings`.
- **`assertions/bmff_hash.rs`**: new `set_default_exclusions_with_options(exclude_free_and_skip: bool)`; existing `set_default_exclusions()` is unchanged (still defaults to `true`) and just delegates to it, so no breaking API change.
- **`builder.rs` / `store.rs`**: the three production call sites now read the setting from `Context`/`Settings` instead of always excluding.

## Tests added
- `set_default_exclusions_excludes_free_and_skip` — default behavior unchanged.
- `set_default_exclusions_with_options_false_keeps_free_and_skip_hashed` — opt-out correctly omits `/free`/`/skip` while leaving other mandatory exclusions intact.

## Checklist

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -635,6 +635,17 @@ impl BmffHash {
 
     // Adds default exclusion ranges for BMFF hashes.  Add as needed.
     pub fn set_default_exclusions(&mut self) -> &[ExclusionsMap] {
+        self.set_default_exclusions_with_options(true)
+    }
+
+    /// Like [`set_default_exclusions`](Self::set_default_exclusions), but lets
+    /// the caller control whether `/free` and `/skip` are excluded from the
+    /// hash. Mirrors
+    /// [`BuilderSettings::bmff_hash_exclude_free_and_skip_boxes`](crate::settings::builder::BuilderSettings::bmff_hash_exclude_free_and_skip_boxes).
+    pub fn set_default_exclusions_with_options(
+        &mut self,
+        exclude_free_and_skip: bool,
+    ) -> &[ExclusionsMap] {
         let exclusions = &mut self.exclusions;
 
         let cp2a_id: [u8; 16] = [
@@ -668,16 +679,18 @@ impl BmffHash {
             exclusions.push(mfra);
         }
 
-        // /free exclusion
-        if !exclusions.iter().any(|e| e.xpath == "/free") {
-            let free = ExclusionsMap::new("/free".to_owned());
-            exclusions.push(free);
-        }
+        if exclude_free_and_skip {
+            // /free exclusion
+            if !exclusions.iter().any(|e| e.xpath == "/free") {
+                let free = ExclusionsMap::new("/free".to_owned());
+                exclusions.push(free);
+            }
 
-        // /skip exclusion
-        if !exclusions.iter().any(|e| e.xpath == "/skip") {
-            let skip = ExclusionsMap::new("/skip".to_owned());
-            exclusions.push(skip);
+            // /skip exclusion
+            if !exclusions.iter().any(|e| e.xpath == "/skip") {
+                let skip = ExclusionsMap::new("/skip".to_owned());
+                exclusions.push(skip);
+            }
         }
 
         /*  no longer mandatory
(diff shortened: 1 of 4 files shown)


sdk/src/builder.rs
```diff
@@ -2518,7 +2518,12 @@ impl Builder {
                 self.bmff_hasher.alg = ph_alg.to_string();
 
                 let mut placeholder_bmff = BmffHash::new("jumbf manifest", ph_alg, None);
-                placeholder_bmff.set_default_exclusions();
+                placeholder_bmff.set_default_exclusions_with_options(
```
tmathern: Does this play well with the embeddable APIs (those: https://github.com/contentauth/c2pa-rs/blob/main/docs/embeddable-api.md) which split up the work?
Like if the setting changes in between those calls (shouldn't, but, well, if I could imagine it...)?

sdk/src/assertions/bmff_hash.rs
```diff
@@ -2506,6 +2519,29 @@ mod bmff_hash_tests {
     use super::*;
     use crate::asset_handlers::bmff_io::{BoxInfoLite, C2PABmffBoxes};
 
+    /// `set_default_exclusions` (no args) must keep excluding `/free`/`/skip`,
+    /// matching its existing, documented default behavior.
+    #[test]
+    fn set_default_exclusions_excludes_free_and_skip() {
```
tmathern, line 2524: I believe this new setting warrants and e2e test with signing + re-reading to make sure the whole flow is covered and works, with both the setting on (default, should work) and off. Wdyt?

sdk/src/store.rs
```diff
@@ -2196,12 +2196,14 @@ impl Store {
         Ok(hashes)
     }
 
-    fn generate_bmff_data_hash_for_stream(alg: &str) -> Result<BmffHash> {
+    fn generate_bmff_data_hash_for_stream(alg: &str, settings: &Settings) -> Result<BmffHash> {
         // The spec has mandatory BMFF exclusion ranges for certain atoms.
         // The function makes sure those are included.
```
tmathern: Maybe update the comment to explain why some others are governed by a setting? (The `spec-permitted, not spec-required` part of the PR), otherwise the comment is confusing.

sdk/src/settings/builder.rs
```diff
@@ -591,6 +591,10 @@ pub struct BuilderSettings {
     ///
     /// [`TimeStamp`]: crate::assertions::TimeStamp
     pub auto_timestamp_assertion: TimeStampSettings,
+    /// Whether `/free` and `/skip` boxes are excluded from the BMFF/MP4 hard-binding hash.
+    ///
+    /// The default value is `true`.
```
tmathern, line 614: It may also need to explain what happens when this is false, and what adding those two to hashing means.

sdk/src/assertions/bmff_hash.rs
```diff
@@ -635,6 +635,17 @@ impl BmffHash {
...
+        self.set_default_exclusions_with_options(true)
+    }
+
+    /// Like [`set_default_exclusions`](Self::set_default_exclusions), but lets
+    /// the caller control whether `/free` and `/skip` are excluded from the
+    /// hash. Mirrors
+    /// [`BuilderSettings::bmff_hash_exclude_free_and_skip_boxes`](crate::settings::builder::BuilderSettings::bmff_hash_exclude_free_and_skip_boxes).
+    pub fn set_default_exclusions_with_options(
+        &mut self,
+        exclude_free_and_skip: bool,
```
tmathern: I would use an options object containing the relevant settings (if not the context, who holds settings) listing the relevant settings here, since I could see it in future that this extends.

tmathern: (For instance, imagine if one day `/free` and `/skip` would be controlled separately).

tmathern: (`generate_bmff_data_hash_for_stream(alg: &str, settings: &Settings)` that one already gets all settings...).

sdk/src/builder.rs
ssanthosh: Confirmed that there is no risk — the setting is read once in placeholder() and baked into the assertion's exclusions list; `update_hash_from_stream`/`sign_embeddable` just reuse what's stored. Added a comment to indicate  that

sdk/src/assertions/bmff_hash.rs
ssanthosh: Added new E2E test with https://github.com/contentauth/c2pa-rs/pull/2511/commits/e5569ed9e08ae0ddffe125c44c51514034c9b1e0

sdk/src/store.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2511/commits/e5569ed9e08ae0ddffe125c44c51514034c9b1e0

sdk/src/settings/builder.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2511/commits/e5569ed9e08ae0ddffe125c44c51514034c9b1e0

sdk/src/assertions/bmff_hash.rs
ssanthosh: Threaded the Settings object to make the changes future proof with https://github.com/contentauth/c2pa-rs/pull/2511/commits/e5569ed9e08ae0ddffe125c44c51514034c9b1e0

tmathern (approved):

gpeacock (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32745704212](https://github.com/contentauth/c2pa-rs/actions/runs/32745704212).

PR #2312, opened 2026-07-15
fix: Updates missing 2.2 validation steps

mauricefisher64:
Prep for 2.3-2.4 support
Add validation for soft binding and cloud data
Support timestamp accuracy
Parsing of soft binding and conforming product registries (in preparation of fetching these dynamically)

## Changes in this pull request
_Give a narrative description of what has been changed._

## Checklist
- [x] This PR represents a single feature, fix, or change.
- [x] All applicable changes have been documented.
- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment.

PR #2512, opened 2026-08-18
fix: Don't abort verification on an undecodable soft-binding/cloud-data assertion

### LINKED ISSUE #43: Failed to process file Error
Im seeing this error message: 

Failed to process file: Failed to read C2PA data: claim could not be converted from CBOR. 

It appears on all side-car files as well as embedded assets we're producing at least. I checked this against older media and its persists. Running the prior deployed version i dont see the error. so it seems this was recently introduced.

andyparsons:
## Summary
- `verify_soft_binding_alg` and `verify_cloud_data` (added in #2312 for C2PA 2.2 validation) already logged a `claim.malformed` failure via `failure_no_throw()` when a `c2pa.soft-binding`/`c2pa.cloud-data` assertion failed to decode, but then immediately re-threw the same error via `?`, hard-aborting the entire manifest read/report over a single malformed assertion. Both now `continue` past the malformed assertion instead, so the failure is still recorded in `validation_status` but verification and reporting complete.
- `Error::ClaimDecoding(String)`'s `Display` impl previously discarded its string field, so every use (assertion URI, missing-field name, etc.) rendered as the same generic "claim could not be converted from CBOR" message. Now includes the field so the error names what actually failed.
- Added `test_verify_soft_binding_alg_undecodable_does_not_abort`, reproducing a `c2pa.soft-binding`-labeled assertion with JSON content instead of the expected CBOR structure, asserting verification still succeeds with a logged `CLAIM_MALFORMED` status.

Fixes contentauth/c2pa-conformance-tool#43

## Test plan
- [x] `cargo test -p c2pa --lib claim::tests` — all pass, including new regression test
- [x] `cargo test -p c2pa --lib` — full suite passes (3 pre-existing failures due to a missing test fixture in this environment, confirmed present on unmodified `main` too, unrelated to this change)
- [x] `cargo clippy -p c2pa --lib` — no new warnings in touched files
- [x] `cargo fmt --check -p c2pa` — clean
- [x] Manually verified with `c2patool` against a file containing a non-conformant `c2pa.soft-binding` assertion (JSON content instead of CBOR): previously hard-errored with `Error: claim could not be converted from CBOR`; now exits 0 and produces the full manifest report with the malformed assertion correctly surfaced as a `claim.malformed` validation failure.


tmathern:
Does this relate maybe also to incoming changes from https://github.com/contentauth/c2pa-rs/pull/2388 ?

sdk/src/claim.rs
```diff
@@ -3452,17 +3452,22 @@ impl Claim {
...
+                        "verify_soft_binding_alg"
+                    )
+                    .validation_status(validation_status::CLAIM_MALFORMED)
+                    .failure_no_throw(validation_log, Error::ClaimDecoding(label.clone()));
 
-                Error::ClaimDecoding(label.clone())
-            })?;
+                    // Malformed, but don't abort the whole verification over one
+                    // bad soft binding assertion -- move on to the next one.
+                    continue;
```
tmathern, line 3467: This is now non fatal, but other checks are, especially the alg checks. Is this because there are different levels of severity?

sdk/src/claim.rs
```diff
@@ -3518,16 +3523,22 @@ impl Claim {
...
+                Err(_) => {
+                    log_item!(
+                        label.clone(),
+                        "cloud-data assertion could not be decoded",
+                        "verify_cloud_data"
+                    )
+                    .validation_status(validation_status::ASSERTION_CLOUD_DATA_MALFORMED)
+                    .failure_no_throw(validation_log, Error::ClaimDecoding(label.clone()));
+
+                    // Malformed, but don't abort the whole verification over one
```
tmathern, line 3536: This is now non fatal, but other checks are, especially the alg checks. Is this because there are different levels of severity?

sdk/src/error.rs
```diff
@@ -85,8 +85,10 @@ pub enum Error {
     #[error("claim could not be converted to CBOR")]
     ClaimEncoding,
 
-    /// The attempt to deserialize the claim from CBOR failed.
-    #[error("claim could not be converted from CBOR")]
+    /// The claim (or one of its assertions) failed to decode or is structurally
+    /// invalid. The string carries the assertion URI or a description of what
+    /// was wrong.
+    #[error("claim could not be converted from CBOR: {0}")]
```
tmathern, line 91: 👍 

Out of an abundance of caution, we'll need to check error handling in downstream SDKs (at least that unit tests don't break in case they were testing this path and expecting an exact error).

sdk/src/claim.rs
```diff
@@ -5030,6 +5041,75 @@ pub mod tests {
         assert!(validation_log.logged_items().is_empty());
     }
 
+    #[test]
+    fn test_verify_soft_binding_alg_undecodable_does_not_abort() {
```
tmathern, line 5044: You could add more assertions and verify the second and/or third assertions get checked. For instance: second assertion is correct, third has an invalid algorithm. Normally, this should hit the 3rd assertion then, and report the algorithm error (that makes sure checks continue past the soft binding check).

tmathern (commented):
Worth cross-checking with what is done in https://github.com/contentauth/c2pa-rs/pull/2388.
(And extending the tests).

sdk/src/error.rs
andyparsons: good point. If we take this PR, the release process should include documenting this error handling change. How do we handle this kind of thing for pending release notes ?

ok-nick (approved):

sdk/src/claim.rs
andyparsons: it's the soft binding that is broken and we've seen this a few times in the wild. I don't think a soft-binding failure should be fatal.

sdk/src/error.rs
tmathern: @scouten-adobe might have an idea.
At some point we'll ahve nightly builds/pre-flight checks to downstream SDKs we own to flag that.
Might still be worth documenting, but not sure how release-plz picks up that changeset (should it maybe go in the PR title?).

tmathern (approved):
(Extending the test would still be a nice touch, see https://github.com/contentauth/c2pa-rs/pull/2512#discussion_r3801092438).

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33199764192](https://github.com/contentauth/c2pa-rs/actions/runs/33199764192).

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) failed to backport this pull request in [workflow run 33200541671](https://github.com/contentauth/c2pa-rs/actions/runs/33200541671).

| Target | Status |
|--------|--------|
| `stable` | :x: Failed |

```bash
git fetch origin stable
git worktree add -d .worktree/backport-2512-to-stable origin/stable
cd .worktree/backport-2512-to-stable
git switch --create backport-2512-to-stable
git cherry-pick -x 4dee370d9d74b1b1a8a5c02ca769a055599ebe77
```

</details>

PR #2513, opened 2026-08-18
feat: Add support for general boxes hash exclusions

ssanthosh:

Adds support for the `exclusions` field on the General Box Hash assertion (`c2pa.hash.boxes`, spec §18.7), letting a `box-hash-map` entry exclude specific byte ranges within one of its named boxes from the hash — as distinct from `excluded`, which skips a whole box. Also validates *what's inside* an exclusion, per spec §15.12.3, down to the specific byte range permitted within each box (not just which box it targets), and correctly distinguishes structural malformation from content mismatch in the resulting validation status.

`BoxMap` previously had no `exclusions` support at all. Since CBOR/JSON deserialization silently ignores unknown map keys, a manifest from any other spec-compliant tool that uses `exclusions` would parse fine today but then get hashed over the box's *full* content instead of content-minus-exclusions, producing a false `assertion.boxesHash.mismatch` on verification. This closes that interop gap, and adds symmetric support for generating spec-conformant exclusions from c2pa-rs itself.

### 1. `exclusions` support (`sdk/src/assertions/box_hash.rs`)

- New `BoxExclusion` (`start`/`length`/`boxIndex`) type, added as `BoxMap::exclusions`.
- `split_exclusions`: resolves `boxIndex`-relative ranges to absolute file offsets and punches them out of a box-hash-map entry's range. Also guards against a fully-excluded entry silently falling back to hashing the entire stream (an existing trap in the shared hashing helper when passed an empty range list).
- `verify_stream_hash_with_progress` now honors `exclusions` when present.
- New `generate_box_hash_from_stream_with_exclusions` API (plus its `_with_progress` variant) lets callers request exclusions by source box index; the existing `generate_box_hash_from_stream[_with_progress]` are unchanged and delegate through with no exclusions, so no existing caller's behavior changes.
- `BoxExclusion`/`BoxHashExclusionRequest` are re-exported from `assertions::mod` alongside `BoxHash`/`BoxMap`, so external callers can actually construct/name them.

### 2. Validate exclusion content down to the permitted *byte range*, not just the box (`AllowedExclusion`)

Spec §15.12.3 requires a validator to reject an exclusion whose content isn't the C2PA Manifest Store or asset metadata (§9.2.6, e.g. EXIF/XMP/IPTC-equivalent), and §18.7.1 further requires that an exclusion range "shall not overlap with any header or length field associated with that box." Classifying an entire box as excludable isn't enough to satisfy that second constraint — e.g. a PNG chunk's `range_start` points at its 4-byte length field, so a whole-box-level "this box is Metadata" flag would let an attacker rewrite that length field (or a JPEG segment's marker/length, or a JPEG-XL `brob` box's embedded original-type field) without invalidating the hash.

- New `AllowedExclusion { start, length, kind: ExclusionKind }` (box-relative offset/length + `ManifestOrPadding`/`AssetMetadata`), added as `BoxMap::allowed_exclusions: Vec<AllowedExclusion>` (`#[serde(skip)]`, like `range_start`/`range_len` — always freshly derived from the *live* asset by `AssetBoxHash::get_box_map`, never trusted from the assertion itself). Replaces an earlier, less precise `BoxKind` (whole-box classification) design.
- `split_exclusions` now takes `box_allowed_exclusions: &[Vec<AllowedExclusion>]` alongside `box_ranges`, and requires each exclusion's `[start, start+length)` to be fully contained within *one* permitted sub-range for its box (not a union of adjacent ones). This subsumes the separate "exclusion extends past its box's end" check (anything out of bounds can't be contained in an in-bounds range either). Enforced identically by generation (fails fast on a bad request) and verification.
- Per-format population, each computing the actual excludable payload range rather than trusting the box's full span:
  - **JPEG** (`jpeg_io.rs`): `APP1`/`APP13` only get an `AllowedExclusion` after checking the segment's live bytes carry a recognized signature (`Exif\0\0`/XMP for `APP1`, `Photoshop 3.0\0` for `APP13`) — a marker name alone doesn't prove its payload is metadata. `COM` (free text, no signature to check) stays unconditionally excludable, skipping just the 2-byte marker + 2-byte length.
  - **PNG** (`png_io.rs`): `eXIf`/`iTXt`/`tEXt`/`zTXt` get `start: 8, length: range_len - 8` — the 4-byte length + 4-byte type fields stay protected, but the range extends through the trailing 4-byte CRC (which covers type+data, so any legitimate metadata edit necessarily changes it too).
  - **JPEG-XL** (`jpegxl_io.rs`): `Exif`/`xml ` get their data range directly from the box's already-tracked `header_size`/`data_size`. A `brob` wrapper (Brotli-compressed `Exif`/`xml `) only gets an `AllowedExclusion` after peeking its first 4 bytes (the embedded original box type, itself outside the excludable range) and confirming it's a recognized type — otherwise the wrapper is left unclassified (fail closed).
  - **GIF** (`gif_io.rs`): Comment Extension and XMP Application Extension both store their payload as a length-prefixed, 0x00-terminated sub-block stream; `CommentExtension` was refactored onto the same `DataSubBlocks` representation `ApplicationExtension` already used, and a shared `sub_block_data_ranges` helper derives one `AllowedExclusion` per sub-block's data (skipping each block's own length-prefix byte and the terminator) for both.
  - `.c2pa` sidecar (`c2pa_io.rs`): trivial, never populated (its one box is always skip-hashed, never reaches this check).
- `ExclusionKind::ManifestOrPadding` is part of the schema but not populated by any handler today (no handler treats a box as "the C2PA store, for exclusion purposes" outside the `skip_c2pa` path, which never consults `allowed_exclusions` at all).

### 3. Informational status

- New `assertion.boxesHash.additionalExclusionsPresent` status (`ASSERTION_BOXHASH_ADDITIONAL_EXCLUSIONS` in `validation_results.rs`), mirroring the existing `DataHash` equivalent.
- `verify_stream_hash_with_progress` returns `Result<bool>` — true if any exclusion beyond the C2PA store itself was used, either an `exclusions` entry referencing an `AssetMetadata` range, or a whole non-C2PA box skipped via `excluded: true` (that path doesn't classify the excluded box's content the way the `exclusions` sub-range path does, but skipping *any* non-C2PA-store box is still an exclusion beyond the baseline and must surface the signal). This is `pub(crate)`, so not a public API change; the public `verify_stream_hash` wrapper still returns `Result<()>`.
- `claim.rs` logs the informational status on `Ok(true)`, alongside the existing success log.

### 4. Malformed vs. mismatch status codes

Per spec §15.12.3, negative/missing/out-of-range `boxIndex`, arithmetic overflow, unordered/overlapping exclusion ranges, and an empty `exclusions` array (CDDL requires `[1* box-exclusions-map]`, so `Some(vec![])` is invalid, not equivalent to the key being absent) are all `assertion.boxesHash.malformed` — distinct from `assertion.boxesHash.mismatch` (an exclusion range not contained in any permitted sub-range for its box). `split_exclusions`/`verify_stream_hash_with_progress` return `Error::C2PAValidation(ASSERTION_BOXESHASH_MALFORMED)` for the former and `Error::HashMismatch` for the latter, and `claim.rs`'s `BoxHash` verification arm mirrors the pattern already used for BMFF hash to surface that distinction instead of collapsing everything to `.mismatch`. The empty-array check runs unconditionally, before the box is potentially skipped via `excluded: true` or being the C2PA store — `excluded` only says the box's hash is skipped, it doesn't excuse an otherwise-present `exclusions` value from still satisfying the CDDL shape.

## Tests added

`sdk/src/assertions/box_hash.rs`:
- `test_exclusions_round_trip` — CBOR/JSON round-trip preserves `exclusions`/`boxIndex`.
- `test_split_exclusions_basic_example` — worked multi-box, multi-exclusion case.
- `test_split_exclusions_fully_excluded_entry_does_not_fall_back_to_full_range` — the empty-range-list trap.
- `test_split_exclusions_rejects_box_with_no_allowed_exclusions` / `_rejects_range_outside_allowed_exclusion` — containment enforcement.
- `test_split_exclusions_allows_manifest_or_padding_without_metadata_signal` — `ManifestOrPadding` is permitted but doesn't set the informational flag (only `AssetMetadata` does).
- Negative tests for the exclusions array's shape: missing/negative/out-of-range `boxIndex`, unordered/overlapping ranges, a range extending past its box's end (all `Error::C2PAValidation(ASSERTION_BOXESHASH_MALFORMED)` or `Error::HashMismatch` as appropriate).
- `test_generate_box_hash_from_stream_with_exclusions_rejects_disallowed_request` — generation fails fast on a request outside any permitted range.
- `test_generate_and_verify_with_exclusions` — generate then verify against mutated data: a change *inside* a permitted excluded range still verifies; a change *outside* one does not.
- `test_verify_stream_hash_rejects_empty_exclusions_array` / `_on_excluded_box` / `_on_c2pa_box` — the CDDL shape check fires regardless of `excluded`/being the C2PA store.
- `test_verify_stream_hash_flags_excluded_non_c2pa_box` / `_does_not_flag_excluded_c2pa_box` — the informational signal fires for a whole excluded non-C2PA box, but not for the C2PA store's own excluded box.

Per-format:
- `jpeg_io.rs`: `test_app_segment_is_recognized_metadata` (Exif/XMP/Photoshop signature checks, including a wrong-marker and an unrecognized-vendor-payload negative case), `test_jpeg_allowed_exclusions`.
- `png_io.rs`: `test_classify_png_allowed_exclusions` (including the CRC-inclusive range length).
- `jpegxl_io.rs`: `test_classify_jxl_allowed_exclusions` (including `brob` wrapping a recognized type, an unrecognized wrapped type, and an unreadable wrapped type).
- `gif_io.rs`: `test_allowed_exclusions` (Comment Extension, XMP Application Extension, and an unrecognized Application Extension), `test_read_blocks` (real-fixture Comment Extension byte ranges).

## Checklist


gpeacock (commented):
It feels to me like this must be incompatible with minimal form boxhash, Have we looked into that?

The implementation looks good to me from the assertion level alone, except for the above question, but we have no way to actually use this yet.

How would someone use this in practice? We allow someone to pass in a pre-constructed boxhash into builder, but how would you make one? We don't have a public api for getting the boxmap from an asset.

If you could read a boxmap, I guess you would iterate through the box names to find the index of a box you know how to parse and then build exclusions for something you don't want to hash.

My asset_io PR will provide away to read BoxMaps from an asset. Maybe we want to hold off on this until we merge with that? I seriously doubt anyone else has a way to create one of these.

ssanthosh:
> It feels to me like this must be incompatible with minimal form boxhash, Have we looked into that?
> 
> The implementation looks good to me from the assertion level alone, except for the above question, but we have no way to actually use this yet.
> 
> How would someone use this in practice? We allow someone to pass in a pre-constructed boxhash into builder, but how would you make one? We don't have a public api for getting the boxmap from an asset.
> 
> If you could read a boxmap, I guess you would iterate through the box names to find the index of a box you know how to parse and then build exclusions for something you don't want to hash.
> 
> My asset_io PR will provide away to read BoxMaps from an asset. Maybe we want to hold off on this until we merge with that? I seriously doubt anyone else has a way to create one of these.

I have added the test to verify exclusions work fine with minimal form box hash also with commit : https://github.com/contentauth/c2pa-rs/pull/2513/commits/71403c6dc56357f66d58ed637a240aa0d53d8c3a

scouten-adobe:
I'm not super familiar with this part of the spec so I'll defer to @gpeacock and @mauricefisher64 for approval.

sdk/src/assertions/box_hash.rs
```diff
@@ -32,13 +32,60 @@ use crate::{
...
+    /// has one entry. Signed (per CDDL `int`) so a negative wire value is a
+    /// validation error, not a deserialize failure.
+    #[serde(rename = "boxIndex", skip_serializing_if = "Option::is_none")]
+    pub box_index: Option<i64>,
+}
+
+/// A caller-specified byte range to exclude when generating a [`BoxHash`],
+/// targeting the box at `source_box_index` in `AssetBoxHash::get_box_map`'s
+/// output (0-based, file order).
+#[derive(Clone, Debug)]
```
mauricefisher64, line 62: Why do we need a separate structure with the same values?

scouten-adobe (changes requested):
Please consider changes proposed in #2543.

sdk/src/assertions/box_hash.rs
ssanthosh: They're not interchangeable as they use different index spaces. `BoxExclusion.box_index `indexes into the final, possibly-folded `BoxMap::names` (wire format, optional+signed per the CDDL). `BoxHashExclusionRequest.source_box_index` indexes into `get_box_map`()'s raw, pre-folding output (caller-facing, never serialized). Generation translates one into the other once it knows how minimal_form folded things — a caller using the request type never has to know that mapping. Keeping them separate is what lets callers target boxes by their plain, obvious position instead of precomputing a post-folding index themselves.

ssanthosh:
> Please consider changes proposed in #2543.

Changes on PR https://github.com/contentauth/c2pa-rs/pull/2543 are merged. Please check

scouten-adobe (approved):
Sorry, found some more things that I've addressed in #2556.

Unfortunately, I'm out until Monday. Approved if you merge #2556; otherwise, we'll need to discuss when I'm back.

mauricefisher64 (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33376043961](https://github.com/contentauth/c2pa-rs/actions/runs/33376043961).

PR #2516, opened 2026-08-18
fix(sdk)!: Use CAWG identity assertion v1.3 status codes for X.509 signature validation

scouten-adobe:
## Summary

The CAWG identity assertion spec's [§8.2.2, "Validating the COSE signature,"](https://cawg.io/identity/1.3/#_validating_the_cose_signature) was revised in version 1.3 to define its own `cawg.x509.*` validation status codes instead of reusing the C2PA claim-signature status codes (`signingCredential.trusted`, `claimSignature.mismatch`, etc.). This PR updates the SDK's CAWG X.509 identity assertion validation to emit the new codes.

- `sdk/src/identity/x509/mod.rs`: adds `remap_x509_cose_status_codes`, which rewrites the shared `crypto::cose` C2PA-oriented status codes into their `cawg.x509.*` equivalents for log items added during a specific X.509 COSE signature check. The remap only touches log items added after a recorded starting index, so nothing outside that check is affected.
- `sdk/src/identity/x509/x509_signature_verifier.rs` (`X509SignatureVerifier`, the pluggable `SignatureVerifier` implementation) and `sdk/src/identity/identity_assertion/assertion.rs` (`IdentityAssertion::validate_partial_claim`/`_async`, used by `CawgValidator` during normal manifest reading) both call the shared `crypto::cose::Verifier`/`parse_cose_sign1` code and then apply the remap.
- Also adds the two status codes the shared code never emitted for this path:
  - `cawg.x509.signature.validated` (success), required by the spec whenever the cryptographic signature checks out.
  - `cawg.x509.signature.mismatch` (failure) for an actual cryptographic mismatch (`RawSignatureValidationError::SignatureMismatch`) — previously this case was silently swallowed with no status code at all.

**The shared `crypto::cose` module itself (`verifier.rs`, `certificate_profile.rs`, `sign1.rs`, `ocsp.rs`) is untouched** — it is also used to validate the C2PA claim generator signature, which must keep emitting the existing C2PA status codes. Confirmed via `git diff --stat` that no files under `sdk/src/crypto/` changed.

### Mapping applied

| Old (C2PA) code | New (`cawg.x509.*`) code |
|---|---|
| `algorithm.unsupported` | `cawg.x509.algorithm.unsupported` |
| `signingCredential.trusted` | `cawg.x509.credential.trusted` |
| `signingCredential.untrusted` / `signingCredential.invalid` | `cawg.x509.credential.untrusted` |
| `signingCredential.expired` | `cawg.x509.signature.outside_validity` |
| `claimSignature.mismatch` (malformed COSE_Sign1) | `cawg.x509.signature.mismatch` |
| _(previously unlogged)_ raw signature mismatch | `cawg.x509.signature.mismatch` |
| _(previously unlogged)_ signature validated | `cawg.x509.signature.validated` |

This covers every status code actually reachable today given the current implementation (OCSP and timestamp validation are not yet implemented for CAWG X.509 assertions, so the corresponding `cawg.x509.ocsp.*` / `cawg.x509.time_stamp.*` / `cawg.x509.time_of_signing.*` codes from the spec aren't wired up in this PR).

## Test plan

- [x] `cargo test -p c2pa --lib identity::` — all 80 non-ignored identity tests pass (updated 6 test assertions across `x509_signature_verifier.rs` and `continue_when_possible.rs` to expect the new codes / the new `cawg.x509.signature.validated` success log entry).
- [x] `cargo test -p c2pa --lib` — full suite passes except 3 pre-existing, unrelated failures (`libpng-test.png` fixture not available in this sandbox; confirmed failing identically on `main`).
- [x] `cargo check --workspace` and `cargo clippy -p c2pa --lib --tests --features file_io -- -D warnings` clean.
- [x] `cargo +nightly fmt --all -- --check` clean.
- [x] Confirmed no files under `sdk/src/crypto/` changed (shared C2PA signature validation path untouched).


sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -346,35 +346,67 @@ impl IdentityAssertion {
...
+                            .map(|cert_info| (cose_sign1, cert_info))
+                            .map_err(|e| match e {
+                                CoseError::RawSignatureValidationError(
+                                    RawSignatureValidationError::SignatureMismatch,
+                                ) => {
+                                    log_current_item!(
+                                        "signature mismatch",
+                                        "validate_partial_claim"
+                                    )
+                                    .validation_status("cawg.x509.signature.mismatch")
```
ok-nick: Can we add these validation codes as constants as we do for C2PA validation.

sdk/src/identity/x509/mod.rs
```diff
@@ -28,4 +28,144 @@ pub use x509_credential_holder::X509CredentialHolder;
...
+///
+/// The `crypto::cose` validation code is shared with C2PA claim signature
+/// validation and must keep emitting the original C2PA status codes, so this
+/// function rewrites the codes after the fact instead. Only log items added to
+/// `status_tracker` at index `first_new_item` or later are considered, so that
+/// log items unrelated to this X.509 signature check (for example, an earlier
+/// step in identity assertion validation) are left untouched.
+///
+/// [§8.2.2, "Validating the COSE signature,"]: https://cawg.io/identity/1.3/#_validating_the_cose_signature
+pub(crate) fn remap_x509_cose_status_codes(
```
ok-nick: Ah, this doesn't seem great but it looks like it's our only option right now.

sdk/src/identity/identity_assertion/assertion.rs
scouten-adobe: Good call — added named constants (`CAWG_X509_CREDENTIAL_TRUSTED`, `CAWG_X509_CREDENTIAL_UNTRUSTED`, `CAWG_X509_SIGNATURE_VALIDATED`, `CAWG_X509_SIGNATURE_MISMATCH`, `CAWG_X509_ALGORITHM_UNSUPPORTED`, `CAWG_X509_SIGNATURE_OUTSIDE_VALIDITY`) in `validation_results.rs` alongside the existing C2PA codes and `CAWG_ICA_UNTRUSTED_ISSUER`, and swapped every raw string literal (production code and tests) for the constant. Also wired the two success codes into `log_kind()`. See 17b9a6d.

ok-nick (approved):

sdk/src/identity/identity_assertion/assertion.rs
ok-nick, line 1: Do these new validation codes affect validation state in any way different than it does now?

scouten-adobe: Yes. Previously (old CAWG codes == C2PA codes) would cause somebody reading the overall manifest to think the entire manifest was invalid/untrusted if the CAWG X.509 cert failed validation. This PR is intended to avoid that conflict. C2PA failure status codes are now only issued if the claim generator signature fails validation.

scouten-adobe:
Following up on a question from @ok-nick about whether these new status codes change validation state: they do, in one specific case, and it's now fixed.

`ValidationResults::validation_state()` only tolerates one exact literal string (`signingCredential.untrusted`) as "failure, but the manifest is still `Valid`"; any other code in the failure list makes the whole manifest `Invalid`. Before this PR, the CAWG X.509 untrusted-credential case reused that exact C2PA string, so it was *accidentally* tolerated by that check. Renaming it to `cawg.x509.credential.untrusted` removed that accidental tolerance -- verified empirically (same test, old code string → `Valid`, new code string → `Invalid`).

Fixed in f30efa4: `validation_state()` and `ValidationFailureSummary::fmt` now treat `cawg.x509.credential.untrusted` the same way as `signingCredential.untrusted`, via a shared `is_untrusted_credential_code()` helper, restoring the original tolerance. Added a regression test at the `ValidationResults` unit level and an end-to-end assertion in the `test_cawg_signing_via_settings` integration test so this can't silently regress again.

sdk/src/identity/identity_assertion/assertion.rs
tmathern: The case of trust: So if only CAWG assertions would be untrusted for some reason, only those assertions are reported as untrusted, but the whole C2PA-manifest itself might still be trusted if C2PA-wise everything is trusted?

Other cases: What when the CAWG error codes are `cawg.x509.algorithm.unsupported` or `cawg.x509.signature.mismatch`? They also stay CAWG-only and do not propagate to the validation state?

sdk/src/identity/x509/mod.rs
```diff
@@ -28,4 +28,176 @@ pub use x509_credential_holder::X509CredentialHolder;
...
+                validation_status::CAWG_X509_ALGORITHM_UNSUPPORTED
+            }
+
+            validation_status::SIGNING_CREDENTIAL_TRUSTED => {
+                validation_status::CAWG_X509_CREDENTIAL_TRUSTED
+            }
+
+            validation_status::SIGNING_CREDENTIAL_UNTRUSTED
+            | validation_status::SIGNING_CREDENTIAL_INVALID => {
+                validation_status::CAWG_X509_CREDENTIAL_UNTRUSTED
```
tmathern: `Untrusted` and `Invalid` map to the same now?
(Isn't untrusted a failure in the CAWG trust chain? And invalid an invalid CAWG claim?).

sdk/src/identity/x509/x509_signature_verifier.rs
```diff
@@ -301,7 +337,142 @@ mod tests {
...
+            .unwrap_err();
+        assert!(matches!(malformed_err, ValidationError::SignatureError(_)));
+        assert_eq!(malformed_st.logged_items().len(), 1);
+        assert_eq!(
+            malformed_st.logged_items()[0]
+                .validation_status
+                .as_ref()
+                .unwrap()
+                .as_ref() as &str,
+            CAWG_X509_SIGNATURE_MISMATCH
```
tmathern, line 482: I don't understand this check here.

Is `CAWG_X509_SIGNATURE_MISMATCH` not considered a cryptographic mismatch? Because just above the test says this in comments:
https://github.com/contentauth/c2pa-rs/blob/f30efa41f3a8596e5f10f6b6673f36c5f5adb6b2/sdk/src/identity/x509/x509_signature_verifier.rs#L456-L457

sdk/src/identity/x509/x509_signature_verifier.rs
```diff
@@ -116,25 +101,72 @@ impl SignatureVerifier for X509SignatureVerifier<'_> {
         c2pa_cbor::to_writer(&mut signer_payload_cbor, signer_payload)
             .map_err(|_| ValidationError::InternalError("CBOR serialization error".to_string()))?;
 
-        let cose_sign1 = parse_cose_sign1(signature, &signer_payload_cbor, status_tracker)?;
+        let first_new_item = status_tracker.logged_items().len();
```
tmathern: Isn't that implicit coupling with the internals, because it may depend on the call orders? to find the first new one?

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -346,35 +347,67 @@ impl IdentityAssertion {
...
-                    )
-                    .await
-            }
-            .map_err(|e| match e {
-                CoseError::RawSignatureValidationError(
-                    RawSignatureValidationError::SignatureMismatch,
-                ) => ValidationError::SignatureMismatch,
-                e => ValidationError::SignatureError(e.to_string()),
-            })?;
+            let first_new_item = status_tracker.logged_items().len();
```
tmathern: Isn't that implicit coupling with the internals, because it may depend on the call orders? to find the first new one?

cdmurph32 (approved):
LGTM from a c2pa-js perspective

sdk/src/identity/identity_assertion/assertion.rs
scouten-adobe: Good question, and it generalizes further than I'd first implemented. Fixed in 677482b6: `is_tolerated_manifest_failure_code` now excludes *any* `cawg.x509.*` failure code (not just the untrusted-credential one) from the overall `validation_state()` computation, via a prefix match. So yes: `cawg.x509.algorithm.unsupported`, `cawg.x509.signature.mismatch`, etc. are now also scoped to the identity assertion and won't flip the enclosing manifest to `Invalid` on their own.

sdk/src/identity/x509/mod.rs
scouten-adobe: You're right, and this was a real bug: `SIGNING_CREDENTIAL_INVALID` (a malformed certificate) and `SIGNING_CREDENTIAL_UNTRUSTED` (chain just isn't trusted) shouldn't have shared a code, especially once I made that code tolerated by `validation_state()`. Fixed in 677482b6 -- `SIGNING_CREDENTIAL_INVALID` now maps to its own new `CAWG_X509_CREDENTIAL_INVALID` code instead. This isn't in the ratified v1.3 spec table (which has no code for this distinction), so I filed CAI-13385 to propose adding it upstream.

sdk/src/identity/x509/x509_signature_verifier.rs
scouten-adobe: Sorry, that was a wording bug on my part, not a code bug. A malformed COSE_Sign1 and a genuine cryptographic mismatch are different Rust error variants internally (`SignatureError` vs `SignatureMismatch`), but the CAWG spec doesn't distinguish them at the status-code level -- both are cases where "validation of the signature fails," so both intentionally share `CAWG_X509_SIGNATURE_MISMATCH`. Reworded the comment in 677482b6 to say that explicitly instead of just "distinct."

sdk/src/identity/x509/x509_signature_verifier.rs
scouten-adobe: Agreed, that was fragile. Replaced it in 677482b6 with an `X509StatusRemapGuard` RAII guard (now in its own file, x509_status_remap.rs) -- you construct it, do the signature check inside its scope, and the remap runs automatically when it drops. That works for both the sync and async paths without needing to wrap the branching logic in a closure, and it can't be pulled apart by an intervening log call the way the index-tracking could.

sdk/src/identity/identity_assertion/assertion.rs
scouten-adobe: Same fix as the equivalent comment in x509_signature_verifier.rs -- replaced with the `X509StatusRemapGuard` RAII guard in 677482b6, which removes the implicit index-order coupling entirely.

sdk/src/identity/x509/x509_status_remap.rs
```diff
@@ -0,0 +1,229 @@
...
+/// constructed in. The remap then applies to exactly the log items logged
+/// during that scope, however the check exits it (normal return, early
+/// return, or `?`), without relying on a separately tracked index that a
+/// later, unrelated statement could accidentally be separated from.
+///
+/// [§8.2.2, "Validating the COSE signature,"]: https://cawg.io/identity/1.3/#_validating_the_cose_signature
+pub(crate) struct X509StatusRemapGuard<'a> {
+    status_tracker: &'a mut StatusTracker,
+    first_new_item: usize,
+}
```
ok-nick, line 37: Hm, this seems complicated for something simple. May I suggest an alternative? We can create a new status tracker for the code that needs it, map all the C2PA codes collected to their respective CAWG codes, then append those logs to the main status tracker.

scouten-adobe: Hrm, no, I don't see the simplicity win and I see an increased risk of forgetting to add the remapped status codes to the original status tracker. Keeping it as is.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32420655172](https://github.com/contentauth/c2pa-rs/actions/runs/32420655172).

PR #2517, opened 2026-08-18
chore: Simplify sign_manifest and expose BmffHash Merkle construction

gpeacock:
## Summary
This ensures that Bmff assertions can be constructed using the Bmff assertion and a stream.

Moved some logic from Builder into Store::sign_manifest to make it more consistent
and streamlined - there are no other callers. Updated associated tests/
Fixes an issue where status was not being reported if dynamic assertions were used.
.
- Move the Merkle-leaf-accumulator-to-`MerkleMap` conversion out of
  `Builder::update_hash_from_stream` and onto
  `BmffHash::add_merkle_maps_from_accumulator`, so BMFF hard-binding
  construction logic lives with the assertion type instead of being
  duplicated in `Builder`'s orchestration code.
- Expose `BmffHash::merkle_uuid_boxes()`, returning a new `MerkleUuidBoxes`
  struct that bundles the proof-box bytes, insertion point, and replacement
  range together. 
- Add `test_build_hard_binding_externally_with_merkle_proof_boxes`,
  constructing a full BMFF hard binding entirely through `BmffHash`'s public
  API (no `Store`/`Builder` involved) to prove this now works end-to-end.


sdk/src/store.rs
```diff
@@ -2318,74 +2330,50 @@ impl Store {
...
+        context.check_progress(ProgressPhase::Signing, 1, 1)?;
 
-            if !missing.is_empty() {
-                return Err(Error::BadParam(format!(
-                    "no placeholder slots were reserved for dynamic assertions [{}]; \
-                     call add_dynamic_assertion_placeholders() before signing",
-                    missing.join(", ")
-                )));
-            }
+        // Reserve placeholder slots for any dynamic assertions exposed by the
```
tmathern, line 2335: I assume slotting happens with `label__number`? We are still good if that call is hit multiple times, no "phantom" increases/relabelling?

sdk/src/store.rs
```diff
@@ -2318,74 +2330,50 @@ impl Store {
...
         pc.set_signature_val(sig);
 
-        let jumbf_bytes = self.to_jumbf_internal(signer.reserve_size())?;
+        let mut jumbf_bytes = self.to_jumbf_internal(signer.reserve_size())?;
 
         if context.settings().verify.verify_after_sign {
             self.verify_store_strict(None, context)?;
         }
 
+        if let Some(len) = target_len {
```
tmathern, line 2371: Could it be possible we ever get a jumbf that manages to sneak in here and be larger then the len? Should it error?
(Disclaimer: I did not trace the whole flow, but since this was nicely in the changeset, asking!)

sdk/src/assertions/bmff_hash.rs
```diff
@@ -621,6 +637,54 @@ impl BmffHash {
         Ok(())
     }
 
+    /// Fills in real Merkle maps from an accumulator that collected leaf
```
tmathern, line 640: Once the loop in this function has processed, should the accumulator be marked as cleared/consumed/used? Or is it multi-use safe?

sdk/src/store.rs
mauricefisher64: In general it should not change.  There could some special cases here Gavin is trying to cover.  It would be nice if they were specified.

sdk/src/store.rs
gpeacock: This works the same way it has always worked. We call the same logic to add the dynamic assertion placeholders and write_dyanamic_assertions always replaces, requiring the identical label to exist already.

sdk/src/store.rs
gpeacock: Maybe this isn't necessary and I can just remove it. I'll check

sdk/src/assertions/bmff_hash.rs
gpeacock: I don't think we would ever use it again, but I can clear it just be safe

sdk/src/store.rs
gpeacock: good catch, I'll added an error for that case (except for with BMFF merkel where it is allowed)

sdk/src/assertions/bmff_hash.rs
```diff
@@ -603,6 +619,55 @@ impl BmffHash {
...
+    /// this supports proof rows (not just the Merkle root), since the
+    /// accumulator's leaves retain their per-leaf lengths needed to rebuild
+    /// the tree shape.
+    ///
+    /// The `/mdat` exclusion is added so mdat content is excluded from the
+    /// flat hash (it is covered by the Merkle trees instead). No-op if the
+    /// accumulator has no leaves (i.e. `add_merkle_leaf` was never called).
+    pub fn add_merkle_maps_from_accumulator(
+        &mut self,
+        accumulator: &mut MerkleAccumulator,
```
ok-nick, line 636: Is `MerkleAccumulator` actually publicly accessible?

tmathern (approved):

PR #2523, opened 2026-08-19
ci: run `cargo-deny` async (or block if dep change) and report errors as issues

ok-nick:
We run into issues every so often that blocks all PRs because a new general vulnerability appears, even if it is not specific to that PR. This change runs vulnerability audits (1) nightly, (2) on merge to main/stable/backport branches, and (3) when changing dependencies in a PR, then creates a GitHub issue detailing the problem.

I created an example repo to demonstrate how issues are reported: https://github.com/ok-nick/deny-test/issues

Made with the help of Claude.

### DIFF
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -83,7 +83,7 @@ jobs:
           cargo deny --format json check advisories bans licenses sources \
             2> deny.ndjson 1>/dev/null || true
 
-            # Advisories
+          # Advisories.
           jq -c 'select(.type == "diagnostic" and .fields.severity == "error" and .fields.advisory != null)
                  | {id: .fields.advisory.id, title: .fields.advisory.title, package: .fields.advisory.package, url: .fields.advisory.url}' \
             deny.ndjson \
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -1,0 +1,189 @@
+name: Dependency audit
+
+# Dependency license / vulnerability audit.
+#
+# cargo-deny checks advisories (RUSTSEC), banned crates, licenses, and sources.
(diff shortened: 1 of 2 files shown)


scouten-adobe (approved):
Nice change — verified the demo repo (ok-nick/deny-test) issue history and it shows the title-based dedup working correctly across repeated runs (no duplicate issues for the same advisory/violation), and confirmed `cargo-deny --format json` writes to stderr as the workflow assumes, so the core plumbing checks out.

A few minor points worth considering (not blockers):

1. **Docs/implementation mismatch**: the top-of-file comment says the post-merge/nightly path "opens or updates a GitHub issue," but `ensureIssue()` only creates — if the title already matches an open issue it just logs `Already tracked` and returns, without refreshing the run link or the `cargo-deny` output in the body. If that's intentional (avoid comment spam on unfixed advisories), might be worth tweaking the comment to say "opens" only; otherwise this could use an `issues.update`/`createComment` call so the issue reflects the latest run.

2. **No spam cap for advisories**: `SPAM_CAP` only groups `bans`/`licenses`/`sources` violations into one issue when there are more than 5. Advisories always get one issue each with no cap, so a batch of newly-published RUSTSEC advisories in one run could open many issues at once. Might be fine since each advisory needs its own fix, but worth a sanity check.

3. **Issue listing isn't paginated**: `issues.listForRepo` for existing open `dependency-audit` issues uses `per_page: 100` with no pagination, so once that label accumulates more than 100 open issues, dedup could silently stop working for the overflow.

4. **Redundant `cargo deny check` run**: on failure, the check runs once in "Audit crate dependencies" and again in "Collect failing checks" to capture output. Harmless (this job runs infrequently), just slightly wasteful.

Overall this looks solid and the design is well validated by the linked demo repo.

PR #2524, opened 2026-08-19
fix: Hash a range of 4 GiB or more without hanging on 32-bit targets (CAI-13325)

### LINKED ISSUE #2449: Hashing a range of 4 GiB or more never terminates on 32-bit targets (wasm32)
### Component

c2pa-rs

### Version

0.91.0-dev (`main` at 4444aed), and earlier versions carrying this code

### Platform

Other: 32-bit targets, `wasm32` in particular. 64-bit is unaffected.

### What happened?

Hashing a range of 4 GiB or more never terminates. Signing and verifying spin at 100% CPU with no error and never return.

`hash_stream_by_alg_with_progress`, which the public `hash_stream_by_alg` delegates to, sizes each read in `usize`:

```rust
let mut chunk = vec![0u8; std::cmp::min(chunk_left as usize, MAX_HASH_BUF)];
data.read_exact(&mut chunk)?;
hasher_enum.update(&chunk);
chunk_left -= chunk.len() as u64;
```

`usize` is 32 bits on `wasm32`, so `chunk_left as usize` truncates. For a range of exactly 4 GiB it is 0:

- the minimum is 0, so the buffer is empty
- `read_exact` on an empty buffer returns `Ok(())` immediately, touching no data
- `chunk_left -= 0` leaves the counter unchanged

The loop then repeats forever. It performs no I/O and allocates nothing, so it is a pure spin. For other lengths above 4 GiB the truncation picks the wrong branch of the minimum instead.

The progress denominator truncates the same way and reports zero total chunks for a range of exactly 2^32:

```rust
(len as usize).div_ceil(MAX_HASH_BUF) as u32
```

The range is hashed and the call returns, on any target width.

### Reproducing

The arithmetic is the whole bug, so this needs no 4 GiB file. Running the loop above verbatim on `wasm32-wasip1` under wasmtime, against a reader that yields zeros for a declared length of `1 << 32`:

```
usize is 32 bits; (1u64<<32) as usize = 0
BUGGY : DID NOT TERMINATE
FIXED : terminated in 1024 iterations
```

Through the public API it needs an asset of 4 GiB or more, because the `data_len < range_end` check rejects a smaller file that declares a large range. So this is a reliability bug on large media, not a small-input denial of service.

### Fix

#2448. Same class on 32-bit/wasm as `test_cai_chunk_length_near_u32_max_returns_error` in `png_io`.

scouten-adobe:
## Summary

Fixes [CAI-13325](https://jira.corp.adobe.com/browse/CAI-13325), the internal mirror of #2449: hashing a range of 4 GiB or more never terminated on 32-bit targets (wasm32).

Fixes #2449.

`hash_stream_by_alg_with_progress_impl` sized each read chunk with `chunk_left as usize`, truncating the `u64` byte count on targets where `usize` is 32 bits. For a range of exactly 4 GiB (2^32), the truncated value was `0`, so:
- the chunk buffer was empty,
- `read_exact` on an empty buffer returned `Ok(())` immediately without touching any data,
- `chunk_left -= 0` left the counter unchanged,

producing a pure infinite spin (no I/O, no allocation). For ranges above 4 GiB, the truncation picks the wrong branch of `min` instead. The progress-total computation had the same truncation, reporting zero total chunks for a range of exactly 2^32.

## Fix

Compute chunk sizes (and the progress denominator) in `u64` and narrow to `usize`/`u32` only *after* bounding by `max_hash_buf`, which is always well within range for both types. This is the same pattern already used in `png_io.rs` (widen, then narrow with a checked conversion) rather than narrowing first.

The repeated chunk-size expression is extracted into a single `chunk_size(chunk_left, max_hash_buf)` helper (used at all 3 call sites), and the progress-total narrowing uses a checked `u32::try_from(...).unwrap_or(u32::MAX)` instead of an unchecked `as u32`.

Note: a third-party PR ([#2448](https://github.com/contentauth/c2pa-rs/pull/2448)) independently diagnosed the same root cause with a similar `chunk_size` helper, but it predates a since-landed refactor that made the hash buffer size configurable for tests (`hash_stream_by_alg_with_progress_impl`'s `NonZeroUsize` parameter), so it's currently in conflict with `main`. This PR's `chunk_size` helper takes `max_hash_buf` as a parameter rather than hardcoding the module constant, so it stays compatible with that test-configurability.

## Test plan

- [x] `chunk_size_does_not_truncate_on_32_bit_targets` unit-tests the helper against edge values (0, 1, cap, cap+1, `2^32`, `u64::MAX`).
- [x] `hash_range_of_exactly_4gib_terminates` drives the real hashing loop across a virtual 4 GiB range (via a synthetic zero-filled `Read + Seek` stream that doesn't allocate 4 GiB) so the regression is caught for real under the repo's `wasm32-wasip2` CI job, where `usize` is genuinely 32 bits.
- [x] `cargo test -p c2pa --lib utils::hash_utils` passes (9/9) on the host target.
- [x] `cargo clippy --lib --tests` clean for the changed file.
- [x] `cargo +nightly fmt --all -- --check` clean.


tmathern (approved):

cdmurph32 (approved):
LGTM. Thanks

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32889492660](https://github.com/contentauth/c2pa-rs/actions/runs/32889492660).

stupside:
@scouten-adobe Thanks for the fix, I didn't get any feedbacks on my contribution (I opened a PR some weeks ago to solve the issue), I will close my PR.

PR #2527, opened 2026-08-20
fix: Enforce zero-content rule for all redacted assertion slots

mpgopala:
## Summary

Fixes [CAI-12895](https://jira.corp.adobe.com/browse/CAI-12895) — a manifest-integrity bypass (H1 report 3856552 / VULN-36847, CVSS 6.5 Medium, CWE-345) where forged content planted in a redacted assertion slot is accepted as a genuine, `Trusted` assertion.

## The vulnerability

Redaction enforcement gated the spec's "a redacted slot must contain only zero bytes" rule on the box at that URI being the `C2PA_REDACTION_UUID` placeholder type (`store.rs`, `get_assertion_from_jumbf_store`). An attacker can byte-edit an already-signed asset to plant an ordinary CBOR/JSON/binary assertion box at a URI listed in a claim's `redacted_assertions`:

- The zeros check never runs — a non-placeholder box takes a different parse arm.
- The ingredient manifest hash mismatch is suppressed — spec-required for redactions (§15.11.3.3.1), and the compensating control was incomplete.
- The forged box surfaces as a real assertion; the manifest still validates `Trusted` with zero failures.

**No signing key required:** the attacker touches only the ingredient's assertion-store bytes; the ingredient claim and the redacting parent claim (and their signatures) are untouched, and the integrity bindings that would catch the edit are exactly the ones redaction declaratively switches off.

## The fix

Per spec 2.x §15.11.3.3.1, the rule is **driven by the redaction list and is type-agnostic**: *any* box present at a redacted URI must contain only `0x00` bytes. Added a single pass in `Store::verify_store` (after `svi` is built, shared by sync/async) over `svi.redactions` that resolves each redacted URI and rejects a present, non-zero box with `assertion.notRedacted`, regardless of box type.

- Absent assertions (the honest "removed entirely" form the c2pa-rs builder produces) are skipped — so legitimate redactions are unaffected.
- The spec-mandated `activeManifest` hash-mismatch suppression is left untouched (it's correct, not a second bug).
- Placement is store-level rather than in the per-box load-time parser, because redactions are declared by a *parent* claim that may load after the ingredient — the full set is only known once `get_store_validation_info` aggregates it.

## Performance

Near-zero overhead: the check runs once over the redaction list (empty for the overwhelming majority of assets) against already-parsed, in-memory boxes. No extra asset pass, no hashing, no re-parse — just a linear `is_zero` scan of a few small boxes on assets that actually use redaction.


sdk/src/store.rs
```diff
@@ -2039,6 +2039,31 @@ impl Store {
...
+            // A fully removed assertion (resolution fails) is a valid form of redaction;
+            // only a *present* box with non-zero content is a violation.
+            if let Ok(claim_assertion) = store.get_claim_assertion_from_uri(redacted_uri) {
+                if !is_zero(claim_assertion.assertion().data()) {
+                    log_item!(
+                        redacted_uri.clone(),
+                        "redacted assertion data must be zeros or empty",
+                        "verify_store"
+                    )
+                    .validation_status(validation_status::ASSERTION_NOT_REDACTED)
```
tmathern, line 2058: Is this one here considered a failure due to suspected tampering?
Previously, in the Reader there was a change to make it informational:
https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/reader.rs#L1065-L1070

sdk/src/store.rs
```diff
@@ -2039,6 +2039,31 @@ impl Store {
...
+        // Per spec §15.11.3.3.1, any box present at a redacted URI must contain only 0x00
+        // bytes, regardless of its JUMBF type — reject with `assertion.notRedacted` otherwise.
+        // This is driven by the redaction list (not the box type), and is the compensating
+        // control for suppressing the ingredient manifest hash mismatch in `ingredient_checks`:
+        // without it, forged content planted in a redacted slot would surface as a genuine,
+        // hash-unverified assertion. See CAI-12895.
+        for redacted_uri in &svi.redactions {
+            // A fully removed assertion (resolution fails) is a valid form of redaction;
+            // only a *present* box with non-zero content is a violation.
+            if let Ok(claim_assertion) = store.get_claim_assertion_from_uri(redacted_uri) {
```
tmathern, line 2051: Do databoxes need similar care, or is this only affecting assertions?
https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/claim.rs#L1861-L1873

tmathern (commented):
2 questions on redaction behavior, but already looking good.

sdk/src/store.rs
mpgopala: The redacted assertions are put into a special data box with `C2PA_REDACTION_UUID` (CAA98EEE-9D4D-F80E-86AD-4DFFCA263973). c2pa makes sure to check the contents of such data box is zero ([Reference](https://github.com/contentauth/c2pa-rs/blob/24d17555beafb70c15e1e1e4054ac3c06fbba1c0/sdk/src/store.rs#L971)).

sdk/src/store.rs
mpgopala: This is according to §15.11.3.3.1 ("Claim Signature Hash Validation Method") of the spec.

tmathern: But isn't that inconsistent then, if we have two different handlings depending on where the error is found/reported?

sdk/src/store.rs
tmathern: I was asking about redactions that target databoxes. Could the vulnerability here also be hit through databoxes? What happens when a redacted URI points in the databox store rather than the assertion store? Do these changes also handle that?

mpgopala: My understanding here is that the redacted URI points to data boxes and if someone changes that data box. This should not be an issue as the contents of a data-box that is not selected for hashing is similar to closing your eyes to a theft and then complaining that theft happened. @ssanthosh is handling a couple of vulnerabilities related to this and such issues should be rejected.

sdk/src/store.rs
mpgopala: Even without this change, there is already a mismatch in handling ASSERTION_NOT_REDACTED. claim.rs makes it failure and reader.rs makes it informational. I think it's OK because the error handling depends on the severity of the issue. Another example is ASSERTION_MISSING.

tmathern: Let's confirm with @gpeacock too then... ?

tmathern (approved):
Please make sure there is a confirmation regarding the reported status differing, the thread starting with https://github.com/contentauth/c2pa-rs/pull/2527/changes#r3825670544.

sdk/src/store.rs
gpeacock: I suspect we will end up adding these kinds of assertions twice. 

This code will now, correctly flag an existing non-zeroed, redacted assertion as a ASSERTION_NOT_REDACTED failure.

The check in reader doesn't have a zero check, so it will report a ASSERTION_NOT_REDACTED as informational regardless if it was zeroed or not. That is incorrect, and we probably don't have a test for that, since we always remove them entirely.

With the redaction checks already in the validation code, we can probably remove them from reader, except that we still need to check for missing assertions, and we can't count redacted assertions as missing. So maybe the reader check needs to add the existing, but zeroed out redacted assertions to redacted so it can correctly count missing.

gpeacock (changes requested):

sdk/src/store.rs
tmathern: (Having a test here for this would actually prevent a regression in the future too if this code moves around. Switching review to ask for the test.).

mpgopala: @gpeacock @tmathern : I have addressed the review comments. Please check again.

gpeacock (approved):

PR #2529, opened 2026-08-20
fix: Exclude RIFF header length field from data hash exclusions

byh-trufo:
## Problem

The RIFF container's L0 chunk length field (bytes 4–8) is rewritten whenever the embedded C2PA manifest store changes size. RIFF data-hash exclusions currently cover only the manifest store chunk, so any in-place store growth invalidates the parent manifest's hard binding — the recomputed hash includes the rewritten length field.

This makes update manifests unusable on RIFF assets, even though the spec explicitly sanctions the operation (C2PA 2.4, update manifests): *“In the case of a file offset hash (c2pa.hash.data), the C2PA Manifest Store has to continue to start at the same file offset after updating - only its length may change.”* On RIFF, the length change itself mutates a hashed byte, so a spec-compliant update always produces `assertion.dataHash.mismatch` on the parent and the asset grades invalid.

Byte-exact confirmation: on a WebP re-embed, the only bytes that differ outside the (excluded) manifest store are offsets 4–5; restoring the original length field makes the parent's recorded digest match exactly.

## Fix

Report the L0 length field as an `OtherExclusion` in `RiffIO::get_object_locations_from_stream` — the same mechanism `tiff_io` uses for its mutable structures — so newly signed RIFF assets record a `{start: 4, length: 4}` data-hash exclusion alongside the store exclusion. Validators honor recorded exclusions as-is, so no validation-side change is needed; the extra exclusion only surfaces the existing informational `assertion.dataHash.additionalExclusionsPresent` code. This parallels the spec's own handling of mutable derived fields elsewhere (e.g. font `checkSumAdjustment` is treated as zero when hashing).

## Verification

- New unit test asserts the exclusion is reported for a RIFF stream; `cargo test -p c2pa --lib riff`: 14 passed.
- Full lib suite: 911 passed; the 3 failures (`test_png_jumbf_generation` etc., missing `libpng-test.png` in the embedded fixture registry) are pre-existing on main.
- End-to-end: WebP signed with this branch records both exclusions; a second signing on top (store grows, length field rewritten) validates with **no hash failures on an unmodified 0.78-based validator** — previously the identical operation produced `assertion.dataHash.mismatch` on the parent manifest.


scouten-adobe (approved):
**One thing worth flagging, not a blocker:** excluding this field does slightly weaken the hard binding — before this change, tampering with the declared top-level RIFF size would invalidate the C2PA hash; after, it won't. A forged size (e.g., shrunk to end before the manifest) wouldn't help an attacker hide content _from c2pa-rs itself_ (its own reader also walks children using that same declared size, so it would just fail to find the manifest — same "no C2PA data" outcome as today). But other RIFF consumers (media players, browsers) that respect the declared size independently could render the trailing bytes differently while the asset still validates. This is the same accepted trade-off the spec sanctions for TIFF's IFD count and font checkSumAdjustment, so I'd call it in-line with precedent rather than a defect — worth a one-line mention in the PR discussion in case a reviewer wants CAI's security team to weigh in, but not something I'd hold up merging over.

Overall, the fix is small, well-scoped, matches an existing pattern elsewhere in the codebase, is backward-compatible, and is covered by a targeted unit test. :+1:

Requesting a second review from @gpeacock.

scouten-adobe:
@byh-trufo on further review, I think this PR runs the risk of getting ahead of the spec. Would you please raise this in C2PA TWG and reopen this when it's accepted there?

byh-trufo:
yup will do!

on a related note, zip file formats also don't support update manifests for a similar reason, but the additional exclusions required for zip would be more substantial (likely not worth doing?)

PR #2532, opened 2026-08-21
feat: Simplify error handling in `IcaSignatureVerifier` methods

imtherealnaska:
The cross_check_signer_payload function previously communicated its result through a mutable output parameter (&mut bool), which the caller had to declare beforehand and pass in alongside the other arguments. This meant the function's signature didn't actually reflect that it produced a meaningful result — a reader had to look inside the function body to discover that ok was being written to.

This has been refactored so that the function now returns its boolean result directly as part of its Result type (Result<bool, E>), following Rust's convention of communicating outcomes through return values rather than through mutation of caller-owned state. At the call site, the result is now accumulated into the running ok flag using 
ok &=self.cross_check_signer_payload(...)?;, where the ? operator still handles the error case by propagating it immediately, and the &= folds the boolean outcome into the overall validation state.

## Checklist
- [x] All applicable changes have been documented. ( refactor only)


scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32889248913](https://github.com/contentauth/c2pa-rs/actions/runs/32889248913).

PR #2535, opened 2026-08-21
fix(sdk): Prevent ingredient assertions from canceling active-manifest failures (CAI-12751)

mpgopala:

Fixes a manifest validation bypass (CAI-12751) in `ValidationResults::from_store`.

### Problem

When building a validation report, `from_store` de-duplicates the live validation
log against statuses embedded in ingredient assertions, so that a failure already
disclosed by an ingredient isn't reported twice. It decided whether a given status
belonged to the **active manifest** (and must never be suppressed) by parsing the
status URL via `manifest_label_from_uri`.

That URL check does not recognize the **bare manifest-label form** (`urn:c2pa:...`),
which is exactly how a genuine active-manifest failure is logged (e.g.
`assertion.action.malformed`). Such a failure was therefore misclassified as an
ingredient status and dropped whenever an ingredient assertion attested a status
with a matching `code` + `url` + `kind`.

Because ingredient assertion content is fully attacker-controlled, an attacker could
author a manifest whose active claim violates a mandatory rule (e.g. a `c2pa.actions`
assertion whose first action is not `created`/`opened`) and add an ingredient that
"attests" that same failure against the active claim's bare label. On validation the
real failure was removed from both the top-level `validation_status` list and
`validationResults.activeManifest.failure`, so the asset read back as **Valid** — or
**Trusted**, when the signing certificate is in the verifier's trust list — with an
empty error list. The suppression also defeated the SDK's own pre-sign self-check, so
the malformed asset could be produced through the public `Builder`/`sign` API.

### Fix

Decide active-manifest membership using `ingredient_uri` instead of URL parsing. The
status tracker sets `ingredient_uri` only while inside an ingredient's recursive-
validation scope, so a `None` value reliably identifies a status the validator
produced for the active manifest and cannot be influenced by ingredient assertion
content. Both the pre-filter gate and the `retain` now key on `ingredient_uri`, and
the now-unused URL-based helper is removed. Legitimate ingredient-delta
de-duplication (statuses with `ingredient_uri` set) is unchanged.

### Tests

Adds `from_store_ingredient_cannot_suppress_active_manifest_failure`, which attests an
active-manifest failure from within an ingredient and asserts the failure is retained
and the state stays `Invalid`. Verified to fail on the pre-fix code and pass after the
fix; the existing legitimate-dedup test
(`from_store_attested_ingredient_failure_after_serde_roundtrip`) still passes. Full
`c2pa` lib suite, `make fmt`, and `make clippy` (`-D warnings`) are green.


tmathern (approved):

sdk/src/validation_results.rs
```diff
@@ -223,9 +193,20 @@ impl ValidationResults {
...
+                // tracker sets only while inside an ingredient's recursive-validation scope. A
+                // `None` value therefore reliably identifies a status the validator produced for
+                // the active manifest itself, and cannot be influenced by ingredient assertion
+                // content. Deciding membership by parsing the status URL would be unsafe: an
+                // active-manifest failure can carry the bare manifest-label form (`urn:c2pa:...`),
+                // which does not resolve to the active manifest, so it would be misclassified as
+                // an ingredient status and could be cancelled by a matching attestation.
                 statuses.retain(|s| {
-                    is_active_manifest(s.url()) || !ingredient_statuses.iter().any(|i| i == s)
+                    s.ingredient_uri().is_none() || !ingredient_statuses.iter().any(|i| i == s)
```
tmathern: Can we make sure there are no side-effects due to early exits (any ? propagated errors, for instance)? Asking because this seems to be part of check loops, in case the loop has early exits and those need handling (eg.  state of intermediary variables).

sdk/src/validation_results.rs
```diff
@@ -199,9 +193,20 @@ impl ValidationResults {
...
```
dcondrey: Swapping the `url` predicate for `ingredient_uri` changes which axis is being tested, and I think it opens a different suppression window than the one it closes.

`ingredient_uri` records that the validator was inside an ingredient recursion scope — not that the status is *about* an ingredient. `Store::ingredient_checks` pushes an ingredient URI at `store.rs:1650` and then logs failures whose `log_item!` label is `to_assertion_uri(claim.label(), &i.label())`, i.e. a URI inside the **parent** manifest. When the parent is the active manifest, those statuses have `ingredient_uri == Some` together with an active-manifest url, so this change makes them dedup-eligible where `is_active_manifest(s.url())` retained them unconditionally.

The one that matters is `assertion.ingredient.malformed` at `store.rs:1659-1668` ("ingredient V3 must have validation results"). It is a Failure and it gates `validation_state`. `ingredient_statuses` is collected across every claim in the store, `PartialEq` compares only code/url/kind, and `to_absolute_uri` returns an already-absolute `self#jumbf=/c2pa/...` verbatim — so a second ingredient assertion carrying that exact code+url cancels it.

I ran your new test's shape with the failure logged the way `ingredient_checks` actually logs it (inside `push_ingredient_uri`, url = an active-manifest assertion URI). On this branch the results contain no failure anywhere and `validation_state()` is `Valid`; on `main` the same input lands in `ingredientDeltas` and stays `Invalid`.

Would keeping both guards cover it — `s.ingredient_uri().is_none() || is_active_manifest(s.url()) || !ingredient_statuses.iter().any(|i| i == s)`?

scouten-adobe (changes requested):
Blocking merge until @mpgopala is back from PTO (this week) and has a chance to review @dcondrey's feedback.

sdk/src/validation_results.rs
mpgopala: It's not an early exit but stricter filter. Also, these statuses are added to the results vector in the end. So, it should be fine?
I have made sure all the tests are passing.

sdk/src/validation_results.rs
mpgopala: Good catch! I agree with your analysis and have made the change you suggested (Brought back the previous check is_active_manifest).

mpgopala:
> Blocking merge until @mpgopala is back from PTO (this week) and has a chance to review @dcondrey's feedback.

@scouten-adobe : I have addressed @dcondrey's comment.

scouten-adobe (approved):
All looks good to me and I think the feedback from @tmathern and @dcondrey has been resolved.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33413533757](https://github.com/contentauth/c2pa-rs/actions/runs/33413533757).

PR #2538, opened 2026-08-24
fix: Harden against invalid  labels as CAWG identity hard bindings

ssanthosh:
### Vulnerability
CAWG identity assertion validation accepted any label starting with `c2pa.hash.` as proof a hard binding was present — including a fake one Core never actually treats as a real hash binding. This let an attacker replay a victim-signed identity assertion onto unrelated content and still get a `Trusted` result.

### Fix
Extracted the correct predicate (already used by `CloudData::is_hard_binding()`) into a shared `labels::is_hard_binding_label()` function, and used it at both CAWG check sites in `signer_payload.rs` instead of the naive prefix match. 

### Tests added
`test_is_hard_binding_label` — covers all real hard-binding labels (including BMFF's versioned variants) plus the exact `c2pa.hash.fake` exploit label and other near-misses.

## Checklist

### DIFF
--- a/sdk/src/assertions/cloud_data.rs
+++ b/sdk/src/assertions/cloud_data.rs
@@ -134,12 +134,7 @@ impl CloudData {
     ///
     /// [`assertion.cloud-data.hardBinding`]: crate::validation_results::validation_codes::ASSERTION_CLOUD_DATA_HARD_BINDING
     pub(crate) fn is_hard_binding(&self) -> bool {
-        let l = self.label.as_str();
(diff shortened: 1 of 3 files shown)


tmathern (approved):

scouten-adobe (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32867934239](https://github.com/contentauth/c2pa-rs/actions/runs/32867934239).

PR #2540, opened 2026-08-25
fix: x.509 cert with missing org should not be a reporting failure

### LINKED ISSUE #2262: Cryptographically valid claim signature reported as claimSignature.mismatch when signing certificate has no Organization (O) subject attribute
### Component

c2pa-rs

### If Other, please specify

_No response_

### Version

0.26.68

### Platform

macOS

### What happened?

Running validation on my signed RAW image (Sony Alpha 1 with CAWG Individual Creator certificate from ssl.com) fails with claimSignature.mismatch.

I believe what's happening is that verify_signature (https://github.com/contentauth/c2pa-rs/blob/bdc32480d6bddb3d8943a35c9849f67ec41eb515/sdk/src/crypto/cose/verifier.rs#L159) returns an Error when the signing cert doesn't include an O (Organization) field in its subject. (This is the case for Individual Creator certificates from ssl.com - they understandably don't include an O field.) This seems to be checked _after_ the signature is validated (and only if the validation is successful).

Then, any kind of error during signature checking is translated into an CLAIM_SIGNATURE_MISMATCH in https://github.com/contentauth/c2pa-rs/blob/bdc32480d6bddb3d8943a35c9849f67ec41eb515/sdk/src/claim.rs#L3007. Which is the error code that I receive when running the c2patool on my image.

When I deploy Sony's "default certificate" (which _does_ have an O attribute in its subject) to my camera, the image verifies.

The photo should have verified.

Under CA/Browser Forum S/MIME rules, individual-validated profiles cannot include an O= attribute (it is reserved for organization-validated/sponsored profiles). These are exactly the certificate types used for CAWG individual creator credentials. These types of certificate _shouldn't_ include an O= attribute, but they _should_ be usable as signing certificates.

Even if such certificates aren't meant to be used as signing certificates, the failure is still misclassified. It should be enforced in check_certificate_profile alongside the other profile checks and surface as signingCredential.invalid with a descriptive message — not as a post-verification failure reported under claimSignature.mismatch, which falsely tells the user the signature failed cryptographic validation.

mauricefisher64:
Fix issue where informational data was being logged as a failure.

## Checklist


ssanthosh (commented):
Can you please add a regression test for the fix ?

scouten-adobe (approved):

mauricefisher64:
> Can you please add a regression test for the fix ?

There is a test for this:  **_test_signature_structure_**

The None case should have been allowed since this is just an informational structure.  Notice the value is optional.  It was never meant for validation.

ssanthosh (approved):

sdk/src/crypto/cose/verifier.rs
```diff
@@ -141,15 +141,13 @@ impl Verifier<'_> {
             .iter_organization()
             .map(|attr| attr.as_str())
             .last()
-            .ok_or(CoseError::MissingSigningCertificateChain)?
-            .map(|attr| attr.to_string())
-            .map_err(|_| CoseError::MissingSigningCertificateChain)?;
+            .and_then(|attr| attr.map_or(None, |a| Some(a.to_string())));
```
scouten-adobe: Requesting change: this line fixes the reported bug (cert with no Organization attribute no longer hard-fails signature verification), but there's no test exercising it. Please add a unit test that calls `verify_signature` with a signing cert lacking an `O=` attribute and asserts it succeeds with `issuer_org: None`, so a future refactor of this chain can't silently regress it.

scouten-adobe: Minor/non-blocking, could be follow-up issues instead:
1. `.and_then(|attr| attr.map_or(None, |a| Some(a.to_string())))` can be simplified to `.and_then(|attr| attr.ok()).map(|a| a.to_string())`, matching the `.ok()`-based "ignore error" idiom already used twice above in this function.
2. This now silently treats "no Organization attribute" and "Organization attribute present but not valid UTF-8" identically as `None`. A short comment noting that's a deliberate simplification would help the next reader.

scouten-adobe (dismissed):
Thanks for the fix — this correctly stops treating a cert's missing/non-UTF8 Organization attribute as a hard verification failure. Requesting one change before merge: please add a test covering `Verifier::verify_signature` with a signing cert that has no Organization attribute, since that's the exact scenario this PR fixes and there's currently no test guarding it. Two minor, non-blocking notes inline — happy to have those filed as follow-up issues instead if you'd rather not fold them into this PR.

sdk/src/crypto/cose/verifier.rs
mauricefisher64: Please file an issue for that.  It will require a new cert chain to be generated since all of our test certs have org.

mauricefisher64: Done

sdk/src/utils/io_utils.rs
```diff
@@ -22,7 +22,9 @@ use std::{
 #[allow(unused)] // different code path for WASI
 use tempfile::{tempdir, Builder, NamedTempFile, SpooledTempFile, TempDir};
 
-use crate::{asset_io::rename_or_move, utils::path_utils::sanitize_archive_path, Error, Result};
+#[cfg(feature = "file_io")]
```
tmathern, line 25: Did that sneak in?

tmathern (approved):

sdk/src/utils/io_utils.rs
mauricefisher64: I guess.  I got a build error so I fixed it.

ok-nick (approved):

scouten-adobe (approved):

PR #2543, opened 2026-08-25
Suggested fixes for box-hash exclusions review (#2513)

scouten-adobe:
## Summary

Follow-up on the review of #2513 (general box-hash exclusions). Four small, independently-buildable commits addressing findings from a code review of that PR:

- **`5847df0d`** — `split_exclusions` dropped an independent check (present in the earlier `BoxKind` design) that an exclusion range must not reach past its box's real length, relying instead on every format handler's `AllowedExclusion` happening to already be self-bounded. Restores that check as a bound on the permitted range itself, so a future/third-party `AssetBoxHash` implementor can't have an out-of-bounds `AllowedExclusion` silently accepted. Also collapses the `box_ranges`/`box_allowed_exclusions` parallel-vector bookkeeping (previously guarded only by a `debug_assert_eq!`, absent in release builds) into a single slice, removing the invariant a future caller could violate.
- **`b0275dbc`** — The box-hash verification failure arm in `claim.rs` was logging/returning only the coarse validation-status constant instead of the actual error message, losing the real failure reason. Restores the underlying error's `Display` text in both the log and the returned error, matching this arm's own pre-PR behavior.
- **`d4d1e5f2`** — `make_box_maps` read each JPEG segment's payload via `jfifdump` and discarded it, only for `get_box_map`'s later pass to seek back and re-read a prefix of the same bytes for the metadata-signature check. Threads the APP1/APP13 prefix through instead of re-reading from the stream. Also extracts the one XMP-signature check shared with the pre-existing `extract_xmp` into a single `is_xmp_signature` helper so the two can't drift.
- **`bc1c584f`** — `jpeg_io.rs`/`png_io.rs`/`jpegxl_io.rs` (and GIF's whole-box case) each hand-rolled the same two `AllowedExclusion` shapes. Adds `AllowedExclusion::whole_box`/`after_header` constructors and adopts them everywhere.

All existing tests pass, plus a new test (`test_split_exclusions_rejects_allowed_exclusion_wider_than_box`) covering the restored bound. `cargo +nightly fmt --all -- --check` and `cargo clippy --features file_io -- -D warnings` are clean.

## Test plan
- [x] `cargo test -p c2pa --features file_io --lib` — 1010 passed, 0 failed
- [x] Each commit verified to compile independently (`cargo check` at each SHA)
- [x] `cargo clippy -p c2pa --features file_io --lib -- -D warnings`

### DIFF
--- a/sdk/src/assertions/box_hash.rs
+++ b/sdk/src/assertions/box_hash.rs
@@ -86,6 +86,31 @@ pub struct AllowedExclusion {
     pub kind: ExclusionKind,
 }
 
+impl AllowedExclusion {
+    /// The box's entire content is excludable - e.g. the C2PA store itself,
+    /// or a format's dedicated padding box.
+    pub fn whole_box(len: u64) -> Self {
+        AllowedExclusion {
+            start: 0,
+            length: len,
+            kind: ExclusionKind::ManifestOrPadding,
+        }
+    }
+
+    /// Everything past a fixed-size structural header is excludable asset
+    /// metadata (EXIF/XMP/IPTC-equivalent per spec §9.2.6). `total_len` is
+    /// the box's own full length, and `header_len` the number of leading
+    /// bytes (marker/length/type/CRC-adjacent fields, as applicable per
+    /// format) that must stay hashed.
+    pub fn after_header(header_len: u64, total_len: u64) -> Self {
+        AllowedExclusion {
+            start: header_len,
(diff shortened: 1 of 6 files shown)


caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 32998333088](https://github.com/contentauth/c2pa-rs/actions/runs/32998333088).

PR #2545, opened 2026-08-25
feat!: Update for C2PA 2.3 spec

mauricefisher64:

Support for multiple independent named trust lists.
Validation of soft-binding assertions
Support for specVersion and trust_list_uri
Support for soft binding algorithm list
Unified trust list model (manifest, CAWG, TSA) are configured in a common manor
More detailed reporting for trust sources
New settings support to automatically handle multiple trust anchors
Per trust list EKU configuration
Hardening of trust checks

## Checklist


scouten-adobe:
Marked as breaking (`!` in commit title) since the settings TOML format is changing

cli/src/main.rs
```diff
@@ -247,6 +251,10 @@ enum Commands {
         /// URL or path to file containing configured EKUs in Oid dot notation
         #[arg(long = "trust_config", env="C2PATOOL_TRUST_CONFIG", value_parser = parse_resource_string)]
         trust_config: Option<TrustResource>,
+
+        /// UR used to identify the trust list
```
scouten-adobe:
```suggestion
        /// URI used to identify the trust list
```

docs/cawg-identity.md
```diff
@@ -6,13 +6,17 @@ The CAI Rust library includes an implementation of the [Creator Assertions Worki
 
 An identity claims aggregation (ICA) credential is signed by an issuer identified by a DID (for example a `did:web` or `did:jwk`). A valid signature only proves that the credential was signed by whoever controls that DID: it does **not**, on its own, establish that the issuer is trustworthy. In particular, a self-issued `did:jwk` can be minted by anyone, so accepting it as trusted would let an attacker assert any identity.
 
-For this reason, the library treats ICA issuers as untrusted unless they appear on an explicit allow-list. Configure the allow-list through the `cawg_trust.trusted_ica_issuers` setting, which is a list of exact DID strings (any DID method):
+For this reason, the library treats ICA issuers as untrusted unless they appear on an explicit allow-list. Configure the allow-list through the by adding a new CAWG specific trust anchor. `trust.anchors[].trusted_ica_issuers` setting, which is a list of exact DID strings (any DID method):
```
scouten-adobe:
```suggestion
For this reason, the library treats ICA issuers as untrusted unless they appear on an explicit allow-list. Configure the allow-list by adding a new CAWG specific trust anchor. `trust.anchors[].trusted_ica_issuers` setting, which is a list of exact DID strings (any DID method):
```

docs/context-settings.md
```diff
@@ -170,12 +170,12 @@ For a complete reference to all the `Settings` properties, see the [SDK object r
...
-| [`cawg_trust`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#trust) | Configuration for CAWG trust lists. |
 | [`cawg_x509_signer`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#signersettings) | Configuration for the [CAWG x.509 signer](https://docs.rs/c2pa/latest/c2pa/struct.Settings.html#structfield.signer). |
 | [`core`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#core) | Configuration for core features. |
 | [`signer`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#signersettings) | Configuration for the base [C2PA signer](https://docs.rs/c2pa/latest/c2pa/struct.Settings.html#structfield.signer). |
-| [`trust`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#trust) | Configuration for C2PA trust lists. |
+| [`trust`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#trust) | Configuration for manifestn trust lists, CAWG trust lists & TSA time-stamp trust lists |
```
scouten-adobe:
```suggestion
| [`trust`](https://opensource.contentauthenticity.org/docs/manifest/json-ref/settings-schema#trust) | Configuration for C2PA claim generator trust lists, CAWG trust lists & TSA time-stamp trust lists |
```

docs/context-settings.md
```diff
@@ -351,8 +363,16 @@ Here's the `Settings` JSON with all default values:
...
   "version": 1,
   "trust": {
-    "trust_anchors": "-----BEGIN CERTIFICATE-----\n...",
-    "trust_config": "1.3.6.1.5.5.7.3.4\n1.3.6.1.5.5.7.3.36"
+      "anchors": [
+          {
+            "trust_anchors": "-----BEGIN CERTIFICATE-----\n...",
+            "trust_config": "1.3.6.1.5.5.7.3.4\n1.3.6.1.5.5.7.3.36",
+            "trust_kind": "manifest",
+            "trust_uri": "https://c2pa-rs/test_cert_trust_list,
```
scouten-adobe:
```suggestion
            "trust_uri": "https://c2pa-rs/test_cert_trust_list",
```

sdk/src/store.rs
```diff
@@ -166,23 +168,48 @@ impl Store {
         let mut store = Store::new();
         let settings = context.settings();
 
-        // load the trust handler settings, don't worry about status as these are checked during setting generation
-        if let Some(ta) = &settings.trust.trust_anchors {
-            let _v = store.add_trust(ta.as_bytes());
-        }
+        // use the incoming trust settings
+        store.ctp.clear();
```
tmathern, line 172: If custom settings (or any adjustment was made by the code for EKUs for instance) are configured, are they all in the incoming trust settings?

sdk/src/crypto/time_stamp/verify.rs
```diff
@@ -520,38 +520,54 @@ pub fn verify_time_stamp(
...
+                        format!("timestamp cert untrusted: {}", &common_name),
+                        "verify_time_stamp"
+                    )
+                    .validation_status(TIMESTAMP_UNTRUSTED)
+                    .informational(&mut current_validation_log);
+
+                    last_err = TimeStampError::Untrusted;
+                    continue;
+                }
+                Ok((_trust_type, trust_uri)) => {
```
tmathern, line 540: I see above that only timestamping certs should be there (in theory), but maybe here it also deserves a check to make sure we verified against a timestamping cert chain? Mistakes can happen, what if someone would misconfigure?

sdk/src/claim.rs
```diff
@@ -714,38 +720,10 @@ impl Claim {
             ? "alg": tstr .size (1..max-tstr-length),
             ? "alg_soft": tstr .size (1..max-tstr-length),
             ? "metadata": $assertion-metadata-map,
+            ? "spec_version": semver-string,
+            ? any unknown fields
             */
 
-            static V2_FIELDS: [&str; 10] = [
```
tmathern, line 719: Is that one not needed anymore because you have the spec version now? What about previously signed files which will not know a spec version?

sdk/src/settings/mod.rs
```diff
@@ -1110,21 +1213,44 @@ impl SettingsValidate for Settings {
...
+                // If both are arrays, merge and deduplicate. This is a design choice.
+                let target_array = target.as_array().cloned().unwrap_or_default();
+                let overlay_array = overlay.as_array().cloned().unwrap_or_default();
+                if !target_array.is_empty() {
+                    let mut unique = HashSet::new();
+                    unique.extend(target_array);
+                    if !overlay_array.is_empty() {
+                        unique.extend(overlay_array);
+                    }
+                    *target = Value::Array(unique.into_iter().collect());
```
tmathern: Is order going to be important when converted into the array? I think HashSet does not guarantee ordering, so multiple runs with same data may end up with different orderings here.

c2pa_c_ffi/src/c_api.rs
```diff
@@ -3306,17 +3306,29 @@ mod tests {
     fn test_c2pa_reader_from_stream_cawg() {
         // This fixture's identity claims aggregation (ICA) credential is signed
         // by a `did:jwk` issuer. ICA issuers are untrusted by default, so we must
-        // add that issuer to `cawg_trust.trusted_ica_issuers` for the credential
-        // to be reported as valid.
+        // add that issuer to `trust.anchors[n].trusted_ica_issuers` for the credential
+        // to be reported as valid. You can do this by added to the trust list anchors.
```
tmathern:
```suggestion
        // to be reported as valid. You can do this by adding to the trust list anchors.
```

sdk/src/claim.rs
mauricefisher64: This is not spec version this is claim version.   Assertions and claims have their own versioning.

sdk/src/crypto/time_stamp/verify.rs
mauricefisher64: Not sure what you mean here.  Only the timestamp signined cert must have the timestamp EKU.  The rest of the chain validated by line 522:

`
            match ctp.check_certificate_trust(
                &ordered_cert_ders[0..],
                &ordered_cert_ders[0],
                Some(signing_time),
            ) {`

checks the certificate chain.  If it is misconfigured you will get an error.

c2pa_c_ffi/src/c_api.rs
mauricefisher64: fixed

sdk/src/store.rs
mauricefisher64: Currently that is the design.  If you are configuring via context the context takes precedence.  If we don't do that the user may get undefined behavior since they may not know that their settings are overlayed on top of our default settings.   Now a user can do that just by using the default Context or Settings then adding their own customizations.  What I didn't want is for them to think the set something and in reality we were using defaults.

mauricefisher64: Note this is only for trust settings.   What I found is that there was no way for a user to not use our defaults for say EKUs unless we let them set them explicitly.  This puts the control back on the caller.  It is still easy to do since all they have to do is use default Settings or Context.  But now they can have a way to NOT use those.

sdk/src/settings/mod.rs
mauricefisher64: Order does not matter.  These are trust lists and we don't care or have a way to prefer one trust list over another.  We only care that we found a valid trust anchor for the cert we are inspecting.

scouten-adobe (changes requested):

scouten-adobe (dismissed):
## Review summary

Automated review pass over this trust-list rework. 10 findings below, ranked most severe first. Two are flagged as behavior-affecting correctness issues; the rest are efficiency/reuse/simplification cleanups surfaced by the same pass.

### Correctness

1. **`sdk/src/validation_results.rs:130`** — `timestampTrustListUri` is never serialized to JSON.
   The new `timestamp_trust_list_uri` field uses `#[serde(rename = "timestampTrustListUri", skip_serializing)]` (unconditional skip) instead of `skip_serializing_if = "Option::is_none"` like its sibling `trust_list_uri`. `ValidationResults::from_store` (~line 199-201) does compute and set this value when a timestamp is validated against a named TSA trust list, but it's silently dropped on serialization — so `timestampTrustListUri` never appears in the JSON output (c2patool, the C API, or any other consumer), defeating this PR's stated goal of "more detailed reporting for trust sources" for the TSA case.

2. **`sdk/src/crypto/cose/certificate_trust/openssl.rs:34`** (and the equivalent loop in `rust_native.rs:64`) — trust anchors from separate named trust lists are no longer pooled.
   Pre-PR, all configured anchors (system + user) were loaded into one `X509Store` and the chain was validated against the pooled set in a single pass. Post-PR, each anchor set gets its own isolated store built only from that set's `trust_anchor_ders`, and the loop returns success only if one single list alone completes the chain. A deployment that splits anchors across two lists (e.g. an intermediate in one, the root that issued it in another) would see chains that used to validate now fail, since no single list contains everything needed. This may be the intended "independent trust lists" semantics, but given it's a change from the old pooled-anchor behavior, please confirm it's deliberate.

### Efficiency

3. **`sdk/src/crypto/cose/certificate_trust/openssl.rs:37`** — the certificate chain and end-entity cert are re-parsed from DER on every iteration of the per-anchor-set loop instead of once per verification, i.e. O(L×T) DER parses instead of O(L) for L chain certs and T configured trust lists, on the signature-verification hot path.

4. **`sdk/src/crypto/cose/certificate_trust/rust_native.rs:64`** — same pattern in the rust-native backend: each chain certificate is re-parsed and re-checked for expiry on every anchor-set iteration, and each anchor set's own DER anchors are re-parsed every time too.

5. **`sdk/src/crypto/cose/verifier.rs:198`** — `let mut adjusted_ctp = ctp.clone();` deep-clones the entire `CertificateTrustPolicy`, including the raw DER bytes for every configured trust list, just to add a couple of EKU OID strings via `add_valid_ekus` when a matched anchor set carries a `trust_config` override.

### Reuse

6. **`sdk/src/crypto/cose/certificate_trust_policy.rs:451`** — `verify_mandatory_ekus` and `has_allowed_eku` (line 496) each independently hardcode the same three well-known EKU→field mappings (email-protection, timestamping, OCSP-signing) instead of sharing one lookup. A future EKU added to one and not the other would silently desync mandatory-EKU enforcement from allowed-EKU detection.

7. **`sdk/src/crypto/cose/ocsp.rs:384`** (also `ocsp.rs:520` and `sdk/src/crypto/time_stamp/verify.rs:495`) — the same 3-line "build a policy scoped to one mandatory EKU" sequence (`CertificateTrustPolicy::default()` → `clear_ekus()` → `add_mandatory_ekus(...)`) is copy-pasted three times rather than factored into a shared constructor.

8. **`sdk/src/crypto/cose/certificate_trust_policy.rs:149`** — `signing_trust_anchors`, `tsa_trust_anchors`, and `cawg_trust_anchors` each triplicate a filter-by-kind that the same struct's generic `get_anchor_set(kind, uri)` (line 444) already parameterizes. These three public methods have no production caller in the repo — only a unit test uses them.

### Simplification

9. **`sdk/src/settings/mod.rs:146`** — `impl Default for TrustAnchor` builds an identical 6-field, all-empty struct literal twice (once per `cfg!(not(test))` branch), one of five near-identical `TrustAnchor { .. }` literals scattered across this file. A new field requires hand-updating all five.

10. **`sdk/src/validation_results.rs:186`** — the new loop in `ValidationResults::from_store` nests `if`/`if let` four levels deep to extract the signing-credential trust-list URI; could be flattened with early continues or a `let-else` guard.

---
🤖 Automated review generated with Claude Code.

cli/src/main.rs
```diff
@@ -53,10 +55,12 @@ mod signer;
 /// Official C2PA conformance trust list (PEM bundle).
 const TRUST_LIST_OFFICIAL_URL: &str =
     "https://raw.githubusercontent.com/c2pa-org/conformance-public/refs/heads/main/trust-list/C2PA-TRUST-LIST.pem";
+const TRUST_LIST_OFFICIAL_URI: &str = "https://c2pa.org/trustlist";
```
ok-nick, line 58: Is this supposed to point to something?

sdk/src/crypto/cose/certificate_trust_policy.rs
```diff
@@ -135,16 +188,16 @@ impl CertificateTrustPolicy {
...
-            return Ok(TrustAnchorType::NoCheck);
+            return Ok((TrustAnchorType::NoCheck, String::new()));
         }
 
         // First check to see if the certificate appears in the allowed set of
         // end-entity certificates.
         let cert_hash = base64_sha256_cert_der(end_entity_cert_der);
         if self.end_entity_cert_set.contains(&cert_hash) {
-            return Ok(TrustAnchorType::EndEntity);
+            return Ok((TrustAnchorType::EndEntity, String::new()));
```
ok-nick, line 200: Can the string returned be optional instead?

tmathern (approved):

scouten-adobe:
Thanks for the context, @mauricefisher64 — following up on the review above with that in mind:

1. **`timestampTrustListUri` not serialized** — makes sense; it's not a spec field yet and is populated for the human-readable manifest report, not JSON. No change needed.
2. **Trust lists no longer pooled** — confirmed by design: the spec now wants the specific matching trust list reported, so independent (non-merged) checking is correct.
3. **OpenSSL re-parsing per trust list** — agreed the `X509Store` itself can't be reused once built. Given the typical count is 2-3 lists, not worth restructuring further.
4. **rust_native re-parsing per trust list** — agreed; dynamic loading plus the small typical list count makes this a non-issue in practice.
5. **`verifier.rs` clone for EKU overlay** — agreed, that's the accepted cost of supporting per-trust-list EKU overrides.

Resolving those 5 points as by-design/no-op. The remaining findings from the review are still open and unaddressed:

- `verify_mandatory_ekus` / `has_allowed_eku` (`certificate_trust_policy.rs`) independently hardcode the same OID→field mapping.
- The single-mandatory-EKU scratch-policy construction is copy-pasted 3x (`ocsp.rs` x2, `time_stamp/verify.rs`).
- `signing_trust_anchors`/`tsa_trust_anchors`/`cawg_trust_anchors` (`certificate_trust_policy.rs:149`) duplicate `get_anchor_set`'s generic pattern and have no production caller.
- `TrustAnchor::default()` (`settings/mod.rs:146`) builds an identical struct literal twice.
- The 4-level nested `if`/`if let` in `ValidationResults::from_store` (`validation_results.rs:186`).

None of these affect trust-decision correctness — they're all reuse/simplification cleanups. Happy to leave them as-is if you'd rather not touch this PR further, or file them as a fast-follow.

scouten-adobe (approved):

sdk/src/crypto/cose/certificate_trust/openssl.rs
```diff
@@ -24,70 +24,56 @@ pub(crate) fn check_certificate_trust(
...
-
-    let cert = X509::from_der(cert_der)?;
-
-    let mut builder = openssl::x509::store::X509StoreBuilder::new()?;
-    builder.set_flags(X509VerifyFlags::X509_STRICT)?;
-
-    let mut verify_param = openssl::x509::verify::X509VerifyParam::new()?;
-    verify_param.set_flags(X509VerifyFlags::X509_STRICT)?;
-    verify_param.set_flags(X509VerifyFlags::PARTIAL_CHAIN)?; // allow intermediates to be on anchor list
+    for anchor_set in ctp.anchor_sets() {
```
ok-nick, line 34: I did a pass through Claude and it flagged this line, that it iterates all trust anchors (C2PA, TSA, CAWG) where the caller may only of meant to validate C2PA anchors. We should double check at the call site that the anchors in the certificate trust policy only contain the ones specific to the job being done.

sdk/src/spec_versions.rs
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing,
+// this software is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR REPRESENTATIONS OF ANY KIND, either express or
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+/// C2PA Generator Version
+#[allow(dead_code)]
+pub const C2PA_GENERATOR_VERSION: &str = "2.2.0";
```
ok-nick, line 16: Are we able to hardcode this? From my understanding the generator version depends on what the user is actually putting in the manifest, which we can't control.

sdk/src/settings/mod.rs
```diff
@@ -895,16 +954,62 @@ impl Settings {
     ///
     /// This overlays the parsed configuration on top of the current Settings
     /// instance without touching thread-local state.
+    #[allow(deprecated)]
```
ok-nick, line 946: I don't think we should deprecate this since the other APIs (in c2pa-rs) rely on it. Can you instead leave a note in the docs that it is supporting legacy behavior?

sdk/src/settings/mod.rs
```diff
@@ -1110,21 +1213,44 @@ impl SettingsValidate for Settings {
...
-        (target, overlay) => *target = overlay,
+        (target, overlay) => {
+            // Currently only allow merging of arrays for the trust.anchors.
+            // This is a design choice to avoid unexpected behavior when merging arrays
+            // in other parts of the settings.
+            if target.is_array() && overlay.is_array() && depth == 2 && outer_key == "anchors" {
+                // If both are arrays, merge and deduplicate. This is a design choice.
+                let target_array = target.as_array().cloned().unwrap_or_default();
+                let overlay_array = overlay.as_array().cloned().unwrap_or_default();
+                if !target_array.is_empty() {
```
ok-nick: Hm, this seems fallible. Not a blocker but I'd prefer if after calling `merge_json_depth`, in `merge_json` we pull the anchors array directly and merge it rather than doing that logic in `merge_json_depth`.

sdk/src/claim.rs
```diff
@@ -714,38 +720,10 @@ impl Claim {
...
-                GATHERED_ASSERTIONS_F,
-                DC_TITLE_F,
-                REDACTED_ASSERTIONS_F,
-                ALG_F,
-                ALG_SOFT_F,
-                METADATA_F,
-            ];
-
-            // make sure only V2 fields are present
-            if let c2pa_cbor::Value::Map(m) = &claim_value {
```
ok-nick, line 733: 👍

sdk/src/store.rs
```diff
@@ -166,23 +168,48 @@ impl Store {
...
+
+        // Add all of the trust anchors
+        if let Some(anchors) = &settings.trust.anchors {
+            for anchor in anchors {
+                let trust_list_uri = match &anchor.trust_uri {
+                    Some(uri) => uri.to_string(),
+                    None => format!(
+                        "https://c2pa-rs/unknown_trust_list_{}",
```
ok-nick, line 180: Why does this have to match a URL scheme? If it doesn't point to av alid URL can we just use the hash directly?

sdk/src/settings/mod.rs
```diff
@@ -895,16 +954,62 @@ impl Settings {
...
+        }
+
+        // if there are any legacy anchors, add them to the anchors list
+        if !legacy_anchors.is_empty() {
+            if let Some(anchors) = &mut settings.trust.anchors {
+                anchors.extend(legacy_anchors);
+            } else {
+                settings.trust.anchors = Some(legacy_anchors);
+            }
+        }
```
ok-nick: Clear the legacy anchors fields after adding them to the new location?

sdk/src/crypto/cose/certificate_trust/openssl.rs
mauricefisher64: I thought about that but did not want to risk breaking people who labeled wrong or not at all.

mauricefisher64: The ability is there is you want to test the hypothesis later.

sdk/src/spec_versions.rs
mauricefisher64: It is not used at the moment.  I am letting Gavin decide.

sdk/src/settings/mod.rs
mauricefisher64: Well the who thing is fallible since it did not handle arrays at all.

mauricefisher64: How is it fallible?  It checks for a specific case.  All others go through the old code?

mauricefisher64: The reason I did not do it outside is there maybe other settings that do need to be merged as well.  Outside got hokey quickly.

sdk/src/store.rs
mauricefisher64: Read the spec.  It is a URI of this form.

mauricefisher64: It is not a URL.

sdk/src/settings/mod.rs
mauricefisher64: Hmm, you will have to keep supporting this then.  @gpeacock what do you think/want?

sdk/src/settings/mod.rs
mauricefisher64: Probably a good thing to do.  Let me change this logic a little so that the legacy values are part of the merge.

sdk/src/crypto/cose/certificate_trust/openssl.rs
mauricefisher64: You will have to plumb through from all the calling sites the correct TrustKind value which hopefully you know the anchor kind you are looking for.

mauricefisher64: Note that TrustKind is not a C2PA spec thing.  I added it so that we know and can choose if we need that.  We have one case that we need that and it is for CAWG.  In that case we do limit the search to CAWG trust lists.

ok-nick: My concern is that it isn't possible to validate C2PA certs against solely the official C2PA trust list because it always uses all trust lists. The same applies for the TSA trust list.

sdk/src/store.rs
ok-nick: Interesting..

sdk/src/settings/mod.rs
mauricefisher64: Changed to merge anchors outside of merge_json

cli/src/main.rs
mauricefisher64: No, it is a URI not a URL

ok-nick (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33213170865](https://github.com/contentauth/c2pa-rs/actions/runs/33213170865).

PR #1998, opened 2026-03-31
ci: add aarch64-unknown-linux-gnu target to release builds

Cyrah2R:
This PR adds support for automatically building and releasing `aarch64-unknown-linux-gnu` binaries for Linux. 

With the growing adoption of ARM-based cloud instances (AWS Graviton, Oracle Cloud, etc.), having pre-compiled `aarch64` releases will save users from having to install the Rust toolchain on their production servers just to compile `c2patool`.

### Changes made
1. **`.github/workflows/release.yml`**: Added the `aarch64` target to the release matrix and included a step to install `gcc-aarch64-linux-gnu` for cross-compilation on the Ubuntu runner. Passed the `ARCH=aarch64` flag to the `make release` command when needed.
2. **`cli/Makefile`**: Updated the Linux release targets to check for the `ARCH=aarch64` flag. If present, it adds the appropriate rustup target, uses the aarch64 linker, and packages the tarball with the correct `_linux_aarch64` naming convention so it doesn't conflict with the Intel build.

Closes #1997

PR #2553, opened 2026-08-27
ci: add aarch64-unknown-linux-gnu target to release builds

httpoz:
This PR adds support for automatically building and releasing `aarch64-unknown-linux-gnu` binaries for Linux.

With the growing adoption of ARM-based cloud instances (AWS Graviton, Oracle Cloud, etc.), having pre-compiled `aarch64` releases will save users from having to install the Rust toolchain on their production servers just to compile `c2patool`.

### Changes made
1. **`.github/workflows/c2patool-release.yml`**: Added the `aarch64` target to the release matrix and included a step to install `gcc-aarch64-linux-gnu` for cross-compilation on the Ubuntu runner. Passed the `ARCH=aarch64` flag to the `make release` command when needed.
2. **`cli/Makefile`**: Updated the Linux release targets to check for the `ARCH=aarch64` flag. If present, it adds the appropriate rustup target, uses the aarch64 linker, and packages the tarball with the correct `_linux_aarch64` naming convention so it doesn't conflict with the Intel build.

Closes #1997

---

This is a rebase of #1998 (originally opened from `Cyrah2R/c2pa-rs`) onto current `main`. Since `main` refactored `release.yml` to be fully tag-driven (binary builds moved out to `c2patool-release.yml`), the aarch64 changes from that PR no longer applied to `release.yml` and were re-homed to `c2patool-release.yml` instead. Also fixes a `cli/Makefile` merge issue from the original patch where `build-release-win` was accidentally overwritten instead of `build-release-linux`, which would have broken the Windows release build.


scouten-adobe:
Hi @httpoz, thank you for submitting this PR. Unfortunately, there are a couple of significant issues that prevent us from merging this:

* **Blocker: Divergent toolchain between CI and release** — tier-1b validates aarch64 with `cargo cross` (Docker + full sysroot). This release workflow uses bare `gcc-aarch64-linux-gnu` without setting up an aarch64 sysroot. If any crate in the dependency graph links against a system library (OpenSSL, libz, etc.), the release build may fail with linker errors that tier-1b never caught. tier-1b validates the c2pa _library_ for aarch64, but there's no analogous test for the _c2patool binary_ that this PR is releasing. The release workflow produces a c2patool tarball with no verification that the cross-compiled binary even loads, let alone runs correctly. I'm not sure this can be fixed given the currently-available set of GitHub Actions runners.

* **Artifact name collision** — `name: c2patool-${{ matrix.os }}` resolves to `c2patool-ubuntu-latest` for both Ubuntu jobs. `upload-artifact@v7` rejects duplicate names by default, so the second Ubuntu job to finish will error, `needs: build` propagates the failure, and the `release` job will never run. This can be fixed by appending the target triple to the artifact name, e.g. `c2patool-${{ matrix.os }}-${{ matrix.target || 'default' }}`. The `download-artifact` step already uses `merge-multiple: true` so it will pick up all variants regardless of name.

PR #2555, opened 2026-08-27
fix: do not log::error! validation error when hashed URI is dangling, instead log::debug!

ok-nick:
It is already added as a validation error, we should not log general errors if these may be caught in the wild.

I am indifferent to removing these lines completely.

### DIFF
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -17,7 +17,7 @@ use std::path::Path;
 use std::{borrow::Cow, io::Cursor, sync::Arc};
 
 use async_generic::async_generic;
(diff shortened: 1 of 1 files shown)


tmathern (approved):
Do the debug lines give more helpful info to debug than just the validation result? That would answer the question keep vs don't keep.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 33408675372](https://github.com/contentauth/c2pa-rs/actions/runs/33408675372).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2565 |

PR #2556, opened 2026-08-27
Fix JPEG XL Exif box exclusion; more box-hash cleanup (#2513 re-review)

scouten-adobe:
## Summary

Follow-up on a re-review of #2513 after #2543 was merged into this branch. All of #2543's fixes verified as correctly landed with no regressions (via a fresh 8-angle review pass). This PR fixes one new correctness bug found in the original feature code, plus several cleanup items.

- **`25942785`** (fix) — `classify_jxl_allowed_exclusions` treated a bare (non-`brob`) `Exif` box's entire payload as excludable `AssetMetadata` right after the box header. Per the JPEG XL/HEIF Exif-in-box convention, that payload actually begins with a mandatory 4-byte big-endian `exif_tiff_header_offset` structural field before the real TIFF/Exif data — the same class of leading structural field the neighboring `"brob"` arm already correctly skips. A signed JPEG XL asset's Exif box offset field could be altered post-signing without invalidating the box hash. Verified against libjxl's own documented box format. Also applies the existing `BoxMap` `..Default::default()` pattern to this file's two literals.
- **`40274d42`** (cleanup) — `c2pa_io.rs` still hand-constructed the `AllowedExclusion` shape that `AllowedExclusion::whole_box(0)` was introduced to replace.
- **`3f411148`** (cleanup) — the box-hash and BMFF-hash error-classification arms in `claim.rs` had already drifted from duplicating the same "classify as MALFORMED/MISMATCH" logic (one discarded the original error, one didn't). Extracted a shared helper; both arms now keep the original error's detail.
- **`464a1ee0`** (cleanup) — `gif_io.rs`'s XMP and Comment Extension sub-block-to-`AllowedExclusion` mapping was duplicated; extracted `metadata_exclusions_from_sub_blocks`.
- **`bbd9ee4c`**, **`ceffa245`** (cleanup) — `png_io.rs`/`jpeg_io.rs` `BoxMap` literals now use `..Default::default()` instead of spelling out every boilerplate field (`BoxMap` already derives `Default`); `jpeg_io.rs` also fixed a spot where APP1/APP13 were identified two different ways (numeric marker vs. resolved name) that could silently drift apart.
- **`2ac80c4c`** (cleanup) — extracted `AllowedExclusion::is_bounded_by`/`contains` so the box-length defense-in-depth check (added in #2543) is a named, harder-to-accidentally-drop step instead of one clause in a larger `&&` chain; turned `BoxRangeInfo` from a `(u64, u64, Vec<AllowedExclusion>)` tuple into a named struct to remove a start/len transposition risk.

**Deliberately not changed:** `jpeg_io.rs`'s `AppSegmentPrefixes` (a `HashMap` keyed by `range_start`) was flagged as possibly over-engineered vs. a parallel `Vec`, but `make_box_maps` has ~15 independent `box_maps.push()` call sites — a parallel `Vec` would require every one of them to push a matching entry in lockstep, reintroducing exactly the parallel-vector fragility fixed elsewhere in this PR. The `HashMap` keyed by a stable identifier is the safer choice here.

## Test plan
- [x] `cargo test -p c2pa --features file_io --lib` — 1027 passed, 0 failed (includes an updated `test_classify_jxl_allowed_exclusions` case covering the Exif fix)
- [x] Each commit verified to compile independently (`cargo check` at each SHA)
- [x] `cargo clippy -p c2pa --features file_io --lib -- -D warnings`

### DIFF
--- a/sdk/src/assertions/box_hash.rs
+++ b/sdk/src/assertions/box_hash.rs
@@ -109,6 +109,26 @@ impl AllowedExclusion {
             kind: ExclusionKind::AssetMetadata,
         }
     }
+
+    fn end(&self) -> Option<u64> {
+        self.start.checked_add(self.length)
+    }
+
+    /// Whether this permitted range itself stays within its box's real
+    /// length - a defense against a buggy or malicious `AssetBoxHash`
+    /// implementor reporting a range that reaches past the box it's
+    /// attached to. `AllowedExclusion` is otherwise trusted as already
+    /// self-bounded, so this is checked independently rather than assumed.
+    fn is_bounded_by(&self, box_len: u64) -> bool {
+        self.end().is_some_and(|end| end <= box_len)
+    }
+
+    /// Whether `[range_start, range_end)` is fully contained within this
+    /// permitted range.
+    fn contains(&self, range_start: u64, range_end: u64) -> bool {
+        self.end()
+            .is_some_and(|end| range_start >= self.start && range_end <= end)
+    }
 }
 
 #[derive(Serialize, Default, Deserialize, Debug, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


ssanthosh (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) found no target branches to backport this pull request to in [workflow run 33199536755](https://github.com/contentauth/c2pa-rs/actions/runs/33199536755).

PR #2559, opened 2026-08-28
fix: Use scrambled opaque IDs instead of real addresses for c_ffi handles

gpeacock:
PointerRegistry keyed tracked handles by their real pointer address, so a handle freed on one thread could be silently reissued to an unrelated object if another thread's allocation reused that same address before a stale caller used the old handle — same type, wrong live instance (ABA).

Handles (are now tracked under a scrambled, always-odd id derived from a monotonic counter via a bijective multiplicative hash, so ids are never reused and can't collide with the real, even-aligned addresses still used for buffers C dereferences directly (to_c_string/to_c_bytes).

Also fixes several existing call sites that bypassed the tracked-pointer macros and dereferenced a handle's raw value directly — the new id scheme turns those into an immediate, obvious crash instead of silent corruption.


c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -33,48 +36,113 @@ use crate::{cimpl::cimpl_error::CimplError, error::Error, maybe_send_sync::Maybe
...
+compile_error!("PointerRegistry's handle scrambling needs a 32- or 64-bit usize");
+
+/// Turns a plain sequential counter value into a scrambled, always-odd
+/// handle id.
+///
+/// - **Always odd (and therefore never zero)**: real Rust allocations
+///   always land on at least a 2-byte boundary, so their addresses are
+///   always even. An odd id can therefore never collide with a real tracked
+///   buffer address (see `track_by_address`).
+/// - **Bijective, not just well-mixed**: multiplying by a fixed odd constant
```
tmathern, line 55: Distinct within that (likely 64bits large) window is strong enough here, I don't think it needs to be actually bijective (because the ids have that large period and repeat. See comment at `scramble_to_odd_id` signature about 32 bits, where it may be more relevant to ponder).

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -33,48 +36,113 @@ use crate::{cimpl::cimpl_error::CimplError, error::Error, maybe_send_sync::Maybe
...
+///   is invertible modulo `2^usize::BITS`, and the product of two odd numbers
+///   is always odd regardless of truncation — so distinct counter values are
+///   mathematically guaranteed to scramble to distinct odd ids, not just
+///   unlikely to collide the way a hash could.
+fn scramble_to_odd_id(counter: usize) -> usize {
```
tmathern, line 60: We also have support for wasm32 (so 32 bits), which will wrap around the ids space earlier (way earlier) than 64 bits architectures. I would fail it there when it wraps in this case, so a holding app/service would be able to detect it and restart.
(It must be fairly long-running to naturally hit it, or churn a lot of allocations).

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -33,48 +36,113 @@ use crate::{cimpl::cimpl_error::CimplError, error::Error, maybe_send_sync::Maybe
...
             // Silently ignore poisoned mutex - this is a best-effort tracking system
         }
     }
 
-    /// Validate that a pointer is tracked and has the expected type
-    pub fn validate(&self, ptr: usize, expected_type: TypeId) -> Result<(), Error> {
-        if ptr == 0 {
+    /// Resolve a handle id to its real address, validating it is tracked
+    /// with the expected type.
+    pub fn resolve(&self, id: usize, expected_type: TypeId) -> Result<usize, Error> {
```
tmathern, line 128: To consider: enforce the caller uses the returned address to do something with it? (`must_use`)

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -93,8 +161,8 @@ impl PointerRegistry {
     /// - `c2pa_context_builder_build`: builder is consumed to produce a context
     /// - `c2pa_reader_with_stream`: reader is consumed to produce a new reader
     /// - `c2pa_builder_with_definition`: builder is consumed to produce a new builder
-    pub fn untrack(&self, ptr: usize, expected_type: TypeId) -> Result<(), Error> {
-        if ptr == 0 {
+    pub fn untrack(&self, id: usize, expected_type: TypeId) -> Result<usize, Error> {
```
tmathern, line 164: To consider: enforce the caller uses the returned address to do something with it? (`must_use`)

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -235,13 +306,16 @@ pub fn track_arc_mutex<T: 'static + MaybeSend>(ptr: *mut Mutex<T>) -> *mut Mutex
...
+    id as *mut Mutex<T>
 }
 
-/// Validate that a pointer is tracked and has the expected type
-pub fn validate_pointer<T: 'static>(ptr: *mut T) -> Result<(), Error> {
-    get_registry().validate(ptr as usize, TypeId::of::<T>())
+/// Validate that a pointer is tracked and has the expected type, returning
+/// the real pointer to dereference (the value passed in is an opaque handle
+/// id, not the real address — see `PointerRegistry::track_by_id`).
+pub fn validate_pointer<T: 'static>(ptr: *mut T) -> Result<*mut T, Error> {
```
tmathern, line 316: To consider: enforce the caller uses the returned address to do something with it? (`must_use`)

tmathern (commented):
Thanks for the change, that should alleviate some pain points!
Just waiting to see what you think on the 32 bits case (how to handle it)...
I also have a few follow-up ideas I want to try out and maybe will PR (later).

c2pa_c_ffi/src/cimpl/utils.rs
scouten-adobe: Probably worth updating the comments here to say that it's guaranteed unique until 2^63 allocations have occurred. (Claude flagged this during my review of this PR.)

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -33,48 +36,113 @@ use crate::{cimpl::cimpl_error::CimplError, error::Error, maybe_send_sync::Maybe
...
+    fn track_by_id(&self, real_addr: usize, type_id: TypeId, cleanup: CleanupFn) -> usize {
+        if real_addr == 0 {
+            return 0;
+        }
+        let counter = self.next_id.fetch_add(1, Ordering::Relaxed);
+        let id = scramble_to_odd_id(counter);
+        if let Ok(mut tracked) = self.tracked.lock() {
+            tracked.insert(id, (real_addr, type_id, cleanup));
         }
+        // Silently ignore poisoned mutex - this is a best-effort tracking system
```
scouten-adobe, line 110: When `self.tracked.lock()` returns `Err` (poisoned mutex), the `if let Ok` body is skipped and `cleanup` is dropped. Dropping `Box<dyn FnMut()>` does **not** call the closure — the captured `ptr_val: usize` is just a number, so `drop(Box::from_raw(ptr_val))` is never invoked and the heap allocation leaks permanently.

The function still returns a non-zero `id`, so the caller gets a handle that looks valid but will fail every subsequent `resolve()` call with "untracked pointer." This is a regression from the old `track()`, which returned `()` — callers still held the real pointer and could use it directly. Here, the opaque id is the *only* path to the allocation; if it never lands in the map, the allocation is permanently inaccessible.

scouten-adobe (changes requested):

tmathern:
I am (slowly) working on adding some cross-thread support for the registry here: https://github.com/contentauth/c2pa-rs/pull/2575. That WIP may contain fixes of interest for this one.

tmathern:
Pulling in main to get a CI/CD fix running jobs.

c2pa_c_ffi/src/cimpl/utils.rs
gpeacock: Good catch. We now recover from the poisoned lock by returning the value. Since the lock cannot be un-poisoned, we do the same thing at all access points. This was already done by drop. The map will continue to work in this state, so it is our best way to handle things.

c2pa_c_ffi/src/cimpl/utils.rs
gpeacock: Must use would require us wrapping results in a NewType so it adds some overhead and complexity. It doesn't work for usize or other basic types. These methods are only accessed through macros, and the macros always do the right thing, so we are not at much risk.

tmathern (approved):

tmathern:
(Note that this will likely need careful integration in downstream SDKs).

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -21,70 +21,637 @@
...
+    borrow_state: AtomicUsize,
+    /// `Option` so `untrack` can take the closure out through a shared `Arc`,
+    /// leaving nothing for `Drop` to run.
+    cleanup: Mutex<Option<CleanupFn>>,
+    /// Process creating the registry, written on process creation.
+    owner_pid: u32,
+}
+
+/// Current process id.
+#[cfg(not(target_arch = "wasm32"))]
```
gpeacock, line 100: do we ever build cffi for wasm at all, let alone wasm32?

tmathern: There is a wasm32 build job: https://github.com/contentauth/c2pa-rs/actions/runs/34873169732/job/104073898411. Since wasm is special and I did not want to fight builds in my PR, I may have added it just for the job to pass...

scouten-adobe (approved):

Issue #2507, opened 2026-08-14
Experimental feature proposal: C2PA Live Video (spec §19), flag unstable_live_video


N1Knight:
## Summary

This issue proposes **C2PA Live Video** (section 19 of the [C2PA Technical Specification 2.4](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#_live_video)) as an experimental feature in `c2pa-rs`, gated behind the cargo flag `unstable_live_video`.

Per [docs/experimental-features.md](https://github.com/contentauth/c2pa-rs/blob/main/docs/experimental-features.md), this issue is the required discussion issue and will stay open for the lifetime of the feature.

## Motivation

Live streaming workflows (DASH/HLS with fMP4 segments) cannot use the standard C2PA embedding flow: the asset is unbounded, segments are produced and published continuously, and there is no point at which the full asset can be hashed and signed. Section 19 of the spec defines two methods to address this, and neither is currently implemented in any C2PA SDK.

By implementing this in `c2pa-rs`, we give the ecosystem a reference implementation for live provenance, so broadcasters and streaming platforms can adopt Content Credentials without having to build a bespoke stack.

## Scope

Both methods defined by the spec, each with signing **and** validation:

**§19.3, Per-segment C2PA Manifest Box**
Every media segment carries its own C2PA Manifest with a `c2pa.livevideo.segment` assertion, chained back to the init segment manifest.

**§19.4, Verifiable Segment Info (VSI)**
The init segment manifest carries a `c2pa.session-keys` assertion (COSE_Key, RFC 9052). Media segments carry a COSE_Sign1 over the Verifiable Segment Info in an ISO BMFF `emsg` box, avoiding a full manifest per segment. Includes `signerBinding` verification against the SPKI of the end-entity certificate from the init manifest (§19.7.3) and per-segment `bmffHash` validation (§19.4.1).

We propose contributing the feature as a single PR covering both methods.

### Out of scope for the initial contribution

- **Network-facing components.** CMAF ingest (the DASH-IF protocol for pushing segments from encoder to packager) and anything else that moves segments over the wire. The API operates on in-memory segment buffers.
- **Key management.** For §19.4 the session keys used to sign each segment are supplied by the caller. Generating, rotating, storing or distributing them is out of scope.

## Compliance with the experimental features policy

| Requirement | Plan |
|---|---|
| **Feature flag** | `unstable_live_video`, off by default, not in any default feature set |
[...]

PR #2562, opened 2026-08-28
feat(experimental): C2PA Live Video (spec §19) signing and validation

N1Knight:
## Summary

Implements [C2PA Technical Specification §19 (Live Video)](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#live-video) for DASH/HLS fMP4 streams: signing and validation for both methods, per-segment Manifest Box (§19.3) and Verifiable Segment Info (§19.4).

This is an experimental feature. Everything here is gated behind the `unstable_live_video` Cargo feature flag, off by default and exempt from the crate's usual semver stability guarantees; a build without the flag behaves identically to one without this code at all. See [docs/experimental-features.md](docs/experimental-features.md) for the crate's experimental-features policy and the registered entry for this flag.

Spec: https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#live-video (section 19: §19.3 per-segment Manifest Box, §19.4 Verifiable Segment Info)

Tracking issue: https://github.com/contentauth/c2pa-rs/issues/2507

## Background

A live stream is an unbounded sequence of fragmented MP4 segments, each of which must be provenance-verifiable as it's produced. §19 defines two methods for this; this PR implements both:

- §19.3, per-segment C2PA Manifest Box: each media segment carries its own full C2PA manifest in a `uuid` box, with a `c2pa.livevideo.segment` assertion (`sequenceNumber`, `streamId`, `previousManifestId`) chaining it to the prior segment. Simple and self-contained per segment, at the cost of a full manifest's worth of overhead on every segment.
- §19.4, Verifiable Segment Info (VSI): the full manifest is signed once, in the init segment, which publishes one or more ephemeral Ed25519 session keys via a `c2pa.session-keys` assertion. Each media segment then carries only a small COSE_Sign1 in an `emsg` box, signed by a session key, covering the segment's `sequenceNumber`, `bmffHash`, and the init manifest's id. Session keys are bound to the manifest signer's certificate via a detached `signerBinding` COSE_Sign1 (§18.25.2), so trust still traces back to the same signing identity without a full manifest per segment.

## What's included

- `sdk/src/live_video/`: `LiveVideoValidator` (both methods), `LiveVideoSigner` (§19.3), `LiveVideoVsiSigner` (§19.4), COSE_Key to DER conversion for `signerBinding` verification, BMFF box parsing for VSI's `emsg`/`moof`/`mfhd`.
- `sdk/src/assertions/`: `LiveVideoSegment` (§19.3) and `SessionKeys`/`SessionKey` (§18.25) assertion types.
- `cli/src/live_video*.rs`: `c2patool live-video-sign` (both methods) and `c2patool live-video` (validates a stream, auto-detecting the method from whether the init manifest carries a `c2pa.session-keys` assertion).
- New public helper `c2pa::live_video::moof_sequence_number` for callers that need to read a segment's own `moof/mfhd.sequence_number` (e.g. to resume a VSI session at the right starting point).
- `c2patool live-video` writes a JSON validation report to stdout, in the same shape the rest of the CLI uses, with per-segment states and the §19.7 failure codes; human-readable progress goes to stderr.

## Usage

```bash
# Build
cargo build --release -p c2patool --features unstable_live_video
C2PATOOL=target/release/c2patool

# Sign (§19.3 per-segment Manifest Box)
$C2PATOOL <segments_dir> live-video-sign \
  --segments_glob "chunk_*.m4s" \
  -o output/ \
  -m manifest.json \
  --init init.mp4

# Sign (§19.4 Verifiable Segment Info; needs a 32-byte Ed25519 seed file)
$C2PATOOL <segments_dir> live-video-sign \
  --segments_glob "chunk_*.m4s" \
  -o output/ \
  -m manifest.json \
  --method vsi \
  --session-key session.key \
  --init init.mp4

# Validate (auto-detects the method from the init segment's manifest)
$C2PATOOL output/init.mp4 live-video --segments_glob "chunk_*.m4s"

# Progress goes to stderr and the JSON report to stdout, so it can be redirected alone
$C2PATOOL output/init.mp4 live-video --segments_glob "chunk_*.m4s" > report.json
```

`manifest.json` must include a `c2pa.livevideo.segment` assertion with a `streamId` (§19.3); see `cli/docs/manifest.md`.

Validation writes a JSON report giving the stream's overall `validation_state`, the detected method, each segment's state, and any failures under their §19.7 status codes:

```json
{
  "validation_state": "Valid",
  "method": "19.3 (per-segment C2PA Manifest Box)",
  "init_segment": "output/init.mp4",
  "segments": [{ "path": "output/chunk_001.m4s", "state": "Valid" }],
  "validation_results": { "failure": [] }
}
```

A stream is reported at the level of its weakest segment, per the nesting of manifest states in §14.3.2, and the command exits non-zero when the state is `Invalid`.

### Trust lists

§19.7.1 defers to the general validation rules of chapter 15, whose acceptance threshold is a `Valid` manifest (§14.3.5), not a `Trusted` one: §15.7 makes verifying a chain of trust a `should`, and §14.3.3 treats an asset as valid when its active manifest is "either Valid or Trusted". A stream signed by a certificate that doesn't chain to an anchor is therefore reported as `Valid` and accepted, which matters because the SDK's X.509 trust lists are empty by default. Content failures such as `assertion.bmffHash.mismatch` still make a segment `Invalid` either way.

When the operator does provision trust material, §15.7 requires the claim to be rejected if a chain of trust cannot be verified, so `signingCredential.untrusted` becomes fatal and the stream is `Invalid`. Trust material is taken from the settings directory, either as the sidecar files `c2patool init trust` writes there or as inline PEM in the settings file's own `[trust]` section, so no extra flag is needed on `live-video` itself. `cli/docs/usage.md` documents both.

Full flag reference:

`live-video-sign` (signing, both methods):
- `--segments_glob <GLOB>` (required): glob to find media segments, matched in natural
  (numeric-aware) filename order.
- `-o, --output <DIR>` (required): output directory for signed segments.
- `-m, --manifest <FILE>` (required): manifest definition JSON.
- `--init <FILE>`: also sign the init segment (optional per §19.2.3).
- `--previous-segment <FILE>`: resume the continuity chain from a prior invocation's last
  signed segment, for a process that restarts mid-stream. With `--method vsi`, this also
  skips re-signing the init segment.
- `--method <manifest|vsi>` (default `manifest`): §19.3 or §19.4.
- `--session-key <FILE>`: Ed25519 session key, 32-byte raw seed. Required with
  `--method vsi`, reused across all invocations for the session.
- `--min-sequence-number <N>`: VSI's starting `minSequenceNumber` (§18.25.2). Only used on
  the first invocation of a VSI session (no `--previous-segment`); if omitted, inferred
  from the first media segment's own `moof/mfhd.sequence_number`.

`live-video` (validation, auto-detects method):
- `--segments_glob <GLOB>` (required): glob to find media segments, resolved relative to
  the init segment's directory.

## Testing

- `cargo test --workspace --all-targets`, with the feature on and off: green, no failures (1492 and 1351 tests respectively), plus doctests and the `rust_native_crypto` feature set.
- `cargo clippy --workspace --all-targets -- -Dwarnings` and `cargo fmt --check`, with the feature on and off: clean.
- Unit tests cover the validation-state policy, the trust-material detection, the state ordering, and the §19.7 status-code reporting.
- Manual sign+validate runs of both methods against fMP4 fixtures, checking the emitted report in four configurations: no trust list (`Valid`), a trust list whose anchors do chain (`Trusted`), anchors that do not chain (`Invalid`, `signingCredential.untrusted` under `livevideo.manifest.invalid`), and a modified `mdat` (`Invalid`, reported as `assertion.bmffHash.mismatch` under §19.3 and as a `bmffHash` failure under §19.4).
- Both methods signed and validated end to end through a live multi-service pipeline, rebuilt from scratch for each run so the signer carries the binary under test: all 5 ABR representations `Trusted` under §19.3 and under §19.4. Flipping one byte inside a segment's `mdat` rejects it in both, with `assertion.bmffHash.mismatch` cascading through the §19.3 chain and a `bmffHash` failure isolated to the segment under §19.4, exit status 1 either way. Those pipeline and browser runs were made just before this branch was merged with `main`; the merge brought no changes to live video code, and the sign-and-validate matrix above was re-run after it.
- Cross-checked with the browser player, which validates through `@svta/cml-c2pa`, an independent TypeScript implementation: 22 segments, 0 invalid, across several ABR renditions.

## Out of scope (for now)

- Mixed/combined method validation: `c2patool live-video` (and `LiveVideoValidator`) auto-detects and validates a stream as either §19.3 (Manifest Box) or §19.4 (VSI), not both at once; nothing here validates a stream that switches method mid-session.
- Custom/vendor-specific `continuityMethod`s for §19.3 (the spec allows them; this PR only implements the standard `c2pa.manifestId` method; others are detected and reported as unsupported, not validated).
- TSA timestamping/OCSP revocation specific to live segments: the existing (non-live) trust/revocation machinery still applies to the manifest signer's certificate as usual, but there's no live-specific extension here.
- VSI session key rotation: a session key's validity window is fixed at creation, and `kid` and `validityPeriod` are currently hardcoded in `live-video-sign` rather than configurable. There is no mechanism to rotate to a new key before expiry, so a live session longer than the window needs a fresh signing invocation.

**Known limitation:** VSI's `iat` (claimed time of signing, §19.4.1) is self-asserted by the segment signer, and is currently preferred over wall-clock time when checking a session key's `validityPeriod`. That is deliberate, since it lets archival/VOD validation of a recorded live stream check a segment against the time it was actually produced, but it means a party holding a leaked session key could claim an `iat` anywhere inside the original validity window and keep signing past the point the key should have stopped being usable. The lower bound is enforced (`iat` cannot predate the key's own `createdAt`), but the broader trade-off between live and archival validation posture is not resolved in this PR.


sdk/src/live_video/cose_key.rs
```diff
@@ -0,0 +1,523 @@
...
+    if bytes.len() > len {
+        return None;
+    }
+    let mut padded = vec![0u8; len - bytes.len()];
+    padded.extend_from_slice(bytes);
+    Some(padded)
+}
+
+// ── DER encoding helpers ────────────────────────────────────────────────────
+
```
mauricefisher64, line 207: Have you considered RASN for this?

sdk/src/builder.rs
```diff
@@ -269,6 +269,12 @@ pub struct AssertionDefinition {
     /// True if this assertion is attributed to the signer (defaults to false)
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub created: bool,
+    /// Pre-encoded CBOR bytes that bypass the `Value` intermediate representation.
```
mauricefisher64, line 272: I would work with @gpeacock to fix this.

mauricefisher64 (approved):
I am OK with the PR.  I would however to work with @ok-nick on a better interface to c2patool since he is reworking that interface.   Also be sure to confirm with @gpeacock to make sure the Builder changes fit into his API vision.  Also make sure the CBOR stuff get addressed.  I also think you could simplify a little by using existing code and features.  All in all nice work!

sdk/src/builder.rs
gpeacock: There should be no need for this after the c2pa_cbor fixes. We really don't want this worked around in our own code!

sdk/src/builder.rs
```diff
@@ -895,6 +905,31 @@ impl Builder {
...
     }
 
+    /// Adds a CBOR assertion using pre-encoded bytes that preserve CBOR features
+    /// (such as tags) that `c2pa_cbor::Value` cannot represent.
+    ///
+    /// The data is first serialized to CBOR via the binary encoder (which handles
+    /// CBOR tags), and those bytes are stored as an override. A `Value` copy is
+    /// also kept for metadata/inspection, though it may lose tag information.
+    #[cfg(feature = "unstable_live_video")]
+    pub(crate) fn add_assertion_cbor<S, T>(&mut self, label: S, data: &T) -> Result<&mut Self>
```
gpeacock: There should be no need for this, and add_assertion already means add_assertion cbor, lets fix it instead of workign around  a bug

sdk/src/assertions/session_keys.rs
```diff
@@ -0,0 +1,253 @@
...
+        deserialize_with = "deserialize_date_tagged"
+    )]
+    pub created_at: String,
+    /// Seconds from `created_at` for which this key is valid.
+    pub validity_period: u64,
+    /// COSE_Sign1_Tagged binding this key to the signer's certificate.
+    ///
+    /// Stored internally as the inner COSE_Sign1 content (without tag 18).
+    /// The CBOR tag 18 is added/stripped transparently during serialization/deserialization.
+    #[serde(
```
gpeacock: we should replace this logic that is trying work around c2pa_cose support

sdk/src/live_video/cose_key.rs
valentinamgiusti: Good call, addressed in [5ce53df](https://github.com/contentauth/c2pa-rs/pull/2562/commits/5ce53dfdb6262c758124d0591254be1883cb62e6). The hand-rolled DER helpers are gone and the `SubjectPublicKeyInfo` is built with `rasn-pkix` instead, the same way `sdk/src/utils/ephemeral_cert.rs` already did.

sdk/src/builder.rs
valentinamgiusti: We looked into this and the reason we needed the workaround is already fixed in `c2pa_cbor`. [Issue #27](https://github.com/contentauth/c2pa-cbor/issues/27) was closed by [PR #29](https://github.com/contentauth/c2pa-cbor/pull/29), released in `0.77.4`, while `c2pa-rs` is still pinned to `0.77.2`.

We tried the upgrade locally and it does remove the need for it. Both `cbor_override` and `add_assertion_cbor` can go, and `builder.rs` ends up with no live-video-specific code, with the suite passing.

Once the version bump happens we can remove the override.

scouten-adobe:
FYI in future, we would appreciate it if you were to check the box that allows maintainers to push to the PR. In this case and since we're close to an upcoming release, we've created #2631 which contains this and some edits we needed to make to land this PR.

scouten-adobe:
Closing in favor of #2631.

PR #2563, opened 2026-08-28
fix: consistently use xmp.iid: instead of xmp:iid:

gpeacock:
The sdk was generating xmp:iid: prefixes when it creates an UUID for xmpMM:InstanceID. 

The Adobe convention for these ids in the xmpMM namespaces is to use a period instead of a colon as a seperator. This makes that consistent and upates all the examples.


tmathern (approved):

ok-nick (approved):
Looks like CI is erroring as well for formatting.

PR #2575, opened 2026-09-02
feat: Update opaque ids handling in C FFI (add checkout mechanism)

tmathern:
Build on top of https://github.com/contentauth/c2pa-rs/pull/2559 to add some cross-thread safety (a thread can't release something it doesn't own). A registry entry is now reference counted and carries a borrow state, so a handle is borrowed for the duration of a call and cannot be released underneath a thread holding it.

Added features:

- Reference counted entries: a registry entry is shared rather than owned by the map. A thread using an object holds a reference, and cleanup runs when the last reference goes rather than when C calls free (relies on Arc types). A free arriving mid-call is therefore deferred rather than refused or ignored. Note: A reader count cannot reach it, on 32-bit or 64-bit, because every reader holds a live guard and that many guards would need more memory than the address space has.
- A checkout mechanism: `checkout_shared` and `checkout_exclusive` hand out guards that dereference to a shared or exclusive reference. Any number of shared borrows may coexist. At most one exclusive borrow may exist, and it excludes all others.
- Pair untracking: two handles are removed together or not at all, leaving the registry exactly as it was on failure. This is needed especially for the case when identity signing is done, since there is a signer and an identity signer to handle.
- Each entry records the process that created it.
- Panic handling: unwinding should not cross the C boundary.

Unfortunately, there is an ABI break I had to introduce with the C2paStream classes (since this switches to hand out ids rather than addresses).

## Checklist


c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -21,21 +21,394 @@
...
+    /// The `Arc`'s refcount frees the object; this field and the refcount
+    /// move independently.
+    /// It can't be inferred from `Arc::strong_count` either,
+    /// since any transient clone increases the strong count without a borrow existing.
+    borrow_state: AtomicUsize,
+    /// `Option` so `untrack` can take the closure out through a shared `Arc`,
+    /// leaving nothing for `Drop` to run.
+    cleanup: Mutex<Option<CleanupFn>>,
+    /// Process creating the registry, written on process creation.
+    owner_pid: u32,
```
gpeacock: is is safe to assume that pid always fits in a u32?

c2pa_c_ffi/src/cimpl/utils.rs
```diff
@@ -21,21 +21,394 @@
...
+}
+
+/// Current process id.
+#[cfg(not(target_arch = "wasm32"))]
+fn current_pid() -> u32 {
+    // SAFETY: getpid takes no arguments and cannot fail.
+    unsafe { libc::getpid() as u32 }
```
gpeacock: This should use std:proccess::id() which supports windows.

gpeacock (commented):
I see you already integrated my branch. it looks like we are getting close. We will need some stress tests on each platform.

c2pa_c_ffi/src/cimpl/utils.rs
tmathern: Done

PR #2578, opened 2026-09-02
fix!: C2PA 2.4 validation

mauricefisher64:
Relax validation on soft bindings assertions
Relax validation on metadata assertions
Updated Cloud Data assertion validation (no fetching)
External Reference assertion validation (no fetching)
Updated Action assertion rules
Fix unit tests in violation of 2.4

## Checklist

### DIFF
--- a/sdk/src/assertions/cloud_data.rs
+++ b/sdk/src/assertions/cloud_data.rs
@@ -16,7 +16,7 @@ use serde::{Deserialize, Serialize};
 use crate::{
     assertion::{Assertion, AssertionBase, AssertionCbor},
     assertions::{assertion_metadata::AssetType, labels, AssertionMetadata},
-    error::Result,
+    error::{Error, Result},
 };
 
 /// The `location` field of a [`CloudData`] assertion: a URL with its
@@ -147,6 +147,81 @@ impl CloudData {
     pub(crate) fn is_actions(&self) -> bool {
         labels::base(&self.label) == labels::ACTIONS
     }
+
+    /// Validates the Cloud Data assertion according to the C2PA 2.4 rules for
+    /// `c2pa.cloud-data` assertions.
+    pub fn validate(&self) -> Result<()> {
+        if self.label.trim().is_empty() {
+            return Err(Error::ValidationRule(
+                "cloud data assertion label must not be empty".to_owned(),
+            ));
+        }
+
+        if self.size < 1 {
+            return Err(Error::ValidationRule(
+                "cloud data assertion size must be >= 1".to_owned(),
+            ));
+        }
+
+        if self.location.url.trim().is_empty() {
+            return Err(Error::ValidationRule(
+                "cloud data assertion location.url must not be empty".to_owned(),
+            ));
+        }
+
+        if self.location.alg.trim().is_empty() {
+            return Err(Error::ValidationRule(
+                "cloud data assertion location.alg must not be empty".to_owned(),
+            ));
+        }
+
+        if self.location.hash.is_empty() {
+            return Err(Error::ValidationRule(
+                "cloud data assertion location.hash must not be empty".to_owned(),
+            ));
+        }
+
+        if self.is_forbidden().is_err() {
+            return Err(Error::ValidationRule(format!(
+                "cloud data assertion must not reference forbidden label '{}', per C2PA 2.4",
+                self.label
+            )));
+        }
+
+        if self.is_hard_binding() {
+            return Err(Error::ValidationRule(format!(
+                "cloud data assertion must not reference hard binding assertions '{}', per C2PA 2.4",
+                self.label
+            )));
+        }
+
+        Ok(())
+    }
+
+    /// check for any forbidden label types, including hard bindings and the cloud-data assertion itself.
+    pub fn is_forbidden(&self) -> Result<()> {
+        // Per C2PA 2.4 §15.10.3.2.1, the referenced assertion must not be any
+        // of the forbidden label types, including hard bindings and the cloud-data
+        // assertion itself.
+        let forbidden = [
+            "c2pa.action",
+            "c2pa.actions",
+            "c2pa.actions.v2",
+            labels::CLOUD_DATA,
+            "c2pa.ingredient",
+            "c2pa.ingredient.v2",
+            "c2pa.ingredient.v3",
+        ];
+
+        if forbidden.contains(&self.label.as_str()) {
+            return Err(Error::ValidationRule(format!(
+                "cloud data assertion must not reference forbidden label '{}', per C2PA 2.4",
+                self.label
+            )));
+        }
+
+        Ok(())
+    }
 }
 
 impl AssertionCbor for CloudData {}
@@ -234,5 +309,19 @@ pub mod tests {
         assert!(make(labels::ACTIONS).is_actions());
         assert!(make("c2pa.actions.v2").is_actions());
         assert!(!make("c2pa.metadata").is_actions());
+    }
+
+    #[test]
+    fn test_validate_rejects_invalid_structure() {
+        assert!(CloudData::new("", 1, make_location()).validate().is_err());
+        assert!(CloudData::new("c2pa.metadata", 0, make_location())
+            .validate()
+            .is_err());
+        assert!(CloudData::new("c2pa.hash.data", 1, make_location())
+            .validate()
+            .is_err());
+        assert!(CloudData::new("c2pa.metadata", 1, make_location())
+            .validate()
+            .is_ok());
     }
 }
--- a/sdk/src/assertions/cloud_data.rs
+++ b/sdk/src/assertions/cloud_data.rs
@@ -148,9 +148,10 @@ impl CloudData {
         labels::base(&self.label) == labels::ACTIONS
     }
 
-    /// Validates the Cloud Data assertion according to the C2PA 2.4 rules for
+    /// Validates structure of the Cloud Data assertion according to the C2PA 2.4 rules for
     /// `c2pa.cloud-data` assertions.
-    pub fn validate(&self) -> Result<()> {
+    #[allow(unused)]
+    pub(crate) fn validate(&self) -> Result<()> {
         if self.label.trim().is_empty() {
             return Err(Error::ValidationRule(
                 "cloud data assertion label must not be empty".to_owned(),
@@ -203,7 +204,7 @@ impl CloudData {
         // Per C2PA 2.4 §15.10.3.2.1, the referenced assertion must not be any
         // of the forbidden label types, including hard bindings and the cloud-data
         // assertion itself.
-        let forbidden = [
+        const FORBIDDEN: [&str; 7] = [
             "c2pa.action",
             "c2pa.actions",
             "c2pa.actions.v2",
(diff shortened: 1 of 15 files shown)


sdk/src/assertions/cloud_data.rs
```diff
@@ -147,6 +147,81 @@ impl CloudData {
     pub(crate) fn is_actions(&self) -> bool {
         labels::base(&self.label) == labels::ACTIONS
     }
+
+    /// Validates the Cloud Data assertion according to the C2PA 2.4 rules for
+    /// `c2pa.cloud-data` assertions.
+    pub fn validate(&self) -> Result<()> {
```
ok-nick: Nice, I like this pattern of storing validation behavior with the assertion.

sdk/src/assertions/cloud_data.rs
```diff
@@ -147,6 +147,81 @@ impl CloudData {
...
+        if self.label.trim().is_empty() {
```
ok-nick, line 155: IMO, we shouldn't trim. If the user input it wrong, it's wrong and should be fixed.

sdk/src/assertions/cloud_data.rs
ok-nick: How come this doesn't take a log tracker as parameter so we can log all the validation errors we come across instead of short circuiting on the first one?

ok-nick: Also I think we should keep this `pub(crate)` for now.

sdk/src/assertions/cloud_data.rs
```diff
@@ -147,6 +147,81 @@ impl CloudData {
...
+        // assertion itself.
+        let forbidden = [
+            "c2pa.action",
+            "c2pa.actions",
+            "c2pa.actions.v2",
+            labels::CLOUD_DATA,
+            "c2pa.ingredient",
+            "c2pa.ingredient.v2",
+            "c2pa.ingredient.v3",
+        ];
```
ok-nick: Any chance we can use const labels for all of them that we defined? Also can we define the list as a const?

sdk/src/assertions/ingredient.rs
```diff
@@ -444,6 +442,16 @@ impl Ingredient {
...
+        /*
+        For now we will not enforce this rule, but we should consider enforcing it in the future.
+        This is caught in the verify_ingredient function, but it would be better to prevent it here.
+        if self.active_manifest.is_some() && self.digital_source_type.is_some() {
+            return Err(serde::ser::Error::custom(
+                "Ingredient v3 activeManifest and digitalSourceType cannot both be present",
+            ));
+        }
+        */
+
```
ok-nick: Is it better to prevent it here? IMO I think we should remove this comment and go with the verify_ingredient function.

sdk/src/assertions/ingredient.rs
```diff
@@ -184,8 +184,6 @@ impl Ingredient {
             && self.validation_status.is_none()
             && self.c2pa_manifest.is_none()
             && self.validation_results.is_some()
-            && self.active_manifest.is_some()
```
ok-nick, line 187: 👍

sdk/src/assertions/ingredient.rs
```diff
@@ -539,6 +550,9 @@ impl Ingredient {
         if let Some(sba) = &self.soft_binding_algorithms_matched {
             ingredient_map.serialize_field("softBindingAlgorithmsMatched", sba)?;
         }
+        if let Some(dst) = &self.digital_source_type {
+            ingredient_map.serialize_field("digitalSourceType", dst)?;
+        }
```
ok-nick: This code is already being done on line 532 (just above).

sdk/src/assertions/ingredient.rs
```diff
@@ -490,6 +498,9 @@ impl Ingredient {
         if self.metadata.is_some() {
             ingredient_map_len += 1
         }
+        if self.digital_source_type.is_some() {
+            ingredient_map_len += 1
+        }
```
ok-nick: This is already being done as well.

sdk/src/assertions/soft_binding.rs
```diff
@@ -117,6 +124,24 @@ pub struct SoftBindingTimespan {
     pub end: u64,
 }
 
+/// Soft binding metadata.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SoftBindingMetdata {
```
ok-nick:
```suggestion
pub struct SoftBindingMetadata {
```

sdk/src/assertions/soft_binding.rs
```diff
@@ -117,6 +124,24 @@ pub struct SoftBindingTimespan {
...
+    /// Contact information for the implementation or author of the binding or the algorithm.
+    pub contact: Option<String>,
+
+    /// A web page containing more details about the implementation or author of the binding or the algorithm.
+    #[serde(rename = "informationalUrl")]
+    pub informational_url: Option<String>,
+
+    /// Uses flatten to allow these fields to be serialized at the same level as known fields.
+    #[serde(flatten, skip_serializing_if = "HashMap::is_empty")]
+    additional_fields: HashMap<String, Value>,
```
ok-nick: Mark this as pub since it's defined in the CDDL?

ok-nick: Soft binding assertion is CBOR so this should be `c2pa_cbor::Value` instead of JSON.

sdk/src/assertions/cloud_data.rs
mauricefisher64: The checks creation.  Validation happens elsewhere.

sdk/src/assertions/ingredient.rs
mauricefisher64: Must be a merge issue since I did this independently.

sdk/src/assertions/ingredient.rs
mauricefisher64: Fixed

sdk/src/assertions/ingredient.rs
mauricefisher64: Fixed

sdk/src/assertions/soft_binding.rs
mauricefisher64: Fixed

scouten-adobe (approved):

sdk/src/assertions/cloud_data.rs
mauricefisher64: **validate** is used to check creation not the manifest validation path

sdk/src/assertions/external_reference.rs
```diff
@@ -0,0 +1,297 @@
...
+    #[serde(rename = "dc:format", skip_serializing_if = "Option::is_none")]
+    pub dc_format: Option<String>,
+
+    /// Optional size of the referenced data in bytes.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub size: Option<u64>,
+
+    /// Optional asset type classifications for the referenced data.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub data_types: Option<Vec<crate::assertions::assertion_metadata::AssetType>>,
```
gpeacock: use crate::assertions::assertion_metadata::AssetType;
up top

sdk/src/assertions/ingredient.rs
gpeacock: Yes I think it is more consistent to enabled this here and remove my test for now. We need to figure out how to create test cases rather than permit mistakes!

gpeacock (approved):
approved with a couple of minor suggestions (not required)

sdk/src/claim.rs
```diff
@@ -1096,7 +1097,7 @@ impl Claim {
         &self.instance_id
     }
 
-    /// set title
+    /// set title, deprecated in C2PA 2.4.  Claim generators should create a c2pa.metadata assertion containing the dc:title field.
```
tmathern, line 1100: 👍 Will the whole method go eventually? Should it already be deprecated to be in the deprecation flow for the releases?

sdk/src/claim.rs
```diff
@@ -2716,6 +2764,26 @@ impl Claim {
...
+                    if claim.soft_binding_assertions().is_empty() {
+                        log_item!(
+                            label.clone(),
+                            "watermark action missing soft binding assertion",
+                            "verify_actions"
+                        )
+                        .validation_status(validation_status::ACTION_ASSERTION_SOFTBINDING_MISSING)
+                        .failure_no_throw(
+                            validation_log,
+                            Error::ValidationRule("the assertion was not redacted".into()),
```
tmathern, line 2783: Why "the assertion was not redacted"? Isn't this a soft binding missing error (validation_status::ACTION_ASSERTION_SOFTBINDING_MISSING)?

sdk/src/assertions/mod.rs
```diff
@@ -98,3 +98,6 @@ pub use soft_binding::{SoftBinding, SoftBindingBlock, SoftBindingScope};
 
 mod cloud_data;
 pub use cloud_data::{CloudData, HashedExtUri};
+
+mod external_reference;
+pub use external_reference::{ExternalReference, ExternalReferenceLocation, UnhashedExtUri};
```
tmathern, line 103: No exports for soft bindings related types? They could be useful elsewhere. Especially `SoftBindingMetadata`.

sdk/src/assertions/soft_binding.rs
```diff
@@ -117,6 +124,24 @@ pub struct SoftBindingTimespan {
     pub end: u64,
 }
 
+/// Soft binding metadata.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SoftBindingMetadata {
```
tmathern, line 129: For the optional fields: Does this need to skip serialization when the option is none, or will they gracefully be handled to not be here at all when serialized?

sdk/src/claim.rs
```diff
@@ -1251,9 +1252,20 @@ impl Claim {
...
+                if self.version() > 1 && cgi.len() == 1 {
+                    return cgi[0].get_spec_version();
+                }
+                None
+            }
+            None => self.spec_version.as_ref(), // legacy case where SpecVersion was including in Claim map
+        }
     }
 
+    /// Decpricated in  C2PA 2.4 or greater compatibile manifests.  Replaced by equivelent value in ClaimGeneratorInfo.
```
tmathern:
```suggestion
    /// Deprecated in  C2PA 2.4 or greater compatible manifests. Replaced by equiveaent value in ClaimGeneratorInfo.
```

tmathern (commented):
Can you double check the validation error linked to `validation_status::ACTION_ASSERTION_SOFTBINDING_MISSING`? It is about redactions, but I think it should be about soft bindings (this [comment](https://github.com/contentauth/c2pa-rs/pull/2578#discussion_r3920534791)).

sdk/src/claim.rs
mauricefisher64: It is correct:

`If the action field is either c2pa.watermarked [DEPRECATED] or c2pa.watermarked.bound:

Verify that there is also a [c2pa.soft-binding assertion](https://spec.c2pa.org/specifications/specifications/2.4/specs/C2PA_Specification.html#soft_binding_assertion) present in the manifest. If there is not, the claim shall be rejected with a failure code of assertion.action.softBindingMissing.`

sdk/src/assertions/mod.rs
mauricefisher64: Done

sdk/src/claim.rs
tmathern: I meant the error text: `the assertion was not redacted`. That is interestingly confusing.

PR #2581, opened 2026-09-03
fix: Harden against invalid  labels as CAWG identity hard bindings (backport #2538)

ssanthosh:

## Checklist

### DIFF
--- a/sdk/src/assertions/labels.rs
+++ b/sdk/src/assertions/labels.rs
@@ -234,14 +234,11 @@ pub const NON_REDACTABLE_LABELS: [&str; 5] =
 /// Returns `true` if `label` names a hard binding assertion (`c2pa.hash.data`,
 /// any `c2pa.hash.bmff.*`, `c2pa.hash.boxes`, or `c2pa.hash.collection.data`).
 ///
-/// `BMFF_HASH` is matched by prefix since it's the only one of these still at
-/// creation version > 1, so it's the only one that carries a version suffix
-/// (e.g. `.v3`) in practice; the others are compared exactly.
+/// `BMFF_HASH` is matched on its base label (version/instance suffix stripped)
+/// so it accepts `.v2`/`.v3` variants without accepting a fake like
+/// `c2pa.hash.bmfffake`; the others are compared exactly.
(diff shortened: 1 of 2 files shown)


tmathern (approved):

sdk/src/assertions/labels.rs
```diff
@@ -231,6 +231,19 @@ pub const HASH_LABELS: [&str; 4] = [DATA_HASH, BOX_HASH, BMFF_HASH, COLLECTION_H
...
+/// any `c2pa.hash.bmff.*`, `c2pa.hash.boxes`, or `c2pa.hash.collection.data`).
+///
+/// `BMFF_HASH` is matched by prefix since it's the only one of these still at
+/// creation version > 1, so it's the only one that carries a version suffix
+/// (e.g. `.v3`) in practice; the others are compared exactly.
+pub(crate) fn is_hard_binding_label(label: &str) -> bool {
+    label == DATA_HASH
+        || label == BOX_HASH
+        || label == COLLECTION_HASH
+        || label.starts_with(BMFF_HASH)
```
scouten-adobe: I'll approve, but please go back to `main` and also submit a fix for this finding that also exists in #2538:

`is_hard_binding_label` checks BMFF hash labels with a raw `label.starts_with(BMFF_HASH)`, so a fake label like `"c2pa.hash.bmfffake"` is wrongly accepted as a hard binding — the exact class of naive-prefix bug this PR is meant to fix, just left open for the BMFF case. It should use `labels::base(label) == BMFF_HASH` (already exists in this file) or check that the character following the prefix is `.` or end-of-string. Since it's already on main, it may be worth a small follow-up fix there and re-backporting, rather than blocking this PR.

scouten-adobe (approved):

sdk/src/assertions/labels.rs
ssanthosh: Performed the necessary fix using https://github.com/contentauth/c2pa-rs/pull/2581/commits/d87bbe20ee2c50e4bb4ca5e3da73f63296d5c3ec . I will update the main branch also with the similar fix in subsequent PR

PR #2596, opened 2026-09-04
feat: Add ZIP support (+ EPUB, Office Open XML, Open Document, and OpenXPS) and Collection Data Hash assertion (backport #499)

ok-nick:
Manually backports #499 due to failed cherry pick.


sdk/src/error.rs
```diff
@@ -353,6 +353,9 @@ pub enum Error {
 
     #[error("error parsing TIFF: {0}")]
     TiffError(#[from] TiffError),
+
+    #[error(transparent)]
```
tmathern, line 357: I wonder if that error should have been more verbose to propagate the original error message.

tmathern (approved):

sdk/src/error.rs
ok-nick: Good point. It will be removed in #2491 anyways.

Issue #2599, opened 2026-09-05
IdentityAssertion::validate_partial_claim reads Context::new() settings, so caller-configured cawg_trust anchors are never used


RobertsRobots:
What happens. A cawg.identity assertion with sig_type: cawg.x509.cose whose certificate chains to a root supplied in cawg_trust.trust_anchors (or user_anchors, or whose leaf is in allowed_list) still validates as cawg.identity.well-formed plus signingCredential.untrusted on the assertion URL. Configuring the anchors through Context::with_settings, through load_settings, or through c2patool --settings makes no difference.

Why. src/identity/identity_assertion/assertion.rs, validate_partial_claim, 0.90.20 line 309 (the function already receives context: &Context):

rust
let settings = Context::new().settings().clone();

Context::new() is Context::default(), whose settings are Settings::default(), so settings.cawg_trust.trust_anchors is always None and check_certificate_trust fails with CertificateNotTrusted before it looks at the chain. 0.85.2 (line 308) has no context parameter and does the same.

Fix. let settings = context.settings().clone();. With that one change a real SSL.com S/MIME identity chain (anchor: SSL.com Client RSA Root CA 2022) reports signingCredential.trusted on the cawg.identity URL in a local build of 0.90.20.

Related. Even with the fix the assertion logs only cawg.identity.well-formed; the // TO DO (CAI-7980): Should instead issue cawg.identity.trusted note a few lines below is the other half of making X.509 CAWG trust visible to readers. Happy to open a PR for the one-line fix if welcome.

Repro. c2patool 0.27.16: c2patool --settings settings.json file.jpg --detailed with cawg_trust.trust_anchors set to the issuing root's PEM → signingCredential.untrusted on c2pa.assertions/cawg.identity. Same with the Python bindings 0.37.7 via ContextBuilder().with_settings(...).

RobertsRobots:
For the stable line (0.90.x, fe07693): the one-line fix plus the same regression test adapted to the pre-v1.3 status codes and settings model. The test fails on current stable ("found in User trust anchors", the caller's settings ignored) and passes with the change, under both rust_native_crypto and openssl. Regression test for main is in #2600.
```diff
diff --git a/sdk/src/identity/identity_assertion/assertion.rs b/sdk/src/identity/identity_assertion/assertion.rs
index e2abe88..2cdd3f6 100644
--- a/sdk/src/identity/identity_assertion/assertion.rs
[...]

PR #2600, opened 2026-09-05
chore: Add regression test for CAWG trust settings reaching X.509 identity validation (#2599)

RobertsRobots:
Changes in this pull request

Adds a regression test for #2599. IdentityAssertion::validate_partial_claim used to build its CAWG trust policy from Context::new().settings() rather than the caller's context, so cawg_trust anchors and allowed lists were never consulted and every X.509 identity certificate came back untrusted. main already carries the one-line fix (#2209); this test pins it: it signs an asset with the Ed25519 test identity credential, reads it back with that certificate on a CAWG TrustAnchor.allowed_list, and asserts the reader reports cawg.x509.credential.trusted through the end-entity list (empty trust-list URI). It fails if the old Context::new() line is reintroduced.

The stable (0.90.x) line still has the bug. A backport there needs the one-line change plus this test adapted to the pre-v1.3 codes and settings model; a tested patch for stable is attached to #2599 for whoever picks up the cherry-pick.

### DIFF
--- a/sdk/src/identity/x509/x509_signature_verifier.rs
+++ b/sdk/src/identity/x509/x509_signature_verifier.rs
@@ -250,10 +250,10 @@ mod tests {
         },
         status_tracker::{LogKind, StatusTracker},
         validation_status::{
-            CAWG_X509_CREDENTIAL_UNTRUSTED, CAWG_X509_SIGNATURE_MISMATCH,
-            CAWG_X509_SIGNATURE_VALIDATED,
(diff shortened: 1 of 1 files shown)


RobertsRobots:
Update: the stable backport mentioned above is now #2604 (the one-line change plus this test adapted to the 0.90.x codes and settings model), confirmed on current stable by @bquinn in #2599 against the IPTC end-entity list. This PR stays the main-side regression test only.

scouten-adobe (approved):

PR #2604, opened 2026-09-06
fix: identity assertion validation uses the caller's cawg_trust settings (stable backport, #2599)

### LINKED ISSUE #2599: IdentityAssertion::validate_partial_claim reads Context::new() settings, so caller-configured cawg_trust anchors are never used

RobertsRobots:
Backport to the 0.90.x line of the one-line change that landed on main inside #2209 (4497b8b6), as identified by @bquinn in #2599: `validate_partial_claim` read `Context::new().settings()`, so caller-configured `cawg_trust` anchors and allow-lists were never consulted and every X.509 CAWG identity reported `signingCredential.untrusted`. Adds the regression test adapted to stable's status codes and settings model; it fails on current stable and passes with the change under both `rust_native_crypto` and `openssl`. The equivalent test for main is #2600.

Fixes #2599 on stable.

https://claude.ai/code/session_01Sw12ZYvjjzQXATcZkpnTZ9

### DIFF
--- a/sdk/src/identity/identity_assertion/assertion.rs
+++ b/sdk/src/identity/identity_assertion/assertion.rs
@@ -306,7 +306,7 @@ impl IdentityAssertion {
         status_tracker: &mut StatusTracker,
         context: &Context,
     ) -> Result<serde_json::Value, ValidationError<String>> {
-        let settings = Context::new().settings().clone();
+        let settings = context.settings().clone();
         self.check_padding(status_tracker)?;
 
         self.signer_payload
--- a/sdk/src/identity/x509/x509_signature_verifier.rs
+++ b/sdk/src/identity/x509/x509_signature_verifier.rs
@@ -221,7 +221,7 @@ mod tests {
             IdentityAssertion,
         },
         status_tracker::{LogKind, StatusTracker},
-        Builder, SigningAlg,
+        Builder, Context, Reader, SigningAlg,
     };
 
     const TEST_IMAGE: &[u8] = include_bytes!("../../../tests/fixtures/CA.jpg");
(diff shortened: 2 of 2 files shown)


domguinard:
This one would be quite important to fix as currently this makes it impossible to validate a CAWG identity signature. Looking forward to a merge of that one.

scouten-adobe (approved):

PR #2606, opened 2026-09-07
fix: Reject BMFF assets with a C2PA box that resolves to no manifest

ssanthosh:
### Vulnerability
Follow-up hardening on the BMFF `box_purpose` retag class from #2471 (VULN-36807). The 8-byte `box_purpose` tag on C2PA uuid boxes is deliberately excluded from the C2PA hash (needed for update workflows), so it can be retagged in place with no hash check catching it. #2471 closed the `update`-without-`original` permutation; a systematic review of the remaining permutations on `BmffIO::read_cai` found that a manifest box retagged to an **unrecognized purpose** (no parse branch exists for it) — or to `merkle` — is silently dropped, and `read_cai` returns `JumbfNotFound`, i.e. the asset appears to carry no C2PA manifest at all rather than a tampered one. A signed manifest silently disappears.

### Fix
Track whether any C2PA box (of any purpose) was present during parsing (`C2PABmffBoxes::c2pa_box_present`), and in `read_cai`'s fall-through return `Error::C2PAValidation("C2PA box present without a readable manifest")` (mirroring #2471's original/update pairing errors) when a C2PA box was present but resolved to no `manifest`/`original`/`update` — instead of returning `JumbfNotFound`. A genuinely unsigned asset (no C2PA box at all) still returns `JumbfNotFound`, unchanged. Enforcement is scoped to `read_cai`; the shared parser only gains the tracking field, so the many hashing callers of `read_bmff_c2pa_boxes` are unaffected.

### Tests added
- `test_read_cai_rejects_c2pa_box_with_unrecognized_purpose` — manifest retagged to an unknown purpose.
- `test_read_cai_rejects_merkle_box_without_manifest` — a valid `merkle` box with no accompanying manifest.
- `test_read_cai_unsigned_asset_still_reports_not_found` — no C2PA box still yields `JumbfNotFound` (guard doesn't over-fire).

### Verification
Full lib suite (openssl) 1064 passed / 0 failed; bmff_io + bmff_hash + BMFF update-manifest e2e 60 passed; both clippy invocations (openssl all-targets, native-crypto lib+tests) clean; `fmt --check` clean.

## Checklist


mpgopala (approved):

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1941,7 +1947,17 @@ impl C2paReader for BmffIO {
...
+        match c2pa_boxes.manifest_bytes {
+            Some(manifest_bytes) => Ok(manifest_bytes),
+            // A C2PA box was present but did not resolve to a manifest - e.g.
+            // a manifest box retagged to an unrecognized or merkle purpose
+            // (the purpose tag is hash-excluded). Reject rather than report
+            // the asset as carrying no manifest.
+            None if c2pa_boxes.c2pa_box_present => Err(Error::C2PAValidation(
+                "C2PA box present without a readable manifest".to_string(),
+            )),
+            None => Err(Error::JumbfNotFound),
```
tmathern, line 1959: How does this impact remote manifests, possibly (since they are not embedded)? Can this change the fallback error flows (JumbfNotFound falling back to looking for a remote manifest) for instance?

sdk/src/asset_handlers/bmff_io.rs
```diff
@@ -1941,7 +1947,17 @@ impl C2paReader for BmffIO {
...
 
-        c2pa_boxes.manifest_bytes.ok_or(Error::JumbfNotFound)
+        match c2pa_boxes.manifest_bytes {
+            Some(manifest_bytes) => Ok(manifest_bytes),
+            // A C2PA box was present but did not resolve to a manifest - e.g.
+            // a manifest box retagged to an unrecognized or merkle purpose
+            // (the purpose tag is hash-excluded). Reject rather than report
+            // the asset as carrying no manifest.
+            None if c2pa_boxes.c2pa_box_present => Err(Error::C2PAValidation(
+                "C2PA box present without a readable manifest".to_string(),
```
scouten-adobe, line 1957: **[MEDIUM]** `Store::load_jumbf_from_stream` matches `Err(Error::JumbfNotFound)` specifically to fall back to an XMP-embedded remote-manifest URL. Since `BmffIO::read_c2pa` (changed in this PR) now returns `Error::C2PAValidation` instead of `JumbfNotFound` whenever any C2PA box is present but unreadable (unrecognized purpose, merkle-only), that fallback is skipped for those cases.

This isn't a new risk *class* — the original/update-pairing errors from #2471 already bypass this same fallback — but this PR broadens the set of cases that skip it. `BmffIO` does implement `remote_manifest_url_ref`/`write_remote_manifest_url` (bmff_io.rs:1981), so BMFF + remote-manifest is a real, supported combination.

This directly answers @tmathern's open review question above (bmff_io.rs:1959) — confirmed via code trace, not yet addressed in the PR. Worth an explicit test and a one-line confirmation that skipping remote-manifest fallback here is intentional (it very likely should be, for a tampered box, but it's currently a side effect rather than a decision).

scouten-adobe (commented):
## PR Review

### Summary
Follow-up hardening to the BMFF `box_purpose` retag vulnerability (VULN-36807, prior fix in #2471). Previously, a C2PA UUID box retagged to an unrecognized purpose or to `merkle` (the purpose tag is outside the C2PA hash) would silently vanish, and `BmffIO::read_c2pa` would report `JumbfNotFound` — i.e. a tampered, previously-signed asset would look unsigned. The fix tracks "any C2PA box seen" (`C2PABmffBoxes::c2pa_box_present`) and returns `Error::C2PAValidation` instead of `JumbfNotFound` when a box was present but resolved to no manifest/original/update. Small, well-scoped diff (2 files, +86/-1); scope matches the description exactly, no creep.

### Change Surface
- 1 new struct field (`c2pa_box_present`), 1 new `match` arm in `BmffIO::read_c2pa`, 3 new tests, 4 test-fixture struct-literal updates.
- Surfaces touched: `sdk/src/asset_handlers/bmff_io.rs`, `sdk/src/assertions/bmff_hash.rs` (test fixtures only).
- No new public API; `C2PABmffBoxes` is `pub(crate)`.

### Blast Radius
- Confirmed `original`/`update` pairing (from #2471) resolves before the new check, so `manifest_bytes`/`original_bytes`/`update_bytes` are mutually exclusive by the time the new `match` runs — no double-counting.
- Confirmed the many hashing callers of `read_bmff_c2pa_boxes` (`bmff_hash.rs`, `builder.rs`) only destructure fields they use and don't construct `C2PABmffBoxes` themselves outside the one production site — unaffected by the new field, as the PR description claims.
- Found an unaddressed reviewer concern (see inline comment) about the interaction with `Store::load_jumbf_from_stream`'s remote-manifest fallback.

### Test Gaps
- No test covers a BMFF asset that has both (a) a C2PA box resolving to no manifest and (b) an XMP remote-manifest provenance reference, to pin down the fallback interaction noted inline.
- Otherwise coverage is good: all three new tests use precise `matches!` assertions on the exact error variant/message, plus a negative test confirming the guard doesn't over-fire on a genuinely unsigned asset.

### Affected-Test Run
- Ran the affected area in an isolated worktree off `origin/ssanthosh/cai-13282`: `cargo test -p c2pa --lib bmff` → 73 passed, 0 failed, 1 pre-existing ignored (unrelated expired-cert test).
- Specifically confirmed the three new tests pass alongside the pre-existing sibling tests from #2471 (`test_read_cai_rejects_original_without_update`, `test_read_cai_rejects_update_without_original`, `test_read_cai_accepts_ordinary_manifest_alone`).

### Lint
- `cargo clippy -p c2pa --lib --tests -- -D warnings` → clean.
- `cargo +nightly fmt --all -- --check` → clean.

### Security
- This is itself a security-hardening fix; no new external-input, injection, or resource-exhaustion concerns introduced. No new dependencies.

### Suggestions Worth Supporting
- @tmathern's inline comment asking how this impacts remote manifests and whether it changes the `JumbfNotFound`-triggered fallback: agreed, and confirmed concretely — see the inline comment on `store.rs`. This is a real, reachable code path, not a theoretical one (`BmffIO` does implement `remote_manifest_url_ref`/`write_remote_manifest_url`), and it hasn't been answered or addressed yet.

### Verdict
**Approve with nits.** The core fix is correct, minimal, well-tested, and consistent with the #2471 precedent; verified green (tests/clippy/fmt) directly on the PR branch. The one open item — confirming the remote-manifest fallback interaction @tmathern flagged — is worth a reply/test before merge but is a design-confirmation gap, not a demonstrated bug.

---

sdk/src/asset_handlers/bmff_io.rs
ssanthosh: Addressed via https://github.com/contentauth/c2pa-rs/pull/2606/commits/1c4075091343c7b0f723b6e64512dc5e6d4bf48d

sdk/src/asset_handlers/bmff_io.rs
ssanthosh: Addressed via https://github.com/contentauth/c2pa-rs/pull/2606/commits/1c4075091343c7b0f723b6e64512dc5e6d4bf48d

ssanthosh: Remote-only assets (no embedded C2PA box) are unaffected: `c2pa_box_present` is false for them, so `read_c2pa` still returns `JumbfNotFound` and the XMP remote-manifest fallback still fires as before. The behavior change only applies when an embedded C2PA box is present but doesn't resolve to a manifest (tampered/unrecognized purpose) — in that case we now deliberately return C2PAValidation instead of `JumbfNotFound`, specifically so `Store::load_jumbf_from_stream` does not fall back to a remote manifest and mask the tampering.

Added `test_load_jumbf_tampered_box_skips_remote_fallback`, which builds an asset with both a tampered C2PA box and a genuinely-readable remote-manifest XMP reference, and asserts it's rejected rather than silently resolved via the remote manifest.

sdk/src/asset_handlers/bmff_io.rs
ssanthosh: Added a one-line comment at the read_c2pa match arm documenting this as intentional.

ssanthosh:

scouten-adobe (approved):

PR #2608, opened 2026-09-07
fix: Resolve OCSP issuer from trust anchor when x5chain omits it

ssanthosh:
### Vulnerability
A signed asset whose protected `x5chain` contains only the **revoked leaf** — with its issuing CA configured as the trust anchor and omitted from the chain (a spec-recommended topology: C2PA §14.5 permits a single certificate and says the anchor should not be included; the official trust list has 11 `pathLen=0` issuing anchors) — is reported as **trusted**. OCSP revocation checking silently does nothing because every OCSP path needs the issuer at `chain[1]`: the fetch path can't build the OCSP request (`process_ocsp_responders` returns `None` when `certs.len() < 2`) so `fetch_and_check_ocsp_response` logs an *informational* `signingCredential.ocsp.inaccessible` and succeeds; the stapled/embedded path can't match the response's CertID (`cert_id_matches_signer` needs `chain[1]`). Net: a revoked leaf goes unchecked. 

### Fix
Add `ocsp_signing_chain(certs, ctp)`: when the x5chain doesn't already carry the issuer, resolve the issuing CA from the configured signing trust anchors (match anchor subject == leaf issuer) and insert it at index 1. Apply it right after the chain is extracted in `fetch_and_check_ocsp_response` and `check_stapled_ocsp_response` (the latter also backs the external-`CertificateStatus` path). Normal multi-cert chains and untrusted lone leaves are returned unchanged, so only the vulnerable topology changes behavior.

### Tests added
- `ocsp_signing_chain_resolves_issuer_from_anchor` — issuer inserted for a lone leaf; chain unchanged when the issuer is already present or no anchor matches.
- `revoked_lone_leaf_detected_after_issuer_resolved` — end-to-end: a revoked lone-leaf is **not** flagged with the leaf alone, but **is** flagged (`signingCredential.revoked`) once the issuer is resolved from the anchor.

## Checklist


sdk/src/crypto/cose/ocsp.rs
```diff
@@ -363,6 +363,8 @@ fn check_stapled_ocsp_response(
     // The OCSP response must pertain to the certificate that signed this
     // manifest, so bind it to that signer's certificate chain.
     let signing_cert_chain = cert_chain_from_sign1(sign1)?;
+    // Ensure the issuing CA is present (§14.5 lets the x5chain omit it).
+    let signing_cert_chain = ocsp_signing_chain(&signing_cert_chain, ctp);
```
tmathern, line 367: Should a more end-to-end test be added here to verify this? E.g. one going through `Store::get_store_validation_info`?

scouten-adobe, line 367: **[MEDIUM]** Agreeing with @tmathern's comment on this line: the current tests exercise `ocsp_signing_chain` and `OcspResponse::from_der_checked` directly, but neither this call site nor the one in `fetch_and_check_ocsp_response` (line ~466) is driven through an integration path like `Store::get_store_validation_info`. A wiring mistake here — e.g. forgetting to shadow `signing_cert_chain` with the resolved value, or passing the pre-resolution chain to `extend_ocsp_cert_chain`/`check_certificate_trust` below — would not be caught by the current unit tests, since they'd still pass by testing the helper in isolation. Worth adding one end-to-end test (or a fast follow-up) that goes through `check_ocsp_status`/`Store::get_store_validation_info` with the lone-leaf-plus-anchor fixture already added, to lock in that the fix is actually wired up at both call sites.

sdk/src/crypto/cose/ocsp.rs
```diff
@@ -446,6 +448,44 @@ fn extend_ocsp_cert_chain(ocsp_certs: &[Vec<u8>], signing_cert_chain: &[Vec<u8>]
...
+    // Issuer already in the chain?
+    if let Some(next_der) = certs.get(1) {
+        if let Ok((_, next)) = X509Certificate::from_der(next_der) {
+            if next.subject().as_raw() == leaf_issuer_raw {
+                return certs.to_vec();
+            }
+        }
+    }
+
+    for anchor in ctp.signing_trust_anchors() {
```
scouten-adobe, line 475: **[LOW]** Anchor resolution here matches purely on subject-name (`anchor_cert.subject().as_raw() == leaf_issuer_raw`) and takes the *first* matching anchor in `ctp.signing_trust_anchors()`. If the trust list ever contains two Manifest anchors sharing a subject name but different keys (e.g. during CA reissuance/rotation), the wrong anchor could be inserted, producing an incorrect `issuerKeyHash` and a false non-match against a legitimate OCSP response — silently reverting to the pre-fix "OCSP checking skipped" behavior for that cert. This mirrors the existing pattern in `extend_ocsp_cert_chain` above, so it's not a regression introduced by this PR, but it's worth a short comment noting the assumption (single anchor per subject name) or a follow-up issue if the trust-list format could ever contain duplicates.

scouten-adobe (commented):
## PR Review

### Summary
This is a genuine security fix for a real bypass: a signed asset whose `x5chain` contains only the revoked leaf certificate (issuer omitted, as C2PA §14.5 explicitly permits) was being reported as trusted because both OCSP paths (fetch and stapled/embedded) need `chain[1]` = issuer to build the request / match the response's `CertID`, and silently no-op otherwise. The fix adds `ocsp_signing_chain()`, which inserts the issuer at index 1 by resolving it (subject-name match) from the configured signing trust anchors when it's missing, and applies it at both call sites (`fetch_and_check_ocsp_response`, `check_stapled_ocsp_response`). I traced both call sites, the downstream `cert_id_matches_signer` / `process_ocsp_responders` consumers, and confirmed `signing_trust_anchors()` (Manifest-type only) is the right filter since every caller of `check_ocsp_status` passes a manifest-signing `CertificateTrustPolicy` (no separate CAWG OCSP path exists in this codebase). Scope matches the PR description exactly — one file, additive only, no deletions.

### Validation performed
- `cargo +nightly fmt --all -- --check`: clean
- `cargo clippy --all-targets --no-deps -- -D warnings` (sdk crate): clean
- Ran `crypto::cose` (28 tests) and `crypto::ocsp` (6 tests) modules plus the 3 tests new to this PR in an isolated worktree at `0be93f37`: all 34 passed
- Confirmed via code reading that both patched call sites correctly use the shadowed/resolved chain downstream

### Security
Verified the fix actually closes the hole: `revoked_lone_leaf_detected_after_issuer_resolved` demonstrates revocation is missed pre-resolution and caught post-resolution using real fixture data. No secrets/injection/resource-exhaustion/path-traversal/deserialization concerns — the trust-anchor list isn't attacker-controlled and all cert parsing fails closed. No new dependencies.

### Suggestion worth supporting
@tmathern asked on line 367: *"Should a more end-to-end test be added here to verify this? E.g. one going through `Store::get_store_validation_info`?"* — I agree, see my inline note below; it's the one real gap in an otherwise thorough test suite.

### Verdict
**Approve with nits.** The fix is correct, well-scoped, and well-tested at the unit level. Nothing here blocks merge — the two inline notes are non-blocking follow-ups.

---

sdk/src/crypto/cose/ocsp.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2608/commits/68682b751fe1b3cb2411ef45909c99713d6fd50c

sdk/src/crypto/cose/ocsp.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2608/commits/68682b751fe1b3cb2411ef45909c99713d6fd50c

ssanthosh: Added one End to End test via `check_ocsp_status`

ssanthosh:

scouten-adobe (approved):

PR #2609, opened 2026-09-07
fix: Reject fragmented BMFF fragments matching a Merkle map without an initHash

ssanthosh:
### Vulnerability
In `BmffHash::verify_stream_segment` / `verify_stream_segments`, fragment hashing and the Merkle check were nested under `if let Some(init_hash)`, so a fragment matching a `MerkleMap` with `init_hash == None` skipped all verification and returned `Ok` — tampered fragment content verified as trusted. Such maps are reachable via public `Builder` APIs, and fragments select their map by attacker-editable `(uniqueId, localId)`. In scope per SECURITY.md (validation bypass).

### Fix
Both multi-file verifiers now reject a fragment matching an `initHash`-less `MerkleMap`. A missing required field is a malformed assertion, so this returns `Error::C2PAValidation(ASSERTION_BMFFHASH_MALFORMED)` (per review) — surfacing as `assertion.bmffHash.malformed` rather than `.mismatch`, consistent with `verify_self`'s structural checks. Legitimate fragmented verification is unaffected; the single-file path already verified the Merkle tree unconditionally.

### Tests added
`fragment_with_no_init_hash_is_rejected` / `fragment_files_with_no_init_hash_is_rejected` — assert the tampered fragment is rejected with `Error::C2PAValidation`; fail on the old code (returned `Ok`), pass with the fix. Positive round-trip stays covered by `test_builder_fragmented`.

## Checklist

### DIFF
--- a/sdk/src/assertions/bmff_hash.rs
+++ b/sdk/src/assertions/bmff_hash.rs
@@ -1682,6 +1682,12 @@ impl BmffHash {
                             {
                                 return Err(Error::HashMismatch("Fragment not valid".to_string()));
                             }
+                        } else {
+                            // A fragmented BMFF MerkleMap must carry an initHash; without
+                            // it the fragment would go entirely unverified.
+                            return Err(Error::HashMismatch(
+                                "BMFF fragment MerkleMap missing required initHash".to_string(),
+                            ));
                         }
                     } else {
                         return Err(Error::HashMismatch("Fragment had no MerkleMap".to_string()));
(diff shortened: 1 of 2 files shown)


scouten-adobe (approved):

sdk/src/assertions/bmff_hash.rs
```diff
@@ -1682,6 +1682,12 @@ impl BmffHash {
                             {
                                 return Err(Error::HashMismatch("Fragment not valid".to_string()));
                             }
+                        } else {
+                            // A fragmented BMFF MerkleMap must carry an initHash; without
+                            // it the fragment would go entirely unverified.
+                            return Err(Error::HashMismatch(
```
tmathern: If something that should be ehre is missing, is it a hash mismatch or a `ASSERTION_BMFFHASH_MALFORMED` (that error is already used at other locations, e.g. https://github.com/contentauth/c2pa-rs/blob/e7424abe6adb50f47d6daf38459472bc6852ae11/sdk/src/assertions/bmff_hash.rs#L1170-L1176).

tmathern (approved):

sdk/src/assertions/bmff_hash.rs
ssanthosh: @tmathern : Thanks for the suggestion.  ASSERTION_BMFFHASH_MALFORMED definitely looks to be a better error mapping for the current scenario. Updated the changes with https://github.com/contentauth/c2pa-rs/pull/2609/commits/0cccca5135da13ddc9c2471df3bd0a4ce6e053ec

PR #2610, opened 2026-09-07
fix: Reject non-contiguous C2PA APP11 segments when computing JPEG manifest-store exclusion

ssanthosh:
### Vulnerability
When an active update manifest inherits a data hash, c2pa-rs replaces the signed manifest-store exclusion with a range recomputed from the current asset. The format scanners over-measured that range: `JpegIO::get_object_locations` summed the lengths of every same-EN APP11 segment with **no contiguity/sequence check** (unlike the reader, which enforces the `z` sequence), and `TiffIO::get_object_locations` used the attacker-declared IFD `value_count` (tag `0xCD41`) verbatim. Either lets an attacker place foreign, decoder-effective bytes inside the recomputed exclusion (unhashed) while it still validates — reporting a tampered asset as `Trusted` with no key. Missing spec check: C2PA 15.12.1 (an exclusion range must contain only the Manifest Store + padding). In scope per SECURITY.md (validation bypass).

### Fix
Bound each format scanner to the true manifest-store extent:
- **JPEG** — reject a same-EN CAI continuation segment that is out of `z` sequence (mirroring `read_c2pa`) or not physically contiguous with the running range, so foreign bytes can't sit inside the exclusion.
- **TIFF** — bound the IFD `value_count` to the actual JUMBF superbox length (LBox/XLBox, big-endian) and reject a mismatch; skipped for the zero placeholder written during signing (not yet a JUMBF box).

Legitimate manifests (contiguous / `value_count` == manifest length) are unaffected.

### Tests added
- `noncontiguous_cai_segments_rejected` (JPEG) — foreign APP11 between two CAI segments is rejected; the untampered file still resolves.
- `noncontiguous_cai_injection_rejected_by_reader` (JPEG, end-to-end) — injects `[foreign APP11, CAI copy]` into a signed update-manifest asset (`update_manifest.jpg`) and asserts the `Reader` reports `ValidationState::Invalid` (baseline is not `Invalid`).
- `test_get_object_location` (TIFF) — updated to store a real minimal JUMBF box so it exercises the new `value_count` bound.

## Checklist


gpeacock (commented):
This looks good for jpeg, but the issue also calls out a similar vulnerability in TIFF and says we need a fix in claim.rs as well:

> **Recommended fix:**
> 
> 1. In `jpeg_io.rs::get_object_locations_from_stream`, measure the C2PA extent as a **span**
>    (`last_end - first_start`) with an explicit **contiguity check** instead of accumulating
>    `+= seg.len_with_entropy()`. `read_cai` at `:195-199` already implements the `z <= cai_seg_cnt`
>    test the calculator omits -- apply the same rejection there.
> 2. In `claim.rs::verify_hash_binding` (`:2691`), refuse to widen the signed exclusion by anything
>    other than the measured update-manifest delta, and enforce spec 15.12.1 by rejecting with
>    `assertion.dataHash.mismatch` when the widened window contains bytes other than the manifest store
>    and permitted padding.
> 
> A JPEG-only fix would be incomplete: `sdk/src/asset_handlers/tiff_io.rs` reaches the same
> `claim.rs:2691` substitution by a different length primitive (an attacker-declared IFD `value_count`
> on tag `0xCD41`), so both measurement sites need bounding.

scouten-adobe (changes requested):
In `sdk/src/store.rs` (which is not touched by this diff, but interacts with it), the silent `if let Ok(locations)` swallow means a rejected asset surfaces at validation time only as an indirect hash mismatch, not a clear diagnostic. Worth a follow-up to surface a dedicated validation status here.

The new test (`noncontiguous_cai_segments_rejected`) only exercises `JpegIO::get_object_locations` in isolation. There's no end-to-end test that reproduces the actual attack (update manifest inheriting a data hash + non-contiguous injection) through `Store`/`Reader` asserting the final validation status isn't `Trusted`. Given the indirect failure path traced above, an integration-level test would give much stronger assurance the full vulnerability is closed.

I agree with @gpeacock: the disclosed vulnerability isn't fully closed by a JPEG-only fix. Their suggestion to harden `claim.rs::verify_hash_binding` directly (refuse to widen the signed exclusion beyond the measured update-manifest delta) would close this bug class at the shared choke point instead of format-by-format, and would also avoid the indirect-hash-mismatch diagnostic gap noted above. Recommend treating the JPEG (and TIFF) scanner checks as useful additional hardening on top of that, not a substitute for it.

ssanthosh:
@gpeacock / @scouten-adobe : Performed the following changes with https://github.com/contentauth/c2pa-rs/pull/2610/commits/0ff8c27a5baa634ae7da042a0ac21f7b6a8af15c : 

1. Added TIFF scanner bound for the similar issue
2. Added the z- check for JPEG IO suggested
3. EndToEnd Reader Test

I will add a follow up task for hardening the generic claim.rs::verify_hash_binding directly and the current hardening would fix the existing reported vulnerability issue

ssanthosh:
@gpeacock / @scouten-adobe : Addressed the review comments. Please check

scouten-adobe (approved):
The JPEG contiguity check and TIFF `value_count` bound close both concrete attack vectors from the vuln report, and `noncontiguous_cai_injection_rejected_by_reader` gives real end-to-end coverage through `Reader`/`Store` (the gap I'd flagged). Thanks for adding that.

Filed the two remaining structural items as follow-ups rather than blocking this PR on them:
- #2645 — harden `claim.rs::verify_hash_binding` at the shared choke point (both @gpeacock and I noted format-by-format bounding isn't a substitute for this)
- #2646 — `store.rs`'s `if let Ok(locations)` swallow turns a scanner rejection into a silent, indirect hash mismatch instead of a clear diagnostic

Approving on that basis.

PR #2611, opened 2026-09-08
fix: Preserve numeric arrays in JSON output

### LINKED ISSUE #2570: Custom assertion: JSON integer array silently coerced to CBOR byte string, values >255 truncated mod 256
### Component

c2pa-rs / c2patool

### Version

c2patool 0.27.16 (c2pa-rs 0.90.16, reported by the tool itself as `org.contentauth.c2pa_rs`)

### Platform

macOS (arm64), installed via Homebrew

### What happened?

A custom assertion whose `data` is (or contains) a JSON array of small integers gets silently corrupted when embedded: it is coerced into a CBOR **byte string**, and any element that doesn't fit in a single byte is truncated modulo 256 instead of being preserved or rejected.

**Repro**

1. A tiny test image, `test-image.jpg` (any JPEG works).
2. `es256_private.key` / `es256_certs.pem` — the project's own test fixtures from `cli/sample/` in this repo (unmodified).
3. `manifest.json`:

```json
{
  "alg": "es256",
  "private_key": "es256_private.key",
  "sign_cert": "es256_certs.pem",
  "claim_generator_info": [
    { "name": "cbor_bug_repro", "version": "0.1.0" }
  ],
  "assertions": [
    {
      "label": "org.example.test",
      "data": { "values": [96, 384] }
    }
  ]
}
```

4. Sign and read back:

```sh
c2patool test-image.jpg -m manifest.json -o signed.jpg -f
c2patool signed.jpg -d
```

**Observed**

The manifest signs and validates cleanly (`"validation_state": "Valid"`, only the expected `signingCredential.untrusted` status for a test/self-signed cert), but the `org.example.test` assertion comes back as:

```json
"org.example.test": {
  "values": "YIA="
}
```

`base64.b64decode("YIA=")` is the two bytes `[96, 128]`. The input array was `[96, 384]`: `96` survived, but `384` (`0x180`) was silently truncated to `0x80` = `128` — i.e. each element got cast to `u8` (`value % 256`) and the whole array was re-typed as a CBOR byte string (major type 2) instead of a CBOR array of integers (major type 4). There is no warning, error, or validation failure anywhere in the sign or verify output — the data loss is completely silent.

The same thing happens whether the array is the assertion's `data` directly (`"data": [96, 384]`) or nested one level inside an object field (`"data": {"values": [96, 384]}`), so this isn't specific to a particular assertion shape.

**Workaround**

Replacing the bare integer array with a named-key object avoids the coercion entirely and round-trips exactly:

```json
"data": { "value_0": 96, "value_1": 384 }
```

Signing with that manifest instead produces `"org.example.test": {"value_0": 96, "value_1": 384}` on read-back — correct.

A JSON array of integers in a custom assertion's `data` should round-trip as a CBOR array of integers (or, at minimum, values that don't fit in a byte should cause a validation error rather than being silently wrapped/truncated). Silent, unbounded-looking-but-actually-lossy numeric truncation in signed, hash-bound manifest content seems like it undermines the integrity guarantees C2PA manifests are meant to provide — a consumer reading back `"values": "YIA="` / `[96, 128]` has no way to know the signer actually asked to store `[96, 384]`.

### Where I looked

I didn't chase this all the way to a single line, but for what it's worth: the JSON manifest's custom assertions appear to go through `AssertionDefinition::deserialize` in `sdk/src/builder.rs`, which for the default (CBOR) kind calls `c2pa_cbor::value::to_value(helper.data)` on the raw `serde_json::Value` before it becomes `AssertionData::Cbor(...)`. That call, or something further down in the `c2pa-cbor` crate's serialization path, is where I'd guess the array-to-bytes coercion happens (similar in spirit to the `is_human_readable()`-driven byte/text confusion already reported in #2509, though the direction and mechanism here look different: numbers → bytes with modulo truncation, rather than bytes → base64 text).

### Additional context

- I checked for a fix in the latest release before filing: `c2patool 0.27.16` / `c2pa 0.90.16` (2026-08-27) is current as of this report, and its release notes don't mention this.
- I searched existing issues for anything matching (CBOR, byte string, integer array, custom assertion, truncation) and didn't find a match. #2319 (CBOR map key ordering), #2333 (crJSON b64 delimiter, merged), #2508/#2509 (OCSP `certificate-status` bytes serializing as base64 text) are all CBOR-adjacent but describe different bugs.

AlvaroBalbin:

Custom assertion data such as `[96, 384]` currently appears as `"YIA="` in JSON reports. The array is stored correctly in CBOR; the report formatter mistakes it for bytes and truncates its values while encoding them. This change keeps `[96, 384]` as an array in Reader's normal and detailed JSON reports and in Builder's display output.

Replace the numeric-array heuristic with formatting that uses serialization types. Actual byte strings, including typed claim hashes, remain base64 strings. For assertions already normalized into JSON, use the original CBOR types to identify matching byte fields while preserving the normalized representation. Assertion instances are resolved independently, and post-validation results are kept as supplied. Stored assertion data and the typed API representation remain unchanged. There are no new public APIs or dependencies.

Existing ambiguity when distinct CBOR map keys collapse to the same JSON property remains outside this fix. This does not provide lossless conversion of arbitrary CBOR maps.

Fixes #2570.

### Validation

- Eight new regression tests pass, covering numeric arrays in JSON and CBOR assertions, actual byte strings, Builder display, assertion instances, post-validation values, unsigned integers and numeric keys, normalized fields, and supported nesting depth.
- The original three regression tests fail against the baseline, reproducing the incorrect report output.
- All 148 selected SDK tests pass: 147 in the offline affected-module run and the one remote-fixture test when rerun with network access. This includes Builder, Reader, detailed report, and new helper tests.
- Before/after CLI comparison uses the same signed image with repository test credentials. Normal and detailed reports now preserve all input arrays. The original and fixed crJSON controls agree, the signed file's SHA256 is unchanged, and validation remains `Valid`. Signing and both sets of reads ran without network access.
- Linux builds use Rust 1.88.0. `cargo build -p c2patool --locked --offline` and `cargo +nightly fmt --all -- --check` pass.
- `cargo clippy -p c2pa --locked --all-targets --features file_io,fetch_remote_manifests,add_thumbnails -- -D warnings` was run and fails on existing upstream diagnostics. Every emitted source span was compared with the base commit and is unchanged. No emitted diagnostic points to this patch. This is not a clean Clippy pass.
- Full upstream CI across platforms and feature configurations has not been run locally.

## Checklist

- [x] Any `TO DO` items (or similar) have been entered as GitHub issues and the link to that issue has been included in a comment. No new TODO items were added.

### DIFF
--- a/sdk/src/builder.rs
+++ b/sdk/src/builder.rs
@@ -56,7 +56,7 @@ use crate::{
     settings::{builder::TimeStampFetchScope, MAX_ASSERTIONS},
     store::Store,
     utils::{
-        hash_utils::hash_to_b64, merkle::MerkleAccumulator, mime::format_to_mime,
+        json_report, merkle::MerkleAccumulator, mime::format_to_mime,
         path_utils::sanitize_archive_path, xmp_inmemory_utils::XmpInfo,
     },
     AsyncSigner, ClaimGeneratorInfo, EphemeralSigner, HashRange, HashedUri, Ingredient,
@@ -3786,8 +3786,7 @@ impl Builder {
 
 impl std::fmt::Display for Builder {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let mut json = serde_json::to_value(self).map_err(|_| std::fmt::Error)?;
-        json = hash_to_b64(json);
+        let json = json_report::to_value(self).map_err(|_| std::fmt::Error)?;
         let output = serde_json::to_string_pretty(&json).map_err(|_| std::fmt::Error)?;
         f.write_str(&output)
     }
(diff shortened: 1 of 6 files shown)


scouten-adobe (approved):
Also requesting review from @gpeacock.

gpeacock:
This is a great idea, I've now built it into the lower levels of the SDK. c2pa_cbor is decoded directly into json, allowing us to preserve the types much better than we did before

PR #2612, opened 2026-09-08
fix: Reject signing certificate whose key algorithm does not match the COSE signing algorithm

ssanthosh:
### Vulnerability
COSE `-8` (EdDSA) selects `SigningAlg::Ed25519`, but the OpenSSL raw validator (`c2pa_raw_crypto`) parses whatever key the certificate carries and calls `verify_oneshot` with no key-type check — so an **Ed448** certificate (SPKI OID `1.3.101.113`) verifies under the Ed25519 selection. With an order-four Ed448 key, one fixed 114-byte signature validates arbitrary claims and the SDK reports `Trusted` **without any end-entity private key**. The rust_native validator enforces the OID and is not affected. In scope per SECURITY.md ("bypass C2PA manifest validation" / "errors in cryptographic implementations").

### Fix
At the verifier, bind the COSE-declared `SigningAlg` to the end-entity certificate's subject-key algorithm before validating the signature; a mismatch is rejected as `SIGNING_CREDENTIAL_INVALID`. The check is driven by the `SigningAlg` enum (single source of truth) rather than a static OID allowlist, and a future non-exhaustive variant defers to its own validator. This is a defense-in-depth fix; the root cause (the OpenSSL validator's missing key-type check) should also be fixed upstream in `c2pa_raw_crypto` to match rust_native. Both the C2PA claim path and the CAWG identity X.509 path route through this verifier, so both are covered.

### Tests added
`ed448_key_rejected_under_ed25519` (the reported bypass), `matching_keys_accepted` (Ed25519/ES256/PS256 still match), and `cross_algorithm_key_rejected`. Adds an `ed448.pub` fixture (CA-issued Ed448 end-entity cert).

## Checklist


sdk/src/crypto/cose/verifier.rs
```diff
@@ -317,3 +357,56 @@ fn dump_cert_chain(certs: &[Vec<u8>]) -> Result<Vec<u8>, CoseError> {
...
+    #[test]
+    fn cross_algorithm_key_rejected() {
+        assert!(!key_matches(
+            include_bytes!("../../../tests/fixtures/crypto/raw_signature/es256.pub"),
+            SigningAlg::Ed25519
+        ));
+        assert!(!key_matches(
+            include_bytes!("../../../tests/fixtures/certs/ed25519.pub"),
+            SigningAlg::Es256
+        ));
```
mpgopala: Add cases for Ps256 using Es256 and others?

sdk/src/crypto/cose/verifier.rs
```diff
@@ -58,6 +60,25 @@ pub enum Verifier<'a> {
...
+
+// Is the certificate's key algorithm the one the COSE `SigningAlg` implies?
+// Blocks algorithm substitution (e.g. an Ed448 key under the Ed25519 selection).
+// A future non-exhaustive variant defers to its own validator.
+fn spki_matches_signing_alg(alg: SigningAlg, spki_alg: &Oid) -> bool {
+    match alg {
+        SigningAlg::Es256 | SigningAlg::Es384 | SigningAlg::Es512 => *spki_alg == EC_PUBLICKEY_OID,
+        SigningAlg::Ps256 | SigningAlg::Ps384 | SigningAlg::Ps512 => {
+            *spki_alg == RSA_OID || *spki_alg == RSASSA_PSS_OID
+        }
```
scouten-adobe, line 93: **[MEDIUM]** The `_ => true` fallback fails **open** for any future `SigningAlg` variant not added to this match. `SigningAlg` is `#[non_exhaustive]` specifically to allow new algorithms later (e.g. Ed448, a PQC scheme), so this match will eventually need a corresponding update — and if it's forgotten, we're back to "trust the validator to check key type by itself," which is exactly the assumption that caused this bug. Given this function exists purely as a defense-in-depth backstop, I'd lean toward failing closed (reject unknown/unmapped algorithms) rather than open, or at least a loud comment/tracking mechanism tied to `SigningAlg` additions so this doesn't get missed.

scouten-adobe (approved):
## Approve with concerns

This correctly fixes the reported algorithm-substitution bypass. I independently verified the root cause in `c2pa_raw_crypto` 0.1.2: the OpenSSL `Ed25519Validator` (`openssl/validators/ed25519_validator.rs`) parses whatever key type is in the DER and calls `verify_oneshot` with no OID check, while the `rust_native` `Ed25519Validator` does check the SPKI OID. The new guard in `verify_signature` closes that gap at the one shared entry point — I grepped the whole workspace and `validator_for_signing_alg`/`.validate(...)` is only called from this one site, so both the C2PA claim path and the CAWG identity path are covered as the description claims.

**Verification performed:**
- Confirmed `ed448.pub` is a genuine CA-issued Ed448 end-entity cert (`openssl x509 -text` shows `Public Key Algorithm: ED448`).
- Ran the 3 new tests plus the full `crypto::cose::*` suite (29 tests) under the default (openssl) feature set — all pass.
- Re-ran the 3 new tests under `--no-default-features --features rust_native_crypto` — all pass, confirming the check is backend-agnostic.
- `cargo +nightly fmt --all -- --check` and `cargo clippy -p c2pa --all-targets --no-deps -- -D warnings` — both clean.

One concern noted inline below (non-blocking): the fallback for algorithms outside the current match fails open, which is the same class of assumption that caused the original bug. Also seconding @mpgopala's suggestion above to add RSA↔EC cross-cases to `cross_algorithm_key_rejected` for fuller coverage — non-blocking, but worth a follow-up.

Approving — this is a solid, narrowly-scoped, well-tested fix for a real vulnerability and I don't think either point should hold up landing it.

---

sdk/src/crypto/cose/verifier.rs
tmathern: +1

sdk/src/crypto/cose/verifier.rs
```diff
@@ -144,6 +165,25 @@ impl Verifier<'_> {
         let pk = sign_cert.public_key();
         let pk_der = pk.raw;
 
+        // Reject a certificate whose key algorithm does not match the declared
+        // COSE signing algorithm (e.g. an Ed448 key under the Ed25519 selection).
+        if !spki_matches_signing_alg(alg, &pk.algorithm.algorithm) {
```
tmathern: Do you also need to check the curve used in the certificate is the expected one, aka `pk.algorithm`'s algorithm + parameters?

tmathern (approved):

sdk/src/crypto/cose/verifier.rs
ssanthosh: Yes, It is needed. Done with https://github.com/contentauth/c2pa-rs/pull/2612/commits/f7b2ac2a95240239df6b9e4e546de3aa8b33b6b3

sdk/src/crypto/cose/verifier.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2612/commits/f7b2ac2a95240239df6b9e4e546de3aa8b33b6b3

sdk/src/crypto/cose/verifier.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/pull/2612/commits/f7b2ac2a95240239df6b9e4e546de3aa8b33b6b3

scouten-adobe (approved):
Appreciate the follow-up. Approved with no concerns.

PR #2614, opened 2026-09-08
fix: Report unrecognized CAWG identity sig_type instead of silently ignoring it

ssanthosh:
### Vulnerability
CAWG identity validation is on by default (`decode_identity_assertions: true`). When an identity assertion's `signer_payload.sig_type` is unrecognized, `IdentityAssertion::validate_partial_claim` fell through to a terminal `else` that returned a bare `Err` with **no logged validation status**, and `Manifest::from_store` discarded that `Err` with `.ok()`. So an assertion the SDK cannot validate was silently accepted: the asset is reported `Valid` (and `Trusted` where the claim signer is trust-listed) with a null CAWG status and empty failure array. CAWG Identity Assertion §7.1 says such assertions SHALL be reported as `cawg.identity.sig_type.unknown`. 

### Fix
Log `cawg.identity.sig_type.unknown` (failure) in the terminal `else` before returning the error, mirroring the per-verifier emitters that were unreachable from this path. The function is `#[async_generic]`, so both sync and async paths are covered.

Per review, a CAWG identity failure's blast radius is only that assertion, never the enclosing manifest: added the `cawg.identity.` prefix to `is_tolerated_manifest_failure_code` (alongside the existing `cawg.x509.`). So `cawg.identity.sig_type.unknown` (and its siblings) surface in the validation results but do not flip `validation_state()` to `Invalid`.

### Tests added
Updated the `invalid_sig_type` tests (`continue_when_possible`, `stop_on_error`) and the two `identity_assertion_builder` naive-credential tests: their default-`Reader` reads now assert the `cawg.identity.sig_type.unknown` failure is surfaced (previously asserted `None`, which encoded the bug). They fail without the fix and pass with it.

## Checklist


scouten-adobe (approved):
Looks great. Thank you!

sdk/src/identity/identity_assertion/assertion.rs
```diff
@@ -447,6 +447,15 @@ impl IdentityAssertion {
             serde_json::to_value(result)
                 .map_err(|e| ValidationError::UnknownSignatureType(e.to_string()))
         } else {
+            // Per CAWG Identity Assertion §7.1, an unrecognized sig_type SHALL be
+            // reported rather than silently ignored.
+            log_current_item!("unsupported signature type", "validate_partial_claim")
+                .validation_status("cawg.identity.sig_type.unknown")
```
tmathern, line 453: There were recent changes to "tolerate" some errors. Is this a tolerated one, or would it flip the overall manifest's valdiation status?

ssanthosh: It is not tolerated and flips the overall manifest validation status to INVALID. `is_tolerated_manifest_failure_code` only covers `signingCredential.untrusted` + `cawg.x509.*`, so `cawg.identity.sig_type.unknown` invalidates. It is consistent with other sibling `cawg.identity.*` codes  
cc : @scouten-adobe

scouten-adobe: @tmathern is right here and I'm glad she caught it. In general, the blast radius for a CAWG validation failure should be _only_ the CAWG identity assertion in question. It _should not_ cause the entire manifest to be invalidated.

scouten-adobe (changes requested):
See @tmathern's comment about scope of validation findings for CAWG assertions.

sdk/src/identity/identity_assertion/assertion.rs
ssanthosh: @tmathern / @scouten-adobe : Fixed the issue with https://github.com/contentauth/c2pa-rs/pull/2614/commits/21f181fe4088be53baff7bee6cfc5149df6c5083. Added the cawg.identity. prefix to is_tolerated_manifest_failure_code (alongside the existing cawg.x509.), so a CAWG identity failure's blast radius is only that assertion and never invalidates the enclosing manifest. cawg.identity.sig_type.unknown (and its siblings) still surface in the validation results, but validation_state() stays Valid. Existing tests that assert the code is reported continue to pass.

ssanthosh:

tmathern (approved):

scouten-adobe (approved):
Re-reviewed after the blast-radius fix. The `cawg.identity.` prefix addition to `is_tolerated_manifest_failure_code` correctly scopes the failure to the identity assertion, matching the existing `cawg.x509.` handling.

Added a unit test (`not_trusted_state_with_cawg_identity_failure`, mirroring `not_trusted_state_with_cawg_x509_failure`) to lock in that `cawg.identity.*` failures don't flip `validation_state()` to `Invalid` — verified it fails without the fix and passes with it.

PR #2615, opened 2026-09-08
fix: Check issuing CA revocation via AIA on the online OCSP fetch path

ssanthosh:
### Vulnerability
With `verify.ocsp_fetch = true`, the online OCSP path only queried the **signing leaf's** responder: `process_ocsp_responders`/`build_ocsp_request` read AIA and fixed the OCSP subject to `certs[0]`. No issuing CA was ever an OCSP request subject, so a `good` leaf beneath a CA revoked for key compromise stayed `Trusted` reporting `signingCredential.ocsp.notRevoked`, even though a signed `revoked` response was published at the CA's AIA endpoint. C2PA 2.4 §15.9 requires determining CA revocation from AIA and reporting `signingCredential.untrusted` when a CA is revoked. 

### Fix
Parameterize the fetch path by subject index so each certificate is queried against its own AIA responder (subject = `certs[i]`, issuer = `certs[i+1]`). In `fetch_and_check_ocsp_response`, query each issuing CA and, if one is validly reported revoked, emit `signingCredential.untrusted` and reject. The per-response validation (CertId binding, responder EKU + path authorization) is factored into a shared helper reused for the leaf and each CA, so no security check is weakened. 

### Tests added
`each_cert_is_queried_against_its_own_responder` — with a `[leaf(AIA), issuing-CA(AIA), root]` chain, the leaf is queried against its responder and the issuing CA against its own (previously the CA was never contacted). Fails without the fix. Adds fixture `ocsp_ca_chain.pem`. 

## Checklist


sdk/src/crypto/cose/ocsp.rs
```diff
@@ -542,7 +579,7 @@ pub(crate) fn fetch_and_check_ocsp_response(
         // EKU check above does not establish it: `check_certificate_profile`
         // inspects a single certificate and builds no path, so a self-signed
         // certificate carrying id-kp-OCSPSigning satisfies it.
-        let ocsp_cert_chain = extend_ocsp_cert_chain(ocsp_certs, &certs);
+        let ocsp_cert_chain = extend_ocsp_cert_chain(ocsp_certs, full_chain);
```
scouten-adobe: **[CRITICAL]** This should pass `subject_chain`, not `full_chain`.

`full_chain` is always the leaf-anchored `certs` regardless of which certificate is being validated. For the CA-revocation loop above (line ~485), that means `extend_ocsp_cert_chain`'s `signing_cert_chain[1]` — meant to be the *direct issuer* of the subject under test — ends up being the CA cert itself instead of its real issuer (`certs[i+1]`).

Concretely, for chain `[leaf, CA, root]` where the CA is revoked and its OCSP response embeds only a delegated responder cert issued by `root` (the standard delegation pattern, and the exact case this function's doc comment says it's meant to handle): the issuer comparison here checks against the CA instead of `root`, the match fails, `extend_ocsp_cert_chain` returns just the lone responder cert with no path to root, `check_certificate_trust` fails closed, and `validate_fetched_ocsp` returns `OcspResponse::default()`. Back in the caller, `ca_response.revoked_at.is_some()` is `false`, so the loop just continues — the revoked CA is never flagged.

I confirmed this empirically with the `ocsp_ca_chain.pem` fixture from this PR: `extend_ocsp_cert_chain(&[root_der], &certs[1..])` (correct) extends to length 2, while `extend_ocsp_cert_chain(&[root_der], &certs)` (as shipped) stays at length 1 — no extension.

---

scouten-adobe (changes requested):
## PR Review

Thanks for tracking down and fixing this — parameterizing the fetch path by subject index and reusing a shared validation helper for the leaf and each CA is the right shape for this fix, and the added test correctly demonstrates that the issuing CA is now queried (which it wasn't before).

I think there's a bug in the refactor that silently defeats the fix for a common configuration, though — see the inline comment on `ocsp.rs:582`. I verified it empirically against the `ocsp_ca_chain.pem` fixture added in this PR: `extend_ocsp_cert_chain` gets called with the wrong chain reference, so when a revoked CA's OCSP response only embeds a delegated responder cert (issued by the CA's own issuer — the standard delegation pattern), the path-completion silently fails and the revocation is dropped rather than reported.

Everything else looks solid: no other callers of the changed `pub(crate)` functions exist outside these two files, `cargo +nightly fmt --all -- --check` and `cargo clippy -p c2pa --all-targets --no-deps -- -D warnings` are both clean on the branch, and the existing OCSP test suite (7 tests, including the new one) passes.

**Test gap**: no test currently exercises `fetch_and_check_ocsp_response` end-to-end for a *revoked* issuing CA — the new test only checks that the right AIA host gets queried, not that a revoked-CA response is recognized and rejected. That gap is exactly why the bug below wasn't caught; a regression test covering it would be worth adding alongside the fix.

---

sdk/src/crypto/cose/ocsp.rs
ssanthosh: `validate_fetched_ocsp` now builds the responder path from `subject_chain` (dropped the redundant full_chain param), so `extend_ocsp_cert_chain` sees the subject's real issuer (certs[i+1]) instead of the CA itself. Added a regression test for the above mentioned case.

scouten-adobe (approved):
Looks good. Thanks for the follow up!

PR #2616, opened 2026-09-08
fix(sdk): Correct two bugs in online OCSP responder validation (#2542 follow-up)

scouten-adobe:
## Summary

This is a follow-up to #2542, which introduced two bugs relative to the stapled-path it was intended to mirror.

**Bug 1 — trust check used wrong policy (security regression)**

`new_ctp.check_certificate_trust(...)` was called instead of `ctp.check_certificate_trust(...)`. `CertificateTrustPolicy::default()` produces an empty trust-anchor list in production (the only `add_trust_anchors` call is inside `#[cfg(test)]`), so `check_certificate_trust` iterates zero anchors and always returns `Err`. The result: every online OCSP response from a delegated responder is silently dropped as `OcspResponse::default()`, making online revocation checking permanently broken in that case — the exact attack surface #2542 was trying to close.

Fix: replace `new_ctp.check_certificate_trust(...)` with `ctp.check_certificate_trust(...)` (matching line 401 of `check_stapled_ocsp_response`).

**Bug 2 — EKU check wrote to caller's validation log**

The EKU check passed `validation_log` (the caller's `StatusTracker`) instead of `&mut current_validation_log` (the scratch tracker). A responder whose certificate fails the EKU check would write failure entries into the real log even though the response was being rejected without appending the scratch log — bypassing the log-hygiene protection that is the whole point of #2542's scratch-log approach.

Fix: pass `&mut current_validation_log` to `check_end_entity_certificate_profile`, matching the guard everywhere else in the function.

**Style nits (5)**

Same block, same diff: en dash instead of `--`, two comment openers capitalised, two comments given terminal periods, blank lines added before multi-line statements — all per the project's CLAUDE.md conventions.

## Test plan

- [x] `cargo +nightly fmt --all -- --check` clean
- [x] `cargo build -p c2pa --features openssl` clean
- [x] `cargo test -p c2pa --features openssl --lib -- ocsp` — 8/8 pass


tmathern:
(THis likely needs to be revised after changes in main, see merge conflict...).

gpeacock (approved):
looks good

ssanthosh (approved):
@scouten-adobe : Fixes look good. We need to fix the compilation issues before merging

PR #2617, opened 2026-09-08
chore: Remove c2patool from this repo and simplify release-plz workflows

scouten-adobe:
## Summary

* Removes the `cli` workspace member (c2patool source, docs, tests, fixtures) and its dedicated `c2patool-release.yml` binary-release workflow — `c2patool` now lives in its own repository, [contentauth/c2patool](https://github.com/contentauth/c2patool).
* `c2pa-c-ffi` continues to publish to crates.io exactly as before (no change to its `release-plz.toml` entry or to `semver-checks.yml`).
* With `c2patool` gone, `release-rc.yml` / `release-train-cut.yml` no longer need to compute and bump a second crate's independent version number in lockstep (`CUR_TOOL`/`RC_TOOL`/`cargo set-version -p c2patool`, the `c2patool-v*` tag, etc.) — those workflows now only track `c2pa` (which `c2pa-c-ffi` continues to follow automatically via the shared workspace version, same as before).
* **Stops tracking `Cargo.lock`.** This is a library workspace (`c2pa`, `c2pa-c-ffi`, plus internal tools), not an application, so the lockfile is no longer committed and is now in `.gitignore`. This also required:
  * Removing `tier-1a.yml`'s "Cargo.lock is up to date" job (`cargo check --locked`), since there's no committed lockfile left for it to validate.
  * Dropping `--locked` from `beta-preflight.yml`'s wasm-bindgen-version step, which otherwise fails outright on a fresh checkout (cargo refuses to generate a lock file under `--locked`).
  * Updating `reconciliation.yml`'s bookkeeping-path filter to drop the now-dead `Cargo.lock` path.
* Updates `docs/release-process.md`, `docs/support-tiers.md`, `SECURITY.md`, `README.md`, `.commitlintrc.yml`, and the bug report template to drop stale `c2patool` references.

## Not in scope

The deeper `release-plz set-version` simplification (replacing the hand-rolled promotion-changelog script, dropping `cargo-edit`) discussed as a possible follow-on once c2patool moved out is tracked separately on [#2315](https://github.com/contentauth/c2pa-rs/pull/2315) and isn't touched here. (That simplification was originally evaluated assuming c2pa-c-ffi would also stop publishing to crates.io; since c2pa-c-ffi remains published, `c2pa`/`c2pa-c-ffi` are still two crates release-plz manages, and #2315's own work is unaffected by this PR either way.)

## Test plan

- [x] `cargo check --workspace` and `cargo test --workspace --no-run` succeed with `cli` removed from the workspace and `Cargo.lock` untracked.
- [x] `cargo +nightly fmt --all -- --check` passes.
- [x] `release-plz.toml` parses correctly; the `c2pa-c-ffi` package entry is unchanged from `main`.
- [x] Verified locally that `cargo metadata --locked` fails on a checkout with no `Cargo.lock` present, confirming the `beta-preflight.yml` fix was necessary.
- [x] Rebased onto latest `main` (which had picked up a dependabot `Cargo.lock`-only commit after this branch was created) and re-resolved the resulting delete/modify conflict in favor of keeping `Cargo.lock` deleted. PR is confirmed mergeable, no conflicts.
- [ ] Maintainer review of the RC-build/train-cut workflow changes (release-rc.yml, release-train-cut.yml) before this leaves draft, since they aren't exercised by normal PR CI.


tmathern:
The c2pa_c_ffi crate holds the C FFI, which is used and extended by other projects. 
Why should it move out at the same time the cli tool does?

There are projects and downstream SDKs that depend on it (and not only pre-built libraries), and extend it for their own FFIs and extensions of language bindings based on what is provided here. The C FFI interfaces ands its utilities (e.g. macros) need then either to be reintegrated in c2pa-rs' published SDK, or become its own crate. Where is it going to go?

scouten-adobe (commented):
This looks right to me — hold until Friday in advance of next breaking-changes release.

gpeacock (approved):
We should also get rid of c2pa-c-ffi releases, but that could be in a separate PR.
What should we do with the previous c2patool releases in c2pa-rs? Do they stay there or get moved?

scouten-adobe:
> We should also get rid of c2pa-c-ffi releases, but that could be in a separate PR.

I'd like to do that, but there are clients (including our own internal clients) that rely on those.

> What should we do with the previous c2patool releases in c2pa-rs? Do they stay there or get moved?

My preference is that they remain where they are and new releases will go into the new(old) c2patool repo.

PR #2639, opened 2026-09-11
docs: Reject vulnerability reports whose only reproduction requires private/internal API access

mpgopala:
Recent HackerOne-sourced ticket triage surfaced reports whose sole demonstrated reproduction was an in-crate `#[cfg(test)]` unit test reaching `pub(crate)` internals -- code no external consumer of the published artifacts could ever call, regardless of how real the underlying defect was. Codify this as an explicit rejection criterion, with a carve-out for cases where the private access was only a replaceable convenience (e.g. building a test fixture) and the actual exploit runs through the public API or CLI.


SECURITY.md
```diff
@@ -20,6 +20,7 @@ Tickets that match the following characteristics will generally be **rejected**:
...
+* **Reproduction that requires calling a private or internal API.** If demonstrating the vulnerability requires code that only compiles as an in-crate test (e.g. `cargo test --lib` against a `#[cfg(test)]` module with `use super::*` insider visibility) to reach items no external consumer's code could ever reference, the report has not shown that someone who can only link against the published artifacts could actually exploit it, regardless of how real the underlying code defect is. It also does not disqualify a report where such access was merely a shortcut for something equally producible without it (e.g. a test fixture buildable just as well with external tooling), provided the private-API step itself could be replaced this way and the actual triggering and observation of the vulnerability goes through the public API or CLI.
```
scouten-adobe, line 23: > It also does not disqualify a report where such access was merely a shortcut for something equally producible without it (e.g. a test fixture buildable just as well with external tooling), provided the private-API step itself could be replaced this way and the actual triggering and observation of the vulnerability goes through the public API or CLI.

I'm not quite sure what you mean here. What I think we should say here is that any report **must** demonstrate how the fault is accessible through the public API surface; it _may_ also include a test case that is wired in at a lower level.

scouten-adobe (changes requested):

SECURITY.md
mpgopala: What I mean here is that people could use private API to produce a manifest faster which otherwise could be produced with other tools like hex editor or openssl. 
Use of private APIs is permissible only for this. As a tool to produce test assets to reproduce a VULN that can be reproduced ONLY using public API.

scouten-adobe (approved):

PR #2643, opened 2026-09-11
fix: Reject data hash exclusions for detached manifests without a matching C2PA box

ssanthosh:
### Vulnerability
A `c2pa.hash.data` assertion's `exclusions` range exists to skip the manifest's own embedded bytes. Presented as a detached manifest (sidecar, remote, or `c2patool --external-manifest`), the verifier applied that exclusion verbatim to whatever asset it was handed, without checking the excluded bytes were ever a C2PA box. An attacker with no signing key can extract a legitimately-signed manifest, overwrite exactly its declared exclusion range in a copy of the asset with unrelated content, and have it validate as `Trusted`. Confirmed via reproduction: a manifest lifted from a signed asset, paired with a copy whose exclusion range was overwritten with a self-contained different image, validated `Trusted` with no failures — while the file actually decoded to the attacker's content. In scope per SECURITY.md (validation bypass).

### Fix
Reject a non-empty (non-zero-length) `c2pa.hash.data` exclusion unless either: the manifest was read out of the asset being validated (`Store.embedded`, now tracked through both the read path and the sign-then-verify path), or the exclusion range matches a C2PA box actually present in the asset right now (via the existing `object_locations`-derived `manifest_store_range`, already used for update-manifest exclusion adjustment). This closes the bypass while still allowing the legitimate case of validating a manifest against the same asset it's still embedded in via the detached-loading API.

### Tests added
`test_detached_manifest_exclusion_hole_rejected` (reader.rs) — signs a real asset, extracts the manifest, overwrites exactly its exclusion range with unrelated bytes, and asserts the detached-manifest validation path (`Reader::from_manifest_data_and_stream`) does not report `Trusted`. Verified this test fails (reports `Trusted`) without the fix.
## Checklist

### DIFF
--- a/sdk/src/claim.rs
+++ b/sdk/src/claim.rs
@@ -2959,8 +2959,33 @@ impl Claim {
                     }
 
                     if !dh.is_remote_hash() {
-                        // there are extra exclusion then log the information code about extra exclusion
                         if let Some(exclusions) = &dh.exclusions {
+                            // Exclusions only make sense for a manifest embedded in this asset;
+                            // a detached manifest's exclusion is a hole for unrelated content.
+                            // If the caller didn't read the manifest out of this asset, fall back
+                            // to checking whether the asset really does have a C2PA box there.
+                            let box_present_in_asset = svi
+                                .manifest_store_range
+                                .as_ref()
+                                .is_some_and(|range| exclusions.contains(range));
+                            let has_real_exclusion = exclusions.iter().any(|e| e.length() > 0);
+                            if has_real_exclusion && !svi.is_embedded && !box_present_in_asset {
+                                log_item!(
+                                    claim.assertion_uri(&hash_binding_assertion.label()),
+                                    "data hash exclusions not valid for a detached manifest",
+                                    "verify_internal"
+                                )
+                                .validation_status(validation_status::ASSERTION_DATAHASH_MISMATCH)
+                                .failure(
+                                    validation_log,
+                                    Error::HashMismatch(
+                                        "data hash exclusions not valid for a detached manifest"
+                                            .to_string(),
+                                    ),
+                                )?;
+                            }
+
+                            // there are extra exclusion then log the information code about extra exclusion
                             if exclusions.len() > 1 {
                                 log_item!(
                                     claim.assertion_uri(&hash_binding_assertion.label()),
--- a/sdk/src/ingredient.rs
+++ b/sdk/src/ingredient.rs
@@ -883,35 +883,39 @@ impl Ingredient {
     ) -> Result<Self> {
         let mut validation_log = StatusTracker::default();
 
-        // retrieve the manifest bytes from embedded or remote and convert to store if found
+        // retrieve the manifest bytes from embedded or remote and convert to store if found;
+        // also track whether the bytes came from parsing `stream` itself (embedded) as
+        // opposed to `manifest_data()` or a remote fallback (not embedded).
         let jumbf_result = match self.manifest_data() {
-            Some(data) => Ok(data.into_owned()),
+            Some(data) => Ok((data.into_owned(), false)),
             None => if _sync {
                 Store::load_jumbf_from_stream(format, stream, context)
             } else {
                 Store::load_jumbf_from_stream_async(format, stream, context).await
             }
-            .map(|(manifest_bytes, _)| manifest_bytes),
+            .map(|(manifest_bytes, remote_url)| (manifest_bytes, remote_url.is_none())),
         };
 
         // We can't use functional combinators since we can't use async callbacks (https://github.com/rust-lang/rust/issues/62290)
         let (mut result, manifest_bytes) = match jumbf_result {
-            Ok(manifest_bytes) => {
+            Ok((manifest_bytes, embedded)) => {
                 let result = if _sync {
-                    Store::from_manifest_data_and_stream(
+                    Store::from_manifest_data_and_stream_with_embedded(
                         &manifest_bytes,
                         format,
                         &mut *stream,
                         &mut validation_log,
                         context,
+                        embedded,
                     )
                 } else {
-                    Store::from_manifest_data_and_stream_async(
+                    Store::from_manifest_data_and_stream_with_embedded_async(
                         &manifest_bytes,
                         format,
                         &mut *stream,
                         &mut validation_log,
                         context,
+                        embedded,
                     )
                     .await
                 };
--- a/sdk/src/reader.rs
+++ b/sdk/src/reader.rs
@@ -1385,6 +1385,70 @@ pub mod tests {
     }
 
     #[test]
+    #[allow(deprecated)]
+    fn test_detached_manifest_exclusion_hole_rejected() {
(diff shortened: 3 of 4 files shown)


sdk/src/claim.rs
```diff
@@ -2959,8 +2959,33 @@ impl Claim {
                     }
 
                     if !dh.is_remote_hash() {
-                        // there are extra exclusion then log the information code about extra exclusion
                         if let Some(exclusions) = &dh.exclusions {
+                            // Exclusions only make sense for a manifest embedded in this asset;
+                            // a detached manifest's exclusion is a hole for unrelated content.
+                            // If the caller didn't read the manifest out of this asset, fall back
+                            // to checking whether the asset really does have a C2PA box there.
+                            let box_present_in_asset = svi
```
scouten-adobe, line 2967: **[MEDIUM]** The new check only verifies that *one* exclusion in the list equals `svi.manifest_store_range` (`exclusions.contains(range)`); it doesn't validate every non-zero-length exclusion against reality. A manifest whose signer legitimately declared more than one exclusion (e.g. TIFF's `ObjectType::OtherExclusion` for the C2PA tag's IFD count field, `sdk/src/asset_handlers/tiff_io.rs:2508`) would satisfy `box_present_in_asset` via the C2PA-box entry alone, while any *other* declared exclusion range is never checked against the asset actually being validated — it's still a blind hash-skip.

Current real-world impact is narrow (TIFF's second exclusion is a fixed 4/8-byte structural field, not attacker-useful content), but this is a gap relative to the PR's own stated goal ("reject a non-empty exclusion unless..."), and JPEG XL code already lists `OtherExclusion` as an expected type for its own hash-verification checks, suggesting more exclusions per manifest may become more common. Worth tightening to check *all* non-zero-length exclusions, not just the one matching the box, or documenting why only the primary exclusion needs this guarantee.

scouten-adobe (approved):
## PR Review: fix: Reject data hash exclusions for detached manifests without a matching C2PA box

### Summary
This is a real, well-scoped fix for a validation-bypass vulnerability: a detached manifest's `c2pa.hash.data` exclusion range was applied to whatever asset was handed to the verifier without checking that the excluded bytes were actually a C2PA box, letting an attacker with no signing key replay a legitimately-signed manifest against unrelated forged content. I independently reproduced the vulnerability and verified the fix (see Affected-Test Run). The core defense is sound; I found one real completeness gap in the new check's scope (flagged inline) and two low-severity consistency gaps in how `Store::embedded` is tracked across less-common code paths (listed below, both outside this PR's diff so not inline-commentable). None of these are blocking — flagging them for follow-on work.

### Change Surface
- Modified: `sdk/src/claim.rs` (new rejection check in `verify_hash_binding`), `sdk/src/store.rs` (new `StoreValidationInfo::is_embedded`, `Store::embedded` now threaded through `from_stream`, the renamed `from_manifest_data_and_stream_with_embedded[_async]`, and the sign-then-verify path), `sdk/src/ingredient.rs` (threads `embedded` through the primary sync/async stream-loading path), `sdk/src/reader.rs` (new regression test only)
- No new public API surface — `Store::from_manifest_data_and_stream` keeps its old signature and now just delegates with `embedded=false`; the new `_with_embedded` variant is `pub(crate)`.
- Scope matches the PR description closely.

### Blast Radius
- Checked all 14 call sites of `verify_store`/`verify_store_strict`. The three calls with `asset_data: None` never reach the new check (guarded by `if let Some(data) = asset_data` in `verify_store`), so they're unaffected.
- Two `Store`-construction paths never set `.embedded` and were not touched by this PR (see Findings below). Neither causes a functional regression today because the new check has a second, independent path to success — `object_locations`-derived `manifest_store_range` matching an exclusion — which covers genuinely-embedded content even when `is_embedded` is wrongly `false`.
- No dangling references or orphaned symbols from the rename; confirmed via repo-wide search that all external callers (`c2pa_c_ffi`, `manifest.rs`, `builder.rs`, `asset_handlers/c2pa_io.rs`) use only the unchanged public signature of `Store::from_manifest_data_and_stream`.

### Findings

**MEDIUM** (flagged inline on `sdk/src/claim.rs`):
- [MEDIUM] sdk/src/claim.rs:2967 — the new check only verifies that *one* exclusion matches `manifest_store_range`; other non-zero-length exclusions in the same manifest are never validated against the current asset.

**LOW** (outside this PR's diff, so noted here rather than inline):
- [LOW] sdk/src/ingredient.rs:990 — `Ingredient::from_stream_async` (deprecated since 0.79.4) has its own copy of the manifest-loading logic, in `from_stream_async_with_settings`, that was never updated to set `embedded`, unlike the sync `from_stream` → `add_stream_internal` path this PR fixed. Functionally safe today (protected by the `manifest_store_range` fallback described above) but `Store::is_embedded()` will misreport for ingredients loaded this way until the deprecated path is removed (scheduled 0.92.0).
- [LOW] sdk/src/store.rs:3866 and 3909 — same gap for `load_from_file_and_fragments`/`load_fragment_from_stream` (fragmented BMFF init-segment loading): `store.embedded` stays at its `Default` (`false`) even though the manifest is read directly from the asset being validated. Currently moot since fragmented BMFF content hard-binds via `BmffHash` rather than `DataHash` (the only assertion type the new check inspects), but worth a follow-up for consistency.

### Test Gaps
- No test exercises the *positive* detached-but-matching-box path — i.e. validating a manifest against the same asset it's still embedded in via the detached-loading API (explicitly called out as a supported case in the PR description) — only the rejection case is tested.
- No test covers the sign-then-verify (`verify_after_sign`) path for `RemoteManifest::SideCar`/`Remote`, which is exactly the code this PR added (`self.embedded = !matches!(...)` in store.rs).
- No test covers the update-manifest + exclusion-adjustment interaction with the new check.

### Affected-Test Run
- Built a temporary worktree at the PR head and ran the new test directly: **passes** with the fix, **fails** (reports `Trusted`, matching the PR's own reproduction claim) when `sdk/src/claim.rs` is reverted to `main` — confirms the test is a real regression guard, not a tautology.
- Full `cargo test --package c2pa --lib`: 1025 passed, 3 failed. All 3 failures (`store::tests::test_png_jumbf_generation`, `test_png_compressed_jumbf_generation`, `test_user_guid_external_manifest_embedded`) reproduce identically on unmodified `main` (missing `libpng-test.png` fixture, file I/O disabled) — pre-existing environment issue, unrelated to this PR.

### Lint
- `cargo clippy --all-targets --no-deps -- -D warnings`: clean.
- `cargo +nightly fmt --all -- --check`: clean.

### Security
- No secrets, injection, resource-exhaustion, path-traversal, or dependency concerns — this is pure validation logic over already-parsed structures.
- Not applicable: no installer/hook/permission/telemetry/network surfaces touched.
- Core fix verified effective by direct reproduction (see Affected-Test Run and the inline finding on claim.rs for the one real completeness gap).

### Verdict
**Approve** (with follow-on work requested — see Findings)

The fix correctly closes the described vulnerability (verified via revert-and-reproduce) and the blast radius is well contained. The findings above are worth follow-up PRs but don't undermine the fix for its stated threat model.

---

PR #2653, opened 2026-09-14
fix: Reject data hash exclusions that don't match the manifest location

ssanthosh:
### Vulnerability
A signed SVG could be tampered to render attacker content while still validating as `Trusted`, with no signing key. The `c2pa.hash.data` exclusion covers the base64 manifest text. An attacker shrinks the manifest by trimming the COSE `pad` (an *unprotected*, unsigned header), patches the JUMBF length fields, and fills the space freed *inside the still-signed exclusion* with renderable SVG markup (closing the manifest early, injecting a `<rect>`, reopening an empty dummy manifest). The signature (protected payload unchanged) and data hash (injected bytes are excluded) both still verify. Confirmed end-to-end: a forged SVG containing `<rect fill='red'/>` validated as `Trusted`. Per C2PA 2.4 §15.12, "a validator shall ensure that the data contained within the exclusion range containing the C2PA Manifest Store consists of only the C2PA Manifest Store and any appropriate padding … If a validator encounters any data other than what is permitted, then the manifest shall be rejected with a failure code of `assertion.dataHash.mismatch`." c2pa-rs did not perform that check. In scope per SECURITY.md (validation bypass).

### Fix
At data-hash validation, require the declared exclusion to match the manifest store's real location, found independently by re-parsing the asset (`manifest_store_range`). If no exclusion equals it, the exclusion covers non-manifest bytes → reject with `assertion.dataHash.mismatch`. This generalizes the existing detached-manifest check (#2643): it now also fires for embedded manifests whose exclusion no longer matches the manifest (the shrink-and-inject case). Legitimate metadata exclusions are unaffected — they are separate ranges and still flow to the informational `additionalExclusionsPresent` code. The check is format-agnostic (shared `c2pa.hash.data` path), so it covers every data-hash format, not just SVG.

### Tests added
`test_svg_shrunk_manifest_injection_rejected` (svg_io.rs) — signs an SVG, shrinks the manifest by trimming the unsigned COSE `pad`, patches JUMBF lengths, injects renderable `<rect>` markup into the freed exclusion space, and asserts the forged asset is not `Trusted` and reports `assertion.dataHash.mismatch`. Fails without the fix (validates as `Trusted`), passes with it.

## Checklist


sdk/src/claim.rs
```diff
@@ -400,6 +400,34 @@ impl Serialize for Claim {
...
+) -> bool {
+    // Zero-length exclusions hash the whole asset; there is nothing to police.
+    let has_real_exclusion = exclusions.iter().any(|e| e.length() > 0);
+    if !has_real_exclusion {
+        return true;
+    }
+
+    match manifest_store_range {
+        // We located the manifest: an exclusion must cover exactly it.
+        Some(range) => exclusions.contains(range),
```
gpeacock: A HashRange can contain an optional bmff field, I'm not sure if that impacts this.

sdk/src/asset_handlers/svg_io.rs
```diff
@@ -720,6 +720,151 @@ pub mod tests {
...
 
+    // A signed SVG's `c2pa.hash.data` exclusion covers the base64 manifest text.
+    // An attacker can shrink the manifest by removing the (unsigned) COSE `pad`
+    // header, then fill the freed space -- still inside the signed exclusion --
+    // with renderable SVG markup. The signature and data hash still verify, so
+    // without the manifest-location check this validates as trusted while
+    // rendering attacker content. Guards against that regression.
+    #[test]
+    #[allow(deprecated)]
+    fn test_svg_shrunk_manifest_injection_rejected() {
```
gpeacock: note that this attack could apply to any text format that embeds base64 manifests.

scouten-adobe (approved):

sdk/src/claim.rs
```diff
@@ -400,6 +400,34 @@ impl Serialize for Claim {
...
+/// space freed by shrinking the manifest.
+fn data_hash_exclusions_match_manifest(
+    exclusions: &[HashRange],
+    manifest_store_range: Option<&HashRange>,
+    is_embedded: bool,
+) -> bool {
+    // Zero-length exclusions hash the whole asset; there is nothing to police.
+    let has_real_exclusion = exclusions.iter().any(|e| e.length() > 0);
+    if !has_real_exclusion {
+        return true;
```
ok-nick: What is a zero length exclusion? I don't see this described in the spec.

sdk/src/claim.rs
```diff
@@ -400,6 +400,34 @@ impl Serialize for Claim {
...
+    }
+
+    match manifest_store_range {
+        // We located the manifest: an exclusion must cover exactly it.
+        Some(range) => exclusions.contains(range),
+        // We could not locate it: accept only when the manifest was read from this
+        // asset (embedded), not supplied separately (detached).
+        None => is_embedded,
```
ok-nick, line 427: If it's not embedded then this function should never run/check anything since there is no manifest range?

sdk/src/claim.rs
ok-nick: I think this is a little more nuanced.

> "a validator shall ensure that the data contained within the exclusion range containing the C2PA Manifest Store consists of only the C2PA Manifest Store and any appropriate padding … If a validator encounters any data other than what is permitted, then the manifest shall be rejected with a failure code of assertion.dataHash.mismatch."

From my understanding, the spec says that an exclusion range that includes any part of the manifest store should emit the validation error. We need to check all exclusion ranges to ensure it doesn't overlap with the manifest store range.

sdk/src/asset_handlers/svg_io.rs
```diff
@@ -720,6 +720,151 @@ pub mod tests {
...
+        use std::io::Cursor;
+
+        use crate::{
```
tmathern: Do you want to keep those inside the test? Why not move them up on top of the mod?

sdk/src/claim.rs
tmathern: +1

ssanthosh: Its not a spec concept — it's our SDK implementation detail. Remote/sidecar signing emits a c2pa.hash.data with a length-0 placeholder exclusion (nothing excluded, whole asset hashed). The guard skips policing when nothing is excluded, keeping those valid (test_external_manifest_embedded_tiff fails without it). Reworded the comment with https://github.com/contentauth/c2pa-rs/pull/2653/commits/060ddfa5ee22d17445912905af2c7b637f00834b

sdk/src/claim.rs
ssanthosh: This case is the detached-manifest fix (#2643): a manifest from a sidecar/remote pointed at an unrelated asset has manifest_store_range == None, and returning is_embedded (false) rejects it. Skipping non-embedded would reopen that bypass. Reworded to reference #2643 with https://github.com/contentauth/c2pa-rs/commit/060ddfa5ee22d17445912905af2c7b637f00834b

sdk/src/asset_handlers/svg_io.rs
ssanthosh: Done with https://github.com/contentauth/c2pa-rs/commit/060ddfa5ee22d17445912905af2c7b637f00834b

sdk/src/claim.rs
ssanthosh: Thats a different check that is already done where we check if there is any overlap between exclusions and fail with `assertion.dataHash.malformed` validation error.  
So the rule constrains the content of the manifest-containing exclusion (manifest + padding only), not that exclusions must avoid the manifest. This check enforces that via geometry: `manifest_store_range` is the manifest's real extent, re-parsed from the asset, and we require an exclusion equal to it. The shrink-and-inject attack bloats that exclusion with non-padding bytes, so it no longer matches → rejected with assertion.dataHash.mismatch, the exact code the spec mandates.

tmathern (commented):
Approving.

Question: Do other formats have similar bypasses? Might be worth considering a sweep for other formats (I would start with images, JPEG and PNG come to mind).

ok-nick (approved):

Issue #2659, opened 2026-09-14
0.91.0-rc.1: direct_cose_handling signer fails with CoseTimeStampGeneration


scouten-adobe:
## Summary

In the `c2pa-rc-v0.91.0-rc.1` release candidate, signing with an `AsyncSigner`/`Signer` where `direct_cose_handling()` returns `true` now fails with `Error::CoseTimeStampGeneration`, even though the same signer implementation succeeds against `c2pa-v0.90.22`.

The `direct_cose_handling` contract is documented as: the signer is responsible for the entire COSE structure, and the SDK should not attempt to interpret or verify the returned bytes ("do not verify remote signers (we never did)" per the comment in `Store::sign_claim`). This RC appears to violate that contract somewhere after signing — likely in the (unconditional) `verify_after_sign` step in `Store::save_to_stream` / `Builder::sign`, which isn't gated on `direct_cose_handling` and now seems to attempt COSE/CBOR parsing of the signer's raw output.

This was found while testing [contentauth/c2pa-js](https://github.com/contentauth/c2pa-js) against the RC ahead of the 0.91.0 release — see [contentauth/c2pa-js#217](https://github.com/contentauth/c2pa-js/pull/217). `c2pa-wasm`'s `WasmSigner` always sets `direct_cose_handling() -> true` (browser/WASM signers return an opaque COSE_Sign1 built by the signing callback), so every c2pa-web test that signs now fails.

## Minimal reproduction

`Cargo.toml`:
```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
c2pa = { git = "https://github.com/contentauth/c2pa-rs.git", tag = "c2pa-rc-v0.91.0-rc.1", features = [
    "unstable_builder_filter",
    "rust_native_crypto",
    "http_reqwest",
], default-features = false }
async-trait = "0.1"
tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
serde_json = "1"
```

`src/main.rs`:
```rust
use std::io::Cursor;

use c2pa::{AsyncSigner, Builder, Result, SigningAlg};
use serde_json::json;

struct NoOpSigner;

#[async_trait::async_trait]
impl AsyncSigner for NoOpSigner {
    async fn sign(&self, _data: Vec<u8>) -> Result<Vec<u8>> {
        // A direct_cose_handling signer may return any opaque byte sequence;
        // the SDK is documented not to interpret or verify it.
[...]

PR #2660, opened 2026-09-14
fix: Skip post-sign store verification for direct_cose_handling signers

scouten-adobe:
## Summary

- `Store::sign_claim()` already skips its own signature check when `signer.direct_cose_handling()` is `true`, since that contract means the SDK must not interpret or verify the signer's opaque COSE bytes ("do not verify remote signers (we never did)").
- However, several callers run an *additional*, unconditional `verify_after_sign -> verify_store_strict()` pass after signing that was **not** gated on `direct_cose_handling()`. That pass tries to parse the signer's opaque bytes as a COSE structure and fails with `Error::CoseTimeStampGeneration` (a lossy mapping of `coset::CoseError::CborParsingError`/`CborGenerationError`).
- This broke any `direct_cose_handling` signer (e.g. `c2pa-wasm`'s `WasmSigner`, used by all of `c2pa-js`) starting with `0.91.0-rc.1`, even though the signer implementation itself was unchanged and worked fine against `0.90.22`.

Fixes the four production call sites that had this unconditional verification:
- `Store::sign_manifest` (both the dynamic-assertions and plain branches)
- `Store::get_data_hashed_embeddable_manifest`
- `Store::save_to_stream`
- `Builder::sign_box_hashed_embeddable` (deprecated)

## Root cause

Reported in https://github.com/contentauth/c2pa-rs/issues/2659, found while testing `c2pa-js` against the `0.91.0` RC (see [contentauth/c2pa-js#217](https://github.com/contentauth/c2pa-js/pull/217)).

## Test plan

- [x] Added `test_direct_cose_handling_signer_skips_post_sign_verification` in `sdk/tests/test_builder.rs`, adapted from the issue's minimal repro. Verified it fails with `CoseTimeStampGeneration` before this fix and passes after.
- [x] `cargo test -p c2pa --lib --features file_io store::` — all pass
- [x] `cargo test -p c2pa --lib --features file_io builder::` — all pass
- [x] `cargo test -p c2pa --test test_builder` — all pass

## Note

The issue also points out that `Error::CoseTimeStampGeneration` is a misleading, lossy conversion for any `CborParsingError`/`CborGenerationError` (not just timestamp-related ones), which made this harder to diagnose. That's a separate, more invasive change (the variant would need to carry the underlying error string, and it's touched from `c2pa_c_ffi` too) and is left out of this fix to keep it targeted; happy to follow up if desired.

This targets `main`; can backport to the `0.91.0` RC branch if needed.


sdk/src/builder.rs
```diff
@@ -3230,13 +3230,17 @@ impl Builder {
                 .get_box_hashed_embeddable_manifest_async(signer, &self.context)
                 .await
         }?;
-        if self.context.settings().verify.verify_after_sign {
+
+        // A `direct_cose_handling` signer returns an opaque, signer-owned COSE
+        // structure that the SDK is documented not to interpret or verify.
+        if self.context.settings().verify.verify_after_sign && !signer.direct_cose_handling() {
```
gpeacock: This is at the wrong level. There was a time when verify_after_sign only applied to the cose structure, but now it does much more than that.

sdk/src/store.rs
```diff
@@ -2423,7 +2423,9 @@ impl Store {
 
             let jumbf_bytes = self.to_jumbf_internal(signer.reserve_size())?;
 
-            if context.settings().verify.verify_after_sign {
+            // A `direct_cose_handling` signer returns an opaque, signer-owned COSE
+            // structure that the SDK is documented not to interpret or verify.
```
gpeacock: The SDK documents that the signer is responsible for the cose processing, but I don't know of any promise to not interpret or verify it. But we should assume that the signer has done all the work to properly construct it. I'm curious what the error is and if it is  valid.

sdk/src/store.rs
gpeacock: The sign_claim operation is where we first check verify_after_sign and do not verify the structure. Nothing has changed there and I presume this call is passing ok.

sdk/src/store.rs
```diff
@@ -3105,7 +3110,9 @@ impl Store {
                     Some(RemoteManifest::SideCar) | Some(RemoteManifest::Remote(_))
                 );
 
-                if context.settings().verify.verify_after_sign {
+                // A `direct_cose_handling` signer returns an opaque, signer-owned COSE
+                // structure that the SDK is documented not to interpret or verify.
+                if context.settings().verify.verify_after_sign && !signer.direct_cose_handling() {
                     let output_len = stream_len(output_stream)?;
                     let validate_hash = context.settings().verify.verify_after_sign_hash;
```
tmathern, line 3117: So here we skip hash checks too?

scouten-adobe:
Per discussion with team, closing as incorrect.

PR #2666, opened 2026-09-17
feat: Implement pdf manifest writing

mpgopala:
Completes the C2PA writer for PDF (previously read-only):

- Add `get_object_locations` for PdfIO, locating the embedded manifest
  stream's raw byte range by re-parsing the written PDF's xref table,
  so the sign pipeline can correctly exclude it when hashing.
- Add a same-length in-place patch fast path to `write_c2pa`: the sign
  pipeline embeds a placeholder manifest, then rewrites it with the
  final same-length manifest once the real hash is known. A second
  full lopdf rewrite isn't guaranteed to reproduce the same object
  layout, which was shifting the manifest out from under the already
  computed hash exclusion.
- Fix `remove_manifest_from_annotations` to skip pages with no
  `/Annots` entry instead of failing, so re-signing a PDF whose
  manifest is attached only via `/AF` (no `/Names` or per-page
  annotations) no longer errors.
- Add end-to-end `Builder::sign` -> `Reader` tests for a fresh PDF and
  for re-signing an already-signed PDF, plus unit tests for object
  location correctness and offset stability across the two write
  passes.
- Update docs/supported-formats.md and asset_handlers/README.md to
  drop the "read-only" PDF notes.

Note: this is a full rewrite on each sign, not an incremental
(byte-preserving) update per C2PA Annex A.4.2.1 — see
docs/pdf-oxide-incremental-update-write.md for that follow-up.


sdk/src/asset_handlers/pdf_io.rs
```diff
@@ -68,6 +152,123 @@ impl PdfIO {
...
+            .map_err(|_| Error::EmbeddingError)?;
+
+        let mut out_bytes = Vec::new();
+        pdf.save_to(&mut out_bytes).map_err(Error::IoError)?;
+        output_stream.rewind()?;
+        output_stream.write_all(&out_bytes)?;
+        Ok(())
+    }
+
+    fn get_object_locations(
```
tmathern, line 206: It is a tad weird, but `get_object_locations` is I believe consulted on read paths too. So we'll probably need to make sure there is no regression, especially with all the variations PDF can have...

gpeacock (changes requested):
We should add support for XMP so we can read and write remote urls. I don't think we only want to support embedding for pdfs

sdk/src/asset_handlers/pdf_io.rs
mpgopala: Good question. `locate_manifest_content` only ever returns `Ok(Some(range))` when it can positively verify the located bytes match the manifest's own content. Anything it's not sure about, comes back as `Err`. And before this PR, PDF's `get_writer` returned `None` entirely, so `object_locations` for PDF always hit that same Err/fallback path anyway (via `Error::UnsupportedType`). So the worst case for any variation this can't handle is "no worse than before this PR," not a new silent failure mode.
I have ran the tests to make sure it is passing all the tests and also added a `test test`_manifest_object_offset_errors_on_compressed_xref_entry which has a compressed manifest. Do let me know if we have any other _types_ of manifests we can check on!

tmathern (approved):
Pending the XMP part review by @gpeacock

gpeacock (approved):
This addresses my concern about support for remote.

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 35888292421](https://github.com/contentauth/c2pa-rs/actions/runs/35888292421).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2700 |

PR #2667, opened 2026-09-17
fix: migrate legacy trust anchors in Settings::from_string()

### LINKED ISSUE #2663: Settings::from_string() doesn't migrate legacy trust_anchors/user_anchors into the new anchors list
## Summary

As of `0.91.0-rc.1` (tag `c2pa-rc-v0.91.0-rc.1`, commit `7239de5dc318edf32a184673feb396eb67a04f44`), the deprecated, thread-local `c2pa::settings::Settings::from_string()` static method no longer causes trust anchors supplied via the legacy flat `trust.trust_anchors` / `trust.user_anchors` JSON fields to be honored during certificate-chain trust validation. Anchors supplied the same way through the newer `Settings::new().with_json(...)` builder path work correctly.

## Root cause (as far as I can tell)

`sdk/src/settings/mod.rs` has two separate code paths for loading settings from a JSON string:

- `Settings::new().with_json(json)` → `with_string()` (instance method) — this got a new backward-compat shim in the `anchors`/`TrustAnchor` unification work. It pulls `trust.trust_anchors` / `trust.user_anchors` out of the overlay JSON, converts each into a `TrustAnchor { trust_kind: TrustListKind::Manifest, .. }`, and merges them into `settings.trust.anchors`.
- `Settings::from_string(json, "json")` (deprecated static method, sets thread-local `SETTINGS`) — this is a plain `merge_json(&mut merged, overlay)` followed by `serde_json::from_value`. It does **not** run the same legacy-field migration, so `trust.trust_anchors` / `trust.user_anchors` land only in the (now otherwise-inert) deprecated `Trust::trust_anchors` / `Trust::user_anchors` fields, and `Trust::anchors` stays `None`.

Since `Store::from_context` (`sdk/src/store.rs:168`) only loads certificate trust anchors from `settings.trust.anchors`, any consumer that loads trust settings via `Settings::from_string()` now gets **no trust anchors at all** in the resulting store, regardless of what was passed in — even though the call itself succeeds without error.

## Why this matters

`Settings::from_string()` is deprecated (since 0.79.4) but not yet removed (scheduled for 0.92.0), and it's still the only practical option for some consumers — e.g. a C FFI layer that needs thread-local settings semantics and has to work around a bounded string-length limit on the C API surface, so it calls `Settings::from_string()` directly instead of the instance builder. For that kind of consumer, upgrading to `0.91.0` silently breaks trust anchor loading: no error is raised, but every previously-trusted signing certificate now validates as untrusted (`signingCredential.untrusted`).

## Repro sketch

```rust
use c2pa::settings::Settings;

#[allow(deprecated)]
fn main() {
    let json = r#"{"trust":{"trust_anchors":"-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----"}}"#;

    // Works: anchors end up in settings.trust.anchors via the with_string() migration shim.
    let built = Settings::new().with_json(json).unwrap();
    assert!(built.trust.anchors.is_some());

    // Broken: same JSON, but via the deprecated thread-local API.
    let applied = Settings::from_string(json, "json").unwrap();
    assert!(applied.trust.anchors.is_none()); // anchors never populated
}
```

## Suggested fix

Route `Settings::from_string()` through the same legacy-anchor migration logic used by `with_string()` (or have it delegate to `with_string()` on top of the current thread-local value), so both paths treat `trust.trust_anchors` / `trust.user_anchors` consistently until they're removed in 0.92.0.

## Environment

Found while testing `adobe_api` (Adobe's internal C2PA integration) against the `0.91.0-rc.1` release candidate ahead of its promotion to a full release.

scouten-adobe:
## Summary

* `Settings::from_string()` (the deprecated, thread-local settings API) merged the raw `trust.trust_anchors` / `trust.user_anchors` overlay JSON directly into the thread-local value, without running the legacy-anchor migration shim that `Settings::new().with_json()` uses via `with_string()`. As a result, anchors supplied through those deprecated flat fields never made it into `settings.trust.anchors`, and `Store::from_context` silently loaded zero trust anchors (no error raised).
* Factored the shared migration logic out of `with_string()` into two helpers — `take_trust_anchor_overlay_fields()` and `merge_legacy_trust_anchors()` — and now run them from `from_string()` as well, so both paths treat `trust.trust_anchors` / `trust.user_anchors` consistently until they're removed in 0.92.0.
* `from_string()` writes the migrated `trust` settings back into just the `trust` subtree of the thread-local JSON value (rather than replacing the whole value), so "hidden" thread-local fields outside of `trust` that aren't modeled by the `Settings` struct are preserved (see `test_thread_local_hidden_setting`).

Fixes #2663.

## Test plan

- [x] `cargo test -p c2pa --lib settings::tests` — all 22 tests pass, including a new regression test `test_from_string_loads_legacy_trust_anchors` based on the repro in the issue
- [x] `cargo clippy -p c2pa --lib --all-features -- -Dwarnings`
- [x] `cargo test -p c2pa --lib` (full suite, aside from 3 pre-existing unrelated failures due to a missing PNG fixture in this environment)


gpeacock (approved):
Note to self, legacy apis should not change behavior once we deprecate them!

PR #2668, opened 2026-09-17
fix: reject box-hash manifests with a truncated `boxes` list

mpgopala:
## Summary

The box-hash (`c2pa.hash.boxes`) verifier never checked that the `boxes`
array named every box actually present in the asset. A truncated list
(e.g. only `"C2PA"`, omitting trailing scan/pixel-data boxes) left the
unnamed boxes completely unhashed while verification still reported a
match, letting an attacker byte-edit that unhashed content in an
already-signed asset without any key and still get
`validation_state: Trusted`.

Adds `test_verify_rejects_truncated_box_list`, which encodes a two-box
asset with a `boxes` list naming only the first box and asserts the
`unknownBox` rejection.


gpeacock (approved):
This looks like it only adds the box hash check though the comment s say it covers bmff and data hash checks too.
Still this one looks good as far as it goes. Maybe update the comments?

PR #2669, opened 2026-09-17
fix: Bound cumulative TIFF IFD entry byte counts to file size

ssanthosh:
### Vulnerability
A crafted TIFF can list many distinct private IFD tags whose offset/count all reference the same in-file byte range. `clone_ifd_entries` bounded each entry's claimed byte size against the whole file size individually, but never capped the sum across all entries in an IFD. Each entry passes its own check while the cloned buffers retained together sum to a large multiple of the actual file (e.g. 536 tags × 600,000 bytes from a ~600 KB file), OOM-killing the process. The 8-byte element types (Rational/SRational/Double/Long8/SLong8/Ifd8) also had no size check at all before reading.

### Fix
Added a cumulative byte counter in `clone_ifd_entries` that rejects once the running total across all entries in an IFD exceeds the source file size, reusing the existing `accumulate_copy_len` helper already used for the analogous strip/tile amplification fix. Also added the missing per-entry size check for the 8-byte element types.

### Tests added
- `clone_ifd_entries_rejects_cumulative_amplification` — entries individually within the file size but summing over it are rejected (fails without the fix).
- `clone_ifd_entries_accepts_valid_cumulative_size` — legitimate entries summing to <= file size still clone successfully.

## Checklist


sdk/src/asset_handlers/tiff_io.rs
```diff
@@ -1999,6 +1999,9 @@ impl<T: Read + Write + Seek> TiffCloner<T> {
         let file_size = stream_len(asset_reader)?;
         let mut target_ifd: BTreeMap<u16, IfdClonedEntry> = BTreeMap::new();
 
+        // Bound bytes cloned across all entries in this IFD, not just each one alone.
+        let mut cumulative_ifd_bytes: u64 = 0;
```
tmathern, line 2003: Each call restarts from 0, so each IFD gets that. 
Question (non-blocking): Should it be per IFD or per file?

tmathern (approved):

sdk/src/asset_handlers/tiff_io.rs
ssanthosh: It is currently per IFD only. The aggregate is still bounded (each IFD ≤ file size; the number of IFDs is itself bounded by file size), so there's no unbounded amplification. A per-file bound risks false-positives on legitimate multi-page/DNG files whose IFD entries across pages reference overlapping/shared byte ranges (their sizes can legitimately sum past the file length).

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 35906189067](https://github.com/contentauth/c2pa-rs/actions/runs/35906189067).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2706 |

PR #2670, opened 2026-09-17
fix: Harden against integer underflow attacks in XMP trailer parsing

ssanthosh:
## Vulnerability

`add_xmp_key` and `remove_xmp_key` in `sdk/src/utils/xmp_inmemory_utils.rs`
computed the XMP padding target length as `orig_length - XMP_END.len()`
after finding a `<?xpacket end` trailer, assuming the trailer was always the
full 19-byte `<?xpacket end="w"?>`. When the input's XMP was truncated right
after the 13-byte prefix (13-18 total bytes), this `usize` subtraction
underflowed, wrapping to near-`usize::MAX` in release builds and driving an
effectively unbounded padding write — OOM from a few-KB input. This is
reachable via the public `Builder::set_remote_url` signing path (e.g.
`png_io.rs` reads attacker-controlled XMP and passes it to `add_provenance`).

The fix uses the already-matched trailer position (`pos`) instead, which can
never exceed `orig_length` and therefore can't underflow.

## Fix

- `sdk/src/utils/xmp_inmemory_utils.rs`: `target_length = pos` instead of
  `target_length = orig_length - XMP_END.len()`, in both `add_xmp_key` and
  `remove_xmp_key` (same bug duplicated in both).

## Tests added

- `add_xmp_key_truncated_trailer_does_not_underflow` /
  `remove_xmp_key_truncated_trailer_does_not_underflow`: 13-byte truncated
  trailer no longer underflows or hangs.
- `add_xmp_key_full_trailer_still_works`: control test confirming normal,
  well-formed XMP still round-trips correctly.

## Checklist


sdk/src/utils/xmp_inmemory_utils.rs
```diff
@@ -435,6 +435,35 @@ mod tests {
...
+        let truncated = "<?xpacket end";
+        assert_eq!(truncated.len(), 13);
+
+        let xmp = remove_provenance(truncated).expect("removing provenance");
+        assert_eq!(extract_provenance(&xmp), None);
+    }
+
+    // Control: a complete trailer should still round-trip normally.
+    #[test]
+    fn add_xmp_key_full_trailer_still_works() {
```
tmathern, line 459: Is the mentioned xpacket trailer added dynamically? 
I don't see it in XMP_DATA: https://github.com/contentauth/c2pa-rs/blob/dbfbc82d8db51a29a7d05434b59adf081e8fafc0/sdk/src/utils/xmp_inmemory_utils.rs#L386-L400

tmathern (approved):

sdk/src/utils/xmp_inmemory_utils.rs
gpeacock: It's added in the xmp_padding method.

gpeacock (approved):

caiopensrc:
[Backport-action](https://github.com/korthout/backport-action) backported this pull request in [workflow run 35903472717](https://github.com/contentauth/c2pa-rs/actions/runs/35903472717).

| Target | Status |
|--------|--------|
| `stable` | :white_check_mark: Created #2704 |

PR #2672, opened 2026-09-17
fix: report CBOR byte strings correctly in crJSON output

gpeacock:
Assertion::as_json_object now decodes CBOR directly instead of transcoding through serde_json, so byte strings (bstr) are wrapped as crJSON's `b64'<base64>'` strings instead of being flattened into plain JSON number arrays indistinguishable from integer arrays. ManifestAssertion::to_assertion() reverses that wrapping so typed structs with Vec<u8>/serde_bytes fields (e.g. IdentityAssertion's signature/pad1/pad2) still deserialize correctly.

Also adds Reader::crjson()/detailed_json() support to the `show` example, and consolidates ad hoc "b64'...'" formatting into a shared base64::encode_b64_wrapped() helper.

### DIFF
--- a/sdk/src/crypto/base64.rs
+++ b/sdk/src/crypto/base64.rs
@@ -25,6 +25,20 @@ pub fn decode(data: &str) -> Result<Vec<u8>, DecodeError> {
     general_purpose::STANDARD.decode(data)
 }
 
+/// Encode a byte slice into a crJSON `b64'<base64>'`-wrapped string.
+pub fn encode_b64_wrapped(data: &[u8]) -> String {
+    format!("b64'{}'", encode(data))
+}
+
+/// Decode a crJSON `b64'<base64>'`-wrapped string back into raw bytes.
+///
+/// Returns `None` if `value` isn't wrapped in the `b64'...'` delimiters (e.g. it's
+/// an ordinary string), so callers can tell "not base64" apart from "malformed base64".
+pub fn decode_b64_wrapped(value: &str) -> Option<Vec<u8>> {
+    let inner = value.strip_prefix("b64'")?.strip_suffix('\'')?;
+    decode(inner).ok()
+}
+
 #[cfg(test)]
 mod tests {
     #![allow(clippy::expect_used)]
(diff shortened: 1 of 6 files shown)


sdk/src/utils/hash_utils.rs
```diff
@@ -624,8 +633,10 @@ pub fn hash_to_b64(mut value: Value) -> Value {
         match current {
             Value::Object(obj) => {
                 for (_, v) in obj.iter_mut() {
-                    if let Value::Array(hash_arr) = v {
-                        if !hash_arr.is_empty() && hash_arr.iter().all(|x| x.is_number()) {
+                    match v {
+                        Value::Array(hash_arr)
```
tmathern, line 637: If string is the other match arm, what happens with an array of strings? Or is this call recursive somehow and the arm would be reached later?

gpeacock: Yes it is flattened recursive algorithm. That's what the queue is for. A bit confusing but it avoids stack overflows due to recursion.

sdk/src/crypto/base64.rs
```diff
@@ -25,6 +25,20 @@ pub fn decode(data: &str) -> Result<Vec<u8>, DecodeError> {
...
+}
+
+/// Decode a crJSON `b64'<base64>'`-wrapped string back into raw bytes.
+///
+/// Returns `None` if `value` isn't wrapped in the `b64'...'` delimiters (e.g. it's
+/// an ordinary string), so callers can tell "not base64" apart from "malformed base64".
+pub fn decode_b64_wrapped(value: &str) -> Option<Vec<u8>> {
+    let inner = value.strip_prefix("b64'")?.strip_suffix('\'')?;
+    decode(inner).ok()
+}
```
ok-nick: Can we return a result?

sdk/src/crypto/base64.rs
```diff
@@ -56,4 +70,30 @@ mod tests {
             Ok(b"Hello, world".to_vec())
         );
     }
+
+    #[test]
+    #[cfg_attr(
+        all(target_arch = "wasm32", not(target_os = "wasi")),
+        wasm_bindgen_test
+    )]
```
ok-nick: Can  we use the general c2pa test macro that handles all of these cases?

tmathern (approved):

PR #2679, opened 2026-09-18
fix: Cert-expiration check ignoring trusted timestamp for nested-ingredient claims

### LINKED ISSUE #2678: Cert-expiration check ignores trusted timestamp for nested-ingredient claims during full-store validation, breaking verify_after_sign
## Summary

With `verify_after_sign` now defaulting to `true` (#2277), signing a new manifest that includes, as an ingredient, an asset whose *own* ingredient chain contains a claim with a since-expired signing certificate now fails with `signingCredential.expired`, even though that nested claim was validly signed (per its own trusted timestamp) at the time. Under 0.90.x this was never surfaced because `verify_after_sign` defaulted to `false`.

Per the C2PA trust model, a certificate that has since expired should not invalidate a claim that carries a trusted timestamp proving it was signed while the cert was still valid. The certificate-expiration check appears to apply this "valid at time of signing" logic correctly in one code path but not another, causing an inconsistency.

## Evidence

Repro fixture: an asset with a 3-level ingredient chain (top manifest → ingredient A → ingredient B), where ingredient B's claim signature's certificate has since expired, but was signed under a trusted timestamp while still valid.

Running `c2patool <asset>` (full validation, not just read) on this asset shows the inconsistency directly in a single run:

- When ingredient B's manifest (`urn:c2pa:e1d26a14-...:adobe`) is validated **in ingredient context** (as a nested ingredient of the manifest above it), the per-ingredient `validation_results` reports success:
  ```
  claimSignature.insideValidity   self#jumbf=.../urn:c2pa:e1d26a14-.../c2pa.signature
  signingCredential.trusted       self#jumbf=.../urn:c2pa:e1d26a14-.../c2pa.signature
  ```
  i.e. the cert is correctly judged valid *as of the trusted timestamp*.

- But the aggregate, whole-store `validation_status` (the same summary that `Store::verify_store` / `verify_store_strict` produces, and that gates `sign()` when `verify_after_sign` is on) reports a **failure** for that exact same signature:
  ```
  {
    "code": "signingCredential.expired",
    "url": "self#jumbf=/c2pa/urn:c2pa:e1d26a14-.../c2pa.signature",
    "explanation": "certificate expired"
  }
  ```

So the *same claim signature*, validated by the *same tool invocation*, is judged "inside validity" in one code path and "expired" in another.

## Likely root cause

`check_certificate_profile()` in `sdk/crypto/cose/certificate_profile.rs` does the right thing when given a timestamp:

```rust
if let Some(tst_info) = _tst_info_opt {
    // checks against tst_info.gen_time — correct, timestamp-relative check
} else {
    // falls back to SystemTime::now() — wall-clock check
}
```

This suggests that somewhere along the call chain used for whole-store validation (`Store::verify_store` → `Verifier::verify_profile` → `check_end_entity_certificate_profile`), the `tst_info` for a claim that is only reachable as a nested/transitive ingredient isn't being threaded through, so it falls into the "no timestamp" branch and gets compared against today's date instead of the claim's own trusted signing time — even though that same claim's timestamp *is* available and validated elsewhere (the ingredient-context validation path clearly has it, since it reports `claimSignature.insideValidity` as a success).

## Impact

Any consumer that signs a new manifest built on top of an ingredient whose own (nested) provenance chain contains an old-but-legitimately-signed claim will now get a hard signing failure once that claim's certificate's natural expiration date passes — even though nothing about the ingredient's validity has changed. Since `verify_after_sign` is on by default as of 0.91, this affects any signing workflow that incorporates older, previously-valid assets as ingredients (e.g. long-lived asset libraries, provenance chains built up over months/years).

## Environment

- Reproduced against `c2pa` `0.91.0-rc.3` (tag `c2pa-rc-v0.91.0-rc.3`, commit `298f16e1068ece3e9972d712a4200eff08d16839`).
- Also reproduced with plain `c2patool` full validation (not tied to any downstream crate), so this is not specific to our own signing code.

## Ask

Could someone confirm whether `Store::verify_store`'s whole-store validation is intended to apply timestamp-relative expiration checking uniformly to nested-ingredient claims, and if so, look at why `tst_info` isn't reaching `check_certificate_profile` for those claims in that path? Happy to provide the repro fixture privately if useful — it's not something we can attach here directly, but it's a standard C2PA-signed JPEG with a 3-manifest ingredient chain.

scouten-adobe:
Fixes #2678.

## Problem

With `verify_after_sign` now defaulting to `true`, signing a new manifest that includes an ingredient whose own (nested) ingredient chain contains a claim signed by a since-expired certificate would fail with `signingCredential.expired`, even though that nested claim carries a trusted timestamp (or a previously recorded, attested validation) proving it was valid when signed.

## Root cause

`Store::ingredient_checks` always performs a full, live re-validation of every nested ingredient claim during whole-store validation (`Store::verify_store` / `verify_store_strict`, which gates `sign()`'s `verify_after_sign`). That live re-check can regress a nested claim from a genuine, previously-recorded success (`claimSignature.insideValidity`, attested in the ingredient assertion's own `validation_results` at add-time — cryptographically bound to that ingredient's own signature) to a `signingCredential.expired` failure once the certificate's wall-clock expiration date passes, even though nothing about the ingredient's actual validity changed.

`ValidationResults::from_store` already de-duplicates exact re-reports of ingredient-attested statuses (see the existing comment/logic in `sdk/src/validation_results.rs`), but it did not previously recognize this *contradicting*-status case: a freshly-computed `signingCredential.expired` delta for a manifest whose own attested `validation_results` already reports `claimSignature.insideValidity` for that same manifest.

## Fix

Extend the existing reconciliation logic in `ValidationResults::from_store` to also suppress that specific delta. This follows the same security model as the pre-existing dedup mechanism:
- Only ingredient-scoped, non-active-manifest statuses can be suppressed — a status describing the active manifest is always kept.
- Suppression only applies when the *same manifest label* has an ingredient attestation of `claimSignature.insideValidity` recorded in ingredient-assertion content bound to that ingredient's own signature (not forgeable without invalidating the containing claim's signature).

Added a regression test: `from_store_suppresses_ingredient_cert_expiry_attested_inside_validity`.

## Testing

- `cargo test -p c2pa --lib validation_results::tests` — pass (22/22, including the new regression test)
- `cargo test -p c2pa --lib` — 1039 passed, 3 pre-existing/unrelated failures (missing `libpng-test.png` fixture in this sandbox, confirmed to fail identically on the base commit before this change)
- `cargo clippy -p c2pa --all-targets --all-features -- -D warnings` — clean
- `cargo fmt -p c2pa` — applied

## Backport note

@scouten-adobe — this is a single-file, narrowly-scoped change to `sdk/src/validation_results.rs` and doesn't touch the certificate-trust-policy/certificate-profile code that's seen recent refactoring, so it should backport cleanly to a `release-0.91`-style branch for another RC build.


sdk/src/validation_results.rs
```diff
@@ -287,9 +298,18 @@ impl ValidationResults {
...
+                        && s.ingredient_uri().is_some()
+                        && !is_active_manifest(s.url())
+                        && s.url()
+                            .and_then(manifest_label_from_uri)
+                            .is_some_and(|label| attested_inside_validity.contains(&label));
+
                     s.ingredient_uri().is_none()
                         || is_active_manifest(s.url())
-                        || !ingredient_statuses.iter().any(|i| i == s)
+                        || (!ingredient_statuses.iter().any(|i| i == s)
```
tmathern, line 333: Non-blocking, future: Is the fact such a case was hit maybe something to consider logging as informational?

sdk/src/validation_results.rs
```diff
@@ -269,6 +272,14 @@ impl ValidationResults {
                     .flatten()
                     .collect();
 
+                let attested_inside_validity: HashSet<String> = ingredient_statuses
```
tmathern: Can this cross-reference other ingredient/manifests, or is it checked they can't "escape"? (e.g. I have 2 ingredient, A and B, each with a manifest. I can't somehow write ingredient A to say "ignore ingredient B, it was fine"?). My question comes due to potential flattening happening later, especially.

scouten-adobe: Good catch — this was in fact possible before this update. `attested_inside_validity` was built by scanning `store.claims()` globally (every claim in the store, not just ones reachable from the active manifest's real ingredient chain) and only matched on the manifest label named in the attested status's URL. Since manifest labels are visible/guessable and ingredient assertion content is attacker-influenceable, ingredient A could forge a `claimSignature.insideValidity` status whose URL names ingredient B's manifest label, and that alone would suppress a genuine live `signingCredential.expired` finding for B anywhere it was found in the store.

Fixed in d860eac8: I now track each ingredient's own manifest label alongside its reported statuses, and only add a label to the suppression set when the `insideValidity` attestation for that label was reported by the ingredient that *is* that label (self-attested), never by a different ingredient vouching for someone else. Added a regression test (`from_store_cannot_be_suppressed_by_unrelated_ingredient_attesting_foreign_manifest_label`) that reproduces exactly the A/B scenario you described — it fails without the fix and passes with it.

sdk/src/validation_results.rs
scouten-adobe: Worth considering, but not immediately. Filed #2680 to follow up.

tmathern (approved):

scouten-adobe:
Good catch on the memory usage — fixed in 1e96cc0b.

The self-attestation fix (d860eac8) had introduced an extra clone pass: it built an intermediate `Vec<(Option<String>, Vec<ValidationStatus>)>` holding every ingredient's statuses, then cloned all of them again to build the flat `ingredient_statuses` list used by the existing dedup logic. For deeply-nested manifests with many ingredients/statuses, that roughly doubled the memory held during this pass.

Reworked it into a single pass that moves each ingredient's statuses directly into `ingredient_statuses` (no clone) while building the `attested_inside_validity` label set from a borrow of the same slice just before the move. No behavior change — all tests (including the two regression tests added for the suppression/escape logic) still pass, along with clippy and fmt.

gpeacock (approved):
I think this looks ok, but it is very scary. That list of validations results is easy to mess up when adding extra conditions.

PR #2683, opened 2026-09-19
fix: reject direct requests to link-local/cloud-metadata hosts by default

mpgopala:
## Summary

- Adds a `MetadataGuardResolver` that rejects the *initial* HTTP request when its URI directly
  names a link-local or cloud-metadata address (e.g. `169.254.169.254`), independent of any
  redirect handling.
- `Context` installs this guard by default whenever no `allowed_network_hosts` allow-list is
  configured; an explicit allow-list still takes precedence and bypasses the guard.
- Introduces `HttpResolverError::MetadataOrLinkLocalUriDisallowed` and updates the `http`/`settings`
  module docs to reflect the narrowed "accepted risk" (loopback/RFC 1918 hosts are still reachable
  by default; only link-local/metadata addresses are blocked at the initial request).

# Fix for issue #2430 
CAI-12574 closed the SSRF *redirect* vector (blocking redirects to internal addresses). This ticket
tracks the residual gap: under the default config (no `allowed_network_hosts`), the *initial*
request was unrestricted, so an attacker-controlled manifest URL (e.g. via `dcterms:provenance`
XMP) could point directly at a cloud-metadata endpoint with no redirect involved. Since remote
manifest fetch is on by default, any app using default settings (including c2patool) was exposed.


sdk/src/http/restricted.rs
```diff
@@ -462,6 +462,106 @@ fn build_redirected_request(
...
+/// Returns true if `uri`'s host is a link-local address (`169.254.0.0/16`, `fe80::/10`) or a
+/// well-known cloud-metadata endpoint outside those ranges (e.g. AWS's IPv6 IMDS address). Used to
+/// guard the *initial* request by default (CAI-13326); see [`MetadataGuardResolver`].
+///
+/// Unlike [`host_is_non_global`], this does **not** flag loopback or private (RFC 1918) addresses —
+/// those remain reachable by default so `localhost`/intranet manifest hosts and enterprise PKI keep
+/// working. Use [`Core::allowed_network_hosts`] to restrict those as well.
+///
+/// [`Core::allowed_network_hosts`]: crate::settings::Core::allowed_network_hosts
+pub(crate) fn host_is_metadata_or_link_local(uri: &Uri) -> bool {
```
scouten-adobe, line 533: **[LOW] Things to consider:** `host_is_metadata_or_link_local` returns `false` when `uri.host()` is absent, while the sibling redirect classifier fails closed for hostless URIs. Hostless requests cannot normally be dialed by the concrete clients, so this is defense-in-depth consistency rather than a blocking issue.

sdk/src/http/restricted.rs
```diff
@@ -462,6 +462,106 @@ fn build_redirected_request(
...
+
+    if let Ok(ip) = host.parse::<IpAddr>() {
+        return ip_is_metadata_or_link_local(ip);
+    }
+
+    // Obfuscated numeric hosts (e.g. a decimal or hex encoding of 169.254.169.254) have no
+    // legitimate use as a manifest/OCSP/timestamp host; fail closed rather than risk an OS
+    // resolver treating them as link-local.
+    looks_like_obfuscated_ip(&host)
+}
```
scouten-adobe, line 550: **[LOW] Things to consider:** IPv6 zone-ID hosts such as `[fe80::1%25eth0]` do not parse as `IpAddr` and are not recognized by `looks_like_obfuscated_ip`. The concrete clients generally reject these forms, and the redirect classifier has the same limitation, but an explicit regression test or shared normalization rule would make this guard more robust.

scouten-adobe (commented):
## Review feedback

The implementation looks sound and the targeted plus full test suites, clippy, and formatting checks passed. I intend to approve this PR; the following are non-blocking things to consider for additional defense-in-depth:

- Fail-closed handling for hostless URIs in the new metadata classifier, to align it with the redirect classifier.
- Explicit coverage or normalization for IPv6 zone-ID forms such as `[fe80::1%25eth0]`.

These do not block approval: the concrete HTTP clients generally reject these forms, and the existing redirect classifier has the same zone-ID limitation.

---

scouten-adobe (approved):
Approving with a request to consider the minor issues identified.

sdk/src/context.rs
```diff
@@ -1208,6 +1214,30 @@ mod tests {
         redirect.assert_calls(1);
     }
 
+    // Under the default policy, a request whose URI *directly* names a link-local/cloud-metadata
+    // address is rejected as `MetadataOrLinkLocalUriDisallowed` (SSRF – CAI-13326), even though no
+    // redirect is involved. No network call is made — the guard rejects before dialing out.
+    #[test]
+    fn test_default_blocks_direct_metadata_host() {
+        use http::Request;
```
tmathern: Maybe move up those use statements at the top level of the test mod?

tmathern (approved):

sdk/src/http/restricted.rs
mpgopala: Done

sdk/src/http/restricted.rs
mpgopala: Done

sdk/src/context.rs
mpgopala: Done

PR #2686, opened 2026-09-21
fix: Scope ingredient status de-duplication to the recording ingredient

ssanthosh:
## Vulnerability
An attacker-supplied *unreferenced* manifest could suppress a genuine validation
failure on a referenced claim. `ValidationResults::from_store` de-duplicated live
findings against ingredient-recorded statuses pooled from **every** parsed claim
(matched only by `code` + `url` + `kind`), so a forged ingredient assertion inside
an unverified claim could cancel a real failure — `Reader` returned
`ValidationState::Trusted` with `validation_status() == None`, exposing tampered
metadata under the victim's label. No signing key required; default configuration.

## Fix
De-duplication is now scoped by each ingredient assertion's own URI: ingredient
statuses are keyed by `to_assertion_uri(claim, assertion)` (exactly what
`Store::ingredient_checks` pushes as `ingredient_uri`), and a live finding can only
be cancelled by the ingredient that actually recorded it — not a store-wide pool.
This closes the reported unreferenced-claim case and the referenced-claim variant of
the same class. The active-manifest and expiry-attestation guards are unchanged.

## Tests added
- `from_store_unreferenced_claim_cannot_suppress_referenced_failure` — fails without
  the fix, passes with it.
- Corrected `from_store_attested_ingredient_failure_after_serde_roundtrip` to push the
  real v3 ingredient URI (`c2pa.ingredient.v3`) that production emits (it previously
  used an unrealistic `c2pa.ingredient`).

## Checklist


sdk/src/validation_results.rs
```diff
@@ -281,25 +284,31 @@ impl ValidationResults {
...
+                        };
+                        if let Some(own_label) = &own_label {
+                            if group_statuses.iter().any(|status| {
+                                status.code() == validation_status::CLAIM_SIGNATURE_INSIDE_VALIDITY
+                                    && status
+                                        .url()
+                                        .and_then(manifest_label_from_uri)
+                                        .is_some_and(|label| &label == own_label)
+                            }) {
+                                attested_inside_validity.insert(own_label.clone());
```
tmathern, line 303: If you put the labels in the hashset, you don't know anymore from which ingredient it came. Can't there be cross-references then if there is only an inclusion check (that something exists in the hashset)?

sdk/src/validation_results.rs
```diff
@@ -328,10 +337,17 @@ impl ValidationResults {
                             .and_then(manifest_label_from_uri)
                             .is_some_and(|label| attested_inside_validity.contains(&label));
 
+                    // A live finding may only be cancelled by the ingredient that actually
+                    // recorded it: look up statuses under its own `ingredient_uri`, not a
+                    // store-wide pool, so an unrelated/unreferenced claim cannot suppress it.
+                    let attested_by_own_ingredient = s
```
tmathern, line 343: I think that might answer my other comment, but the hashset still makes me somewhat uneasy.

sdk/src/validation_results.rs
ssanthosh: Good catch — you're right that the label set lost the ingredient origin and allowed a cross-reference (an unreferenced claim could claim the victim's label and forge insideValidity to suppress its expiry). I removed the HashSet and scoped the expiry guard through the same per-ingredient_uri map, so an expiry finding is only excused by an insideValidity attestation from the same ingredient. Added a regression test for that exact case.  Changes are done with https://github.com/contentauth/c2pa-rs/pull/2686/commits/ca101f6a1c357f7fa6f23aac95967134ff15f6ac

sdk/src/validation_results.rs
ssanthosh: Yes. Removed the hashset and performed the necessary validation using https://github.com/contentauth/c2pa-rs/pull/2686/commits/ca101f6a1c357f7fa6f23aac95967134ff15f6ac

tmathern (approved):
Provisional, pending green CI/CD.
Suggestion, also: add a test with more deeply nested ingredients (and a mix of ingredient versions, too).

